PR #25, opened 2019-05-16
Use webpack copy plugin for images

teneightfive:
Images were being included via the main JavaScript file which was
causing them to be loaded in the JavaScript when they are actually
only used within the HTML and not wihtin any of the JavaScript.

This switches that method for the webpack copy plugin for copying
the source files to the build directory.


teneightfive:
Closing this PR as it's a little stale now. Will leave the branch to be picked up at a later date.

Have also created a [Jira ticket](https://dsdmoj.atlassian.net/browse/P4-550) to make sure it doesn't get missed.

PR #107, opened 2019-07-01
[P4-417] Add healthcheck endpoints

teneightfive:
This change adds a new healthcheck feature app that is responsible
for that status of the app and its dependencies.

It contains two endpoints:

1) `/healthcheck` - returns the status of the app and its dependencies in JSON format

2) `/healthcheck/ping` - to automate checking the status of the app
with other third party services like Pingdom. Will return `200` status code and the current app version in JSON format if everything is OK or `503` status code and `Service unavailable` if any dependencies are down.

## Optional todo

- [x] The API returns the currently deployed git hash and build date and tag (presumably from Cloud platform). If we know how to get those we can also return those at both endpoints.


teneightfive:
The problem with that though is we'll also use that ping endpoint for 3rd party services like Pingdom to tell us if there is a problem with some part of the app and to report downtime.

Wouldn't we ideally want that to notify us if there is a problem with the API connection, for example?

Unless we have a separate endpoint for internal dependencies to all dependencies?

PR #115, opened 2019-07-04
Bypass SSO

teneightfive:
This change allows SSO to be bypassed during local development.

It also includes some refactoring to the authentication middleware to follow more functional javascript theories and to simplify some of the tests.

It also includes some refactoring to move configuration of auth providers to the configuration module so that they can be swapped out in future if necessary.


teneightfive:
> This was a lot more code than I expected but I think it makes sense. I guess most of this is refactoring.

Yeah, it was a bit of refactoring to split functionality more where it belongs like auth specific middleware being moved into the auth feature. I tried to break it down a bit more in the commits.

PR #116, opened 2019-07-04
[P4-411] Store the person's reference number

teneightfive:
This change sets the current reference field to the specific
custody suite being used for the pilot.

It also adds support to the Person service formatter to correctly
handle any identifiers that are passed to it.

### DIFF
--- a/common/services/person.js
+++ b/common/services/person.js
@@ -11,13 +11,32 @@ function getFullname ({
 
 function format (data) {
   const relationships = ['gender', 'ethnicity']
+  const identifiers = [
+    'police_national_computer',
+    'criminal_records_office',
+    'prison_number',
+    'niche_reference',
+    'athena_reference',
+  ]
 
-  return mapValues(data, (value, key) => {
-    if (relationships.includes(key) && typeof value === 'string') {
-      return { id: value }
+  const formatted = mapValues(data, (value, key) => {
+    if (typeof value === 'string') {
+      if (relationships.includes(key)) {
+        return { id: value }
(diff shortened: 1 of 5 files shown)


common/services/person.js
```diff
@@ -11,13 +11,32 @@ function getFullname ({
...
-  return mapValues(data, (value, key) => {
-    if (relationships.includes(key) && typeof value === 'string') {
-      return { id: value }
+  const identifiers = [
+    'pnc_number',
+    'cro_number',
+    'prison_number',
+    'niche_reference',
+    'athena_reference',
+  ]
```
stevehook: We changed the first two of these keys to avoid the acronyms, see https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/77/commits/65c1eed29126f3b01a885eca1d1668b9754d1b5d

teneightfive: Ah okay. I’ll update them. 

Bit of inconsistency with those names now though :/ 

_number_ was dropped from `pnc_number` and `cro_number` but then _number_ is used in `prison_number` and _reference_ is used in `athena_reference` and `niche_reference`.

teneightfive: Maybe it just be?

```
'police_national_computer'
'criminal_records_office'
'nomis’
'niche'
'athena'
```

stevehook: yes that might be better, if everybody agrees I'm happy to change the others too

teneightfive:
Not one for now but we can decide maybe after next week.

PR #117, opened 2019-07-04
[P4-76] Configure Redis in production using K8s secrets

stevehook:
Cloud Platform outputs the connection details for Elasticache Redis instances as K8s secrets. These are mapped to environment variables to make them available to application code.

These two variables are relevant:

* `REDIS_URL` - is already used in non-production environments to declare the location of the Redis server. In production it just contains the host.
* `REDIS_AUTH_TOKEN` - is the Redis 'password' only relevant in production (Cloud Platform Elasticache instances are password protected).

This PR takes into account the differences between dev and prod.


config/redis-helpers.js
```diff
@@ -0,0 +1,13 @@
+/* eslint no-process-env: "off" */
+
+const buildRedisUrl = () => {
```
teneightfive: I'd be tempted to keep this within the config module as this breaks the rule of only declaring `process.env` within the config file (the eslint rule that you had to disable).

I know it might mean you won't be able to test it right now but I'd say put it in there because I'm doing a piece of work around Redis in #107 which will split building the redis client into a separate service where we can move this to and test it properly.

teneightfive: Also, the check for production might not work fully as expected as if we run in Heroku, it will also expect the `REDIS_URL` as its set in the environment variables so it won't deploy there.

stevehook: @teneightfive I've rolled the helper method into `index.js`. I've exported it so that it can be tested - not sure whether I like that so let me know if you would rather not test that method at all (or do it a better way).

Also changed the check for cloud platform to the presence of `REDIS_AUTH_TOKEN` as you suggested.

teneightfive:
@stevehook nice work. Yeah not sure about exporting it on the config object just to test. It feels a little strange.

Also, I think a neater way of testing it rather than having to set `process.env` variables in the test would be to pass in the values that the method expects as arguments. So it could take two arguments of `redisUrl` and `redisAuthToken` to make it more functional.

I think it works as is though and getting it will see if it'll deploy on Cloud platform. 

Also, I think some of this could be re-worked in #119 that I've created so can look at it again after.

PR #120, opened 2019-07-05
Add sign out link in header

slorek:
The order of events is:

* We destroy the session on our service
* We redirect the user to the HMPPS SSO logout URL, which logs them out of that session as well
* HMPPS SSO redirects back to our service at the URL we specify, or failing that, back to its own login screen

### DIFF
--- a/app/auth/controllers.test.js
+++ b/app/auth/controllers.test.js
@@ -1,4 +1,18 @@
-const controllers = require('./controllers')
+const proxyquire = require('proxyquire').noCallThru()
+
+const logOutUrl = 'http://test/test'
+const mockConfig = {
+  AUTH_PROVIDERS: {
+    test: {
+      logout_url: logOutUrl,
+    },
+  },
+  DEFAULT_AUTH_PROVIDER: 'test',
+}
+
+const controllers = proxyquire('./controllers', {
+  '../../config': mockConfig,
+})
 
 const url = '/test'
 
@@ -40,6 +54,32 @@ describe('Authentication app', function () {
       it('unsets session.postAuthRedirect', function () {
         expect(req.session.postAuthRedirect).to.equal(null)
       })
+    })
+  })
+
+  describe('#signOut()', function () {
+    let req, res
+
+    beforeEach(function () {
+      req = {
+        query: {},
+        session: {
+          destroy: sinon.spy(),
+        },
+      }
+      res = {
+        redirect: sinon.spy(),
+      }
+
+      controllers.signOut(req, res)
+    })
+
+    it('destroys the session', function () {
+      expect(req.session.destroy).to.be.called
+    })
+
+    it('redirects to the auth provider logout URL', function () {
+      expect(res.redirect).to.be.calledWith(logOutUrl)
     })
   })
 })
--- a/app/auth/controllers.js
+++ b/app/auth/controllers.js
@@ -1,3 +1,5 @@
+const { AUTH_PROVIDERS, DEFAULT_AUTH_PROVIDER } = require('../../config')
+
 module.exports = {
   redirect: (req, res) => {
     const url = req.session.postAuthRedirect || '/'
@@ -5,5 +7,11 @@ module.exports = {
     req.session.postAuthRedirect = null
 
     res.redirect(url)
+  },
+  signOut: (req, res) => {
+    const authProviderLogoutUrl = AUTH_PROVIDERS[DEFAULT_AUTH_PROVIDER].logout_url
+
+    req.session.destroy()
+    res.redirect(authProviderLogoutUrl)
   },
 }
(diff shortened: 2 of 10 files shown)


app/sign-out/index.js
```diff
@@ -0,0 +1,14 @@
...
+// Local dependencies
+const { get } = require('./controllers')
+
+// Define routes
+router.get('/', get)
+
+// Export
+module.exports = {
+  router,
+  mountpath: '/sign-out',
```
teneightfive: I think we should add this as a route to the `auth` feature rather than a separate feature. It feels linked to that.

app/sign-out/controllers.js
```diff
@@ -0,0 +1,21 @@
...
+
+function buildRedirectUrl () {
+  const protocol = IS_PRODUCTION ? 'https' : 'http'
+  return `${protocol}://${SERVER_HOST}`
+}
+
+function buildLogoutUrl () {
+  const postLogoutRedirectUrl = buildRedirectUrl()
+  const path = `/auth/logout?client_id=${AUTH.PROVIDER_KEY}&redirect_uri=${postLogoutRedirectUrl}`
+  return new URL(path, AUTH.PROVIDER_URL).href
```
teneightfive: I’ve found this can blow up if this env isn’t set, so good to protect with a ternary like this:

```js
return AUTH.PROVIDER_URL ? new URL(path, AUTH.PROVIDER_URL).href : ‘’
```

app/sign-out/conhtrollers.test.js
```diff
@@ -0,0 +1,31 @@
+const controllers = require('./controllers')
+const config = require('../../config')
+
+describe('Sign out', function () {
+  describe('#get()', function () {
+    let req, res, redirectUrl
+
+    beforeEach(async function () {
```
teneightfive: Don’t need `async` here as the function isn’t doing anything asynchronous.

app/sign-out/conhtrollers.test.js
```diff
@@ -0,0 +1,31 @@
...
+        query: {},
+        session: {
+          destroy: sinon.spy(),
+        },
+      }
+      res = {
+        redirect: sinon.spy(),
+      }
+
+      await controllers.get(req, res)
```
teneightfive: Don’t need `await` here either.

app/sign-out/index.js
slorek: That was my preference too, but the middleware intercepts it first. The only alternative was to move the middleware inline.

teneightfive: I’ve done some refactoring to when and how that middleware actually loads in #115 which should solve that. Maybe I’ll get that in then you can make the changes on top of that. 

You’ll just then need to add the sign out path to the auth whitelist.

app/sign-out/controllers.js
```diff
@@ -0,0 +1,21 @@
+const { IS_PRODUCTION, SERVER_HOST, AUTH } = require('../../config')
+
+function buildRedirectUrl () {
+  const protocol = IS_PRODUCTION ? 'https' : 'http'
+  return `${protocol}://${SERVER_HOST}`
```
teneightfive: I think we may need a more generic way to figure out the protocol too. I’ve used some middleware in other apps to do this so it can be shared by things like building canonical URLs or asset paths.

app/sign-out/controllers.js
teneightfive: Not #115 is also in I’d add the logout path to the hmpps object in the config.

Then you can append the redirect in the controller, where you’ll also have access to the full URL of the app I think so you won’t have to use the `SERVER_HOST` variable.

app/sign-out/controllers.js
slorek: I'm surprised Express doesn't provide this already, but I think it's outside the scope of this feature.

app/sign-out/controllers.js
slorek: 👍

app/sign-out/index.js
slorek: Ok great.

app/sign-out/controllers.js
teneightfive: It does have this: https://expressjs.com/en/api.html#req.protocol

But there are some issues with behind a proxy.

app/auth/controllers.test.js
```diff
@@ -1,4 +1,5 @@
 const controllers = require('./controllers')
+const { AUTH_PROVIDERS, DEFAULT_AUTH_PROVIDER } = require('../../config')
```
teneightfive: As we spoke about yesterday, might be better to mock these values using proxyquire rather than relying on the actual values set within the config module.

PR #121, opened 2019-07-08
[P4-475] Update button text for final create step

teneightfive:
This change allows custom text to be passed to the form wizard
on any step by adding a `buttonText` property to the step
options.

It also includes updating the final step of the creation journey
to change the text to reflect the action that will happen after
clicking it.

### DIFF
--- a/app/moves/steps.js
+++ b/app/moves/steps.js
@@ -74,6 +74,7 @@ module.exports = {
     controller: Assessment,
     next: 'save',
     heading: 'moves:steps.health_information.heading',
+    buttonText: 'actions:schedule_move',
     fields: [
       'health',
       'health__special_diet_or_allergy',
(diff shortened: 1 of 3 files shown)


app/moves/steps.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
     controller: Assessment,
     next: 'save',
     heading: 'moves:steps.health_information.heading',
+    buttonText: 'actions:schedule_move',
```
slorek: It's a little weird passing a key for the view to translate rather than the actual text.

teneightfive: This file doesn't have access to the request object which contains the translation method though.

It's the way the [form wizard](https://github.com/UKHomeOffice/passports-form-wizard) actually handled it too. It's possibly something we could revisit later?

slorek: This was an issue I encountered on #120 as well.`t` should be available outside of the views. I'm happy to leave it for now if we can add this to the issues list.

PR #130, opened 2019-07-15
[P4-439] Validate creation journey

teneightfive:
This adds validation to the form wizard to validate each of the
move creation steps.

### Personal details

localhost_3000_moves_new_personal-details

localhost_3000_moves_new_personal-details (1)

### Move details

localhost_3000_moves_new_move-details

localhost_3000_moves_new_move-details (1)

localhost_3000_moves_new_move-details (2)

localhost_3000_moves_new_move-details (3)

### DIFF
--- a/app/moves/fields.js
+++ b/app/moves/fields.js
@@ -1,23 +1,20 @@
 const { date } = require('./formatters')
 
-function assessmentQuestionComments ({ required = false } = {}) {
-  const labelPath = required ? 'required' : 'optional'
-
-  return {
-    skip: true,
-    component: 'govukTextarea',
-    classes: 'govuk-input--width-20',
-    rows: 3,
-    label: {
-      text: `fields:assessment_comment.${labelPath}`,
-      classes: 'govuk-label--s',
-    },
-  }
+const assessmentQuestionComments = {
+  skip: true,
+  rows: 3,
+  component: 'govukTextarea',
+  classes: 'govuk-input--width-20',
+  label: {
+    text: `fields:assessment_comment.optional`,
+    classes: 'govuk-label--s',
+  },
 }
 
 module.exports = {
   // personal details
   athena_reference: {
+    validate: 'required',
     component: 'govukInput',
     label: {
       text: 'fields:athena_reference.label',
@@ -32,6 +29,7 @@ module.exports = {
     autocomplete: 'off',
   },
   first_names: {
+    validate: 'required',
     component: 'govukInput',
     label: {
       text: 'fields:first_names.label',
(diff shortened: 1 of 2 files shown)


slorek:
What happens if the API throws a validation error which isn't implemented here? Will the error display in the same way?

teneightfive:
> What happens if the API throws a validation error which isn't implemented here? Will the error display in the same way?

At the moment they’ll see the `Page unavailable` error page and it will throw a `422` error which will be logged in Sentry for us to investigate. 

I think long term it would be good to move completely to the server but I think this would be easier having separate validation endpoints. And then we can also extend the reference data to be conditional based on certain previous answer. For example, we don’t need to display the pregnancy question if gender is male.

slorek:
Can we have a stop gap so at least the page renders again (maybe listing the errors) in the event of a 422 error? I think bombing out is potentially going to cause us a bunch of support issues.

teneightfive:
> Can we have a stop gap so at least the page renders again (maybe listing the errors) in the event of a 422 error? I think bombing out is potentially going to cause us a bunch of support issues.

This is the issue with doing full API validation at the moment. It’s a little trickier to go back and list the errors in a good way. The best way would be to have a particular endpoint that either returns errors or true so that we can incorporate that into the validation point of the lifecycle. At the moment it’s right at the end and is also trying to save it with the same information.

I think for MVP this is fine, we’ll spot that immediately and can fix them quickly. We can also double check the validation on the API but I’m pretty confident we won’t see any `422`s with the validation configuration in this PR.

slorek:
I think showing the page again (even if the errors are just in a block at the top and not individually highlighted) would be fairly trivial to do but will be better for the user as an interim until the front end is properly integrated.

teneightfive:
My concern is wasting time on this now if we're going to implement full validation on the API side using validation endpoints.

But even in that case we will still need some level of validation on the client app for conditional fields like selected the to location type first.

If we've set the validation correctly there should also be no case where a `422` should be thrown or we've implemented it incorrectly, in which case I think throwing an error to Sentry for us to investigate feels like an okay solution in the interim.

teneightfive:
@slorek as discussed, have created [a ticket](https://dsdmoj.atlassian.net/browse/P4-215?atlOrigin=eyJpIjoiNmQ2NWE2YmNkMDVkNDQ4NjlkNDU0YTA2NTIzNzBjZWEiLCJwIjoiaiJ9) to track making a change to properly handle `422` errors from the API.

Not the ideal way around to do this but this change should catch the expected validation errors we know about at this stage.

PR #134, opened 2019-07-16
[P4-470] Capture errors in sentry

teneightfive:
This set of changes ensure exceptions are logged in Sentry.

It also includes a change to the form processing to ensure we are capturing any validation errors that we aren't aware of.

### DIFF
--- a/app/moves/controllers/form.js
+++ b/app/moves/controllers/form.js
@@ -1,5 +1,6 @@
 const { map, fromPairs } = require('lodash')
 const { Controller } = require('hmpo-form-wizard')
+const Sentry = require('@sentry/node')
 
 const fieldHelpers = require('../../../common/helpers/field')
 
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -96,6 +96,7 @@ jobs:
               --build-arg APP_BUILD_DATE=${BUILD_DATE} \
               --build-arg APP_BUILD_TAG=${CIRCLE_BRANCH} \
               --build-arg APP_GIT_COMMIT=${CIRCLE_SHA1} \
+              --build-arg APP_SHORT_GIT_HASH=$(echo $CIRCLE_SHA1 | cut -c -7) \
```
slorek: Just wondering if we need both or if the short version can be used for both? Also the naming is inconsistent.

teneightfive: Yeah, this was slightly annoying if I'm honest...I wanted to use the full but Sentry rejected the release because it exceeded the character length!

It also suggests prefixing the release as they're not unique per project:
>Note that releases are global per organization so make sure to prefix them with something project specific if needed

But maybe the long git hash is fine for us?

PR #138, opened 2019-07-17
Include the created date in assessment answers

teneightfive:
This change updates the assessment answers to use the new `created_at`
property the API now supports.

### Before
image

### After
image

### DIFF
--- a/common/lib/api-client/models.js
+++ b/common/lib/api-client/models.js
@@ -52,6 +52,8 @@ function defineModels (jsonApi) {
   })
 
   jsonApi.define('assessment_question', {
+    created_at: '',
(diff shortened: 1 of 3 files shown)


common/lib/api-client/models.js
```diff
@@ -52,6 +52,8 @@ function defineModels (jsonApi) {
   })
 
   jsonApi.define('assessment_question', {
+    created_at: '',
+    updated_at: '',
```
stevehook: We don't have `updated_at` (yet) - we don't track changes, we just overwrite what's there. We do have `expires_at` (which is empty by default).

PR #139, opened 2019-07-17
Fix auth return URL

teneightfive:
Currently with the order in which the files are loaded the
auth logout return URL isn't being populated.

The url helper was being loaded before the config had populated any
values that were loaded into that module and set as defaults.

This change simplifies that to build the base url in the config
file which can be sent to the auth logout as a return URL.


config/index.js
```diff
@@ -1,10 +1,10 @@
 /* eslint no-process-env: "off" */
 require('dotenv').config()
 
-const { buildUrl } = require('../common/helpers/url')
-
 const IS_DEV = process.env.NODE_ENV !== 'production'
 const IS_PRODUCTION = process.env.NODE_ENV === 'production'
+const SERVER_HOST = process.env.SERVER_HOST
+const BASE_URL = `${IS_PRODUCTION ? 'https' : 'http'}://${SERVER_HOST}`
```
stevehook: Could you use a property on the request to do this switch? e.g. https://stackoverflow.com/questions/10348906/how-to-know-if-a-request-is-http-or-https-in-node-js suggests `req.connection.encrypted`

teneightfive: Sooo, yes that would be the ideal but at this point the config has no idea about the request.

I've added [an issue]() to see if we can move the setting of the return URL to the controller that makes the request.

PR #143, opened 2019-07-18
Display a different list template based on permissions

teneightfive:
This change accomdates the supplier role where they cannot view
individual moves. If that permission is missing it renders an
alternative template without the list of moves.

### Supplier user dashboard

localhost_3000_moves (1)

### Police user dashboard

localhost_3000_moves_c249ed09-0cd5-4f52-8aee-0506e2dc7579

### DIFF
--- a/app/moves/controllers/list.js
+++ b/app/moves/controllers/list.js
@@ -1,12 +1,19 @@
 const { format, addDays, subDays } = require('date-fns')
+const { get } = require('lodash')
 
 const { getQueryString } = require('../../../common/lib/request')
+const permissions = require('../../../common/middleware/permissions')
 const presenters = require('../../../common/presenters')
 
 module.exports = function list (req, res) {
   const { moveDate, movesByDate } = res.locals
   const yesterday = format(subDays(moveDate, 1), 'YYYY-MM-DD')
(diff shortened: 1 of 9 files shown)


teneightfive:
I'm not 100% convinced on the logic yet but be good to get a second opinion. I'm trying to not do lots of the logic in the template, but instead, switch which template a user sees based on what they can do.

Ultimately the goal is that suppliers have a different view on the dashboard.

app/moves/controllers/list.js
```diff
@@ -1,12 +1,19 @@
...
 
 module.exports = function list (req, res) {
   const { moveDate, movesByDate } = res.locals
   const yesterday = format(subDays(moveDate, 1), 'YYYY-MM-DD')
   const tomorrow = format(addDays(moveDate, 1), 'YYYY-MM-DD')
+  const canViewMove = permissions.check(
+    'move:view',
+    get(req.session, 'user.permissions')
+  )
+  const template = !canViewMove ? 'moves/views/download' : 'moves/views/list'
```
slorek: Nitpick: prefer truthy ternary e.g. 
```
canViewMove ? 'moves/views/list' : 'moves/views/download'
```

PR #145, opened 2019-07-19
[P4-353] Add locations to the user object

teneightfive:
This change extends the user class to set the locations available
to that user on the user session object.

This PR is a precursor to some future changes:
- ensure user can access location being viewed
- if user only has access to only one location, default to that location
- if user has access to multiple locations, allow them to choose

### DIFF
--- a/app/auth/middleware.js
+++ b/app/auth/middleware.js
@@ -1,30 +1,41 @@
 const User = require('../../common/lib/user')
+const referenceData = require('../../common/services/reference-data')
 
 function _decodeAccessToken (token) {
   const payload = token.split('.')[1]
   return JSON.parse(Buffer.from(payload, 'base64').toString('utf8'))
 }
 
-function processAuthResponse (req, res, next) {
-  const { grant, postAuthRedirect } = req.session
+function processAuthResponse (defaultLocations = []) {
+  return async function middleware (req, res, next) {
+    const { grant, postAuthRedirect } = req.session
 
-  if (!grant) {
-    return next()
-  }
-
-  req.session.regenerate(error => {
-    if (error) {
-      return next(error)
+    if (!grant) {
+      return next()
     }
 
-    const accessToken = _decodeAccessToken(grant.response.access_token)
+    try {
+      const accessToken = _decodeAccessToken(grant.response.access_token)
+      const locations = await referenceData.getLocationsById(defaultLocations)
 
-    req.session.authExpiry = accessToken.exp
-    req.session.postAuthRedirect = postAuthRedirect
-    req.session.user = new User(accessToken)
+      req.session.regenerate(error => {
+        if (error) {
+          return next(error)
+        }
 
-    next()
-  })
+        req.session.authExpiry = accessToken.exp
+        req.session.postAuthRedirect = postAuthRedirect
+        req.session.user = new User({
+          ...accessToken,
+          locations,
+        })
+
+        next()
+      })
+    } catch (error) {
+      next(error)
+    }
+  }
 }
 
 module.exports = {
--- a/common/lib/user.js
+++ b/common/lib/user.js
@@ -1,25 +1,31 @@
-function User (token) {
-  const permissions = []
-  const authorities = token.authorities || []
-
-  if (authorities.includes('ROLE_PECS_POLICE')) {
-    permissions.push(
-      ...[
-        'moves:view:by_location',
-        'moves:download:by_location',
-        'move:view',
-        'move:create',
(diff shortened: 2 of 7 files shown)


app/auth/middleware.js
```diff
@@ -5,26 +5,33 @@ function _decodeAccessToken (token) {
...
+  try {
     const accessToken = _decodeAccessToken(grant.response.access_token)
+    const user = new User()
+
+    await user.init(accessToken)
 
-    req.session.authExpiry = accessToken.exp
-    req.session.postAuthRedirect = postAuthRedirect
-    req.session.user = new User(accessToken)
+    req.session.regenerate(async error => {
```
slorek: Does this require `async`?

common/lib/user.js
```diff
@@ -1,25 +1,63 @@
...
+
+    return permissions
+  },
+  getLocations (locations = []) {
+    // TODO: Remove once we can determine locations from auth
+    locations.push(CURRENT_LOCATION_UUID)
+
+    const locationPromises = locations.map(locationId => {
+      return referenceDataService
+        .getLocationById(locationId)
```
slorek: The value we receive from HMPPS SSO will be the NOMIS agency ID, not the location ID as known by the reference data service. We probably want a service method like `getLocationByNomisAgencyId` to be clear what it is.

Also, the values we get in the token will be prepended with `PECS_` so we'll need to strip these off the value beforehand.

common/lib/user.js
```diff
@@ -1,25 +1,63 @@
...
+const Sentry = require('@sentry/node')
+
+const referenceDataService = require('../../common/services/reference-data')
+// TODO: Remove once we can determine locations from auth
+const { CURRENT_LOCATION_UUID } = require('../../config')
+
+function User () {}
+
+User.prototype = {
+  init (token) {
```
slorek: I'm not sure instantiation with an access token is the best way to separate concerns. Also the structure of the access token content can differ quite a lot between providers. I'd prefer we could instantiate using the user data (name, roles, locations) as named arguments. This would probably improve the clarity of the tests as well.

app/auth/middleware.js
teneightfive: Good spot. That was from a previous location of this change.

common/lib/user.js
teneightfive: Yep, I thought that might be the case. Right now this just moves the location to the user object so that once we get it back we can replace it in one place.

It temporarily still uses the ID from config which we can remove once we get an actual list back from SSO.

The reason for moving this here was so that the user object has some concept of allowed locations so that we can check that they can see the location they might be viewing in the moves feature.

teneightfive: tl;dr I think that will be a future change

common/lib/user.js
teneightfive: I think I see what you're saying.

So should the User class be simpler in that its not concerned with retrieving the full location object from the API (whether that be by ID like now, or by agency code in the future)?

The main purpose of this change is so that our user object has the concept of `locations` that it can access. We can then check parts of the app like viewing moves for a specific location or changing the current default location against this list.

I didn't fully like that this User class became async to be honest, made it a bit trickier to test.

Maybe we should use the reference data service within the auth middleware then and then pass that list of locations in the user object?

slorek: I agree.

common/lib/user.js
slorek: I think given we're introducing the `getLocations` method in this PR we should make sure it's not a knowingly incorrect implementation... we can set the environment variable to `PECS_GDI` now and get that working.

teneightfive: The issue with that is that it also involves an API change so will hold up the other changes that are part of this PR, like being able to associate locations to a user.

teneightfive: We also don't know the exact implementation of how it will come back as part of the auth yet so might be better to leave that change once we have an example implementation of it.

slorek: This is already implemented on the SSO service.

teneightfive: As discussed on slack I think we need to leave the implementation of SSO `agencyId` support for a future change once we also have support in the API to get a location by agencyId.

This is holding up some other changes that we need in soon and the main basis for this is to move the locations to the user object so that other parts of the app can use this.

teneightfive: Backend ticket: https://dsdmoj.atlassian.net/browse/P4-511
Frontend ticket: https://dsdmoj.atlassian.net/browse/P4-512

PR #146, opened 2019-07-19
[P4-353] Set current location

teneightfive:
Since the locations have been [moved to the user object](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/145) we can
now set the current location based on that object.

For now, we are only supporting one location so if there are more
than one it will pass through this middleware.

In a future change we will redirect to a locations feature where
a user can choose which location to view when they first sign in
and this value is not set.


common/middleware/set-current-location.js
```diff
@@ -0,0 +1,16 @@
+const { get } = require('lodash')
+
+module.exports = function currentLocation (req, res, next) {
+  if (req.session.currentLocation) {
+    return next()
+  }
+
+  const userLocations = get(req.session, 'user.locations', [])
+
+  if (userLocations.length === 1) {
```
slorek: Should we just default to the first one in this case? Otherwise the service will be broken until the user can choose.

teneightfive: Not a bad shout

PR #152, opened 2019-07-22
[P4-273] Add "today" shortcut to pagination

teneightfive:
This PR extends the pagination component to support list items in the middle and then adds the today shortcut to the moves listing template.

### DIFF
--- a/app/moves/views/_layout.njk
+++ b/app/moves/views/_layout.njk
@@ -28,7 +28,11 @@
           next: {
             href: pagination.nextUrl,
             text: t("actions:next_day")
-          }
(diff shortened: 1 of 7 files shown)


app/moves/views/_layout.njk
```diff
@@ -28,7 +28,11 @@
           next: {
             href: pagination.nextUrl,
             text: t("actions:next_day")
-          }
+          },
+          items: [{
+            href: pagination.todayUrl,
+            text: "Today"
```
stevehook: Should this string be localised?

PR #165, opened 2019-07-24
[P4-536] Update namespace separator for i18n

teneightfive:
The default namespace separator for the i18n library is currently
`:`.

This has raised an issue with some strings that don't have translations
including a single colon and therefore looking for a particular
namespace first.

When this is not found it removes that first part of the string
and doesn't fallback to the full string.

This caused an issue with NOMIS imported ethnicities which included
a single colon.

This change replaces the single colon with a double with the
assumption that this will less likely occur in fallback strings.

**Note:** open to other suggestions regarding alternative namespace
separator


app/move/steps.js
```diff
@@ -73,7 +73,7 @@ module.exports = {
   '/health-information': {
     controller: Assessment,
     next: 'save',
-    heading: 'moves:steps.health_information.heading',
+    heading: 'moves::steps.health_information.heading',
     buttonText: 'actions:schedule_move',
```
teneightfive: Missed one

PR #168, opened 2019-07-29
Add page title to form wizard

teneightfive:
This change sets the page title for form wizard steps. At the moment
the form wizard falls back to the default page title.

This change adds a new property to the wizard settings to use as
the page title. It also renames the `heading` property for each step
to be more consistent.

## Before
image

## After
image

### DIFF
--- a/app/move/index.js
+++ b/app/move/index.js
@@ -13,6 +13,7 @@ const wizardConfig = {
   controller: Form,
   name: 'new-move',
   journeyName: 'new-move',
+  journeyPageTitle: 'actions::create_move',
   template: 'form-wizard',
 }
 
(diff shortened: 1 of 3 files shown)


app/move/index.js
```diff
@@ -13,6 +13,7 @@ const wizardConfig = {
   controller: Form,
   name: 'new-move',
   journeyName: 'new-move',
+  journeyPageTitle: 'Create a move',
```
stevehook: not sure how it works but it _looks_ like this needs to be localisable?

teneightfive: Good spot. Glad you keep remembering this :)

PR #172, opened 2019-08-01
Create Pull Request template

teneightfive:
Since the recent analytics changes it's worth us creating a template for new Pull Requests to make sure future changes don't affect anything. 

This file can be expanded upon as we think of things that should be considered with new changes.


teneightfive:
What the f! That’s very odd. Haha! I’ll update that!

PR #173, opened 2019-08-01
[P4-281] API timeouts

teneightfive:
This change adds middleware to inject a timeout into the axios library
used to make API requests.


config/index.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
     AUTH_URL: process.env.API_AUTH_URL,
     CLIENT_ID: process.env.API_CLIENT_ID,
     SECRET: process.env.API_SECRET,
+    TIMEOUT: 30000, // in seconds
```
slorek: 30000 seconds is 500 minutes. I think we can safely make this a bit shorter ;)

teneightfive: Doh!

teneightfive: @slorek sorry, it was the correct value, but wrong comment! It's actually milliseconds, not seconds!

PR #177, opened 2019-08-05
[P4-317] Store listing state

teneightfive:
Currently the link back to the dashboard will go back to a non-filtered view.

If a user changes the date and navigates to a move the current date will be lost and they will return to the default dashboard view for "today" on clicking the back to dashboard link.

This change will persist the current filter so that it can be used to go back to the correctly filtered version of the dashboard.

### DIFF
--- a/app/move/views/view.njk
+++ b/app/move/views/view.njk
@@ -16,7 +16,7 @@
   {{ govukBackLink({
     text: t("actions::back_to_dashboard"),
     classes: "app-print--hide",
-    href: "/moves"
+    href: MOVES_URL
   }) }}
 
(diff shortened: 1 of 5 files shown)


app/move/views/view.njk
```diff
@@ -16,7 +16,7 @@
   {{ govukBackLink({
     text: t("actions::back_to_dashboard"),
     classes: "app-print--hide",
-    href: "/moves"
+    href: "/moves" + MOVES_QUERY_STRING
```
slorek: You might consider allowing the middleware to specify the whole path including string. In general I would recommend avoiding paths being hard coded or constructed in the view.

PR #178, opened 2019-08-05
[P4-503] Create cancel option

teneightfive:
This adds a cancel button that can be used by the user to cancel the current create journey.

## What it looks like

localhost_3000_move_new_personal-details

### DIFF
--- a/app/move/controllers/new.form.js
+++ b/app/move/controllers/new.form.js
@@ -10,6 +10,16 @@ class FormController extends Controller {
     this.use(this.checkCurrentLocation)
   }
 
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setCancelUrl)
+  }
+
+  setCancelUrl(req, res, next) {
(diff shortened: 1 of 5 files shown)


app/move/controllers/new.form.js
```diff
@@ -10,6 +10,16 @@ class FormController extends Controller {
     this.use(this.checkCurrentLocation)
   }
 
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setCancelUrl)
+  }
+
+  setCancelUrl(req, res, next) {
+    res.locals.cancelUrl = `/moves${res.locals.MOVES_QUERY_STRING || ''}`
```
slorek: See my [previous comment](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/177#discussion_r310991322) I think setting the full path in the middleware would avoid duplication here.

teneightfive: Will address once the other PR has been merged.

PR #181, opened 2019-08-07
[P4-537] Extra gender information

teneightfive:
This adds a new field to capture any extra gender information necessary
to transport a detainee/prisoner.

### Request journey
image

#### With additional information
image

#### Without additional information
image

### Checklist


stevehook:
By the way we've not deployed the schema change to production on the backend yet - we will need to make sure we get things like this deployed in the right order going forward!

teneightfive:
> By the way we've not deployed the schema change to production on the backend yet - we will need to make sure we get things like this deployed in the right order going forward!

Yes, this is a good point. I think this change wouldn't be a breaking change if either was deployed before the other. I think the backend just ignores extra properties right? 

However, it would cause confusion if a user enters this information and it doesn't get stored.

Issue #82, opened 2019-06-21
Default value for `API.BASE_URL` config is required to run tests


teneightfive:
Whilst trying to update the config items and [removing the default value](#81) for the API base URL it threw up a lot of issues with the tests and possible with the way the services are configured.

Currently it uses that value imported via the config module to use as the URL to mock with nock.

If that value is removed nock will throw an error that an invalid URL has been passed to the class.

To properly solve this we would need to be able to stub the API config object within the service tests but that module is not directly imported, it is imported in the API client module.

This issue was discovered when making updates in #81

PR #182, opened 2019-08-07
Allow reference data service to search for locations by NOMIS agency ID

slorek:
This allows the reference data service to look up locations by both type as currently, and also NOMIS agency ID, as implemented in https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/95.

This is a pre-requisite to look up locations by agency ID, as supplied by HMPPS SSO.

### DIFF
--- a/common/services/reference-data.js
+++ b/common/services/reference-data.js
@@ -18,12 +18,13 @@ function getAssessmentQuestions(category) {
     .then(response => response.data)
 }
 
-function getLocations(type, combinedData, page = 1) {
+function getLocations({ type, nomisAgencyId, combinedData, page = 1 } = {}) {
   return apiClient
     .findAll('location', {
       page,
       per_page: 100,
       'filter[location_type]': type,
+      'filter[nomis_agency_id]': nomisAgencyId,
     })
     .then(response => {
       const { data, links } = response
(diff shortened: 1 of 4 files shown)


common/services/reference-data.js
```diff
@@ -35,7 +36,12 @@ function getLocations(type, combinedData, page = 1) {
         return sortBy(locations, 'title')
       }
 
-      return getLocations(type, locations, page + 1)
+      return getLocations({
+        type: type,
```
teneightfive: Top two lines can be written using ES6 shorthand:

```js
return getLocations({
  type,
  nomisAgencyId,
  combinedData: locations,
  page: page + 1,
})
```

common/services/reference-data.test.js
```diff
@@ -106,30 +106,102 @@ describe('Reference Service', function() {
...
+        it('should call API with the type filter query parameter', function() {
+          expect(nock.isDone()).to.be.true
+        })
       })
+    })
 
-      it('should return a full list of locations', function() {
-        expect(response.length).to.deep.equal(locationsDeserialized.data.length)
-        expect(response).to.deep.equal(locationsDeserialized.data)
+    context('with nomisAgencyId parameter', function() {
```
teneightfive: I've been thinking about how we can write these tests a bit better and I think this highlights we may need a refactor.

I think we should test the serialization/deserialization in the API client and in these services we should move the apiClient response rather than using nock.

But one for later on anyway.

slorek: Yes there’s definitely scope for de-duplication in these tests, but it’s one for another day.

teneightfive: Yeah, also hoping that refactor could fix #82 too

slorek:
Merging since the unit_test CircleCI failure is related to the uploading of test coverage data to 'coverall', otherwise it's working fine.

PR #186, opened 2019-08-12
Locations lookup

slorek:
This allows the frontend to look up the user's locations either on HMPPS SSO or NOMIS depending on the type of user, and resolves those to the IDs represented in the API. This will allow us to implement location switching.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,7 @@ AUTH_PROVIDER_KEY=
 AUTH_PROVIDER_SECRET=
 AUTH_PROVIDER_URL=
 SERVER_HOST=localhost:3000
+NOMIS_ELITE2_API_URL=https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api
 
 ## Development only
 # BYPASS_SSO=true
--- a/app/auth/middleware.js
+++ b/app/auth/middleware.js
@@ -1,12 +1,8 @@
 const User = require('../../common/lib/user')
-const referenceData = require('../../common/services/reference-data')
+const userLocationService = require('../../common/services/user-locations')
+const { decodeAccessToken } = require('../../common/lib/access-token')
 
-function _decodeAccessToken(token) {
-  const payload = token.split('.')[1]
-  return JSON.parse(Buffer.from(payload, 'base64').toString('utf8'))
-}
-
-function processAuthResponse(defaultLocations = []) {
+function processAuthResponse() {
   return async function middleware(req, res, next) {
     const { grant, postAuthRedirect } = req.session
 
@@ -15,19 +11,23 @@ function processAuthResponse(defaultLocations = []) {
     }
 
     try {
-      const accessToken = _decodeAccessToken(grant.response.access_token)
-      const locations = await referenceData.getLocationsById(defaultLocations)
+      const userLocations = await userLocationService.getUserLocations(
+        grant.response.access_token
+      )
 
       req.session.regenerate(error => {
         if (error) {
           return next(error)
         }
 
+        const accessToken = decodeAccessToken(grant.response.access_token)
+
         req.session.authExpiry = accessToken.exp
         req.session.postAuthRedirect = postAuthRedirect
         req.session.user = new User({
-          ...accessToken,
-          locations,
+          name: accessToken.user_name,
+          roles: accessToken.authorities,
+          locations: userLocations,
         })
 
         next()
--- a/app/auth/middleware.test.js
+++ b/app/auth/middleware.test.js
@@ -1,14 +1,22 @@
 const proxyquire = require('proxyquire')
 
-const referenceDataService = require('../../common/services/reference-data')
-
-function UserStub(token) {
-  this.user_name = token.user_name
+const userLocationSuccessStub = {
+  getUserLocations: () => {
+    return Promise.resolve(['TEST'])
+  },
+}
+const userLocationFailureError = new Error('test')
+const userLocationFailureStub = {
+  getUserLocations: () => {
+    return Promise.reject(userLocationFailureError)
+  },
 }
 
-const authentication = proxyquire('./middleware', {
-  '../../common/lib/user': UserStub,
-})
+function UserStub({ name, roles = [], locations = [] } = {}) {
+  this.userName = name
+  this.permissions = []
+  this.locations = locations
+}
 
 const expiryTime = 1000
 const payload = {
@@ -18,17 +26,6 @@ const payload = {
   exp: expiryTime,
 }
 const encodedPayload = Buffer.from(JSON.stringify(payload)).toString('base64')
-const mockLocations = [
-  {
-    id: '1111',
-  },
-  {
-    id: '2222',
-  },
-  {
-    id: '3333',
-  },
-]
 
 describe('Authentication middleware', function() {
   describe('#processAuthResponse', function() {
@@ -46,8 +43,13 @@ describe('Authentication middleware', function() {
     })
 
     context('when there is no grant response', function() {
-      beforeEach(function() {
-        authentication.processAuthResponse()(req, {}, nextSpy)
+      beforeEach(async function() {
+        const authentication = proxyquire('./middleware', {
+          '../../common/lib/user': UserStub,
+          '../../common/services/user-locations': userLocationSuccessStub,
+        })
+
+        await authentication.processAuthResponse()(req, {}, nextSpy)
       })
 
       it('returns next', function() {
@@ -61,7 +63,6 @@ describe('Authentication middleware', function() {
 
     context('when there is a grant response', function() {
       beforeEach(function() {
-        sinon.stub(referenceDataService, 'getLocationsById')
         req.session.grant = {
           response: {
             access_token: `test.${encodedPayload}.test`,
@@ -69,29 +70,29 @@ describe('Authentication middleware', function() {
         }
       })
 
-      context('when getLocationsById rejects', function() {
-        const errorMock = new Error('Session Error')
-
+      context('when the user locations lookup fails', function() {
         beforeEach(async function() {
-          referenceDataService.getLocationsById.rejects(errorMock)
+          const authentication = proxyquire('./middleware', {
+            '../../common/lib/user': UserStub,
+            '../../common/services/user-locations': userLocationFailureStub,
+          })
 
           await authentication.processAuthResponse()(req, {}, nextSpy)
         })
 
-        it('call next with error', function() {
-          expect(nextSpy).to.be.calledOnce
-          expect(nextSpy.args[0][0] instanceof Error).to.be.true
-          expect(nextSpy.args[0][0].message).to.equal('Session Error')
-        })
-
-        it('doesn’t regenerate the session', function() {
-          expect(req.session.regenerate).not.to.be.called
+        it('returns next', function() {
+          expect(nextSpy).to.be.calledOnceWithExactly(userLocationFailureError)
         })
       })
 
-      context('when getLocationsById resolves', function() {
+      context('when the user locations lookup succeeds', function() {
+        let authentication
+
         beforeEach(function() {
-          referenceDataService.getLocationsById.resolves(mockLocations)
+          authentication = proxyquire('./middleware', {
+            '../../common/lib/user': UserStub,
+            '../../common/services/user-locations': userLocationSuccessStub,
+          })
         })
 
         context('when session regeneration throws an error', function() {
@@ -107,37 +108,7 @@ describe('Authentication middleware', function() {
           })
         })
 
-        context('when default locations are set', function() {
-          const locations = ['1', '2', '3']
-
-          beforeEach(async function() {
-            await authentication.processAuthResponse(
-              locations
-            )(req, {}, nextSpy)
-          })
-
-          it('should call reference data with locations', function() {
-            expect(
-              referenceDataService.getLocationsById
-            ).to.be.calledWithExactly(locations)
-          })
-        })
-
-        context('when no default locations are set', function() {
-          beforeEach(async function() {
-            await authentication.processAuthResponse()(req, {}, nextSpy)
-          })
-
-          it('should call reference with empty array', function() {
-            expect(
-              referenceDataService.getLocationsById
-            ).to.be.calledWithExactly([])
-          })
-        })
-
         context('when session regenerates successfully', function() {
-          let user
-
           beforeEach(async function() {
             // Stub the express-session #regenerate function which takes a callback
             req.session.regenerate = callback => {
@@ -146,10 +117,6 @@ describe('Authentication middleware', function() {
               }
               callback()
             }
-
-            user = new UserStub({
-              user_name: payload.user_name,
-            })
 
             await authentication.processAuthResponse()(req, {}, nextSpy)
           })
@@ -163,7 +130,9 @@ describe('Authentication middleware', function() {
           })
 
           it('sets the user info on the session', function() {
-            expect(req.session.user).to.deep.equal(user)
+            expect(req.session.user.userName).to.equal('test')
+            expect(Array.isArray(req.session.user.permissions)).to.be.true
+            expect(req.session.user.locations).to.deep.equal(['TEST'])
           })
 
           it('sets the redirect URL in the session', function() {
--- a/common/lib/user.js
+++ b/common/lib/user.js
@@ -1,14 +1,14 @@
-function User(token = {}) {
-  this.userName = token.user_name
-  this.locations = token.locations || []
-  this.permissions = this.getPermissions(token.authorities)
+function User({ name, roles = [], locations = [] } = {}) {
+  this.userName = name
+  this.permissions = this.getPermissions(roles)
+  this.locations = locations
 }
 
 User.prototype = {
-  getPermissions(authorities = []) {
+  getPermissions(roles = []) {
     const permissions = []
 
-    if (authorities.includes('ROLE_PECS_POLICE')) {
+    if (roles.includes('ROLE_PECS_POLICE')) {
       permissions.push(
         ...[
           'moves:view:by_location',
@@ -20,7 +20,7 @@ User.prototype = {
       )
     }
 
-    if (authorities.includes('ROLE_PECS_SUPPLIER')) {
+    if (roles.includes('ROLE_PECS_SUPPLIER')) {
       permissions.push(...['moves:view:all', 'moves:download:all'])
     }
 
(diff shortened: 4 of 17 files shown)


app/auth/middleware.test.js
```diff
@@ -1,13 +1,13 @@
...
-  this.user_name = token.user_name
+const locationStub = {
+  getUserLocations: token => {
+    return ['TEST']
+  },
 }
 
 const authentication = proxyquire('./middleware', {
-  '../../common/lib/user': UserStub,
+  '../../common/services/user-locations': locationStub,
```
teneightfive: This looks like we're not longer stubbing the User class. Why has this been removed?

common/services/user-locations.js
```diff
@@ -0,0 +1,68 @@
...
+      break
+  }
+
+  const locations = await Promise.all(
+    agencies.map(async agencyId => {
+      return referenceDataService.getLocations({ nomisAgencyId: agencyId })
+    })
+  )
+
+  return locations.map(results => {
```
teneightfive: Could be shortened to:

```js
return locations.map(results => results[0].id)
```

common/services/user-locations.js
```diff
@@ -0,0 +1,68 @@
+const axios = require('axios')
+
+const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
+const { decodeAccessToken } = require('../lib/access-token')
+const referenceDataService = require('./reference-data')
+
+async function getUserLocations(token) {
+  var agencies
```
teneightfive: Better to favour `let` over `var`

common/services/user-locations.js
```diff
@@ -0,0 +1,68 @@
...
+async function _getAuthAgencies(token) {
+  const tokenData = decodeAccessToken(token)
+
+  const response = await axios.get(
+    AUTH_PROVIDERS.hmpps.groups_url(tokenData.user_name),
+    {},
+    _authHeaders(token)
+  )
+
+  return response.data.map(group => {
```
teneightfive: Could be shortened to:

```js
return response.data.map(group => group.groupCode.replace(/^PECS_/, ''))
```

common/services/user-locations.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+async function _getNomisAgencies(token) {
+  const response = await axios.get(
+    NOMIS_ELITE2_API.user_caseloads_url,
+    {},
+    _authHeaders(token)
+  )
+
+  return response.data.map(caseLoad => {
```
teneightfive: Could be shortened to:

```js
return response.data.map(caseLoad => caseLoad.caseLoadId)
```

or, if you use [lodash's map](https://lodash.com/docs/4.17.15#map) function it could be:

```js
return map(response.data, 'caseLoadId')
```

common/services/user-locations.js
```diff
@@ -0,0 +1,68 @@
...
+    {},
+    _authHeaders(token)
+  )
+
+  return response.data.map(group => {
+    return group.groupCode.replace(/^PECS_/, '')
+  })
+}
+
+async function _getNomisAgencies(token) {
```
teneightfive: I wonder if it might be more readable to make these return a promise rather than using async/await.

This method could then become:

```js
function _getNomisAgencies(token) {
  return axios
    .get(NOMIS_ELITE2_API.user_caseloads_url, {
      headers: _getAuthHeaders(token),
    })
    .then(caseLoad => caseLoad.caseLoadId)
}
```

And I think `_authHeaders` should return just the headers object, not the whole config object:

```js
function _getAuthHeaders(token) {
  return {
    Authorization: `Bearer ${token}`,
  }
}
```

teneightfive: `_getAuthAgencies` can also be structured in a similar way.

common/services/user-locations.js
```diff
@@ -0,0 +1,68 @@
+const axios = require('axios')
+
+const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
+const { decodeAccessToken } = require('../lib/access-token')
+const referenceDataService = require('./reference-data')
+
+async function getUserLocations(token) {
```
teneightfive: After looking at the other methods in this module I think this might be neater if its restructured to return a promise as well rather than use async/await.

common/services/user-locations.test.js
```diff
@@ -0,0 +1,94 @@
...
+  getLocations: async args => {
+    return [{ id: 'test' }, { id: 'test2' }]
+  },
+}
+
+const userLocationsService = proxyquire('./user-locations', {
+  './reference-data': referenceDataStub,
+})
+
+const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
```
teneightfive: I think it's better to stub these using `proxyquire` rather than load them into the test file.

This is the problem we have in the other services at the moment and is causing #82. I think it's what is causing the tests to fail on CI too.

config/index.js
```diff
@@ -81,11 +85,19 @@ module.exports = {
       logout_url: _authUrl(
         `/auth/logout?client_id=${AUTH_KEY}&redirect_uri=${BASE_URL}`
       ),
+      groups_url: userName => {
+        return _authUrl(`/auth/api/authuser/${userName}/groups`)
+      },
+      base_url: AUTH_BASE_URL,
```
teneightfive: Do we need to add this? It looks like it's being used by the tests and pulled in direct to create the mock?

I think if we stub the url in the test using `proxyquire` it prevents us needing to add this to the config.

config/index.js
```diff
@@ -81,11 +85,19 @@ module.exports = {
...
       key: AUTH_KEY,
       secret: process.env.AUTH_PROVIDER_SECRET,
     },
   },
   DEFAULT_AUTH_PROVIDER: 'hmpps',
+  NOMIS_ELITE2_API: {
+    user_caseloads_url: _nomisElite2Url('/elite2api/api/users/me/caseLoads'),
+    base_url: NOMIS_ELITE2_API_URL,
```
teneightfive: Same as the other config `base_url` config.

app/auth/middleware.test.js
slorek: It's no longer necessary since the `User` class is now much simpler and has no external dependencies.

common/services/user-locations.js
slorek: Thanks, I forgot about `then`. I couldn't get the fancy `map` syntax above to work but it's certainly better now.

common/services/user-locations.js
slorek: 🤔 I've refactored this as you suggest and it's cleaned it up a bit, though it leaves a somewhat nasty nested `then` which I think makes it a bit less clear what the return value of the function is. Have a look and let me know what you think.

common/services/user-locations.test.js
slorek: 👍

config/index.js
slorek: Good shout.

app/auth/middleware.test.js
teneightfive: I'm a little uncomfortable removing this stub as if the User library changes in the future and gets more complex it will be being pulled into this test and could affect it.

I think it's good to stub services to what you expect in the tests rather than allow it to use all the dependencies.

common/services/user-locations.js
```diff
@@ -0,0 +1,64 @@
...
+      return response.data.map(group => group.groupCode.replace(/^PECS_/, ''))
+    })
+}
+
+function _getNomisAgencies(token) {
+  return axios
+    .get(NOMIS_ELITE2_API.user_caseloads_url, {
+      headers: _getAuthHeaders(token),
+    })
+    .then(response => {
```
teneightfive: Does using lodash map not work here?

Could this work:

```js
.then(response => map(response.data, 'caseLoadId')
```

or maybe even chaining thens to make it easier to read:

```js
.then(response => response.data)
.then(data => map(data, 'caseLoadId'))
```

common/services/user-locations.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function _getAuthAgencies(token) {
+  const tokenData = decodeAccessToken(token)
+
+  return axios
+    .get(AUTH_PROVIDERS.hmpps.groups_url(tokenData.user_name), {
+      headers: _getAuthHeaders(token),
+    })
+    .then(response => {
```
teneightfive: Could do same as other method:

```js
.then(response => response.data)
.then(data => data.map(group => group.groupCode.replace(/^PECS_/, '')))
```

common/services/user-locations.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function _getAgencies(token) {
+  let agenciesPromise
+
+  const tokenData = decodeAccessToken(token)
+
+  switch (tokenData.auth_source) {
+    case 'auth':
+      agenciesPromise = _getAuthAgencies(token)
```
teneightfive: Could just return within the switch to avoid needed to declare the promise:

```js
function _getAgencies(token) {
  const { auth_source: source } = decodeAccessToken(token)

  switch (source) {
    case 'auth':
      return _getAuthAgencies(token)
      break
    case 'nomis':
      return _getNomisAgencies(token)
      break
  }
}
```

Also could destructure the token.

common/services/user-locations.js
```diff
@@ -0,0 +1,64 @@
...
+
+const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
+const { decodeAccessToken } = require('../lib/access-token')
+const referenceDataService = require('./reference-data')
+
+function getUserLocations(token) {
+  return _getAgencies(token).then(agencies => {
+    return Promise.all(
+      agencies.map(agencyId =>
+        referenceDataService.getLocations({ nomisAgencyId: agencyId })
```
teneightfive: Just looking at this, wonder if this should be in the reference data service as `getLocationByAgencyCode` rather than part of the all locations method. It already contains `getLocationById`.

That method could then be responsible for returning the first result if there are multiple by that agency. Removes that responsibility from this library as it shouldn't care. Might make it easier to test too.

teneightfive: In fact, maybe create a `getLocationsByAgencyId` which mimics [`getLocationsById`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/services/reference-data.js#L52-L60) and call that from this library?

slorek: This was my original approach but after our in-person discussion I think we have a way forward.

common/services/user-locations.js
teneightfive: Also just thinking about fallback, maybe we should have a `default` case which returns a successful promise incase the source doesn't exist and it doesn't cause an unhandled promise error.

app/auth/middleware.js
```diff
@@ -15,20 +11,16 @@ function processAuthResponse(defaultLocations = []) {
...
           return next(error)
         }
 
-        req.session.authExpiry = accessToken.exp
-        req.session.postAuthRedirect = postAuthRedirect
-        req.session.user = new User({
-          ...accessToken,
-          locations,
-        })
+        _setSessionData(
```
teneightfive: I'd probably just scrap this. Is it just to beat code climate's warning of line length? If it's just that we can manually ignore that warning for this. I don't think splitting this out helps with re-use or readability to be honest.

If you were to refactor at all it would be neater to separate out the `session.regenerate` callback rather than have half the function inline and the other half separated.

common/lib/user.js
```diff
@@ -1,14 +1,14 @@
-function User(token = {}) {
-  this.userName = token.user_name
-  this.locations = token.locations || []
-  this.permissions = this.getPermissions(token.authorities)
+function User({ name, roles = [], locations = [] } = {}) {
```
teneightfive: This is much nicer 👍

common/services/reference-data.js
```diff
@@ -49,14 +49,23 @@ function getLocationById(id) {
...
-  )
+function getLocationsByAgencyId(ids = []) {
+  return _mapLocationIdsToLocations(ids, getLocationByAgencyId)
+}
+
+function _mapLocationIdsToLocations(ids, callback) {
+  const locationPromises = ids.map(id => {
+    return callback(id).catch(() => false)
+  })
+  return Promise.all(locationPromises).then(result => result.filter(Boolean))
```
teneightfive: Maybe leave `result` as `locations`. Feels more clear to what is returned from those promises.

common/services/reference-data.js
```diff
@@ -49,14 +49,23 @@ function getLocationById(id) {
   return apiClient.find('location', id).then(response => response.data)
 }
 
-function getLocationsById(locations = []) {
-  const locationPromises = locations.map(locationId => {
-    return getLocationById(locationId).catch(() => false)
-  })
+function getLocationByAgencyId(agencyId) {
```
teneightfive: Could also name the argument `nomisAgencyId` so you can use shorthand in the `getLocations` call.

common/services/user-locations.test.js
```diff
@@ -0,0 +1,122 @@
...
+  getLocationsByAgencyId: sinon.spy(args => {
+    return [{ id: 'test' }]
+  }),
+}
+
+const configStub = {
+  AUTH_PROVIDERS: {
+    hmpps: {
+      groups_url: userName => {
+        return 'http://test/'
```
teneightfive: Could shorten this to:

```js
groups_url: () => 'http://test/',
```

app/auth/middleware.js
```diff
@@ -38,6 +30,18 @@ function processAuthResponse(defaultLocations = []) {
...
 
+function _setSessionData(session, token, postAuthRedirect) {
+  const tokenData = decodeAccessToken(token)
+
+  session.authExpiry = tokenData.exp
+  session.postAuthRedirect = postAuthRedirect
+  session.user = new User({
+    name: tokenData.user_name,
+    roles: tokenData.authorities,
+    locations: userLocationService.getUserLocations(token),
```
teneightfive: Does this not require `await` as the call is async?

Probably neater to leave the locations call on line 19, just inside the `try/catch`, like before. And can then leave the line above the same as it was.

app/auth/middleware.js
slorek: I agree it's not any better. I'm not sure it's possible to extract the callback function easily because it'll need scope to access `req` and `next` which are scoped to the parent `middleware` function. Sure it's possible but it's going to be a meaty refactoring (unless you can think of an easy way). Are you happy for me to revert it back and mark it as won't fix?

common/services/reference-data.js
slorek: I was thinking of renaming the function `getLocationByNomisAgencyId` (and everywhere else it is called `agencyId` to be clearer, do you think this is worthwhile?

app/auth/middleware.js
teneightfive: Yeah, given that would be a little messier too I'd just revert to what it was before and we can mark code climate to ignore the line length in this particular instance.

common/services/reference-data.js
teneightfive: I was also thinking the same thing. I think it's clearer to what it's referring to as these are tied directly to NOMIS.

app/auth/middleware.js
teneightfive: Ah, the test is passing because you're not returning a promise, which differs from what `getUserLocations` actually returns.

If you look how the test worked for the reference data stub you should be able to copy that to test promise rejection and success.

slorek: I've fixed the test and added test cases for failure.

app/auth/middleware.test.js
```diff
@@ -46,8 +43,13 @@ describe('Authentication middleware', function() {
     })
 
     context('when there is no grant response', function() {
-      beforeEach(function() {
-        authentication.processAuthResponse()(req, {}, nextSpy)
+      beforeEach(async function() {
+        const authentication = proxyquire('./middleware', {
```
teneightfive: You can leave the proxyquire definition at the top this where it was and pass a `sinon.stub()` into the `getUserLocations` method. Then use `. rejects()` and `.resolves()` in the test before functions.

At top of file:
```js
const getUserLocationsStub = sinon.stub()
const authentication = proxyquire('./middleware', {
  '../../common/lib/user': UserStub,
  '../../common/services/user-locations': {
    getUserLocations: getUserLocationsStub,
  },
})
```

In success `beforeEach`:

```js
getUserLocationsStub.resolves(['TEST'])
```

In failure `beforeEach`:

```js
getUserLocationsStub.rejects(userLocationFailureError)
```

slorek: Ah, clever thanks. I took this implementation from another test.

.env.example
```diff
@@ -8,6 +8,7 @@ AUTH_PROVIDER_KEY=
 AUTH_PROVIDER_SECRET=
 AUTH_PROVIDER_URL=
 SERVER_HOST=localhost:3000
+NOMIS_ELITE2_API_URL=https://gateway.t3.nomis-api.hmpps.dsd.io
```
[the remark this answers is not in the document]
slorek: It’s not considered secret so I think it is helpful to include it here.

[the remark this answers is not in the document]
teneightfive: It doesn't match what we have at the moment is the only thing. I'd say we should either add all of the non-secret URLs if we think it's helpful, or none if we don't want to expose them. 

Bit odd having half a convention.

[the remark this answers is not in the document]
slorek: I think we should add the missing ones, but in a separate PR.

app/auth/middleware.js
```diff
@@ -15,19 +11,23 @@ function processAuthResponse(defaultLocations = []) {
...
+      const userLocations = await userLocationService.getUserLocations(
+        grant.response.access_token
+      )
 
       req.session.regenerate(error => {
         if (error) {
           return next(error)
         }
 
+        const accessToken = decodeAccessToken(grant.response.access_token)
```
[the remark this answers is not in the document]
slorek: I felt that it was a minor improvement to move it as `accessToken` is only used within the scope of the callback function (same reason `userLocations` was moved).

app/auth/middleware.test.js
slorek: I've had a go at this, but I don't think the dynamic changing of the stub works in this way. It seems that the behaviour is fixed once `proxyquire` loads it. I think that's why it was also implemented this way in https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/app/healthcheck/middleware.test.js. Happy for you to have a crack at it.

teneightfive:
Realised with this PR we're introducing another dependency for the frontend app. We should add a ping endpoint to the heathcheck dependencies.

Happy for this to be part of this PR as this is the PR that introduces the dependency but I'd make it another commit to be clearer.

slorek:
Good spot about the health check. I've added this.

PR #192, opened 2019-08-18
Show user's location in header

slorek:
This shows the user's current location on the page header, using the [MoJ organisation switcher](https://moj-design-system.herokuapp.com/components/organisation-switcher) component. This can then be extended in a subsequent PR to allow the user to switch their location.

## What it looks like

Screenshot 2019-08-18 at 14 24 34

### DIFF
--- a/common/assets/scss/application.scss
+++ b/common/assets/scss/application.scss
@@ -29,6 +29,9 @@ $govuk-font-family: "Inter", system-ui, sans-serif;
 @import "govuk-frontend/govuk/utilities/all";
 @import "govuk-frontend/govuk/overrides/all";
 
+// Import MoJ Frontend Components
+@import "@ministryofjustice/frontend/moj/components/organisation-switcher/_organisation-switcher.scss";
+
 // Import app settings/tools/helpers
 @import "helpers/all";
 
--- a/common/assets/scss/overrides/_font-corrections.scss
+++ b/common/assets/scss/overrides/_font-corrections.scss
@@ -8,3 +8,11 @@
   padding-top: 4px;
   padding-bottom: 4px;
 }
+
+.moj-organisation-nav {
+  margin: 0;
+  padding: 10px 0;
+}
+.moj-organisation-nav__link {
+  padding-top: 2px;
+}
--- a/common/templates/layouts/base.njk
+++ b/common/templates/layouts/base.njk
@@ -12,6 +12,8 @@
 {% from "govuk/components/summary-list/macro.njk"      import govukSummaryList %}
 {% from "govuk/components/textarea/macro.njk"          import govukTextarea %}
 
+{% from "moj/components/organisation-switcher/macro.njk" import mojOrganisationSwitcher %}
+
 {# App level components #}
 {% from "card/macro.njk"              import appCard %}
 {% from "data/macro.njk"              import appData %}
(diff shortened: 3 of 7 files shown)


common/assets/scss/application.scss
```diff
@@ -29,6 +29,9 @@ $govuk-font-family: "Inter", system-ui, sans-serif;
 @import "govuk-frontend/govuk/utilities/all";
 @import "govuk-frontend/govuk/overrides/all";
 
+// Import MoJ Frontend Components
+@import "@ministryofjustice/frontend/moj/components/organisation-switcher/_organisation-switcher.scss";
```
teneightfive: Is there a way to add this module to the sass load paths to reduce the amount of folders needed. Similar to how the GOV.UK frontend is loaded.

slorek: I think this is being loaded in the same way, just the path in node_modules is longer in this library.

teneightfive: Ah yes. Tad annoying they've added that `frontend/` folder. Not sure what its use is...

teneightfive:
> Could we double check the spacing top/bottom. The screenshot looks a little uneven. It might need adjusting because we’re not using the GOV.UK Transport font where the baseline is slightly off.

@slorek You can add the line height corrections to the [font-corrections overrides file](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/assets/scss/overrides/_font-corrections.scss).

slorek:
Done. i've also ensured the optional link (used in #193) is aligned to the baseline of the header.

Screenshot 2019-08-19 at 20 12 37

common/assets/scss/overrides/_font-corrections.scss
```diff
@@ -8,3 +8,12 @@
   padding-top: 4px;
   padding-bottom: 4px;
 }
+
+.moj-organisation-nav {
```
teneightfive: You can drop this surrounding definition. One of the ideas behind [BEM](http://getbem.com/) is that you don't need to nest things so heavily so it makes it easier to read.

teneightfive: Also, noticed they've added some annoying margins onto the element which clash with the margins in our main container. 

This should fix the erroneous way spacing is being done:

```scss
.moj-organisation-nav {
  margin: 0;
  padding: 10px 0;
}
.moj-organisation-nav__link {
  padding-top: 2px;
}
```

I might try and do a PR to that repo to fix the spacing too.

teneightfive:
Also, I noticed a strange thing where the banner doesn't show up the first time you sign in. It might be because the `USER` variable isn't available in the view initially.

common/assets/scss/overrides/_font-corrections.scss
slorek: 👍

slorek:
> Also, I noticed a strange thing where the banner doesn't show up the first time you sign in. It might be because the `USER` variable isn't available in the view initially.

Strange, didn't see that. I'll try to reproduce.

teneightfive:
> > Also, I noticed a strange thing where the banner doesn't show up the first time you sign in. It might be because the `USER` variable isn't available in the view initially.
> 
> Strange, didn't see that. I'll try to reproduce.

I can't seem to reproduce either, I think it might have been a strange thing whilst I was switching branches. Will keep an eye on out but seems fine now.

PR #193, opened 2019-08-18
Return error when user has no locations set

slorek:
This shows a more meaningful error message when the user has no locations set, e.g. when they have no groups assigned, or no matching location is found in the back end API.

Screenshot 2019-08-18 at 15 08 09

### DIFF
--- a/app/move/controllers/new.form.js
+++ b/app/move/controllers/new.form.js
@@ -22,9 +22,7 @@ class FormController extends Controller {
 
   checkCurrentLocation(req, res, next) {
     if (!req.session.currentLocation) {
-      const error = new Error(
-        'Current location is not set. Check environment variable is correctly set.'
-      )
+      const error = new Error('Current location is not set.')
       return next(error)
     }
 
--- a/app/move/controllers/new.form.test.js
+++ b/app/move/controllers/new.form.test.js
@@ -120,7 +120,7 @@ describe('Move controllers', function() {
           expect(nextSpy).to.be.calledOnce
           expect(nextSpy.args[0][0] instanceof Error).to.be.true
           expect(nextSpy.args[0][0].message).to.equal(
-            'Current location is not set. Check environment variable is correctly set.'
+            'Current location is not set.'
           )
         })
       })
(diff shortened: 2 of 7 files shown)


slorek:
I think a User#isSupplier method would be a better way to check if they are a supplier. Will do that tomorrow.

teneightfive:
I’m not 100% convinced we actually need this to “error”. 

I was more thinking that by the time is reached the set middleware, if no location in the session it will always redirect to the locations feature to choose a location (unless there is only one available). 

If the list of locations is empty (either they have access to none or the ones they have aren’t in our list of locations) we can have a message saying no available locations and to contact “support” (ie, us) to rectify. 

I think that’s better than the app actually throwing a permission error as it can come with more targeted actions in the message too.

slorek:
I think we can iterate this to display a friendly error page once we’ve got an idea of what that could look like. For now the error which is displayed is misleading and unhelpful.

teneightfive:
> I think we can iterate this to display a friendly error page once we’ve got an idea of what that could look like. For now the error which is displayed is misleading and unhelpful.

But I think what I'm saying is the other task should be the iteration of that. This change just introduces debt that we'll have to remove once we make the other change.

I think we can close this PR and focus on the other feature which will make this change redundant.

slorek:
OK so you want me to redirect to the choose locations page if there are no locations and display an error on that page. 👍

teneightfive:
> OK so you want me to redirect to the choose locations page if there are no locations and display an error on that page. 👍

Yes, pretty much. I don't think it should really be an app "error" though.

The locations feature should list all possible locations at the root of the feature (`/locations`) and we can then use the nunjucks [`foreach/else`](https://mozilla.github.io/nunjucks/templating.html#for) to display a message if there are no locations available.

Feels like that requires less logic around middleware too.

We can work with Leigh and Jason on what that copy should say, but it should be something around contacting the support team to ensure they have access to the correct locations.

slorek:
OK, once #194 is merged we can iterate and hook that logic in. I'll move the ticket back to the backlog and add what we've just agreed.

PR #194, opened 2019-08-19
[P4-506] Choose location

slorek:
This allows the user to switch location as per the prototype. It builds upon #192 so please review only the last commit.

### DIFF
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -1,4 +1,4 @@
-const { get, sortBy } = require('lodash')
+const { find, get, sortBy } = require('lodash')
 
 function locations(req, res) {
   const locations = get(req.session, 'user.locations', [])
@@ -8,6 +8,21 @@ function locations(req, res) {
   })
 }
 
+function setLocation(req, res, next) {
+  const { locationId } = req.params
+  const locations = get(req.session, 'user.locations', [])
+  const location = find(locations, { id: locationId })
+
+  if (!location) {
+    return next()
+  }
+
+  req.session.currentLocation = location
+
+  res.redirect('/')
+}
+
 module.exports = {
   locations,
+  setLocation,
 }
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -1,0 +1,13 @@
+const { get, sortBy } = require('lodash')
+
+function locations(req, res) {
+  const locations = get(req.session, 'user.locations', [])
+
+  res.render('locations/views/locations.njk', {
+    locations: sortBy(locations, 'title'),
+  })
+}
+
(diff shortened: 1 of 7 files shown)


locales/en/locations.json
```diff
@@ -0,0 +1,5 @@
+{
+  "police": "Police Stations",
+  "prison": "Pirsons",
```
slorek: Spotted typo.

app/locations/controllers.js
```diff
@@ -0,0 +1,28 @@
...
+  const userLocations = _getUserLocations(req.session.user)
+  res.render('locations/views/locations.njk', { userLocations })
+}
+
+function setLocation(req, res) {
+  const locationId = req.params.locationId
+  const location = find(req.session.user.locations, ['id', locationId])
+
+  if (locationId && !location) {
+    throw new Error(`Not authorised to switch to location ${locationId}`)
```
teneightfive: This could probably be passed to `next` rather than throwing an error in the middle of this method.

app/locations/controllers.js
```diff
@@ -0,0 +1,28 @@
+const { find, groupBy, sortBy } = require('lodash')
+
+function _getUserLocations(user) {
```
feedmypixel: I don't feel we need to specify functions as private in this way. This isn't a class. 
I'd also agree with Dom I don't feel the `_getUserLocations` is adding to the clarity of the code and this abstraction is only being used in one place and there is a bit of jumping around which is harder to read. Feels like it would be clearer in the `locations` function

app/locations/controllers.js
feedmypixel: How would you like this error shown? Does it need to be in the UI? or is it an operational error?

app/locations/controllers.js
```diff
@@ -0,0 +1,28 @@
...
+
+function setLocation(req, res) {
+  const locationId = req.params.locationId
+  const location = find(req.session.user.locations, ['id', locationId])
+
+  if (locationId && !location) {
+    throw new Error(`Not authorised to switch to location ${locationId}`)
+  } else if (location) {
+    req.session.currentLocation = location
+  }
```
feedmypixel: To make it a little clearer rather than an `if, else if` you could do:

```
if (locationId && !location) {
    throw new Error(`Not authorised to switch to location ${locationId}`)
}

if (location) {
    req.session.currentLocation = location
}
```

app/locations/index.js
```diff
@@ -0,0 +1,15 @@
+// NPM dependencies
```
feedmypixel: Are the comments needed in this file?

PR #195, opened 2019-08-20
[P4-611] Add message when no locations are available

teneightfive:
This display a message to the user when there are no available
locations.

Rather than throw an app level error we want to communicate to the
user what has happened and how to solve the scenario.

### Without email address set
localhost_3000_locations

### With email address set
localhost_3000_locations (4)


teneightfive:
@Andy-Millar I've currently done it using the "info" message as the error message felt too harsh. Have added screenshots of the alternative styles. Let me know your thoughts

### Error
localhost_3000_locations (2)

### Warning
localhost_3000_locations (1)

### Muted
localhost_3000_locations (3)

PR #196, opened 2019-08-20
[P4-505] Display username in header

feedmypixel:
## Feature
Add user name into the internal header.

## ToDo
The `user.username` is currently only available via the `access_token` further work needs to be created to query the api to get the `user.name` which is easier on the eye. This was felt out of scope for this work.

### Large screen
Screenshot 2019-08-20 at 17 10 22

### Small screen
Screenshot 2019-08-20 at 17 14 31

### DIFF
--- a/common/components/internal-header/template.njk
+++ b/common/components/internal-header/template.njk
@@ -62,6 +62,10 @@
                 {{ item.text }}
               </a>
             </li>
+          {% elif item.text %}
+            <li class="app-header__navigation-item">
+              {{ item.text }}
+            </li>
           {% endif %}
         {% endfor %}
       </ul>
--- a/common/components/internal-header/template.test.js
+++ b/common/components/internal-header/template.test.js
@@ -106,6 +106,31 @@ describe('Internal header component', function() {
         'my-attribute-2'
       )
     })
+
+    it('renders navigation items with text only', function() {
+      const $ = render('internal-header', {
+        navigation: [
+          {
+            text: 'Item 1',
+          },
+          {
+            text: 'Item 2',
+          },
+        ],
+      })
+
+      const $component = $('.app-header')
+      const $list = $component.find('ul.app-header__navigation')
+      const $firstItem = $list.find('li.app-header__navigation-item').first()
+      const $lastItem = $list.find('li.app-header__navigation-item').last()
+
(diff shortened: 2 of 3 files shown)


common/components/internal-header/template.njk
```diff
@@ -63,6 +63,11 @@
               </a>
             </li>
           {% endif %}
+          {% if item.text and not item.href %}
```
teneightfive: Could do this as an `elif`

common/components/internal-header/template.test.js
```diff
@@ -80,10 +80,15 @@ describe('Internal header component', function() {
       const $component = $('.app-header')
       const $list = $component.find('ul.app-header__navigation')
       const $items = $list.find('li.app-header__navigation-item')
-      const $firstItem = $items.find('a.app-header__link:first-child')
-      expect($items.length).to.equal(4)
-      expect($firstItem.attr('href')).to.equal('#1')
-      expect($firstItem.text()).to.contain('Navigation item 1')
+      const $anchorNavItem = $items
```
teneightfive: You can probably ignore needing to change this test actually and write a separate test where you pass in a set of items without a `href` property.

common/components/internal-header/template.test.js
```diff
@@ -106,6 +111,17 @@ describe('Internal header component', function() {
         'my-attribute-2'
       )
     })
+
+    it('renders navigation item with text only', function() {
+      const $ = render('internal-header', examples['with navigation'])
```
teneightfive: You could reuse the method used for the test above to pass in a custom set of items rather than using the ones from the example.

The idea behind the examples is to use those configurations to show examples of the component later on similar to how it's done on the [GOV.UK Design System](https://design-system.service.gov.uk/components/)

common/components/internal-header/template.test.js
```diff
@@ -106,6 +111,17 @@ describe('Internal header component', function() {
...
+
+      const $component = $('.app-header')
+      const $list = $component.find('ul.app-header__navigation')
+      const $lastItem = $list.find('li.app-header__navigation-item').last()
+
+      expect($lastItem.text()).to.contain('Navigation item 5')
```
teneightfive: Maybe an assertion to check that there is no `<a>` element present too?

PR #197, opened 2019-08-21
[P4-617] Redirect user to choose location on sign in

teneightfive:
# Feature

This change handles setting the current location when a user signs in.

**If a user has access to multiple locations** - it will redirect to the locations feature to choose a current location

**If a user has access to one location** - it will set that as the current location and bypass the locations feature

### Checklist


common/middleware/ensure-current-location.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function ensureCurrentLocation({
```
feedmypixel: this may benefit, readability wise, from having the `module.exports` at the bottom

app/locations/middleware.test.js
```diff
@@ -0,0 +1,119 @@
...
+    })
+
+    context('when no user exists on session', function() {
+      beforeEach(function() {
+        middleware.setUserLocations(req, res, nextSpy)
+      })
+
+      it('should set empty user locations on request', function() {
+        expect(req).to.have.property('userLocations')
+        expect(req.userLocations).to.deep.equal([])
```
feedmypixel: does this need to be `deep`?

app/locations/middleware.test.js
```diff
@@ -0,0 +1,119 @@
+const middleware = require('./middleware')
+
+describe('Locations middleware', function() {
+  let req, res, nextSpy
+  const userLocations = [
```
feedmypixel: so `userLocations` mock data is used in a couple of places. Could this be dry'd up?

app/locations/controllers.test.js
```diff
@@ -29,67 +29,43 @@ describe('Locations controllers', function() {
...
-            id: 'testprison',
-            title: 'Test Prison',
-            location_type: 'prison',
-          },
-        ])
-      })
+    it('should return locations sorted by title', function() {
+      const params = res.render.args[0][1]
+      expect(params).to.have.property('locations')
+      expect(params.locations).to.deep.equal([
```
feedmypixel: `userLocations` mock data is used in a couple of places. Could this be dry'd up?

PR #200, opened 2019-08-23
Amend no locations message

teneightfive:
This change updates the copy of the message that is displayed when
a user doesn't have access to any locations.


lmoney-moj:
Looks good to me.

PR #201, opened 2019-08-26
[P4-504] Add sticky sidebar to move detail page

teneightfive:
This adds some client side JS to pin the right hand panel
on the move detail page to the top of the window on scroll.

This allows the user to keep the context of the person they
are moving and where they are moving as they scroll the page.

## What it looks like (on scroll)

image

### DIFF
--- a/app/move/views/view.njk
+++ b/app/move/views/view.njk
@@ -45,7 +45,7 @@
     {% endif %}
   </header>
 
-  <div class="govuk-grid-row">
+  <div class="govuk-grid-row sticky-sidebar-container">
     <div class="govuk-grid-column-two-thirds">
       <h2 class="govuk-heading-m">
         {{ t("moves::steps.personal_details.heading") }}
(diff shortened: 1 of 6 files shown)


teneightfive:
> Looks good. One question would be how much weight is this library adding to our client-side work? Compared to writing this ourselves in `JavaScript`? Or just using `position: sticky` (which has reasonable support) https://caniuse.com/#feat=css-sticky and then falling back to not sticky for older browsers?
> 
> Which then asks the question whats our user base using browser wise?

Just checked analytics and looks like Police users are on IE11 (which doesn't support `position: sticky` unfortunately). Prison users are also likely to be on IE and doesn't look support will be added for it.

teneightfive:
> Looks good. One question would be how much weight is this library adding to our client-side work? 

### Raw file size
Before: 69.2kb
After: 79.8kb (~ +10kb)

I can't seem to check the gzipped size for the new file but the current file comes down to **20.7kb when gzipped**.

I'm happy with the small bump, what about you?

feedmypixel:
> I'm happy with the small bump, what about you?

Thats fine with me 👍

PR #202, opened 2019-08-27
Restructure move feature folder structure

teneightfive:
This change groups all controllers, fields and steps related to
the create journey into separate folders to make it easier to maintain.

It is also prep to allow the cancel journey within a move to be
expanded in a future change.


app/move/controllers/create/assessment.js
```diff
@@ -1,11 +1,11 @@
 const { flatten, values } = require('lodash')
 
-const FormController = require('./new.form')
-const fieldHelpers = require('../../../common/helpers/field')
-const referenceDataService = require('../../../common/services/reference-data')
-const referenceDataHelpers = require('../../../common/helpers/reference-data')
+const CreateBaseController = require('./base')
```
feedmypixel: ~do we need `base` in this name? Would `CreateController` do?~

Reading through this I can see why you have used `base`

app/move/controllers/create/assessment.test.js
```diff
@@ -1,8 +1,8 @@
 const FormController = require('hmpo-form-wizard').Controller
 
-const Controller = require('./new.assessment')
-const referenceDataService = require('../../../common/services/reference-data')
-const referenceDataHelpers = require('../../../common/helpers/reference-data')
+const Controller = require('./assessment')
```
feedmypixel: Can this become `AssessmentController`?

app/move/controllers/create/base.js
```diff
@@ -0,0 +1,31 @@
...
+
+  setCancelUrl(req, res, next) {
+    res.locals.cancelUrl = res.locals.MOVES_URL
+    next()
+  }
+
+  checkCurrentLocation(req, res, next) {
+    if (!req.session.currentLocation) {
+      const error = new Error(
+        'Current location is not set. Check environment variable is correctly set.'
```
feedmypixel: I'm wondering if this is the correct place to check this? 
If we need the application to work with this environment variable maybe we should be checking in one place on startup of the app? 
Maybe something that loops through a required list of `envs` and checks they are set in `process.env`?

app/move/controllers/create/base.test.js
```diff
@@ -0,0 +1,127 @@
...
+      })
+
+      context('with no moves url local', function() {
+        beforeEach(function() {
+          controller.setCancelUrl({}, res, nextSpy)
+        })
+
+        it('should set cancel url correctly', function() {
+          expect(res.locals).to.have.property('cancelUrl')
+          expect(res.locals.cancelUrl).to.equal(undefined)
```
feedmypixel: more personal preference but you could aslo use 
```
.to.be.an('undefined')
```

app/move/controllers/create/base.test.js
```diff
@@ -0,0 +1,127 @@
...
+        })
+
+        it('should call next', function() {
+          expect(nextSpy).to.be.calledOnceWithExactly()
+        })
+      })
+
+      context('with moves url local', function() {
+        beforeEach(function() {
+          res.locals.MOVES_URL = '/moves?move-date=2019-10-10'
```
feedmypixel: `'/moves?move-date=2019-10-10'` could go into a `const` here as its used a few times and giving a `const` a name may explain what it is a little more (if needed)

app/move/controllers/create/base.test.js
```diff
@@ -0,0 +1,127 @@
...
+      })
+
+      context('when no current location exists', function() {
+        beforeEach(function() {
+          controller.checkCurrentLocation(req, {}, nextSpy)
+        })
+
+        it('should call next with error', function() {
+          expect(nextSpy).to.be.calledOnce
+          expect(nextSpy.args[0][0] instanceof Error).to.be.true
```
feedmypixel: more personal preference but you also have:
```
.to.be.an('error')
```

app/move/controllers/create/assessment.test.js
teneightfive: I think that was just when I was keeping the tests a little simple. The test is testing that particular controller and it's not importing multiple controllers so didn't see the need to have a more verbose name.

If we want to rename things like this it might be better as a separate change, what do you think?

app/move/controllers/create/base.js
teneightfive: In fact, this may be redundant with the recent set of changes to how locations are handled.

I'll address this in a separate PR.

app/move/controllers/create/base.test.js
teneightfive: Yeah, that's much nicer. Again, may do these kind of tweaks as a separate change to not confuse with what this change is doing.

app/move/controllers/create/base.js
teneightfive: Have updated this in #207

app/move/controllers/create/base.test.js
teneightfive: Have added #208 to amend this across the codebase.

app/move/controllers/create/base.test.js
teneightfive: Have amended this in #208 as well.

app/move/controllers/create/base.test.js
teneightfive: Might leave this one as it only appears twice. I think this logic will be revisited again soon too when we start introducing more complex search to the dashboard.

PR #205, opened 2019-08-28
[P4-505] Display first and last name

feedmypixel:
### Feature
We currently are displaying a users `userName` in the header. This work gets the current authed users name from our API and displays it in the header.

### Other
This work also involves creating a user service that brings in the user locations work 

### Screenshots
Screenshot 2019-08-28 at 11 05 53


app/auth/middleware.js
```diff
@@ -1,5 +1,5 @@
 const User = require('../../common/lib/user')
-const userLocationService = require('../../common/services/user-locations')
+const { getUserLocations } = require('../../common/services/user-locations')
```
teneightfive: Might want to make sure this isn't affecting the test stubs. 

I came across problems when stubbing methods that were destructured in the initial module.

app/auth/middleware.js
```diff
@@ -12,9 +12,7 @@ function processAuthResponse() {
     }
 
     try {
-      const userLocations = await userLocationService.getUserLocations(
-        grant.response.access_token
```
teneightfive: Wonder if it's worth setting `grant.response.access_token` to a `const`? It's called a number of times in this method now.

common/services/user-locations.js
```diff
@@ -1,53 +1,47 @@
...
-      return _getNomisAgencies(token)
-    default:
-      return Promise.resolve([])
+  if (authSource === 'nomis') {
+    return getNomisAgencies(token).then(agencies =>
+      getLocationsByNomisAgencyId(agencies)
+    )
   }
+
+  return []
```
teneightfive: Just noticed this should return the same type like the current method. The other instances return a promise (which ultimately return an array), but this just returns an array. I'd probably keep the `Promise.resolve([])` line to produce a consistent return type.

common/services/user-locations.js
```diff
@@ -1,53 +1,47 @@
 const axios = require('axios')
-const { map } = require('lodash')
 
 const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
 const { decodeAccessToken } = require('../lib/access-token')
-const referenceDataService = require('./reference-data')
+const { getLocationsByNomisAgencyId } = require('./reference-data')
```
teneightfive: Again, just make sure this isn't affecting the test stubs.

app/auth/middleware.test.js
```diff
@@ -15,12 +15,6 @@ const userLocationFailureStub = {
   },
 }
 
-function UserStub({ name, roles = [], locations = [] } = {}) {
```
teneightfive: I wonder if we do want to do this. For me this ties these two unit tests together and makes this test care about how the User library is structured.

Maybe this middleware test should just be checking that the User class is called with the correct arguments that are built during this middleware?

common/services/user-locations.test.js
```diff
@@ -1,4 +1,5 @@
 const proxyquire = require('proxyquire')
+const { encodeToken } = require('../../test/unit/common-helpers')
```
teneightfive: Maybe just `helpers` instead of `common-helpers`?

common/lib/request.js
```diff
@@ -1,8 +1,14 @@
 const queryString = require('query-string')
 
+
+const buildAuthHeader = function(token) {
```
teneightfive: For consistency maybe use the same shorthand syntax that is used for `getQueryString`?

common/lib/request.js
```diff
@@ -1,14 +1,19 @@
 const queryString = require('query-string')
 
+const getQueryString = function(target, source) {
+  const merged = {
```
teneightfive: Could just inline this too?

```js
return `?${queryString.stringify({ ...target, ...source })}`
```

common/templates/layouts/base.njk
```diff
@@ -65,7 +65,7 @@
     containerClasses: "govuk-width-container",
     navigation: [
       {
-        text: USER.userName
+        text: USER.firstName + " " + USER.lastName
```
teneightfive: Might be neater to add a `fullname` property onto the User class when it is built so that the fullname can be referenced anywhere else and we know it will always be consistent.

For example, it could be re-used in the future on a welcome page or a user profile page.

config/index.js
```diff
@@ -86,12 +86,14 @@ module.exports = {
       groups_url: userName => {
         return _authUrl(`/auth/api/authuser/${userName}/groups`)
       },
+      user_url: userName => _authUrl(`/auth/api/authuser/${userName}`),
```
teneightfive: I'm guessing the `/me` didn't work for SSO?

common/lib/request.js
feedmypixel: I've moved this into the user service work

common/services/user-locations.test.js
feedmypixel: have removed this as its now only being used in one file

app/auth/middleware.js
```diff
@@ -11,23 +11,25 @@ function processAuthResponse() {
...
-        req.session.authExpiry = accessToken.exp
+        req.session.authExpiry = decodedAccessToken.exp
         req.session.postAuthRedirect = postAuthRedirect
         req.session.user = new User({
-          name: accessToken.user_name,
-          roles: accessToken.authorities,
-          locations: userLocations,
+          fullName,
+          locations,
+          userName: decodedAccessToken.user_name,
```
teneightfive: I know it's not fully related to this change but `user.userName` should probably be all lowercase as it's one word, not two.

app/auth/middleware.test.js
```diff
@@ -56,7 +51,7 @@ describe('Authentication middleware', function() {
         expect(nextSpy).to.be.calledOnceWithExactly()
       })
 
-      it('doesn’t regenerate the session', function() {
+      it('doesnt regenerate the session', function() {
```
teneightfive: Any reason to remove the apostrophe here?

app/auth/middleware.test.js
```diff
@@ -73,15 +68,28 @@ describe('Authentication middleware', function() {
       context('when the user locations lookup fails', function() {
         beforeEach(async function() {
           const authentication = proxyquire('./middleware', {
-            '../../common/lib/user': UserStub,
```
teneightfive: I think it would be good to keep some kind of stub here to check that the User class is being called with the right properties in the object sent to it.

app/auth/middleware.test.js
```diff
@@ -73,15 +68,28 @@ describe('Authentication middleware', function() {
...
+      context('when the user fullname lookup fails', function() {
+        beforeEach(async function() {
+          const authentication = proxyquire('./middleware', {
+            '../../common/services/user': userFullNameFailureStub,
           })
 
           await authentication.processAuthResponse()(req, {}, nextSpy)
         })
 
         it('returns next', function() {
```
teneightfive: `calls with with error` (although I know it's an existing description but it's doing more than returning next)?

Might also be worth checking that the things that occur after it, don't actually occur as this current test doesn't guarantee that.

common/lib/user.js
```diff
@@ -1,5 +1,6 @@
-function User({ name, roles = [], locations = [] } = {}) {
-  this.userName = name
+function User({ userName, fullName, roles = [], locations = [] } = {}) {
+  this.userName = userName
```
teneightfive: Mentioned earlier but maybe this should be `username` as it's actually only one word.

common/lib/user.test.js
```diff
@@ -21,6 +25,23 @@ describe('User class', function() {
...
+    context('with full name', function() {
+      it('should set full name', function() {
+        fullName = 'Mr Benn'
+        user = new User({ fullName })
+
+        expect(user.fullName).to.equal(fullName)
+      })
+    })
+
+    context('without first name', function() {
```
teneightfive: `without full name`?

common/lib/user.test.js
```diff
@@ -21,6 +25,23 @@ describe('User class', function() {
...
+      it('should not set first name', function() {
```
teneightfive: `should not set full name`?

common/lib/user.js
teneightfive: In fact, do we even need this anymore? It's not being used anywhere in the app. Could just simply things and remove it?

app/auth/middleware.test.js
feedmypixel: being trigger happy :)

common/lib/user.test.js
```diff
@@ -2,22 +2,25 @@ const User = require('./user')
...
     })
 
-    context('without username', function() {
-      it('should not set username', function() {
+    context('without fullname', function() {
+      it('should not set fullname', function() {
         user = new User()
 
-        expect(user.userName).to.be.undefined
+        expect(user.fullname).to.be.an('undefined')
```
teneightfive: This can also be `.to.be.undefined`. I find it reads a little nicer than "an undefined".

common/services/user.test.js
```diff
@@ -55,16 +83,16 @@ describe('User locations service', function() {
...
           nock(configStub.AUTH_PROVIDERS.hmpps.groups_url('test'))
             .get('/')
             .reply(200, JSON.stringify(authGroups))
 
-          result = await userLocationsService.getUserLocations(token)
+          result = await getLocations(token)
         })
 
-        it("requests the user's groups from HMPPS SSO", function() {
+        it('requests the users groups from HMPPS SSO', function() {
```
teneightfive: Another trigger happy removal? ;)

feedmypixel: he he yup :)

PR #206, opened 2019-08-28
[P4-638] Fix moves permissions with no locations case

teneightfive:
If a user has permission to view all moves, regardless of location,
they should be able to view moves even if their user does not currently
have any available locations.

This is currently the case for suppliers. Without this change a
supplier user is only greeted with the no locations message and cannot
view any other parts of the app.

### DIFF
--- a/common/middleware/ensure-current-location.js
+++ b/common/middleware/ensure-current-location.js
@@ -1,9 +1,14 @@
 function ensureCurrentLocation({ locationsMountpath, whitelist = [] } = {}) {
   return (req, res, next) => {
+    const { permissions = [], locations = [] } = req.session.user || {}
+    const canViewAllMovesWithoutLocations =
+      permissions.includes('moves:view:all') && !locations.length
+
     if (
       req.session.currentLocation ||
       whitelist.includes(req.url) ||
-      req.url.includes(locationsMountpath)
+      req.url.includes(locationsMountpath) ||
+      canViewAllMovesWithoutLocations
     ) {
       return next()
     }
--- a/common/middleware/ensure-current-location.test.js
+++ b/common/middleware/ensure-current-location.test.js
@@ -86,4 +86,69 @@ describe('Ensure current location middleware', function() {
       expect(res.redirect).to.be.calledOnceWithExactly(locationsMountpath)
     })
   })
+
+  context('when user has view all moves permission', function() {
+    beforeEach(function() {
+      req.session.user = {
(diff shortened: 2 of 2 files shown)


common/middleware/ensure-current-location.js
```diff
@@ -1,9 +1,12 @@
 function ensureCurrentLocation({ locationsMountpath, whitelist = [] } = {}) {
   return (req, res, next) => {
+    const { permissions = [], locations = [] } = req.session.user || {}
```
feedmypixel: If you move the `ensureCurrentLocation` middleware in `server.js` [server.js#L129-L134](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/544b833225215e50121116c30f2213729cc2afdc/server.js#L129-L134)

```
app.use(
  ensureCurrentLocation({
    locationsMountpath: locationsApp.mountpath,
    whitelist: config.AUTH_WHITELIST_URLS,
  })
)
```

To below the routing middleware [server.js#L139](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/server.js#L139)

There is no need to have the empty object fallback:
```
|| {}
```

common/middleware/ensure-current-location.js
```diff
@@ -1,9 +1,12 @@
...
+
     if (
       req.session.currentLocation ||
       whitelist.includes(req.url) ||
-      req.url.includes(locationsMountpath)
+      req.url.includes(locationsMountpath) ||
+      (permissions.includes('moves:view:all') && !locations.length)
```
feedmypixel: Maybe capturing this in a const would aid in explanation:

```
const supplierWithoutLocations = (permissions.includes('moves:view:all') && !locations.length)

```

or

```
const viewAllMovesWithoutLocations = (permissions.includes('moves:view:all') && !locations.length)

```

But then maybe it's not needed? I think it will become apparent after writing to see if it's too much. Your call and as mentioned after a chat and seeing the tests its become clear what it does.

common/middleware/ensure-current-location.js
teneightfive: Don't think that would work. That would mean all the app routes are served without this middleware ever being run.

common/middleware/ensure-current-location.js
teneightfive: Was thinking about it over lunch and this would be the most descriptive: `canViewAllMovesWithoutLocations`...but is that too verbose?

PR #210, opened 2019-08-28
[P4-629] Capture cancellation reason

teneightfive:
This change adds the initial journey to capture reasons for cancelling a move request.

It makes use of the existing form wizard library to capture form inputs and handle
validation of particular fields.

## What it looks like

localhost_3000_move_00ddf148-f3b8-4189-82ff-7ef26fe611d9_cancel_reason (2)

localhost_3000_move_00ddf148-f3b8-4189-82ff-7ef26fe611d9_cancel_reason (3)

localhost_3000_moves

### DIFF
--- a/app/move/controllers/cancel.js
+++ b/app/move/controllers/cancel.js
@@ -1,24 +1,37 @@
+const { pick } = require('lodash')
+
+const FormWizardController = require('../../../common/controllers/form-wizard')
 const moveService = require('../../../common/services/move')
 
-module.exports = {
-  post: async (req, res, next) => {
-    try {
-      const { move } = res.locals
+class CancelController extends FormWizardController {
+  async successHandler(req, res, next) {
+    const { id: moveId, person } = res.locals.move
 
-      await moveService.cancel(move.id)
+    try {
+      const data = pick(
+        req.sessionModel.toJSON(),
+        Object.keys(req.form.options.allFields)
+      )
+
+      await moveService.cancel(moveId, {
+        reason: data.cancellation_reason,
+        comment: data.cancellation_reason_comment,
+      })
+
+      req.journeyModel.reset()
+      req.sessionModel.reset()
 
       req.flash('success', {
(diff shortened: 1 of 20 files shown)


teneightfive:
@Andy-Millar at the moment we don't have a way to nest other types of HTML inside our form wizard so I couldn't exactly replicate this:

image

@Andy-Millar / @lmoney-moj what about having that at the bottom before the cancel button to warn people that they still need to inform the supplier if they are cancelling the move request?

teneightfive:
@lmoney-moj as we don't have full support for using detainee/prisoner/young person yet I added in the persons name to the bottom warning text rather than "detainee", what do you think about this? Should we stick to "detainee"?

So, this:
image

instead of this:

image

lmoney-moj:
> @Andy-Millar at the moment we don't have a way to nest other types of HTML inside our form wizard so I couldn't exactly replicate this:
> 
> image
> 
> @Andy-Millar / @lmoney-moj what about having that at the bottom before the cancel button to warn people that they still need to inform the supplier if they are cancelling the move request?

Yes - but then we potentially have x2 lines before the button, both quite important.

Andy-Millar:
Just realised that you COULDN'T nest the warning message! I think it's fine as text

config/i18n.js
```diff
@@ -21,6 +21,15 @@ i18next.use(Backend).init({
   backend: {
     loadPath: './locales/{{lng}}/{{ns}}.json',
   },
+  interpolation: {
+    formatSeparator: '|',
+    format: function(value, format) {
```
feedmypixel: As discussed maybe this would be best doing it in the view layer before it hits the `i18n`. So in nunjucks use 
```
{{ t("moves::confirmation.page_title", { name: fullname | upper }) }}
```

teneightfive: Have dropped this commit altogether as after discussing I don't think we need formatting here too.

PR #211, opened 2019-08-28
[P4-629] Set persons fullname in shared middleware

teneightfive:
This extracts the logic for setting the persons fullname into the
shared middleware so that it can be used by all views that deal
with a person on a move model.

### DIFF
--- a/common/presenters/move-to-card-component.js
+++ b/common/presenters/move-to-card-component.js
@@ -1,5 +1,4 @@
 const assessmentToTagList = require('./assessment-to-tag-list')
-const personService = require('../services/person')
 const i18n = require('../../config/i18n')
 const filters = require('../../config/nunjucks/filters')
 
@@ -35,7 +34,7 @@ module.exports = function moveToCardComponent({ id, reference, person }) {
   return {
     href: `/move/${id}`,
     title: {
-      text: personService.getFullname(person).toUpperCase(),
+      text: person.fullname ? person.fullname.toUpperCase() : '',
     },
     caption: {
       text: i18n.t('moves::move_reference', {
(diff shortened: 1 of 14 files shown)


common/presenters/move-to-card-component.js
```diff
@@ -35,7 +34,7 @@ module.exports = function moveToCardComponent({ id, reference, person }) {
   return {
     href: `/move/${id}`,
     title: {
-      text: personService.getFullname(person).toUpperCase(),
+      text: person.fullname ? person.fullname.toUpperCase() : '',
```
feedmypixel: are we going to be in a situation where `person.fullname` doesn't exist?

common/services/person.js
```diff
@@ -2,8 +2,11 @@ const { mapValues, uniqBy } = require('lodash')
 
 const apiClient = require('../lib/api-client')
 
-function getFullname({ first_names: firstNames, last_name: lastName }) {
-  return `${lastName}, ${firstNames}`
+function transform(person) {
+  return {
+    ...person,
+    fullname: `${person.last_name}, ${person.first_names}`,
```
feedmypixel: is the comma in between the first and last name intentional?

teneightfive: Yes, so that's the way our app will always display a "persons" name in this format. 

It was one that always surprised me during research but our users are used to seeing a persons last name first in all their other systems.

common/presenters/move-to-card-component.js
teneightfive: In theory, not anymore that is has been added to the Move service.

We may have to transform it if we extend the Person service to `GET` people.

My ideal place for this would have been in some transformation layer in the model but the devour library doesn't support that.

PR #212, opened 2019-08-29
Tidy up query string method

feedmypixel:
A little bit of a tidyup/refactor around the query string method

### DIFF
--- a/common/lib/request.js
+++ b/common/lib/request.js
@@ -1,8 +1,9 @@
 const queryString = require('query-string')
 
+function getQueryString(target, source) {
+  return `?${queryString.stringify({ ...target, ...source })}`
+}
+
 module.exports = {
-  getQueryString(target, source) {
-    const merged = Object.assign({}, target, source)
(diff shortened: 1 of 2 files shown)


common/lib/request.js
```diff
@@ -1,8 +1,8 @@
...
 
+const getQueryString = (target, source) =>
+  `?${queryString.stringify({ ...target, ...source })}`
+
 module.exports = {
-  getQueryString(target, source) {
-    const merged = Object.assign({}, target, source)
-    return `?${queryString.stringify(merged)}`
-  },
+  getQueryString,
```
teneightfive: I quite like this shorthand style when there's only 1/2 methods being exported, but that's personal preference I suppose. I'll let you decide which is best :)

feedmypixel: > I quite like this shorthand style when there's only 1/2 methods being exported, but that's personal preference I suppose. I'll let you decide which is best :)

I hear yah - changed to a standard function 🙌

PR #214, opened 2019-08-30
Refactor API model definitions

teneightfive:
This change exports the model objects rather than a function that
defines each model on the API client.

This will allow the model and its attributes to be used in other
parts of the app like services to ensure when formatting they follow
the same property structure as the API model.

### DIFF
--- a/common/lib/api-client/json-api.js
+++ b/common/lib/api-client/json-api.js
@@ -3,7 +3,7 @@ const JsonApi = require('devour-client')
 const { API, IS_DEV } = require('../../../config')
 const { errors, requestTimeout } = require('./middleware')
 const { devourAuthMiddleware } = require('./middleware/auth')
-const defineModels = require('./models')
+const models = require('./models')
 
 const jsonApi = new JsonApi({
   apiUrl: API.BASE_URL,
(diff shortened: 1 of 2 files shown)


common/lib/api-client/json-api.js
```diff
@@ -14,6 +14,9 @@ jsonApi.replaceMiddleware('errors', errors)
 jsonApi.insertMiddlewareBefore('axios-request', requestTimeout(API.TIMEOUT))
 jsonApi.insertMiddlewareBefore('axios-request', devourAuthMiddleware)
 
-defineModels(jsonApi)
+// define models
+Object.entries(models).forEach(function([modelName, model]) {
```
feedmypixel: an arrow function would work well here

common/lib/api-client/models.js
```diff
@@ -1,74 +1,71 @@
...
-    move_type: '',
-    additional_information: '',
-    updated_at: '',
-    time_due: '',
-    date: '',
-    person: {
-      jsonApi: 'hasOne',
-      type: 'people',
+module.exports = {
+  move: {
```
feedmypixel: maybe just preference but how come these are being set to empty strings arther than say `null`?

teneightfive: This was just following the [devour docs](https://github.com/twg/devour#quick-start).

teneightfive:
> Looks good. Looking over the `api-client` folder do we need some tests in around this `model` work?

Yep, this has been my concern and one of the reasons for this refactor. I've added some tests in #215 but also been thinking about it again this weekend of how to improve the unit tests around the API client and the services to be a bit more robust and isolated.

PR #215, opened 2019-08-30
Refactor api client

teneightfive:
This change starts to refactor the API client so that it can be tested better and that
it checks the fixtures against the model definitions to ensure they matches.

It also begins to address issue #82.


common/lib/api-client/json-api.test.js
```diff
@@ -40,4 +46,92 @@ describe('Back-end API client', function() {
       expect(apiMock.isDone()).to.be.true
     })
   })
+
+  describe('Models', function() {
+    const testCases = {
```
feedmypixel: 👍

test/fixtures/api-client/assessment_question.findAll.json
```diff
@@ -0,0 +1,184 @@
+{
```
feedmypixel: there's a mix of case in the filename here is this correct? `api-client/assessment_question.findAll.json`

test/fixtures/api-client/ethnicity.findAll.json
```diff
@@ -0,0 +1,46 @@
+{
```
feedmypixel: there's a mix of case in the filename here is this correct? `api-client/ethnicity.findAll.json`

test/fixtures/api-client/gender.findAll.json
```diff
@@ -0,0 +1,35 @@
+{
```
feedmypixel: there's a mix of case in the filename here is this correct? `api-client/gender.findAll.json`

common/lib/api-client/json-api.test.js
```diff
@@ -40,4 +46,92 @@ describe('Back-end API client', function() {
...
+            })
+
+            it('should contain correct attributes', async function() {
+              const { data } = await jsonApi[testCase.method](
+                modelName,
+                testCase.args
+              )
+
+              if (Array.isArray(data)) {
+                data.forEach(item => checkProperties(item, model.attributes))
```
feedmypixel: could just do 
```
flatten([data]).forEach(item => checkProperties(item, model.attributes))
```

But maybe thats a little magic!

test/fixtures/api-client/assessment_question.findAll.json
teneightfive: So the thought was to try and dynamically build these in the test cases. It's currently using this:

`fixtures/api-client/{model name}/{devour client method}.json`

I suppose the mix in case is because with the model names we're trying to match the API paths (which use snakecase) and then match the case used by devour for its methods in the second part (camel case).

Maybe we should always stick to using kebab case for filenames and then we can translate to the relevant case in this method.

But then I don't know if that hides the fact that the ultimately these file names need to be correct as they are dynamically looked up.

teneightfive: Could add a README to the fixtures folder?

test/fixtures/api-client/ethnicity.findAll.json
teneightfive: See above. These all relate to the same problem so going to resolve these ones and continue the discussion on the first file.

test/fixtures/api-client/gender.findAll.json
teneightfive: See above. These all relate to the same problem so going to resolve these ones and continue the discussion on the first file.

common/lib/api-client/json-api.test.js
teneightfive: Ooo, think I prefer that, reads shorted. Could use spread too...

test/fixtures/api-client/assessment_question.findAll.json
feedmypixel: > Could add a README to the fixtures folder?

👍 yeah I think this is a good shout to explain what you just have above

PR #217, opened 2019-09-02
[P4-627] Add completed page create journey

feedmypixel:
## Feature
Add a Move completed page to the create journey

### Of Note
- If a user uses the browser back button whilst on the confirmation page it will redirect them to `/moves`
- Users can go to `move/<move_id>/confirmation` to see individual confirmations

### Screenshot
screencapture-1

### Journey
confirmation-2

### DIFF
--- a/app/move/controllers/confirmation.js
+++ b/app/move/controllers/confirmation.js
@@ -1,0 +1,14 @@
+function confirmation(req, res) {
+  const { move_type: moveType, to_location: toLocation } = res.locals.move
+
+  const locals = {
+    location:
+      moveType === 'prison_recall'
+        ? req.t('fields::move_type.items.prison_recall.label')
+        : toLocation.title,
+  }
+
+  res.render('move/views/confirmation', locals)
+}
+
+module.exports = confirmation
--- a/app/move/controllers/confirmation.test.js
+++ b/app/move/controllers/confirmation.test.js
@@ -1,0 +1,77 @@
+const controller = require('./confirmation')
+
+const mockMove = {
+  move_type: 'court_appearance',
+  to_location: {
+    title: 'Axminster Crown Court',
+  },
+}
+
+describe('Move controllers', function() {
+  describe('#confirmation()', function() {
+    describe('with move_type "court_appearance"', function() {
+      let req, res
+
+      beforeEach(function() {
+        req = {}
+        res = {
+          render: sinon.spy(),
+          locals: {
+            move: mockMove,
+          },
+        }
+
+        controller(req, res)
+      })
+
+      it('should render confirmation template', function() {
+        const template = res.render.args[0][0]
+
+        expect(res.render.calledOnce).to.be.true
+        expect(template).to.equal('move/views/confirmation')
+      })
+
+      it('should contain a location param', function() {
+        const params = res.render.args[0][1]
+        expect(params).to.have.property('location')
+        expect(params.location).to.equal(mockMove.to_location.title)
+      })
+    })
+
+    describe('with move_type "prison_recall"', function() {
+      let req, res
+
+      beforeEach(function() {
+        req = {
+          t: sinon.stub().returnsArg(0),
+        }
+        res = {
+          render: sinon.spy(),
+          locals: {
+            move: {
+              ...mockMove,
+              move_type: 'prison_recall',
+            },
+          },
+        }
+
+        controller(req, res)
+      })
+
+      it('should render confirmation template', function() {
+        const template = res.render.args[0][0]
+
+        expect(res.render.calledOnce).to.be.true
+        expect(template).to.equal('move/views/confirmation')
+      })
+
+      it('should pass correct values to location translation', function() {
+        const params = res.render.args[0][1]
+        expect(params).to.have.property('location')
+        expect(req.t.firstCall).to.have.been.calledWithExactly(
+          'fields::move_type.items.prison_recall.label'
+        )
+      })
+    })
+  })
+})
--- a/app/move/controllers/create/save.js
+++ b/app/move/controllers/create/save.js
@@ -1,10 +1,8 @@
 const { omit } = require('lodash')
 
 const CreateBaseController = require('./base')
-const { mountpath: movesUrl } = require('../../../moves')
 const moveService = require('../../../../common/services/move')
 const personService = require('../../../../common/services/person')
-const filters = require('../../../../config/nunjucks/filters')
 
 class SaveController extends CreateBaseController {
   async saveValues(req, res, next) {
@@ -14,7 +12,6 @@ class SaveController extends CreateBaseController {
         'errors',
         'errorValues',
       ])
-
       const move = await moveService.create(data)
       await personService.update(data.person)
 
@@ -27,31 +24,12 @@ class SaveController extends CreateBaseController {
   }
 
   successHandler(req, res) {
-    const {
-      date,
-      person,
-      move_type: moveType,
-      to_location: toLocation,
-    } = req.sessionModel.get('move')
+    const { id } = req.sessionModel.get('move')
 
     req.journeyModel.reset()
-    req.journeyModel.destroy()
     req.sessionModel.reset()
-    req.sessionModel.destroy()
 
-    req.flash('success', {
-      title: req.t('messages::create_move.success.title'),
-      content: req.t('messages::create_move.success.content', {
-        name: person.fullname,
-        location:
-          moveType === 'prison_recall'
-            ? req.t('fields::move_type.items.prison_recall.label')
-            : toLocation.title,
-        date: filters.formatDateWithDay(date),
-      }),
-    })
-
-    res.redirect(`${movesUrl}?move-date=${date}`)
+    res.redirect(`/move/${id}/confirmation`)
   }
 }
 
--- a/app/move/controllers/create/save.test.js
+++ b/app/move/controllers/create/save.test.js
@@ -126,14 +126,10 @@ describe('Move controllers', function() {
           sessionModel: {
             get: sinon.stub(),
             reset: sinon.stub(),
-            destroy: sinon.stub(),
           },
           journeyModel: {
             reset: sinon.stub(),
-            destroy: sinon.stub(),
           },
-          flash: sinon.stub(),
-          t: sinon.stub().returnsArg(0),
         }
         res = {
           redirect: sinon.stub(),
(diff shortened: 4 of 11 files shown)


app/move/controllers/create/index.js
```diff
@@ -1,13 +1,15 @@
 const Assessment = require('./assessment')
 const Base = require('./base')
 const MoveDetails = require('./move-details')
+const MoveCompleted = require('./move-completed')
```
teneightfive: Wonder if it would be better to call it a _confirmation_ page rather than _completed_ in hindsight.

We will be using the term "completed" for a different action at some point - when the move from A to B has been fully completed.

This is more of a confirmation of the _request_ being booked, which annoyingly also doesn't guarantee it because the supplier could still choose not to move the person (which is basically the cancellation work I'm doing).

app/move/controllers/create/save.js
```diff
@@ -25,34 +23,6 @@ class SaveController extends CreateBaseController {
       next(error)
     }
   }
-
-  successHandler(req, res) {
```
teneightfive: My only concern with removing this altogether is it brings in the case where a user could navigate back with the browser buttons and see the form fully populated and can submit again, creating a duplicate record.

It also brings in the possibility someone having bookmarked the second step and seeing all the details filled in because it wasn't cleared after saving that move to the database.

We probably want a way to separate these two pieces of information. Maybe we create another model for confirmation or make use of the default `req.session` object to store the information we need for the confirmation page?

Trying to think how we've done this in the past with a confirmation page actually 🤔

feedmypixel: This has been added back Re: the discussion around https://design-system.service.gov.uk/patterns/confirmation-pages/#help-users-who-bookmark-the-page

Have a look over and we can catch up tomorrow

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,65 @@
...
+{% endblock %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      {{ govukPanel({
+        titleText: t('moves::confirmation.panel.heading'),
+        html: t('moves::confirmation.panel.content', {
+          moveReference: moveReference
```
teneightfive: Could use `move.reference` seeing as the move object is already available in the view locals.

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,65 @@
...
+        })
+      }) }}
+
+      <p>{{ t("moves::confirmation.detail", {
+          fullname: fullname,
```
teneightfive: Could use `move.person.fullname` as the move object and therefore person object are already available.

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,65 @@
...
+          location: location,
```
teneightfive: Same with location and date I'd say. Just use what is already available via `res.locals.move` and will simplify the controller a lot too.

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,65 @@
...
+          {{ t("moves::confirmation.cta_heading") }}
+        </h2>
+        <ul class="govuk-list">
+          <li>
+            <a href="{{ action.createNewMoveUrl }}">
+              {{ t("actions::create_new_move") }}
+            </a>
+          </li>
+          <li>
+            {{ t("actions::or") }}
```
teneightfive: I wonder if we need "or" at all. A list signifies _or_ options by default, no?

Or if we want or, maybe it could just be a sentence?

[Create a new more](#) or [return to the dashboard]().

teneightfive: One for @lmoney-moj / @Andy-Millar though.

app/move/controllers/confirmation.js
```diff
@@ -0,0 +1,31 @@
...
+    moveReference: reference,
+    fullname: person.fullname,
+    location:
+      moveType === 'prison_recall'
+        ? req.t('fields::move_type.items.prison_recall.label')
+        : toLocation.title,
+    date: filters.formatDateWithDay(date),
+    action: {
+      createNewMoveUrl: '/move/new',
+      dashboardUrl: `${movesUrl}?move-date=${date}`,
```
teneightfive: There is `res.locals.MOVES_URL` which exists on the locals to take the user back to the dashboard view they started on so this could be used in the view to avoid needing to define in the controller.

app/move/controllers/confirmation.js
```diff
@@ -0,0 +1,31 @@
...
+  const locals = {
+    moveReference: reference,
+    fullname: person.fullname,
+    location:
+      moveType === 'prison_recall'
+        ? req.t('fields::move_type.items.prison_recall.label')
+        : toLocation.title,
+    date: filters.formatDateWithDay(date),
+    action: {
+      createNewMoveUrl: '/move/new',
```
teneightfive: I'd probably move this to the view too as that's what we've done elsewhere for now.

We might want to create some kind of URL/path helper in the future to ensure all our URLs are consistent but for now probably move this to the view.

app/move/controllers/create/assessment.js
```diff
@@ -65,6 +66,11 @@ class AssessmentController extends CreateBaseController {
 
     super.saveValues(req, res, next)
   }
+
+  errorHandler(err, req, res, next) {
```
teneightfive: We might want this at a more global level, like the parent form wizard class.

I think we need a way to handle the `prereq` step missing error in that parent class. And then maybe we just redirect to the start of the journey. I remember doing this elsewhere and will try and dig out the code for it.

common/templates/layouts/base.njk
```diff
@@ -11,6 +11,7 @@
 {% from "govuk/components/select/macro.njk"            import govukSelect %}
 {% from "govuk/components/summary-list/macro.njk"      import govukSummaryList %}
 {% from "govuk/components/textarea/macro.njk"          import govukTextarea %}
+{% from "govuk/components/panel/macro.njk"             import govukPanel %}
```
teneightfive: This commit probably needs to be moved to before the commit where you use this macro otherwise it would throw an error at that point in time :)

common/lib/user.js
```diff
@@ -16,6 +16,7 @@ User.prototype = {
           'move:view',
           'move:create',
           'move:cancel',
+          'move:confirmation',
```
teneightfive: Is this being used for something? I'd probably base this "permission" on whether a user can create a move. If they can create a move, they should be able to see this page, rather than having a separate permission for it.

locales/en/actions.json
```diff
@@ -1,10 +1,13 @@
 {
   "create_move": "Create a move",
+  "create_new_move": "Create another move",
   "previous_day": "Previous day",
   "next_day": "Next day",
   "current_day": "Today",
+  "or": "Or",
   "back": "Back",
   "back_to_dashboard": "$t(actions::back) to dashboard",
+  "return_to_dashboard": "Return to dashboard",
```
teneightfive: Maybe we should use the key above so we're not creating another term for the same action?

locales/en/moves.json
```diff
@@ -80,5 +80,14 @@
...
       "warning": "This will delete this move request."
     }
+  },
+  "confirmation": {
+    "page_title": "Move confirmation for {{name}}",
+    "panel": {
+      "heading": "Move scheduled",
+      "content": "Move reference<br><strong>{{moveReference}}</strong>"
+    },
+    "detail": "The move for <strong>{{fullname}}</strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled",
```
teneightfive: `name` instead of `fullname` to be consistent with above and elsewhere? The translation doesn't care what part of a name it is, that's more the concern of what is sending it to the translation.

locales/en/moves.json
```diff
@@ -80,5 +80,14 @@
       "confirmation": "Are you sure you want to cancel the move for <strong>{{name}}</strong> to <strong>{{location}}</strong>?",
       "warning": "This will delete this move request."
     }
+  },
+  "confirmation": {
+    "page_title": "Move confirmation for {{name}}",
+    "panel": {
+      "heading": "Move scheduled",
+      "content": "Move reference<br><strong>{{moveReference}}</strong>"
```
teneightfive: @lmoney-moj Could this just be `Reference number` now as we've used the term "Move" already in this panel?

locales/en/moves.json
```diff
@@ -80,5 +80,14 @@
       "confirmation": "Are you sure you want to cancel the move for <strong>{{name}}</strong> to <strong>{{location}}</strong>?",
       "warning": "This will delete this move request."
     }
+  },
+  "confirmation": {
+    "page_title": "Move confirmation for {{name}}",
```
teneightfive: Maybe we should use the same term as the heading here: `Move scheduled for {{name}}`.

"Confirmation" is more an internal name we use for this step so better to not surface another term to the user in the page content if we can help it.

app/move/views/confirmation.njk
feedmypixel: spoken with @lmoney-moj and have updated the work as Per above

app/move/controllers/create/assessment.js
feedmypixel: done - covered in https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/223

app/move/index.js
```diff
@@ -32,6 +32,9 @@ router
   .route('/:moveId/cancel')
   .get(protectRoute('move:cancel'), cancel.get)
   .post(protectRoute('move:cancel'), cancel.post)
+router
+  .route('/:moveId/confirmation')
```
teneightfive: Can probably use shorthand get:

```js
router.get('/:moveId/confirmation', protectRoute('move:create'), confirmation)
```

The above was using `.route()` because it was binding both `GET` and `POST`.

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,49 @@
+{% extends "layouts/base.njk" %}
+
+{% block customGtagConfig %}
+  gtag('set', {'page_title': 'Move completed'});
```
teneightfive: "Move scheduled"?

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,49 @@
...
+{% endblock %}
+
+{% block pageTitle %}
+  {{ t("moves::confirmation.page_title", { name: fullname | upper }) }}
```
teneightfive: Should this be `move.person.fullname`?

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,49 @@
...
+{% endblock %}
+
+{% block beforeContent %}
```
teneightfive: As discussed earlier, maybe we should override this block for this template as it's not that useful and creates two links to the same destination.

app/move/views/confirmation.njk
```diff
@@ -0,0 +1,49 @@
...
+      {{ govukPanel({
+        classes: "govuk-!-margin-bottom-7",
+        titleText: t("moves::confirmation.panel.heading"),
+        html: t("moves::confirmation.panel.content", {
+          moveReference: move.reference
+        })
+      }) }}
+
+      <p>{{ t("moves::confirmation.detail", {
+          name: move.person.fullname,
```
teneightfive: To be consistent we should uppercase the name.

app/move/controllers/confirmation.test.js
```diff
@@ -0,0 +1,74 @@
+const controller = require('./confirmation')
+
+const {
+  data: mockMove,
+} = require('../../../test/fixtures/api-client/move.get.deserialized.json')
```
teneightfive: This has been deprecated and refactor in #221. 

I think it's better to just mock the mock object with the properties that are required by this file rather than pull in the whole fixture. Have a look at the commits in #221 for what I've done elsewhere.

app/move/controllers/create/save.test.js
```diff
@@ -132,8 +132,6 @@ describe('Move controllers', function() {
             reset: sinon.stub(),
             destroy: sinon.stub(),
```
teneightfive: Can also remove the `.destroy()` stubs.

locales/en/moves.json
```diff
@@ -80,5 +80,14 @@
...
       "warning": "This will delete this move request."
     }
+  },
+  "confirmation": {
+    "page_title": "Move scheduled for {{name}}",
+    "panel": {
+      "heading": "Move scheduled",
+      "content": "Reference number<br><strong>{{moveReference}}</strong>"
+    },
+    "detail": "The move for <strong>{{name}}</strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled",
```
teneightfive: As discussed earlier, maybe we should make the name a link as otherwise there is no way to view the move that was created from this page.

locales/en/moves.json
```diff
@@ -80,5 +80,14 @@
...
     }
+  },
+  "confirmation": {
+    "page_title": "Move scheduled for {{name}}",
+    "panel": {
+      "heading": "Move scheduled",
+      "content": "Reference number<br><strong>{{moveReference}}</strong>"
+    },
+    "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled",
+    "cta_heading": "What to do next"
```
teneightfive: Think we can drop this as it's not being used anymore, right?

PR #225, opened 2019-09-04
Remove need for config item default

### LINKED ISSUE #82: Default value for `API.BASE_URL` config is required to run tests

teneightfive:
Previously we needed to set a default value for `API.BASE_URL` as tests
were relying on this value.

Those tests have now been refactor so that this default can be removed
safely.

This change also removes the config test which currently relies on
the default value. This will throw up a reduction in test coverage
for the config file but this should be addressed differently rather
than relying on environment variables being set in test environments.

This changes means all unit tests can run in any environment without
requiring environment variables to be set.

Fixes #82


teneightfive:
> This pull request will bring the total coverage in the repository to **99.6%** (**-0.1%** change).

This is acceptable at this stage as we don't really want to be testing logic in a configuration file. We need to refactor this to move the logic elsewhere at a later stage.

PR #226, opened 2019-09-04
[P4-636] Display cancellation banner on move detail page

teneightfive:
If a move has been cancelled it will display a banner to signify
that this move is in a cancelled state.

The banner can be used in the future to show the status of a move and any further
information like reason for cancellation.

🚀 **[DEMO](https://book-a-secure-move.herokuapp.com/)** 🚀 

## What it looks like

localhost_3000_move_a53fb8f7-3e34-41ca-8f7d-2b97ae89f785

### DIFF
--- a/app/move/controllers/view.js
+++ b/app/move/controllers/view.js
@@ -2,7 +2,11 @@ const presenters = require('../../../common/presenters')
 
 module.exports = function view(req, res) {
   const { move } = res.locals
-  const { person } = move
+  const {
+    person,
+    cancellation_reason: cancellationReason,
+    cancellation_reason_comments: cancellationComments,
+  } = move
   const locals = {
     moveSummary: presenters.moveToMetaListComponent(move),
     personalDetailsSummary: presenters.personToSummaryListComponent(person),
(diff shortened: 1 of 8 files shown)


app/move/controllers/view.js
```diff
@@ -2,7 +2,11 @@ const presenters = require('../../../common/presenters')
 
 module.exports = function view(req, res) {
   const { move } = res.locals
-  const { person } = move
+  const {
+    person,
+    cancellation_reason: reason,
```
feedmypixel: could this benefit from being `cancelReason`?

app/move/controllers/view.js
```diff
@@ -2,7 +2,11 @@ const presenters = require('../../../common/presenters')
...
+    cancellation_reason_comments: reasonComments,
```
feedmypixel: could this benefit from being `cancelReasonComments`?

common/components/message/_message.scss
```diff
@@ -87,3 +93,22 @@
   border-width: 0 0 0 5px;
   background-color: govuk-colour("light-grey");
 }
+
+.app-message--temporary {
+  background-color: govuk-colour("blue");
+  border-width: 0;
+  color: govuk-colour("white");
+  padding: $_message-padding + 5;
```
feedmypixel: should we be aligning white space and not using magic numbers?

teneightfive: Do you mean the variable or the `5`? The 5 is the border width. 

Maybe that could be a variable so it's easier to understand.

feedmypixel: > Maybe that could be a variable so it's easier to understand.

This sounds good

PR #227, opened 2019-09-04
[P4-497] Create Move right hand summary panel

feedmypixel:
## Feature
Add right-hand summary panel to the create journey

### Screenshots
sidebar-move-2

### DIFF
--- a/app/move/controllers/create/base.js
+++ b/app/move/controllers/create/base.js
@@ -1,4 +1,5 @@
 const FormWizardController = require('../../../../common/controllers/form-wizard')
+const presenters = require('../../../../common/presenters')
 
 class CreateBaseController extends FormWizardController {
   middlewareChecks() {
@@ -9,6 +10,7 @@ class CreateBaseController extends FormWizardController {
   middlewareLocals() {
     super.middlewareLocals()
     this.use(this.setCancelUrl)
+    this.use(this.setMoveSummary)
   }
 
   setCancelUrl(req, res, next) {
@@ -21,6 +23,20 @@ class CreateBaseController extends FormWizardController {
       const error = new Error('Current location is not set in session.')
       return next(error)
     }
+
+    next()
+  }
+
+  setMoveSummary(req, res, next) {
+    const currentLocation = req.session.currentLocation
+    const sessionModel = req.sessionModel.toJSON()
+    const moveSummary = presenters.moveToMetaListComponent({
+      ...sessionModel,
+      from_location: currentLocation,
+    })
+
+    res.locals.person = sessionModel.person
+    res.locals.moveSummary = sessionModel.move_type ? moveSummary : {}
 
     next()
   }
--- a/common/templates/form-wizard.njk
+++ b/common/templates/form-wizard.njk
@@ -1,4 +1,6 @@
-{% extends "layouts/base.njk" %}
+{% extends "layouts/two-column.njk" %}
+
+{% set additionalContainerClass = "sticky-sidebar-container" %}
 
 {% block pageTitle %}
   {{ t(options.pageTitle) }} - {{ t(options.journeyPageTitle) }}
@@ -15,7 +17,7 @@
   {% endif %}
 {% endblock %}
 
-{% block content %}
+{% block contentHeader %}
   {% if errors.errorList.length %}
     {{ govukErrorSummary({
       titleText: t("validation::summary.heading"),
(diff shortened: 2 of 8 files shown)


app/move/controllers/create/base.js
```diff
@@ -24,6 +27,48 @@ class CreateBaseController extends FormWizardController {
 
     next()
   }
+
+  async locals(req, res, callback) {
```
feedmypixel: @teneightfive I am questioning `middlewareLocals` is prefered over this use of `HMPO` `locals` method?

teneightfive: I think if you're adding more locals I ended up preferring extending the `middlewareLocals`. You can break down the things you're adding to be a little more descriptive.

I think the `locals` method is useful if you want to override the locals that are sent to the view.

app/move/controllers/create/base.js
```diff
@@ -24,6 +27,48 @@ class CreateBaseController extends FormWizardController {
...
+      } = req.sessionModel.toJSON()
+      let toLocationDetail
+
+      if (toLocation) {
+        toLocationDetail = await referenceDataService.getLocationById(
+          toLocation
+        )
+      }
+
+      const moveSummary = presenters.moveToMetaListComponent({
```
teneightfive: I'd be tempted to pass the session model in here as the field names should match the expected values of the presenter?

Saves re-writing it twice.

You will need to override the `to_location` property though to be what the reference data service returns. But that could be done using spread.

app/move/controllers/create/base.js
```diff
@@ -24,6 +27,48 @@ class CreateBaseController extends FormWizardController {
...
+        date,
+        time_due: timeDue,
+        move_type: moveType,
+        additional_information: additionalInfo,
+        from_location: currentLocation,
+        to_location: toLocationDetail,
+      })
+      const person = personService.transform({
```
teneightfive: I think there is a `person` property on the session model that is stored after the first step. You could pass that straight in here instead?

common/templates/layouts/two-column.njk
```diff
@@ -0,0 +1,24 @@
...
+
+{% set containerClass = ["govuk-grid-row"].concat(additionalContainerClass) | join(" ") %}
+
+{% block content %}
+
+  {% block contentHeader %}{% endblock %}
+
+  <div class="{{ containerClass }}">
+
+    <article class="govuk-grid-column-two-thirds">
```
teneightfive: Haven't read up on this in a while but I wonder whether we actually need to use `<article>` and `<aside>` elements.

Having a poke around GOV.UK they haven't seemed to use them for article style pages. Don't know if its something that has phased out?

teneightfive:
Only other thing that looked a little weird in the recording was the panel appearing when you don't have a persons name. Maybe it shouldn't appear until we have saved the person?

@Andy-Millar what do you think? 

Maybe that's a problem with the summary panel order. It's hard to build it up as the first thing we actually know is the from location. And with these kind of step summaries things it normally appends each step as you progress. But in our case the name then jumps to the top and then we start appending the information.

Andy-Millar:
@feedmypixel @teneightfive The panel shouldn't be visible on personal details page (because we don't know anything yet!)

On the move details page, name and move from should be visible. From the next page the rest of the info is visible.

Hope this makes sense!

common/templates/layouts/two-column.njk
feedmypixel: So having a read over [html doctor](http://html5doctor.com/aside-revisited/) they are still relevant. I can change them to `<div>` if you like?

app/move/controllers/create/base.js
feedmypixel: Yeah agreed - it sits much nicer. I'm also not a fan of the way the `HMPO` locals works with a callback

common/templates/layouts/two-column.njk
feedmypixel: This may also be due to support right? I see we aren't using `html5shiv` anywhere also for these elments

feedmypixel:
> On the move details page, name and move from should be visible. From the next page the rest of the info is visible.
> 
> Hope this makes sense!

@Andy-Millar sone as described 👆I have also updated the gif in the description to show the revised implementation. Shout if this isn't working for you.

common/templates/layouts/two-column.njk
teneightfive: Ah yes, maybe lets stick to divs for now then.

Andy-Millar:
@feedmypixel I don't think we need 'To: Unknown' on the move details page. Otherwise all good to me

app/move/controllers/create/base.js
```diff
@@ -24,6 +28,39 @@ class CreateBaseController extends FormWizardController {
...
+      }
+
+      const moveSummary = presenters.moveToMetaListComponent({
+        ...sessionData,
+        from_location: currentLocation,
+        to_location: toLocationDetail,
+      })
+
+      res.locals = {
+        ...res.locals,
```
teneightfive: I have a vague memory of reading somewhere that this isn't good. But that might have been `req.session`...

Could the person local also be this:

```js
res.locals.person = personService.transform(person)
```

Saves having to translate the restructured property and feels less to follow.

app/move/controllers/create/base.js
```diff
@@ -24,6 +28,39 @@ class CreateBaseController extends FormWizardController {
...
+        person,
+        moveSummary,
+      }
+
+      return next()
```
teneightfive: Don't need to `return` this. The other block won't be run.

app/move/controllers/create/base.js
```diff
@@ -24,6 +28,39 @@ class CreateBaseController extends FormWizardController {
 
     next()
   }
+
+  async setSummaryData(req, res, next) {
```
teneightfive: `setMoveSummary`? Feels more descriptive of what it's setting. _Data_ is a bit ambiguous.

app/move/controllers/create/base.js
teneightfive: It also feels wrong to need to use the transform method here.

Ideally it would be better to keep this transform function to the services which keeps it closer to the ideal place we wanted it - the API client.

Just looking at the person service and it should probably be called in the `create` and `update` methods there.

I'm happy to extend the change I made to this so that you can just re-use the person object that already exists.

common/templates/form-wizard.njk
```diff
@@ -30,35 +32,47 @@
...
+        <a href="{{ cancelUrl }}" class="govuk-button govuk-button--text">
+          {{ t("actions::cancel") }}
+        </a>
+      {% endif %}
+    {% endblock %}
+  </form>
+{% endblock %}
+
+{% block aside %}
+  {% if person.fullname %}
```
teneightfive: Could just check if there is a `person`. Feels like it reads nicer.

app/move/controllers/create/base.js
teneightfive: Have added [231](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/231) to ensure the person transformation always happens when interacting with the service so should mean you can simplify this a bit more.

app/move/controllers/create/base.js
```diff
@@ -24,6 +27,39 @@ class CreateBaseController extends FormWizardController {
...
+      })
+
+      res.locals = {
+        ...res.locals,
+        person,
+        moveSummary,
+      }
+
+      next()
+    } catch (err) {
```
teneightfive: `error` instead of `err`? Everywhere else uses _error_ I believe and have started to favour this as it just reads a little nicer too.

Not a show-stopper though.

teneightfive: I think the other benefit of `error` is that eslint will tell you off if we don't handle it correctly.

feedmypixel:
> @feedmypixel I don't think we need 'To: Unknown' on the move details page. Otherwise all good to me

@Andy-Millar FYI I've removed any move information from the `move-details` page. Gif updated

Andy-Millar:
@feedmypixel Ahhh - I didn't mean remove 'move from' though - that information is static, so fine to show

feedmypixel:
> @feedmypixel Ahhh - I didn't mean remove 'move from' though - that information is static, so fine to show

@Andy-Millar  The reason I have removed all move info is the `Move from` is already on the left. Its also a little tricky (although not impossible) showing `from` but not `to`.
Let's catch up tomorrow and tweak it so it feels right for you?
 
Screenshot 2019-09-09 at 16 33 45

PR #229, opened 2019-09-06
Fix for locations

feedmypixel:
### Fix

Previously the call to `referenceDataService.getLocations` was bringing back all `location_type` rather than a specific type. This work fixes that.

## Of Note
This work also removes the call to get `prison` locations as the form on this page no longer requires them

### DIFF
--- a/app/move/controllers/create/move-details.js
+++ b/app/move/controllers/create/move-details.js
@@ -15,14 +15,6 @@ class MoveDetailsController extends CreateBaseController {
           'filter[location_type]': 'court',
         },
       })
-      const prisons = await referenceDataService.getLocations({
-        type: 'prison',
-      })
-
-      const {
-        to_location_prison_recall: toLocationPrison,
-        date_type: dateType,
-      } = req.form.options.fields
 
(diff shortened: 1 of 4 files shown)


app/move/controllers/create/move-details.js
```diff
@@ -9,30 +10,24 @@ const referenceDataHelpers = require('../../../../common/helpers/reference-data'
 class MoveDetailsController extends CreateBaseController {
   async configure(req, res, next) {
     try {
-      const courts = await referenceDataService.getLocations({
-        type: 'court',
+      const courtLocations = await referenceDataService.getLocations({
+        filter: {
+          'filter[location_type]': 'court',
```
teneightfive: I’d be tempted to create a method in the service like we talked about to abstract this logic. 

This syntax is specific to the API so it would be better to have it contained within one service rather than living in this controller.

feedmypixel: Totally agreed I wanted to get this simple fix up so I can finish the sidebar work and then do the abstraction we spoke about in a separate PR connected with the sidebar work.

PR #236, opened 2019-09-10
[P4-437] - Use translations for assessment questions

feedmypixel:
## Feature

If translations are available in our locales file then use these in the application. This means we can provide:
- form options `hint`locale translations
- override `title` text sent from the `BE` `API` with `label` locale translations

### Before
Screenshot 2019-09-10 at 17 38 52

### After
Screenshot 2019-09-10 at 17 38 24

### DIFF
--- a/common/helpers/field.js
+++ b/common/helpers/field.js
@@ -1,11 +1,16 @@
 const { cloneDeep, fromPairs, get, set } = require('lodash')
 
 const componentService = require('../services/component')
+const i18n = require('../../config/i18n')
 
-function mapReferenceDataToOption({ id, title, key, conditional }) {
+function mapReferenceDataToOption({ id, title, key, conditional, hint }) {
   const option = {
     value: id,
     text: title,
+  }
+
+  if (hint) {
+    option.hint = { text: hint }
   }
 
   if (key) {
@@ -17,6 +22,18 @@ function mapReferenceDataToOption({ id, title, key, conditional }) {
   }
 
   return option
+}
+
+function mapAssessmentQuestionToTranslation(item) {
+  const { category, key, title } = item
+  const hintKey = `fields::${category}.items.${key}.hint`
+  const labelKey = `fields::${category}.items.${key}.label`
+
+  return {
+    ...item,
+    hint: i18n.exists(hintKey) ? hintKey : undefined,
+    title: i18n.exists(labelKey) ? labelKey : title,
+  }
 }
 
 function mapAssessmentQuestionToConditionalField(item) {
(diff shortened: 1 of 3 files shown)


common/helpers/field.js
```diff
@@ -3,6 +3,7 @@ const { cloneDeep, fromPairs, get, set } = require('lodash')
 const componentService = require('../services/component')
 
 function mapReferenceDataToOption({ id, title, key, conditional }) {
+const i18n = require('../../config/i18n')
```
teneightfive: Looks like this might have been accidentally nudged down a couple of lines?

common/helpers/field.js
```diff
@@ -19,6 +20,25 @@ function mapReferenceDataToOption({ id, title, key, conditional }) {
   return option
 }
 
+function mapAssessmentQuestionToTranslation(item) {
+  const { category, key } = item
```
teneightfive: Was about to say "What about destructuring the arguments instead?" but then realised you need to use `item` later on...

common/helpers/field.js
```diff
@@ -1,12 +1,23 @@
...
+  hint,
+  title,
+  key,
+  conditional,
+  label,
+}) {
   const option = {
     value: id,
-    text: title,
+    text: label || title,
```
teneightfive: I wonder if we should leave this logic as it was as the majority of cases that are mapping reference data to an option will still expect to use title.

And then the logic for finding a "label" translation key should be kept to preceding map functions like the one in the previous commit. And that's where we handle the fallback.

common/helpers/field.js
```diff
@@ -19,6 +20,25 @@ function mapReferenceDataToOption({ id, title, key, conditional }) {
   return option
 }
 
+function mapAssessmentQuestionToTranslation(item) {
+  const { category, key } = item
+  const translations = ['hint', 'label']
+  const translationStrings = {}
+
+  translations.forEach(translation => {
```
teneightfive: Wonder if this needs a `forEach`.

Maybe we handle the label and hint individually where in the case of the _label_ not being present we use `title` as the fallback.

teneightfive: What are your thoughts on something like this?

```js
function mapAssessmentQuestionToTranslation(item) {
  const { category, key, title } = item
  const fieldPath = `fields::${category}.items.${key}`
  const titleKey = `${fieldPath}.title`
  const hintKey = `${fieldPath}.hint`

  return {
    ...item,
    title: i18n.exists(titleKey) ? titleKey : title,
    hint: i18n.exists(hintKey) ? hintKey : undefined,
  }
}
```

feedmypixel: > What are your thoughts on something like this?

Sounds good to me. If it's urgent please feel free to take this over. If not I can pick it up when back.

PR #238, opened 2019-09-11
Analytics improvements

teneightfive:
This change adds a few improvement to how we are capturing things using analytics:

- ensure we are capturing basic metrics even when JS is not available
- capture error fields as events

### DIFF
--- a/common/assets/javascripts/analytics.js
+++ b/common/assets/javascripts/analytics.js
@@ -1,0 +1,42 @@
+const { nodeListForEach } = require('./utils')
+
+function Analytics($module) {
+  this.cacheEls($module)
+}
+
+Analytics.prototype = {
+  init: function() {
+    this.cacheEls()
+    this.render()
+  },
+
+  cacheEls: function() {
+    this.$errors = document.querySelectorAll('.govuk-error-summary__list li a')
+  },
+
+  render: function() {
+    nodeListForEach(this.$errors, this.trackError.bind(this))
+  },
+
+  trackError: function(error) {
+    const fieldId = error.href.substring(error.href.indexOf('#') + 1)
+    const errorMessage = error.text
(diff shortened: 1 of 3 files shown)


common/assets/javascripts/analytics.js
```diff
@@ -0,0 +1,42 @@
...
+
+    this.sendEvent({
+      action: fieldId,
+      category: 'validation_error',
+      label: errorMessage,
+    })
+  },
+
+  sendEvent: function({ action, category, label }) {
+    try {
```
feedmypixel: how come this is in a `try/catch`? its just pushing into the `dataLayer`. Are you expecting this to error?

common/templates/layouts/base.njk
```diff
@@ -60,6 +59,15 @@
   <link rel="apple-touch-icon" href="{{ getAssetPath('images/hmpps-apple-touch-icon.png') }}">
 {% endblock %}
 
+{% block bodyStart %}
+  {% if GA_ID %}
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ GA_ID }}"
```
feedmypixel: this feels a little scrunched up would it look clearer like?
```
<noscript>
  <iframe src="https://www.googletagmanager.com/ns.html?id={{ GA_ID }}" height="0" width="0" 
          style="display:none;visibility:hidden">
  </iframe>
</noscript

```

common/assets/javascripts/analytics.js
teneightfive: This is so that this function doesn't error when `gtag` is not available on the window. For example when a Google Analytics ID hasn't been set as an environment variable and GA isn't available.

PR #240, opened 2019-09-13
Extract API client auth

teneightfive:
This change refactors the way the  auth token is retrieved and set to connect to the backend API.

It refactors most of the logic from the middleware into a self-contained library that handles getting the token which makes the middleware a lot simpler and easier to test/mock elsewhere.

### DIFF
--- a/common/lib/api-client/auth.js
+++ b/common/lib/api-client/auth.js
@@ -1,0 +1,58 @@
+const axios = require('axios')
+
+function getTimestamp() {
+  return Math.floor(new Date() / 1000)
+}
+
+function Auth({ timeout = 10000, authUrl, username, password } = {}) {
+  this.accessToken = null
+  this.tokenExpiresAt = null
+  this.config = {
+    timeout,
+    authUrl,
+    username,
+    password,
+  }
+}
+
+Auth.prototype = {
+  async getAccessToken() {
+    if (this.isExpired()) {
+      return this.refreshAccessToken().then(data => {
+        this.accessToken = data.access_token
+        this.tokenExpiresAt = data.expires_in + getTimestamp()
+
+        return this.accessToken
+      })
+    }
+
+    return Promise.resolve(this.accessToken)
+  },
+
+  refreshAccessToken() {
+    const data = {
+      grant_type: 'client_credentials',
+    }
+    const config = {
+      timeout: this.config.timeout,
+      auth: {
+        username: this.config.username,
+        password: this.config.password,
+      },
+    }
+
+    return axios
+      .post(this.config.authUrl, data, config)
+      .then(response => response.data)
+  },
+
+  isExpired() {
+    if (!this.accessToken || !this.tokenExpiresAt) {
+      return true
+    }
+
+    return this.tokenExpiresAt <= getTimestamp()
+  },
+}
+
+module.exports = Auth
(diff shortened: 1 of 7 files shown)


common/lib/api-client/auth.js
```diff
@@ -0,0 +1,58 @@
+const axios = require('axios')
+
+function Auth({ timeout = 10000, authUrl, username, password } = {}) {
```
feedmypixel: do we need to destructure here? Could you just pass `config` through?

common/lib/api-client/auth.js
```diff
@@ -0,0 +1,58 @@
...
+  this.tokenExpiresAt = null
+  this.config = {
+    timeout,
+    authUrl,
+    username,
+    password,
+  }
+}
+
+Auth.prototype.getCurrentTime = function() {
```
feedmypixel: How about `getTimestamp`?

common/lib/api-client/auth.js
```diff
@@ -0,0 +1,58 @@
...
+  if (this.isExpired()) {
+    return this.refreshAccessToken()
+      .then(data => this.setAccessToken(data))
+      .then(() => this.accessToken)
+  }
+
+  return Promise.resolve(this.accessToken)
+}
+
+Auth.prototype.setAccessToken = function(response) {
```
feedmypixel: not too sure but my spidey senses are saying this needs to resolve the `accessToken`

common/lib/api-client/auth.js
teneightfive: I quite like  destructuring when there's only a few options as it helps you see what's expected. It also makes it neater to use so that you don't need to do `options.{property}`.

common/lib/api-client/auth.js
teneightfive: What do you mean? I tried to split the concerns of `get` and `set`. This sets the token on the class, the other method returns it.

teneightfive:
> I'm questioning a little bit if this actually needs to be prototypal class?

It felt a little cleaner as a class and easier to test/mock when the methods were exposed. Most of them need to be class methods as they use the properties from the class. I've pushed a fixup which tidies it up a little bit. Let me know what you think?

common/lib/api-client/auth.js
feedmypixel: it feels like if you also returned the access key when using set then the above (L21) could become 
```
.then(data => this.setAccessToken(data))
.then((accessToken) => accessToken)
```

But that then feels a bit like concerns are being mixed between OO and Functional

feedmypixel:
> I've pushed a fixup which tidies it up a little bit. Let me know what you think?

Looking good. Is it worth going the whole hog and using a class? Not sure what your preference is on this?

```
class Auth {
  constructor({ timeout = 10000, authUrl, username, password } = {}) {
    this.accessToken = null
    this.tokenExpiresAt = null
    this.config = {
      timeout,
      authUrl,
      username,
      password,
    }
  }
...
```

common/lib/api-client/auth.js
teneightfive: Yeah that's true. I suppose the name didn't suggest it was returning anything but I I see your point about functional should always return something.

Could combine with the get method so it gets set there and is also always returned.

PR #241, opened 2019-09-16
[P4-367] Cache reference data from API

teneightfive:
This change allows API models to be cached rather than serving from the API each time.

This is useful for reference data that doesn't change very often.

### Questions

- Is this the best place to cache this data? 
  - Could we use a caching layer in between rather than using the frontend Redis instance?
- What's a reasonable time to cache this data? 
  - 24 hours?
  - 7 days?

### DIFF
--- a/common/lib/api-client/middleware/request.js
+++ b/common/lib/api-client/middleware/request.js
@@ -1,0 +1,46 @@
+const { get } = require('lodash')
+
+const redisStore = require('../../../../config/redis-store')()
+const models = require('../models')
+
+function cacheResponse(key, expiry) {
+  return async response => {
+    await redisStore.client.setexAsync(
+      key,
+      expiry,
+      JSON.stringify(response.data)
+    )
+    return response
+  }
+}
+
+function requestMiddleware(expiry = 60) {
+  return {
+    name: 'axios-request',
+    req: async function req(payload) {
+      const { req, jsonApi } = payload
+      const pathname = new URL(req.url).pathname
+      const searchString = new URLSearchParams(req.params).toString()
+      const key = `cache:${req.method}.${pathname}${
+        searchString ? `?${searchString}` : ''
+      }`
+      const cacheModel = get(models, `${req.model}.options.cache`)
+
+      if (!cacheModel) {
+        return jsonApi.axios(req)
+      }
+
+      return redisStore.client.getAsync(key).then(response => {
+        if (!response) {
+          return jsonApi.axios(req).then(cacheResponse(key, expiry))
+        }
+
+        return {
+          data: JSON.parse(response),
+        }
+      })
+    },
+  }
+}
+
+module.exports = requestMiddleware
--- a/common/lib/api-client/index.js
+++ b/common/lib/api-client/index.js
@@ -1,7 +1,7 @@
 const JsonApi = require('devour-client')
 
 const { API, IS_DEV } = require('../../../config')
-const { auth, errors, requestTimeout } = require('./middleware')
+const { auth, errors, request, requestTimeout } = require('./middleware')
 const models = require('./models')
 
 let instance
@@ -17,6 +17,7 @@ module.exports = function() {
   })
 
   instance.replaceMiddleware('errors', errors)
+  instance.replaceMiddleware('axios-request', request(API.CACHE_EXPIRY))
   instance.insertMiddlewareBefore('axios-request', requestTimeout(API.TIMEOUT))
   instance.insertMiddlewareBefore('axios-request', auth)
 
--- a/common/lib/api-client/index.test.js
+++ b/common/lib/api-client/index.test.js
@@ -7,6 +7,7 @@ const mockConfig = {
   API: {
     BASE_URL: 'http://api.com/v1',
     TIMEOUT: 1000,
+    CACHE_EXPIRY: 5000,
   },
 }
 const mockModels = {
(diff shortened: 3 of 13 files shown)


feedmypixel:
Looking good a few things:

- it feels like we may have a need to have different cache times when using `setex`? Maybe it would be good to allow an override for specific API calls.

- the `bluebird`, `promisifyAll` feels a little overkill considering that we are only using `setx` and `get` in one place. Is extra Promise dependancy really really needed?

- Part of me wants the `cache` work in redis to be stored with the word `cache` at the front so we differentiate between `sess:` and `cache:` data:
Screenshot 2019-09-19 at 10 45 36

teneightfive:
> > Questions
> > Is this the best place to cache this data?
> > Could we use a caching layer in between rather than using the frontend Redis instance?
> 
> Here or on the backend feels like the correct place. What do the backend guys @stevehook @cesidio @ngw think?

Yeah that's why I thought it best to push and get some thoughts before spending too much more time on it.

> 
> > What's a reasonable time to cache this data?
> > 24 hours?
> > 7 days?
> 
> It feels like we need a default and the ability to override this

Yeah I  was thinking of setting this on each model, but using a default value from the config if one isn't set?

> Looking good a few things:
> 
> * it feels like we may have a need to have different cache times when using `setex`? Maybe it would be good to allow an override for specific API calls.
> * the `bluebird`, `promisifyAll` feels a little overkill considering that we are only using `setx` and `get` in one place. Is extra Promise dependancy really really needed?

The reason I used this as it felt neater than manually turning the ones needed into a promise. Basically where they're used is much neater when using promises rather than callbacks.

> * Part of me wants the `cache` work in redis to be stored with the word `cache` at the front so we differentiate between `sess:` and `cache:` data:
> 
> Screenshot 2019-09-19 at 10 45 36

Yep, good shout. Have pushed a change to add this.

common/lib/api-client/middleware/request-cache.js
```diff
@@ -0,0 +1,42 @@
+const { get } = require('lodash')
+
+const { client } = require('../../../../config/redis-store')()
+const models = require('../models')
+
+module.exports = function requestCache(cacheTimeout) {
```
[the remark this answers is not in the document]
feedmypixel: this one would be good to try and reduce - although complex the method does read well as is.

feedmypixel:
> Yeah I was thinking of setting this on each model, but using a default value from the config if one isn't set?

sounds good 👍 

> The reason I used this as it felt neater than manually turning the ones needed into a promise. Basically where they're used is much neater when using promises rather than callbacks.

cool 👍 

> Yep, good shout. Have pushed a change to add this.

👍

common/lib/api-client/middleware/request.js
```diff
@@ -0,0 +1,39 @@
+const { get } = require('lodash')
```
feedmypixel: So complexity wise I have had a little looks at it. There may be bits here that help:

```
const { get } = require('lodash')

const redisStore = require('../../../../config/redis-store')()
const models = require('../models')

function requestMiddleware(expiry = 60) {
  async function req(payload) {
    const { req, jsonApi } = payload
    const pathname = new URL(req.url).pathname
    const searchString = Object.keys(req.params).length
      ? `?${new URLSearchParams(req.params).toString()}`
      : ''
    const key = `cache:${req.method}.${pathname}${searchString}`
    const hasCacheOption = Boolean(
      get(models, `${req.model}.options.cache`, false)
    )

    if (!hasCacheOption) {
      return jsonApi.axios(req)
    }

    return redisStore.client.getAsync(key).then(response => {
      if (response) {
        return { data: JSON.parse(response) }
      }

      return jsonApi.axios(req).then(async response => {
        await redisStore.client.setexAsync(
          key,
          expiry,
          JSON.stringify(response.data)
        )
        return response
      })
    })
  }

  return {
    name: 'axios-request',
    req,
  }
}

module.exports = requestMiddleware
```

common/lib/api-client/middleware/request.js
```diff
@@ -0,0 +1,39 @@
...
+      if (!get(models, `${req.model}.options.cache`)) {
+        return jsonApi.axios(req)
+      }
+
+      const result = await redisStore.client
+        .getAsync(key)
+        .then(response => JSON.parse(response))
+
+      if (result) {
+        return { data: result }
```
feedmypixel: should this also be a returning a Promise? then the return is always a Promise - same as `jsonApi.axios(req)`?

common/lib/api-client/middleware/request.js
```diff
@@ -20,12 +20,11 @@ function requestMiddleware(expiry = 60) {
     req: async function req(payload) {
       const { req, jsonApi } = payload
       const pathname = new URL(req.url).pathname
-      const searchString = Object.keys(req.params).length
-        ? `?${new URLSearchParams(req.params).toString()}`
-        : ''
-      const key = `cache:${req.method}.${pathname}${searchString}`
+      const searchString = new URLSearchParams(req.params).toString()
+      const key = `cache:${req.method}.${pathname}?${searchString}`
```
feedmypixel: the ternary feels like the better option rather than always having a `?`

common/lib/api-client/middleware/request.js
```diff
@@ -21,7 +21,9 @@ function requestMiddleware(expiry = 60) {
       const { req, jsonApi } = payload
       const pathname = new URL(req.url).pathname
       const searchString = new URLSearchParams(req.params).toString()
-      const key = `cache:${req.method}.${pathname}?${searchString}`
+      const key = `cache:${req.method}.${pathname}${
+        searchString ? `?${searchString}` : ''
+      }`
```
feedmypixel: 👍

PR #242, opened 2019-09-17
[P4-685] Filter expired alerts

teneightfive:
This change applies an expired filter to ensure expired alerts or flags are not displayed on the move detail page or the dashboard.

This is so that any flags that are no longer relevant are not displayed.

### DIFF
--- a/common/helpers/reference-data.test.js
+++ b/common/helpers/reference-data.test.js
@@ -1,6 +1,6 @@
 const { addMonths, subMonths } = require('date-fns')
 
-const { filterDisabled } = require('./reference-data')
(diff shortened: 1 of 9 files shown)


common/helpers/reference-data.test.js
```diff
@@ -163,4 +163,48 @@ describe('Reference data helpers', function() {
...
   })
+
+  context('#filterExpired', function() {
+    context('without expired property', function() {
+      it('should return true', function() {
+        expect(filterExpired({})).to.be.true
+      })
+    })
+
+    context('with expired property is present', function() {
```
feedmypixel: should this be `when` not `with`?

PR #246, opened 2019-09-19
Update connect redis dependency

feedmypixel:
Update connect redis dependency. Of note there is an addition of the `redis` package required by the breaking change upgrade. As described [here](https://github.com/tj/connect-redis#api)
 
## Checks app is working ok
- `npm t` passes
- Application works as in you can login and view the dashboard


config/redis-store.js
```diff
@@ -1,6 +1,8 @@
+const redis = require('redis')
 const session = require('express-session')
 const RedisStore = require('connect-redis')(session)
 
+const client = redis.createClient()
```
teneightfive: I think we'd need to pass `options` into this method now as that is what it was doing previously. But `connect-redis` was creating the store in the background.

Have you tried restarting the server to ensure this is working as expected?

I'd imagine we'd have to do something like this:

```js
...
const client = redis.createClient(options)
store = new RedisStore({ client })
...
```

Basically I think the options we're passing in are for the client, not necessarily the store. Might be nicer to destructure them too thinking about it, be clearer as to what's getting used where.

feedmypixel: > Basically I think the options we're passing in are for the client, not necessarily the store.

Agreed - have pushed a fixup

> Might be nicer to destructure them too thinking about it, be clearer as to what's getting used where.

I hear you on this - I had a look at this there is a lot to destructure. It feels like this is a little overkill and doesnt aid with explanation of whats going where.

feedmypixel:
> Could try the heroku instance as a test first?

all ok on heroku 👍

feedmypixel:
and all ok on staging 👍

PR #250, opened 2019-09-23
Mobile sticky sidebar fix

feedmypixel:
# Summary panel appears over content on mobile

## Problem
The summary panel appears over the content on mobile

screencapture-localhost-3000-move-new-move-details-2019-09-23-13_03_56

## Solution 
Add an override class to stop this happening
screencapture-localhost-3000-move-new-health-information-2019-09-23-13_02_50

### DIFF
--- a/common/assets/scss/components/_sticky-sidebar.scss
+++ b/common/assets/scss/components/_sticky-sidebar.scss
@@ -6,4 +6,8 @@
   transform: translate(0, 0); /* For browsers don't support translate3d. */
   transform: translate3d(0, 0, 0);
   will-change: position, transform;
+
+  @include mq ($until: tablet) {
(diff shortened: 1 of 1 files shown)


common/assets/scss/overrides/_sticky-sidebar.scss
```diff
@@ -0,0 +1,7 @@
+.app-\!-not-sticky-mobile {
```
teneightfive: Do we need a separate override class? Feels like it could be part of the `sticky-sidebar` module as can't see any scenarios where we want it to display how it is on mobile.

feedmypixel: good point 👍

PR #252, opened 2019-09-23
[P4-690] Update a person if they already exist

feedmypixel:
The current create journey creates a new person in the API each time the personal details step is saved. 
So if a user does:
- User completes personal details step
- Hits continue
- Hits "Back" on move details steps
- Hits continue

Then duplicate people will be created in the API

### Solution
If a person already exists in the session then update their details via the personService rather than creating a duplicate. This avoids duplicate people being saved to the API and stops duplicates being created when a user goes back in the browser to update a person details.

### DIFF
--- a/app/move/controllers/create/personal-details.test.js
+++ b/app/move/controllers/create/personal-details.test.js
@@ -135,6 +135,46 @@ describe('Move controllers', function() {
       })
     })
 
+    describe('#savePerson()', function() {
+      const mockId = '1234567'
+
+      beforeEach(function() {
+        sinon.stub(personService, 'create').resolves()
+        sinon.stub(personService, 'update').resolves()
+      })
+
+      context('without id', function() {
+        beforeEach(function() {
+          controller.savePerson(undefined, personMock)
+        })
+
+        it('should create a person via the personService', function() {
+          expect(personService.create).to.be.calledOnceWithExactly(personMock)
+        })
+
+        it('should not update a person via the personService', function() {
+          expect(personService.update).to.not.be.called
+        })
+      })
+
+      context('with id', function() {
+        beforeEach(function() {
+          controller.savePerson(mockId, personMock)
+        })
+
+        it('should update a person via the personService', function() {
+          expect(personService.update).to.be.calledOnceWithExactly({
+            id: mockId,
+            ...personMock,
+          })
+        })
+
+        it('should not create a person via the personService', function() {
+          expect(personService.create).to.not.be.called
+        })
+      })
+    })
+
     describe('#saveValues()', function() {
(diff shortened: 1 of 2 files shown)


app/move/controllers/create/personal-details.test.js
```diff
@@ -135,6 +135,41 @@ describe('Move controllers', function() {
       })
     })
 
+    describe('#savePerson()', function() {
+      const mockId = '1234567'
+
+      context('without id', function() {
+        beforeEach(function() {
+          sinon.spy(personService, 'create')
```
teneightfive: You could move these spies to a shared `beforeEach` to reduce some repetition.

I also wonder if they may be safer as stubs so that the actual function doesn't still get called.

app/move/controllers/create/personal-details.test.js
```diff
@@ -135,6 +135,41 @@ describe('Move controllers', function() {
...
+          sinon.spy(personService, 'update')
+
+          controller.savePerson(undefined, personMock)
+        })
+
+        it('should create a person via the personService', function() {
+          expect(personService.create).to.be.calledOnceWithExactly(personMock)
+          expect(personService.update).to.not.be.called
```
teneightfive: Could break these out into two separate assertions. Might make it easier to debug if one fails.

app/move/controllers/create/personal-details.test.js
```diff
@@ -150,55 +185,31 @@ describe('Move controllers', function() {
...
-            expect(personService.update).to.not.be.called
-          })
-
-          it('should set person response on form values', function() {
-            expect(req.form.values).to.have.property('person')
-            expect(req.form.values.person).to.deep.equal(personMock)
-          })
+      context('when save is successful', function() {
+        beforeEach(async function() {
+          sinon.stub(personService, 'create').resolves(personMock)
```
teneightfive: I think it would be safer to stub `this.savePerson` in this unit test otherwise you're testing that function of `savePerson` twice.

And then you can just check that `this.savePerson` is being called with the correct values here.

app/move/controllers/create/personal-details.test.js
```diff
@@ -135,6 +135,43 @@ describe('Move controllers', function() {
...
+        beforeEach(function() {
+          controller.savePerson(mockId, personMock)
+        })
+
+        it('should update a person via the personService', function() {
+          expect(personService.update).to.be.calledOnceWithExactly({
+            id: mockId,
+            ...personMock,
+          })
+          expect(personService.create).to.not.be.called
```
teneightfive: Maybe do the same as above by splitting these out?

PR #254, opened 2019-09-24
[P4-464] Download and print icons

feedmypixel:
## Download and and Print icons

Add the download and print png icons for the app. This work uses webpack and includes `@2x` images for high resolution devices.

## Of Note

I have push the SVG work to the branch [`download-and-print-icons-svg`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/download-and-print-icons-svg) if this is needed in the future.

## Screenshots
Screenshot 2019-09-26 at 14 58 16

### High resolution devices
Screenshot 2019-09-26 at 14 57 51
Screenshot 2019-09-26 at 14 57 58

#### Dashboard
Screenshot 2019-09-26 at 14 58 13

#### View move page
Screenshot 2019-09-26 at 14 58 05


feedmypixel:
@teneightfive be good to get your thoughts on this approach.

common/assets/scss/overrides/_visibility.scss
```diff
@@ -0,0 +1,3 @@
+.is-hidden {
```
teneightfive: There's a class in the Design System that will already do this for us: `.govuk-!-display-none`

app/move/views/view.njk
```diff
@@ -22,6 +22,7 @@
   }) }}
 
   <a href="javascript:window.print()" class="app-!-float-right app-print--hide govuk-!-margin-top-3">
+    {{ appIcon({ id: "print-icon" }) }}
```
teneightfive: Might be better to try and stick to BEM naming convention and just passing in `print` or `download` here.

Then in the classnames on the element it would build `app-icon app-icon--print`. At the moment it displays as `app-icon print-icon` which feels less consistent with the rest of the classname structure.

common/templates/includes/svg-icons.njk
```diff
@@ -0,0 +1,17 @@
+<div class="is-hidden">
```
teneightfive: As mentioned when we chatted this feels nice in terms of flexibility in potentially adding more icons but it also feels poor in terms of performance to have this loaded on every page. Particularly right now where we are only using these icons on ~20% of the pages we have.

I wonder if it this block could be added to the icon component so that at least it's only ever loaded on pages where an icon has been included?

feedmypixel: > I wonder if it this block could be added to the icon component so that at least it's only ever loaded on pages where an icon has been included?

multiple icons on a page though would mean multiples of the `SVGs`. I don't feel it's too bad size wise and as we are not going to have many icons then file size wise it shouldn't be a problem. At the moment these 2 icons are adding `0.2kb` to the html

## Without inline svg
without-inline-svg

## With inline svg
with-inline-svg

common/assets/scss/overrides/_visibility.scss
feedmypixel: As discussed I have added a `TODO` around this and changed the name to `govuk-!-display-none`. Currently this work has been done but has not yet been released. I am watching hte releases and will get this sorted once its out in the wild.

[commit info GOVUK-frontend](https://github.com/alphagov/govuk-frontend/commit/6bbee4b116525199bc2e07659b9af90722555ba2)

feedmypixel:
> There also seems to be some rogue underline of spacing that we might want to remove:
> image

so the spacing is in the SVG - I could bump up the arrow a little if required. These are basically `24 x 24` SVG and the vectors are centred in that `24x 24` space.

common/assets/scss/components/_icons.scss
```diff
@@ -0,0 +1,25 @@
+$width: 20px;
+
+.app-icon {
+  display: inline-block;
+  vertical-align: middle;
+  background-repeat: no-repeat;
+  background-size: $width $width;
+  padding-left: $width + govuk-spacing(2);
```
teneightfive: Looks like we just need to add `height: $width`. In fact, maybe something like this:

```scss
$_icon-width: 20px;
$_icon-height: 20px;

...
background-size: $_icon-width $_icon-height;
padding-left: $_icon-width + govuk-spacing(2);
height: $_icon-height;
...
```

PR #256, opened 2019-09-25
Update API environment variable configuration

teneightfive:
This change amends the way the API is set in the environment variables.

Currently it requires the host to be defined 3 times for `API_BASE_URL`,
`API_HEALTHCHECK_URL` and `API_AUTH_URL`. If the host needs to be
changed, for example switching between environments when testing this
requires changing 3 variables.

This change just neatens that up so that it only needs to be defined once.

It introduces some temporary logic to make sure the change is backwards
compatible with current releases. Once all envionrments have been
updated this logic can be removed.

### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,9 @@
 SESSION_SECRET=example-secret
 REDIS_URL=//localhost:6379
-API_BASE_URL=http://localhost:4000/api/v1
-API_AUTH_URL=http://localhost:4000/oauth/token
+API_BASE_URL=http://localhost:3000
+API_PATH=/api/v1
+API_HEALTHCHECK_PATH=/ping.xml
+API_AUTH_PATH=/oauth/token
 API_CLIENT_ID=
 API_SECRET=
 AUTH_PROVIDER_KEY=
--- a/README.md
+++ b/README.md
@@ -120,9 +120,10 @@ npm run lint
 | REDIS_URL **(required)** | Redis server URL, including port and protocol | |
 | REDIS_HOST **(required)** | Redis hostname. Can be used instead of `REDIS_URL`. Will override `REDIS_URL` if set | |
 | REDIS_AUTH_TOKEN | Optional auth token for the Redis instance | |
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -1,7 +1,9 @@
 SESSION_SECRET=example-secret
 REDIS_URL=//localhost:6379
-API_BASE_URL=http://localhost:4000/api/v1
-API_AUTH_URL=http://localhost:4000/oauth/token
+API_BASE_URL=http://localhost:3000
```
feedmypixel: worth changing this to `5000`?

README.md
```diff
@@ -120,9 +120,10 @@ npm run lint
 | REDIS_URL **(required)** | Redis server URL, including port and protocol | |
 | REDIS_HOST **(required)** | Redis hostname. Can be used instead of `REDIS_URL`. Will override `REDIS_URL` if set | |
 | REDIS_AUTH_TOKEN | Optional auth token for the Redis instance | |
-| API_BASE_URL **(required)** | Base URL for the backend API server for this service without any path | `http://localhost:3000/api/v1` |
-| API_HEALTHCHECK_URL **(required)** | URL to which healthcheck pings are sent | |
-| API_AUTH_URL **(required)** | URL to which OAuth2 access token requests should be sent | |
+| API_BASE_URL **(required)** | Base URL for the backend API server for this service without any path | `http://localhost:3000` |
+| API_PATH **(required)** | Base base for the API | `/api/v1` |
```
feedmypixel: Not part of this work but I wonder if we need to do an `env` check on required on `envs` when the app starts up. Then error if they are not present?

.env.example
teneightfive: I actually changed it back to 3000 as it's the default that Rails runs on. But then you'd have to run the frontend on a separate port.

This is more just for guidance as to what env vars to set anyway I suppose so maybe not that important?

README.md
teneightfive: Yeah might be worth having. But then we'd need a way to store required envs and maintain that list.

PR #260, opened 2019-09-27
[P4-550] Revision images

feedmypixel:
## Feature

This work: 
- Revision images when `webpack` is ran in `production` mode
- Minify images when `webpack` is ran in `production` mode
- Tidy up of `webpack` `env` config files

### development assets
Screenshot 2019-09-27 at 14 16 43

### production assets
Screenshot 2019-09-27 at 14 16 08

### filesize development
Screenshot 2019-09-27 at 14 20 57

### filesize production
Screenshot 2019-09-27 at 14 23 24

## Of Note
- This work does not revision `GOVUK` or `MOJ` `dependency` assets. This will come in a separate PR if needed?
- The image size options could be tweaked a little more at the moment this just uses the default

### DIFF
--- a/webpack.config.production.js
+++ b/webpack.config.production.js
@@ -1,13 +1,44 @@
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
+const ImageminPlugin = require('imagemin-webpack-plugin').default
 
 module.exports = {
   output: {
     filename: 'javascripts/[name].[contenthash:8].js',
   },
 
+  module: {
+    rules: [
+      {
+        test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              name: '[name].[contenthash:8].[ext]',
+            },
+          },
+        ],
+      },
+      {
+        test: /\.(png|svg|jpg|gif|ico)$/,
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              name: '[name].[contenthash:8].[ext]',
+            },
+          },
+        ],
+      },
+    ],
+  },
+
   plugins: [
     new MiniCssExtractPlugin({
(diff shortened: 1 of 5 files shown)


webpack.config.develop.js
```diff
@@ -4,6 +4,22 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const { PORT } = require('./config')
 
 module.exports = {
+  module: {
```
teneightfive: What's the reason for changing the way this was defined? Using the `IS_DEV` flag and having the config in one place seemed neater?

webpack.config.production.js
```diff
@@ -1,13 +1,55 @@
...
+              hmr: false,
+            },
+          },
+        ],
+      },
+      {
+        test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
+        use: [
+          {
+            loader: 'file-loader',
```
teneightfive: You might not need to define this line as it may be taken from the shared config, maybe try it without?

webpack.config.production.js
```diff
@@ -1,13 +1,55 @@
...
+            options: {
+              name: '[name]-[contenthash:8].[ext]',
```
teneightfive: Maybe use `.` instead of `-` as the hash prefix to be consistent with the other asset types?

webpack.config.production.js
feedmypixel: `merge.smart` needs the `loader` name to match and merge this with the shared config

webpack.config.develop.js
feedmypixel: I was trying to move to just using the files (`config.production` or `config.development`) rather than using the `IS_DEV` variable and the files for specifying different configs for `production` or `development`. But agreed on reflection this seems a little verbose for the sake of it - have reverted to what was there 👍

PR #263, opened 2019-09-30
[P4-709] Add the 'Error:' prefix to cancel error page title

feedmypixel:
## Problem
The Error page when you try and cancel a move and it the form validation errors does not have the `'Error:'` prefix in the page title element.

## Solution
This work adds the `'Error:'` prefix in to the page title element.

### Without form validation error
Screenshot 2019-09-30 at 14 08 34

### With form validation error
Screenshot 2019-09-30 at 14 08 22

### DIFF
--- a/app/move/views/cancel.njk
+++ b/app/move/views/cancel.njk
@@ -4,7 +4,7 @@
   gtag('set', {'page_title': 'Cancel move'});
 {% endblock %}
 
-{% block pageTitle %}
+{% block innerPageTitle %}
   {{ t("moves::cancel.page_title", { name: move.person.fullname | upper }) }}
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


app/move/views/cancel.njk
```diff
@@ -5,6 +5,9 @@
 {% endblock %}
 
 {% block pageTitle %}
```
teneightfive: Only thing I was wondering is whether it was worth creating a separate block for the page title for views that inherit from the form wizard to override the page title but to keep the error prefix always in place and it one place.

It might avoid the same scenario accidentally happening again if we override it in another sub-template. Or do you think that's overkill?

feedmypixel: good idea - done 👍

PR #264, opened 2019-09-30
[P4-567] Update URLs to match new IA

teneightfive:
This change refactors the URL structure to match the new IA and will future-proof some of the upcoming pages.

### URL summary

**Before**
`/moves` - redirect to location if don't have permission
`/moves/:locationId` - moves for today
`/moves/:locationId?move-date=2019-09-20` - moves for a date

**After**
`/moves` - redirect to `/moves/{today}` or `/moves/{today}/{currentLocation}` if location is set
`/moves/2019-09-20/:locationId` - moves for a date

The new structure will allow us to support `/incoming` and `/outgoing` on all sets of URLs in the future.


teneightfive:
@feedmypixel might be worth checking this out and having a play when reviewing to make sure it all makes sense and I haven't missed anything

PR #265, opened 2019-10-01
[P4-545] Mandatory other free text

feedmypixel:
## Make other comments boxes mandatory

This work makes the "other" comment boxes in the assessment section of the move journey mandatory.

### Screenshots
give-details-other-errors

### DIFF
--- a/app/move/controllers/create/assessment.js
+++ b/app/move/controllers/create/assessment.js
@@ -7,9 +7,9 @@ class AssessmentController extends CreateBaseController {
   async configure(req, res, next) {
     try {
       const { fields } = req.form.options
-
-      await fieldHelpers.setupAssessmentQuestions(fields)
-
+      req.form.options.fields = await fieldHelpers.populateAssessmentQuestions(
+        fields
+      )
       super.configure(req, res, next)
     } catch (error) {
       next(error)
--- a/app/move/controllers/create/assessment.js
+++ b/app/move/controllers/create/assessment.js
@@ -2,33 +2,13 @@ const { flatten, values } = require('lodash')
 
 const CreateBaseController = require('./base')
 const fieldHelpers = require('../../../../common/helpers/field')
-const referenceDataService = require('../../../../common/services/reference-data')
-const referenceDataHelpers = require('../../../../common/helpers/reference-data')
 
 class AssessmentController extends CreateBaseController {
   async configure(req, res, next) {
     try {
       const { fields } = req.form.options
 
-      await Promise.all(
-        Object.keys(fields).map(key => {
-          const field = fields[key]
-
-          if (!Object.prototype.hasOwnProperty.call(field, 'items')) {
-            return
-          }
-
-          return referenceDataService
-            .getAssessmentQuestions(key)
-            .then(response => {
-              field.items = response
-                .filter(referenceDataHelpers.filterDisabled())
-                .map(fieldHelpers.mapAssessmentQuestionToConditionalField)
-                .map(fieldHelpers.mapAssessmentQuestionToTranslation)
-                .map(fieldHelpers.mapReferenceDataToOption)
-            })
-        })
-      )
+      await fieldHelpers.setupAssessmentQuestions(fields)
 
       super.configure(req, res, next)
     } catch (error) {
--- a/app/move/fields/create.js
+++ b/app/move/fields/create.js
@@ -11,6 +11,15 @@ const assessmentQuestionComments = {
   },
 }
 
+const requiredAssessmentQuestionComments = {
+  ...assessmentQuestionComments,
+  label: {
+    text: 'fields::assessment_comment.required',
+    classes: 'govuk-label--s',
+  },
+  validate: 'required',
+}
+
 function assessmentCategory(category) {
   return {
     component: 'govukCheckboxes',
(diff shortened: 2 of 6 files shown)


feedmypixel:
@teneightfive be good to get your thoughts on this approach 🙏

app/move/fields/create.js
```diff
@@ -11,6 +11,16 @@ const assessmentQuestionComments = {
   },
 }
 
+const requiredAssessmentQuestionComments = {
+  ...assessmentQuestionComments,
```
teneightfive: Nice 👍

app/move/fields/create.js
```diff
@@ -11,6 +11,16 @@ const assessmentQuestionComments = {
...
+  label: {
+    text: 'fields::assessment_comment.required',
+    classes: 'govuk-label--s',
+  },
+  dependent: null, // This value is populated within the assessment controller
```
teneightfive: I was thinking about this earlier and it being a bit of weird concept and I wonder if actually what we could do is use the presence of `validate: 'required'` to set the dependent object in the map function?

The dependance only occurs when this field is pulled in as a child within the assessment controller so maybe checking for that assessment question being required feels like a more logical check than checking for `null`. What do you think?

app/move/controllers/create/assessment.js
```diff
@@ -2,33 +2,15 @@ const { flatten, values } = require('lodash')
...
 const fieldHelpers = require('../../../../common/helpers/field')
-const referenceDataService = require('../../../../common/services/reference-data')
-const referenceDataHelpers = require('../../../../common/helpers/reference-data')
 
 class AssessmentController extends CreateBaseController {
   async configure(req, res, next) {
     try {
       const { fields } = req.form.options
 
       await Promise.all(
```
teneightfive: Just looking at how this has been refactored to use async/await in the map, does this still need to be a `Promise.all`?

teneightfive: `fieldHelpers.mapAssessmentQuestion` doesn't seem to be returning a promise any more so might not be needed.

The only other thing I'm wondering is if this has actually made what was more of an async set of statements into sync ones?

common/helpers/field.js
```diff
@@ -177,14 +179,59 @@ function insertItemConditional({ key, field }) {
...
+    .filter(referenceDataHelpers.filterDisabled())
+    .map(mapAssessmentQuestionToConditionalField)
+    .map(mapAssessmentQuestionToTranslation)
+    .map(mapReferenceDataToOption)
+
+  return [key, field]
+}
+
+function mapAssessmentValidation([key, field], index, obj) {
+  if (!isNull(field.dependent)) {
```
teneightfive: Based on the earlier comment I think it would make this easier to read too. It would become "If field is not required" rather than "If dependent property is not null".

locales/en/fields.json
```diff
@@ -108,5 +108,14 @@
   },
   "cancellation_reason_comment": {
     "label": "Give details"
+  },
+  "court__other_court": {
+    "label": "Any other information details"
```
teneightfive: Are these being used?

common/helpers/field.js
```diff
@@ -177,14 +179,59 @@ function insertItemConditional({ key, field }) {
...
+    return [key, field]
+  }
+
+  const parentField = Object.values(fromPairs(obj)).find(field =>
```
teneightfive: Wonder if using lodash for this would make it a bit easier to read? We're already using it in other places so don't mind using it for more complicated finds.

common/helpers/field.js
feedmypixel: Agreed - it appears that `validate` can only have the string `'required'` so a check for `field.validate` should suffice here

app/move/controllers/create/assessment.js
feedmypixel: > Just looking at how this has been refactored to use async/await in the map, does this still need to be a `Promise.all`?

so the `fieldHelpers.mapAssessmentQuestion` method:
```
async function mapAssessmentQuestion([key, field]) {
 ...
```
 is returning a promise due to the `async`. Therefore the `map` is returning an array of `Promises`. So yup `Promise.all` is very much needed.

> The only other thing I'm wondering is if this has actually made what was more of an async set of statements into sync ones?

Nope as far as I am aware these are `Promises` so all still `async` but will work as `sync` due to the 
```
await Promise.all(
  ...
```

common/helpers/field.js
feedmypixel: have removed the need for this code

common/helpers/field.js
```diff
@@ -177,14 +179,58 @@ function insertItemConditional({ key, field }) {
...
+  )
+
+  field.dependent = {
+    field: parentField.name,
+    value: conditionalParent.value,
+  }
+}
+
+async function setupAssessmentQuestions(fields) {
+  const parentField = Object.values(fields).find(findParent)
```
teneightfive: I wonder if there's a little too much abstraction here which feels like it makes it a little harder to follow. Lots of this only originally occurred in one place so not sure it warrants abstracting everything out into shared functions as they're never reused anywhere.

I think the new terms `parent`/`child` threw me a little too as its another term we're using. Wonder  if we can maybe make use more of `dependent` which is the term already being used when one field's validation depends on another's. It doesn't necessarily need to therefore be a parent/child relationship.

teneightfive: Wonder if we can use something more like `hasItems` as opposed to calling them parent/child too?

PR #268, opened 2019-10-03
Refactor locations to cater for suppliers

teneightfive:
This change refactors the way locations are retrieved for supplier users.

When a user has a supplier role it will lookup the list of locations using a different filter from the API.

This then allows the user to be treated the same way as other users and to see a list of locations they can access.

This is the first step in filtering moves for suppliers.

### DIFF
--- a/common/services/user.js
+++ b/common/services/user.js
@@ -1,7 +1,10 @@
 const axios = require('axios')
 
+const {
+  getLocationsBySupplierId,
+  getLocationsByNomisAgencyId,
+} = require('./reference-data')
 const { decodeAccessToken } = require('../lib/access-token')
-const { getLocationsByNomisAgencyId } = require('./reference-data')
 const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
 
 const getAuthHeader = token => ({ Authorization: `Bearer ${token}` })
@@ -18,15 +21,15 @@ function getLocations(token) {
 
   switch (authSource) {
     case 'auth':
-      return getAuthAgencies(token)
+      return getAuthLocations(token)
     case 'nomis':
-      return getNomisAgencies(token)
+      return getNomisLocations(token)
     default:
       return Promise.resolve([])
   }
 }
 
-function getAuthAgencies(token) {
+function getAuthGroups(token) {
   const { user_name: userName } = decodeAccessToken(token)
 
   return axios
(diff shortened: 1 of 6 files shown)


app/auth/middleware.js
```diff
@@ -13,10 +16,11 @@ function processAuthResponse() {
     try {
       const accessToken = grant.response.access_token
       const decodedAccessToken = decodeAccessToken(grant.response.access_token)
-      const [locations, fullname] = await Promise.all([
-        getLocations(accessToken),
+      const [groups, fullname] = await Promise.all([
+        getGroups(accessToken),
         getFullname(accessToken),
       ])
+      const locations = await getLocationsByNomisAgencyId(groups)
```
feedmypixel: this feels a little awkward but I cant really see a better way of doing it.

Maybe removing the `Promise.all` and doing 3 awaits?

```
const groups = await getGroups(accessToken)
const fullname =  await getFullname(accessToken)
const locations = await getLocationsByNomisAgencyId(groups)
```

common/services/user.js
```diff
@@ -14,19 +17,27 @@ function getFullname(token) {
 }
 
 function getLocations(token) {
```
feedmypixel: It feels like this method has got a few things going on.

Rather than having the line `if (authorities.includes('ROLE_PECS_SUPPLIER'))` inside of this method and it doing two different things what do you think to having it at a higher level?

so for instance in the `app/auth/middleware.js` file we make a decision on what type of auth user this is and send in different items to the `Promise.all`.

Then either in the `common/services/user.js` file or in a new `common/services/user-supplier.js` file we use the common methods:
• `getAuthGroups`
• `getNomisCaseloads`

But have separate methods:
## `getLocations`
```
function getLocations(token) {
  const { auth_source: authSource } = decodeAccessToken(token)

  switch (authSource) {
    case 'auth':
      return getAuthGroups(token).then(getLocationsByNomisAgencyId)
    case 'nomis':
      return getNomisCaseloads(token).then(getLocationsByNomisAgencyId)
    default:
      return Promise.resolve([])
  }
}
```
## `getSupplierLocations`
```
function getSupplierLocations(token) {
  const { authorities = [], auth_source: authSource } = decodeAccessToken(token)

  switch (authSource) {
    case 'auth':
      return getAuthGroups(token).then(([firstGroup]) => getLocationsBySupplierId(firstGroup))
    case 'nomis':
      return getNomisCaseloads(token).then(getLocationsByNomisAgencyId)
    default:
      return Promise.resolve([])
  }
}
```

This then means that these methods are doing just one thing rather than two. Easier to understand at a higher level and easier to test, Theres is a little repetition but I feel that helps with explanation rather than needing to be dry'd up.

What do you think?

teneightfive: Yeah I wasn't too sure on it to be honest and didn't like that we're introducing role logic into the service. I like the idea of having separate methods for getting supplier/other user locations.

So maybe we keep that logic in the auth middleware to determine which service method to call.

PR #270, opened 2019-10-04
Reverting the removal of cookie-parser dep

feedmypixel:
It turns out that hmpo-form-wizard requires the use of cookie-parser.
This work reverts the previous removal of cookie-parser in #267


teneightfive:
Or were you just trying to catch up on number of commits  👀😉

feedmypixel:
> Or were you just trying to catch up on number of commits 👀😉

this 🤣

PR #271, opened 2019-10-07
Use abbreviation element pnc

feedmypixel:
### Feature
In some instances we are now outputting "**PNC number**" as a label. We should try and use an [abbreviation element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) for PNC to explain what it stands for.


common/presenters/person-to-summary-list-component.js
```diff
@@ -2,10 +2,18 @@ const i18n = require('../../config/i18n')
...
+
+  if (hasHtmlTranslation) {
+    key.html = i18n.t(translationKey, { context: 'html' })
+  } else {
+    key.text = i18n.t(translationKey)
+  }
+
   return {
-    key: {
-      text: i18n.t(`fields::${type}.label`),
```
teneightfive: I think you might just be able to change the `text` property to `html` to output HTML instead of text?

teneightfive: Oh, I see, we don't always know if it has HTML. I'd be tempted to just change it to `html` to all of them to avoid the need for the extra logic and checking against the translations. Would that solve the issue?

PR #272, opened 2019-10-07
Allow users to view all locations at once

teneightfive:
This change updates how a location is checked and set.

This allows users with a role of `moves:view:all` to be able to select all moves available to them.

### DIFF
--- a/common/middleware/ensure-current-location.js
+++ b/common/middleware/ensure-current-location.js
@@ -1,14 +1,11 @@
+const { isUndefined } = require('lodash')
+
 function ensureCurrentLocation({ locationsMountpath, whitelist = [] } = {}) {
   return (req, res, next) => {
-    const { permissions = [], locations = [] } = req.session.user || {}
-    const canViewAllMovesWithoutLocations =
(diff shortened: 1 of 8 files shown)


common/middleware/ensure-current-location.js
```diff
@@ -1,14 +1,11 @@
...
+
 function ensureCurrentLocation({ locationsMountpath, whitelist = [] } = {}) {
   return (req, res, next) => {
-    const { permissions = [], locations = [] } = req.session.user || {}
-    const canViewAllMovesWithoutLocations =
-      permissions.includes('moves:view:all') && !locations.length
-
     if (
-      req.session.currentLocation ||
+      !isUndefined(req.session.currentLocation) ||
```
feedmypixel: this double negative is a bit hard to read could you use: 
```
const currentLocation = get(req, 'session.currentLocation')
 ...
if (
      currentLocation ||
  ...
```

teneightfive: Only problem with that is we need it to check for only `undefined` now as when you are viewing all it is set to `null`.

PR #273, opened 2019-10-07
Update missing location error handling

teneightfive:
This change updates how the app handles a missing location.

If a location is missing it will be treated as an expected error
page.

It will display a message to go to the locations page to set one
before creating a move.

## Before
localhost_3000_move_new_personal-details (1)

## After
localhost_3000_move_new_personal-details

### DIFF
--- a/app/move/controllers/create/base.js
+++ b/app/move/controllers/create/base.js
@@ -20,7 +20,8 @@ class CreateBaseController extends FormWizardController {
 
   checkCurrentLocation(req, res, next) {
     if (!req.session.currentLocation) {
-      const error = new Error('Current location is not set in session.')
+      const error = new Error('Current location is not set in session')
+      error.code = 'MISSING_LOCATION'
       return next(error)
     }
 
--- a/common/controllers/form-wizard.js
+++ b/common/controllers/form-wizard.js
@@ -26,7 +26,11 @@ class FormController extends Controller {
       return res.redirect(err.redirect)
     }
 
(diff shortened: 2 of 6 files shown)


app/move/controllers/create/base.js
```diff
@@ -20,7 +20,8 @@ class CreateBaseController extends FormWizardController {
 
   checkCurrentLocation(req, res, next) {
     if (!req.session.currentLocation) {
-      const error = new Error('Current location is not set in session.')
+      const error = new Error('Current location is not set in session')
+      error.code = 'MISSING_LOCATION'
```
feedmypixel: Maybe for future but these could sit in one place?

common/controllers/form-wizard.js
```diff
@@ -26,7 +26,11 @@ class FormController extends Controller {
       return res.redirect(err.redirect)
     }
 
-    if (err.code === 'SESSION_TIMEOUT' || err.code === 'MISSING_PREREQ') {
+    if (
+      err.code === 'SESSION_TIMEOUT' ||
+      err.code === 'MISSING_PREREQ' ||
+      err.code === 'MISSING_LOCATION'
```
feedmypixel: could do an `.includes` on an array of error codes?

app/move/controllers/create/base.js
teneightfive: The errors? Possibly, although this one is specific to this part of the app. The others are set within the actual form wizard library too so not sure how easy it would be to extract them all and merge them.

PR #274, opened 2019-10-09
Use supplier auth group to retrieve locations

teneightfive:
This change uses the key retrieved from a supplier's auth group to get the list of available locations for that supplier.

### DIFF
--- a/common/services/reference-data.js
+++ b/common/services/reference-data.js
@@ -93,6 +93,18 @@ const referenceDataService = {
       locations.filter(Boolean)
     )
   },
+
+  getSuppliers() {
+    return apiClient.findAll('supplier').then(response => response.data)
+  },
+
+  getSupplierByKey(key) {
+    if (!key) {
+      return Promise.reject(new Error('No supplier key provided'))
+    }
+
+    return apiClient.find('supplier', key).then(response => response.data)
+  },
 }
 
 module.exports = referenceDataService
--- a/common/services/user.js
+++ b/common/services/user.js
@@ -3,6 +3,7 @@ const axios = require('axios')
 const {
   getLocationsBySupplierId,
   getLocationsByNomisAgencyId,
+  getSupplierByKey,
 } = require('./reference-data')
 const { decodeAccessToken } = require('../lib/access-token')
 const { AUTH_PROVIDERS, NOMIS_ELITE2_API } = require('../../config')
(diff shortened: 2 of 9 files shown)


common/services/user.js
```diff
@@ -43,9 +44,16 @@ function getAuthGroups(token) {
 function getAuthLocations(token) {
   const { authorities = [] } = decodeAccessToken(token)
 
-  return getAuthGroups(token).then(groups => {
+  return getAuthGroups(token).then(async groups => {
     if (authorities.includes('ROLE_PECS_SUPPLIER')) {
-      return getLocationsBySupplierId(groups[0])
+      try {
+        const supplierKey = groups[0].toLowerCase()
```
feedmypixel: does this need protecting do we always have an element in the array?

common/services/reference-data.js
```diff
@@ -93,6 +93,18 @@ const referenceDataService = {
...
     )
   },
+
+  getSuppliers() {
+    return apiClient.findAll('supplier').then(response => response.data)
+  },
+
+  getSupplierByKey(key) {
+    if (!key) {
+      return Promise.reject(new Error('No supplier key supplied'))
```
feedmypixel: `'No supplier key provided'`?

common/services/user.js
```diff
@@ -43,9 +44,16 @@ function getAuthGroups(token) {
...
+        const supplier = await getSupplierByKey(supplierKey)
+
+        return getLocationsBySupplierId(supplier.id)
+      } catch (err) {
+        return Promise.resolve([])
```
feedmypixel: do we want the error to be silent here?

common/services/reference-data.js
teneightfive: Haha yeah that does read strange

common/services/user.js
teneightfive: No, we won't necessarily have one, it could return an empty array.

But in those instances it will be because a user doesn't belong to any groups and the `try/catch` will catch that error and then just return an empty array which is the fallback case.

common/services/user.js
teneightfive: So I kind of thought, yes. Mainly because in all the error cases in this instance are mostly expected and we then just want to return an empty array of locations for the app to handle.

Errors would be:
* user has no groups
   * `try/catch` will error trying to convert group to lowercase and return empty array
* user has groups, but no supplier found
  * `try/catch` will catch `404` response from `getSupplierByKey` and return empty array
* user has groups, supplier is found, but no locations
  * `catch` will not be run but `getLocationsBySupplierId` will return an empty array

feedmypixel: if `getSupplierByKey` actually errors though we won't know right? this will just return an empty array.

It feels like we could cater for all these things with `catch` maybe?

```
try {
  const supplierKey = groups[0].toLowerCase()
  const supplier = await getSupplierByKey(supplierKey)

  return getLocationsBySupplierId(supplier.id)
} catch (err instanceof TypeError) { // no elements in array
  return Promise.resolve([])
} catch (err.code === '404') {
  return Promise.resolve([])
} catch (err) { // all otehr errors
  logger.error(error)
}
```

PR #281, opened 2019-10-14
Fix for html in error messages

feedmypixel:
The `abbr` html tag has been brought in previously for labels. This fix
changes the error messages to be html rather than text so that html can
be rendered in an error message.

### Before fix
Screenshot 2019-10-14 at 16 39 21

### After fix
Screenshot 2019-10-14 at 16 37 24


tyom:
Could you rebase please to retest against updated requirements?

PR #286, opened 2019-10-21
Add test for checking move tag creation and nav

tyom:
https://dsdmoj.atlassian.net/browse/P4-698

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,6 +148,9 @@ jobs:
           command: npm run test-e2e:ci
       - store_test_results:
           path: reports/testcafe
+      - store_artifacts:
+          path: artifacts
+
 
   build_staging:
     <<: *cloud_container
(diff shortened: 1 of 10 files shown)


feedmypixel:
🤔 tests are failing in Circle. I'll take a look why in the morning

tyom:
> hmmmm the e2e tests are failing - ill take a look why in the morning

I rerun the workflow and E2E tests came up green. It appears that FF download CSV test can be a little flaky. It's worth keeping an eye on it in case it keeps failing frequently and review it if necessary.

There's a red cross against CodeClimate check (Code Climate was unable to fetch this commit). Could you take a look what it needs @feedmypixel.

feedmypixel:
> I rerun the workflow and E2E tests came up green. It appears that FF download CSV test can be a little flaky. It's worth keeping an eye on it in case it keeps failing frequently and review it if necessary.

will do - thanks

PR #288, opened 2019-10-22
[P4-744] PNC search functionality

feedmypixel:
Work involved around the PNC search functionality

### Search with results
search

### Search with results then cancel
search-results-cancel

### Search with results then create move for different person
search-results-continue

### Search no results
search-no-results

### Search empty PNC
search-no-pnc

## Click continue without pnc
search-continue-no-pnc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ typings/
 
 # App build folder
 .build/
+
+# Test cafe reports
+report*.json
--- a/app/move/controllers/create/index.js
+++ b/app/move/controllers/create/index.js
@@ -3,6 +3,7 @@ const Base = require('./base')
 const MoveDetails = require('./move-details')
 const PersonalDetails = require('./personal-details')
 const Save = require('./save')
+const PncSearchResults = require('./pnc-search-results')
 
 module.exports = {
   Assessment,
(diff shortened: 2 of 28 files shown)


Andy-Millar:
Cheers Ben - all looks good, comments below...

Search
- Spot on

Search no results
- For this example, can we pretend there is a PNC number (rather than searching blank)
- In the case above, there should be no results, should just continue to Personal Details, with PNC prefilled (see screenshots)
- I was thinking that you shouldn't be able to search an empty PNC (instead, the 'I don't have a PNC number' link should be used), but searching an empty PNC number would present results where there are no PNC numbers! So this is good. In the case that there are no results for a blank PNC number, it should just continue to Personal Details, with PNC prefilled (in this case, blank!). If there are previous moves without PNC number, then these results should be presented in the results page, for the user to choose.
- Personal Details, blank PNC number: Since the input is empty, I wonder if we need to indicate that the field is effectively Optional in this instance... Need to be able to allow the user to add it later if they get the PNC number (and can be bothered, likely not)

Search no results, search again
- Spot on

Search no results, create
- Same as 'search no results' above: my view is there should be no results, should just continue to Personal Details, with PNC prefilled (see screenshots)

Search no PNC
- Similar to my comments about blank PNC number in 'search no results': Since the input is empty, I wonder if we need to indicate that the field is effectively Optional in this instance... Need to be able to allow the user to add it later if they get the PNC number (and can be bothered, likely not)

PR #290, opened 2019-10-23
Use book-a-secure-move GitHub group for Cloud Platform deployment through Circle

spikeheap:
Should fix the failing CP deployments

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ references:
     docker:
       - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
         environment:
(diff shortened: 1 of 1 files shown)


feedmypixel:
@spikeheap please don't merge PR's without a review. This is so other people can check your changes and so there is an awareness of what has gone into `master`. I presume something was overridden here as we have the repo setup for all PR's to require a review.
Thank you 🙏

feedmypixel:
For clarity the etiquette is:
• All PR's need 1 approval (team visibility)
• No committing direct to master (this is just plain dangerous)
• Owner merges PR (in case there are other things that need doing)

PR #297, opened 2019-11-12
[P4-788] - Send create a move journey time to GA

feedmypixel:
# Description

This work sends the Create a Move Journey duration to Google analytics.

It can be found in 
`behaviour` 
`-> site speed` 
`-> user timings`

## Screen shot
This is GA from the demo site:
ga-timing-hit

### DIFF
--- a/app/move/controllers/create/base.js
+++ b/app/move/controllers/create/base.js
@@ -1,3 +1,4 @@
+const { get, set } = require('lodash')
 const FormWizardController = require('../../../../common/controllers/form-wizard')
 const presenters = require('../../../../common/presenters')
 
@@ -11,10 +12,19 @@ class CreateBaseController extends FormWizardController {
     super.middlewareLocals()
     this.use(this.setCancelUrl)
     this.use(this.setMoveSummary)
+    this.use(this.setJourneyTimer)
   }
 
   setCancelUrl(req, res, next) {
     res.locals.cancelUrl = res.locals.MOVES_URL
+    next()
+  }
+
+  setJourneyTimer(req, res, next) {
+    if (!get(req, 'session.createMoveJourneyTime')) {
+      set(req, 'session.createMoveJourneyTime', process.hrtime())
+    }
+
     next()
   }
 
--- a/common/lib/analytics.test.js
+++ b/common/lib/analytics.test.js
@@ -1,0 +1,79 @@
+const proxyquire = require('proxyquire')
+const mockHrTime = [44444, 55555]
+
+describe('Analytics', function() {
+  describe('#sendJourneyTime()', function() {
+    context('without GA_ID', function() {
+      let result, analytics
+
+      beforeEach(async function() {
+        analytics = proxyquire('./analytics', {
+          '../../config': {
+            ANALYTICS: {
+              GA_ID: '',
+            },
+          },
+        })
+
+        result = await analytics.sendJourneyTime('111111', {})
+      })
+
+      it('should resolve with empty promise', function() {
+        expect(result).to.equal('No GA ID!')
+      })
+    })
+
+    context('with GA_ID', function() {
+      const mockUUID = '3333'
+      const mockGaID = '11111'
+      let analytics, result, req
+
+      beforeEach(async function() {
+        req = {
+          session: {
+            user: {
+              role: 'robocop',
+            },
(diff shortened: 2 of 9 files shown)


app/move/controllers/create/base.js
```diff
@@ -11,13 +12,22 @@ class CreateBaseController extends FormWizardController {
...
   }
 
   setCancelUrl(req, res, next) {
     res.locals.cancelUrl = res.locals.MOVES_URL
     next()
   }
 
+  setJourneyTimer(req, res, next) {
+    if (!get(req, 'session.createMoveJourneyTimestamp')) {
+      set(req, 'session.createMoveJourneyTimestamp', new Date().getTime())
```
spikeheap: Depending on our browser support, could we use `performance.now` here & for the final calculation ([caniuse](https://caniuse.com/#feat=mdn-api_performance))? `performance.now` is guaranteed to increase at a constant rate, whereas `Date().getTime()` is affected by user clock changes, NTP, etc., but it's not supported on IE8.

feedmypixel:
> how will this behave if a user completes a move journey and then creates a new move? It looks like it might not reset the `session.createMoveJourneyTimestamp`.

@spikeheap good spot. I left this out yesterday as it was being a little tricky. Fresh eyes have fixed it. I have just pushed up this work as a fixup. Have a look and let me know what you think

## Of Note
I have marked a test as pending as I have to pop out (and holiday) will sort this out later

app/move/controllers/create/base.js
feedmypixel: This is server side google analytics.

common/lib/analytics.test.js
```diff
@@ -15,7 +22,7 @@ describe('Analytics', function() {
       })
     })
 
-    context('with GA_ID', function() {
+    xcontext('with GA_ID', function() {
```
feedmypixel: this has been marked as `pending` I will sort this out later today

app/move/controllers/create/base.js
spikeheap: :+1:, ignore the browser comments. The problem I point out still stands though.

A quick Google suggests that the Performance Timing API was added to Node 8.5.0.

If you don't like that solution, I'd suggest something like `process.hrtime()`, which was added in 0.7.6, and (from [the docs](https://nodejs.org/api/process.html#process_process_hrtime_time)):

> These times are relative to an arbitrary time in the past, and not related to the time of day and therefore not subject to clock drift. The primary use is for measuring performance between intervals

feedmypixel: > The problem I point out still stands though.

I don't feel this is a problem TBH. It feels like this is personal preference. I'd prefer to just use `Date` and timestamps as I always have.

> whereas Date().getTime() is affected by user clock changes, NTP, etc

Timing wise it would be like for like so even if the clock changes it won't be a problem

spikeheap: Hi @feedmypixel, I'm afraid this isn't true. Using `Date().getTime()` the time might change locally (e.g. through NTP) *between* the first and second reading, so the elapsed time is unreliable. The purpose of `hrtime()` is to prevent that, and it guarantees it will increment predictably over time.

The Node documentation linked above cites performance benchmarking as the primary use of `process.hrtime()`. This is a great opportunity to learn new approaches, and since I learned about the unreliability of local clock time I've applied this across all the projects I've contributed to.

Is there a compelling reason *not* to use `process.hrtime()`?

feedmypixel: > Is there a compelling reason _not_ to use `process.hrtime()`?

We are measuring the time of a journey in seconds. What you are talking about is a real edge case. If we were measuring process in less than a millisecond and needed that level of accuracy then I may consider this. But for GA and journey times I feel we do not need this.

> the time might change locally (e.g. through NTP) between the first and second reading,

This is unlikely

@spikeheap I feel we are going round in circles here and this has become a personal preference thing. This is not the point of code review IMO.

If you would like to speak about this in person please give me a slack call and we can work something out.

feedmypixel:
> There are a couple of failing checks we might want to address,

The issue on code climate is fine as 5 of the lines are an object being passed to `sendHit`. I have dismissed this IMO refactoring to make this 25 lines will make this method harder to read.

PR #298, opened 2019-11-15
Add preprod to circleci pipeline

spikeheap:
This PR adds preprod to the CircleCI pipeline. It doesn't depend on staging, and doesn't become a dependency on production.

We may want to revisit the workflow later on (e.g. to promote builds), this PR just adds the environment following the current approach.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,16 +204,12 @@ workflows:
       - deploy_staging:
           requires:
             - build_staging
-      - hold_preprod:
-          type: approval
+      - build_preprod:
           requires:
             - build_and_test
           filters:
             branches:
               only: master
-      - build_preprod:
-          requires:
-            - hold_preprod
       - deploy_preprod:
           requires:
             - build_preprod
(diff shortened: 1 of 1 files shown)


feedmypixel:
> We may want to revisit the workflow later on (e.g. to promote builds), this PR just adds the environment following the current approach.

I'd visit this now in this PR otherwise it may get forgotten about. _temporary is rarely temporary_

spikeheap:
Morning @feedmypixel. The rationale behind leaving the workflow updates to later was mostly to keep it consistent with API, and protect that from harming data in preprod while we validate the NOMIS sync changes... but that's a lower risk/cost now. I'll update it to auto-deploy here and in the API, and we can think about the workflow at some other juncture. 

Where I've worked with preprod environments and batched releases before, we've tended to release to preprod at the same intervals as prod, so we get a sense of how the release will change things. I'm not sure that really applies here because we're effectively going to use preprod as "staging but with better data", so there's a real benefit to keeping it up to date.

PR #304, opened 2019-12-16
File size errors

feedmypixel:
This work:

- Increases the filesize limit to `50 MB`
- displays a custom error when files are above `50 MB`
- makes the clientside error reporting more robust. So if the API doesn't provide a response in the expected error format the client displays a generic `Something went wrong error`
- refactor of the upload middleware. This work was abstracted but it becan to share items found in the document controller. So I have brought into this controller. Along with its tests. This may change in the future if `multi-upload` is required elsewhere in the app

#### Documents to large screen capture
uploads-error

#### Documents to large error
documents-to-large-error

#### Generic server error
generic error


feedmypixel:
> 👍 LGTM. One minor thing is the 50Mb limit is set in the controller, template and locales files. I wonder if it could be abstracted so we set it in one place, to reduce the risk of missing one of them in a future change. Nothing to worry about if it's a pain, and it shouldn't block merging this.

Thanks - We appear to be aligned :). I pushed this up 10 mins ago

I have added it to config for the node layer. The component and the validation I have hard coded as they are a little bit trickier.

PR #313, opened 2020-01-15
P4-872 Send supplier ID when requesting moves for locations

electrified:
### Checklist

### DIFF
--- a/common/services/user.js
+++ b/common/services/user.js
@@ -53,7 +53,6 @@ function getAuthLocations(token) {
       try {
         const supplierKey = groups[0].toLowerCase()
         const supplier = await getSupplierByKey(supplierKey)
-
         return getLocationsBySupplierId(supplier.id)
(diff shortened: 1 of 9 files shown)


common/services/user.js
```diff
@@ -76,7 +75,32 @@ function getNomisLocations(token) {
     .then(agencies => getLocationsByNomisAgencyId(agencies))
 }
 
+function getSupplierId(token) {
```
electrified: I am not happy about this function - it's basically a copy of getAuthLocations, except it returns the supplier ID. Therefore introduces another API call to getSupplierByKey.

Any thoughts on the best way to proceed?

teneightfive:
@electrified let's catch up on this over a call. I wonder if there's a better way to solve this using the original patterns we tried to set out.

electrified:
Closing this PR as going for a different approach of making multiple requests

PR #314, opened 2020-01-16
P4 558 display status

barbaracassani:
The layout in app-card might be seen as a bit cheeky -- happy to look into alternatives.

### Dashboard card
image
image

### Move detail
image

**Questions:**

- are the banners displayed for every possible status?
- will the statuses be translated?
- the explanation at the bottom of the big banner: which property of the move does it correspond to? (for example, is it additional_information?

### Checklist

### DIFF
--- a/app/move/views/view.njk
+++ b/app/move/views/view.njk
@@ -25,19 +25,15 @@
     {{ t("actions::print_move_detail") }}
   </a>
 
-  {% if move.status == "cancelled" %}
+  {% if move.status %}
     {{ appMessage({
       allowDismiss: false,
-      classes: "app-message--temporary",
+      classes: "app-message--info",
       title: {
-        text: t("moves::cancel.panel.heading")
+        text: messageTitle
       },
       content: {
-        text: t("moves::cancel.panel.content", {
-          reason: cancellationReason
-        }) if cancellationReason else t("statuses::" + move.status, {
-            context: "description"
-        })
+        text: messageContent
       }
     }) }}
   {% endif %}
@@ -48,7 +44,6 @@
     <h1 class="govuk-heading-xl govuk-!-margin-bottom-1">
       {{ move.person.fullname | upper }}
     </h1>
-
     <span class="govuk-caption-xl">
       {{ t("moves::move_reference", {
         reference: move.reference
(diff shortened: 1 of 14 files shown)


common/components/card/_card.scss
```diff
@@ -7,6 +7,11 @@
   padding-bottom: govuk-spacing(3);
 }
 
+strong.app-card__status {
```
teneightfive: Might be better to remove the element selector. [BEM](http://getbem.com/) allows you to target elements using specific patterns so we can avoid using element selectors. 

The other benefit is it means they can be applied to any element in the markup without needing to update the styles.

common/components/card/_card.scss
```diff
@@ -7,6 +7,11 @@
...
+  margin-top: -15px;
```
teneightfive: I think we can remove this margin so that the label aligns with the card title.

common/components/card/_card.scss
teneightfive: In fact, we can probably remove this declaration altogether and use `app-tag app-tag--inactive` classnames to give the same appearance.

common/components/card/template.njk
```diff
@@ -62,4 +58,11 @@
       </ul>
     {% endif %}
   </div>
+
+    {% if params.status %}
```
teneightfive: Drop this back one indentation to be consistent with the rest of the file?

app/move/views/view.njk
```diff
@@ -38,6 +38,17 @@
         }) if cancellationReason
       }
     }) }}
+  {% else %}
```
teneightfive: I wonder if we just use the same banner style? The grey banner in the ticket looks like an old version of the design that @Andy-Millar and I ended up switching to the blue variation to convey status.

I know we talked about using it for "completed" but the app didn't have that concept at the time.

teneightfive:
@barbaracassani after speaking with Jason and Andy I think these are the answers to the question:

> are the banners displayed for every possible status?

Yes, I've made it use the translation library and where a direct translation doesn't exist, it will use the status from the API

> will the statuses be translated?

Yep, I've added this functionality so we can override it in the frontend

> the explanation at the bottom of the big banner: which property of the move does it correspond to? (for example, is it additional_information?

I'm going to have this come from the translation too as I don't feel it's something that should be associated with every move object from the API.

PR #317, opened 2020-01-20
P4-872 - Batch requests for moves

electrified:
### Checklist


app/moves/middleware.js
```diff
@@ -74,13 +81,20 @@ module.exports = {
...
     try {
-      const userLocations = get(req.session, 'user.locations', [])
-      const locationIds =
-        fromLocationId || userLocations.map(location => location.id).join(',')
-      const [requestedMoves, cancelledMoves] = await Promise.all([
-        moveService.getRequested({ moveDate, fromLocationId: locationIds }),
-        moveService.getCancelled({ moveDate, fromLocationId: locationIds }),
-      ])
+      const locationIds = fromLocationId
+        ? fromLocationId.split(',')
```
teneightfive: Does this need to be split? I thought the logic around `fromLocationId` in  the session  is that it will only ever be one. So if set, only display moves from that location, if not, display all moves from the locations 
 that the user has access - which will be an array.

app/moves/index.js
```diff
@@ -11,6 +11,7 @@ const {
   setFromLocation,
   setPagination,
   setMovesByDate,
```
teneightfive: Maybe could rename this to be more verbose too? `setMovesByDateAndLocation`?

app/moves/middleware.js
```diff
@@ -5,12 +5,19 @@ const {
...
   isValid: isValidDate,
 } = require('date-fns')
-const { find, get } = require('lodash')
+const { find, get, chunk } = require('lodash')
 
 const moveService = require('../../common/services/move')
 
 const moveDateFormat = 'yyyy-MM-dd'
 
+const batchSize = 50
```
teneightfive: As you mentioned in our quick call - I think it could be useful to set this in the config and have it overridable by an env var.

app/moves/middleware.js
```diff
@@ -90,4 +94,34 @@ module.exports = {
...
+      return next()
+    }
+
+    try {
+      const userLocations = get(req.session, 'user.locations', []).map(
+        loc => loc.id
+      )
+
+      if (userLocations.length === 0) {
+        userLocations.push(null)
```
teneightfive: What's the purpose of this? Is this to handle when the user has no locations?

electrified: Yes, that's correct. If the user hits view all, but there are no user locations configured.

teneightfive: In theory, that shouldn't even happen as the global middleware will redirect the user to the locations sub-app if there are no locations so we can probably get rid of this and assume this will always be an array.

.env.example
```diff
@@ -21,3 +21,4 @@ E2E_POLICE_USERNAME=
 E2E_POLICE_PASSWORD=
 E2E_SUPPLIER_USERNAME=
 E2E_SUPPLIER_PASSWORD=
+#MOVES_BATCH_SIZE=40
```
teneightfive: Missing new line it looks like

app/moves/middleware.js
```diff
@@ -5,12 +5,19 @@ const {
...
   isValid: isValidDate,
 } = require('date-fns')
-const { find, get } = require('lodash')
+const { find, get, chunk } = require('lodash')
 
 const moveService = require('../../common/services/move')
 
 const moveDateFormat = 'yyyy-MM-dd'
 
+const { MOVES_BATCH_SIZE } = require('../../config')
```
teneightfive: Maybe move this up to line 11.

The pattern for files we've tried to use is:

- load external libraries
- new line
- load internal libraries
- new line
- define module constants

It just helps keep consistency and helps with knowing where to look for things within the files across the app.

app/moves/middleware.js
```diff
@@ -5,12 +5,19 @@ const {
...
+
+const makeMultipleRequests = (service, moveDate, locationIdBatches) =>
```
teneightfive: Maybe make this a function declaration rather than a function expression.

app/moves/middleware.js
```diff
@@ -90,4 +94,36 @@ module.exports = {
...
+  setMovesByDateAllLocations: async (req, res, next) => {
+    const { moveDate } = res.locals
+
+    if (!moveDate) {
+      return next()
+    }
+
+    try {
+      const userLocations = get(req.session, 'user.locations', []).map(
+        loc => loc.id
```
teneightfive: Maybe use more verbose variable names like we've been using so far: `location => location.id`. I find more verbose names help when returning to code after long periods of time.

app/moves/middleware.js
teneightfive: I wonder if returning `next()` if there are no locations rather than pushing `null` into the array and running the rest of the function is neater if we feel we still need this here?

config/index.js
```diff
@@ -112,4 +112,5 @@ module.exports = {
       sortOrder: 2,
     },
   },
+  MOVES_BATCH_SIZE: process.env.MOVES_BATCH_SIZE || 40,
```
teneightfive: Could we add this to the readme too please :)

I wonder if `LOCATIONS_BATCH_SIZE` might also be a better a name as its referring to when we have lots of locations rather than moves

README.md
```diff
@@ -169,6 +169,7 @@ npm run lint
 | E2E_POLICE_PASSWORD | Police user password used for acceptance testing | |
 | E2E_SUPPLIER_USERNAME | Supplier user username used for acceptance testing | |
 | E2E_SUPPLIER_PASSWORD | Supplier user password used for acceptance testing | |
+| LOCATIONS_BATCH_SIZE | Maximum number of location IDs to send in one request when requesting moves for all locations | |
```
teneightfive: We should probably add the default as the third column

PR #319, opened 2020-01-23
Feture flag document upload feature

teneightfive:
This change adds the ability to enable or disbale the document upload
feature using environment variables.

### Checklist

### DIFF
--- a/app/move/controllers/create/document.js
+++ b/app/move/controllers/create/document.js
@@ -4,10 +4,11 @@ const { get } = require('lodash')
 const documentService = require('../../../../common/services/document')
 const moveService = require('../../../../common/services/move')
 const CreateBaseController = require('./base')
-const { API } = require('../../../../config')
+const { API, FEATURE_FLAGS } = require('../../../../config')
 
 class DocumentUploadController extends CreateBaseController {
   middlewareChecks() {
+    this.use(this.checkFeatureEnabled(FEATURE_FLAGS.DOCUMENTS))
     this.use(this.parseMultipartForm)
     super.middlewareChecks()
   }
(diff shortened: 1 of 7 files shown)


app/move/steps/create.js
```diff
@@ -122,7 +123,7 @@ module.exports = {
   '/save': {
     skip: true,
     controller: Save,
-    next: 'document',
+    next: FEATURE_FLAGS.DOCUMENTS ? 'document' : 'final',
```
electrified: Should we have a test around the feature flag on/off behaviour?

teneightfive: I was even wondering if this was the best place for it. Because I didn't then feel comfortable testing a config file.

I might move this to the controller actually, that way it can be tested more neatly.

app/move/controllers/create/document.js
```diff
@@ -17,6 +18,16 @@ class DocumentUploadController extends CreateBaseController {
     this.use(this.populateDocumentUpload)
   }
 
+  checkFeatureEnabled(enabled) {
+    return function functionName(req, res, next) {
```
electrified: Perhaps use a function expression rather than giving it a generic name?

app/move/steps/create.js
```diff
@@ -108,7 +108,6 @@ module.exports = {
     controller: Assessment,
     next: 'save',
     pageTitle: 'moves::steps.health_information.heading',
-    buttonText: 'actions::continue',
```
electrified: Putting the comment here - not necessarily applicable to this line - I think the last button in the process should say "Schedule Move", at the moment it still says "Continue"

app/move/controllers/create/document.js
teneightfive: Doh! That was me trying to figure out a failing test. Good spot :)

app/move/steps/create.js
teneightfive: Ah! Good spot! I'll add this logic to the last step.

PR #326, opened 2020-01-28
Ensure download users correct middleware

teneightfive:
The download all moves route was using the middleware for
a single location rather than all locations and was there not
filtering the moves.

This fix ensures the routes use the correct controller.

### DIFF
--- a/app/moves/index.js
+++ b/app/moves/index.js
@@ -43,7 +43,7 @@ router.get(
 router.get(
   '/:date/download.:extension(csv|json)',
   protectRoute('moves:download:all'),
-  setMovesByDateAndLocation,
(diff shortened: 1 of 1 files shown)


app/moves/index.js
```diff
@@ -43,7 +43,7 @@ router.get(
 router.get(
   '/:date/download.:extension(csv|json)',
   protectRoute('moves:download:all'),
-  setMovesByDateAndLocation,
+  setMovesByDateAllLocations,
```
starswan: I know this is an emergency fix, but shouldn't we be writing a failing test? If we had had a test for this scenario originally, then the regression might not have happened?

PR #329, opened 2020-01-31
[P4-938] Remove deletion of move on cancel action

teneightfive:
This work removes the changes that were introduced to allow a move to be deleted via the API.

It reverts some commits for simplicity and then uses another commit to revert work that was added as part of another commit.


teneightfive:
> I imagine this is in the context of the work around the uploads?

Yep, at the moment this feature is hidden behind a feature flag so this change won't affect any current journeys. But this is something we need to change for the re-work to how documents are uploaded.

PR #330, opened 2020-02-03
[P4-935] Update multi field component

teneightfive:
This change is linked to P4-935 which will allow documents to be uploaded without the need for a move ID.

As part of this work some updates to the multi file upload component were required so this work is being split out to make it easier to review.

See breakdown of commits for reasoning behind changes.

### DIFF
--- a/common/components/multi-file-upload/multi-file-upload.js
+++ b/common/components/multi-file-upload/multi-file-upload.js
@@ -293,10 +293,7 @@ MultiFileUpload.prototype = {
         )
 
         $uploadActions.appendChild(
-          this.buildDeleteButton(
-            documentDetail.id,
-            documentDetail.attributes.filename
-          )
+          this.buildDeleteButton(documentDetail.id, documentDetail.filename)
         )
       })
       .catch(errors => this.renderErrors(errors, $fileUploadRow))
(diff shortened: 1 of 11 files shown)


common/components/multi-file-upload/multi-file-upload.js
```diff
@@ -315,7 +315,7 @@ MultiFileUpload.prototype = {
     if (errorData.length === 0) {
       errorData.push({
         text: 'Something went wrong',
-        href: '#document_upload',
+        href: '#' + this.$fileInput.id,
```
barbaracassani: Extremely minor, but maybe let's use a string template? `#${this.$fileInput.id}` for consistency's sake? If you don't want to do it no problem, string concatenation is arguably more readable.

common/components/multi-file-upload/template.njk
```diff
@@ -4,39 +4,38 @@
...
+    {% for file in params.value %}
+    <div class="govuk-summary-list__row app-row js-upload-row" data-document-id="{{ file.id }}">
       <dt class="govuk-summary-list__key app-row__key">
-        {{ document.filename }}
+        {{ file.filename }}
       </dt>
       <dd class="govuk-summary-list__value app-row__value js-upload-message">
         <span class="app-multi-file-upload-message app-multi-file-upload-message__success">
           <span class="app-icon app-icon--tick"></span>File Uploaded!
         </span>
```
barbaracassani: I understand we don't do i18n in templates?

PR #331, opened 2020-02-03
[P4-935] Refactor documents controller to upload documents without move ID

teneightfive:
This change is a refactor to the way the documents controller
was handling file uploads.

It tries to simplify some of the methods and make more use of the
HMPO Form Wizard lifecycle.

Some major changes:
- replace [formidable](https://github.com/node-formidable/node-formidable) with [multer](https://github.com/expressjs/multer)
  Multer can be used as part of middleware which simplifies how it is loaded and how errors are handled
- remove DELETE handler in place of normal POST
- reduce logic within POST


app/move/controllers/create/document.js
```diff
@@ -1,24 +1,46 @@
...
-    this.use(this.checkFeatureEnabled(FEATURE_FLAGS.DOCUMENTS))
-    this.use(this.parseMultipartForm)
+    this.use(this.isEnabled(FEATURE_FLAGS.DOCUMENTS))
     super.middlewareChecks()
   }
 
-  middlewareLocals() {
-    super.middlewareLocals()
-    this.use(this.populateDocumentUpload)
+  setNextStep(req, res, next) {
```
barbaracassani: Could this be let to HOF to handle, as a fork?

app/move/controllers/create/document.js
```diff
@@ -28,167 +50,64 @@ class DocumentUploadController extends CreateBaseController {
...
+    let documents = []
 
     try {
-      if (id) {
-        res.locals.move = await moveService.getById(id)
-        documentUpload.documents = res.locals.move.documents
+      if (req.files) {
+        const uploaded = await Promise.all(
+          req.files.map(file => documentService.create(file))
+        )
```
barbaracassani: V. nice

app/move/controllers/create/document.js
```diff
@@ -1,24 +1,46 @@
...
+
+const upload = multer({
+  dest: FILE_UPLOADS.UPLOAD_DIR,
+  limits: {
+    fileSize: FILE_UPLOADS.MAX_FILE_SIZE,
+  },
+})
+
+// TODO:
+// - bug with simulteaneous xhr uploads
```
barbaracassani: ? It's a bit cryptic / scary :-) Maybe add a reference to a Jira ticket?

common/services/document.js
```diff
@@ -14,6 +14,17 @@ const logErrorMessage = error => {
...
+      filename: data.originalname,
+      id: data.filename,
+    })
+  },
+
+  destroy(id) {
+    return Promise.resolve(id)
+  },
+
   async upload(file, moveId) {
```
electrified: Does anything delete the files out of the temporary directory once they are uploaded? Also thinking that ./tmp/ will be a path inside the container and wondering how much space we will have?

config/index.js
```diff
@@ -44,13 +44,16 @@ module.exports = {
...
-    MAX_FILE_UPLOAD_SIZE: 50 * 1024 * 1024, // 50 mega bytes
   },
   DATE_FORMATS: {
     SHORT: 'd M yyyy',
     LONG: 'd MMM yyyy',
     WITH_DAY: 'EEEE d MMM yyyy',
   },
+  FILE_UPLOADS: {
+    UPLOAD_DIR: '.tmp/uploads/',
+    MAX_FILE_SIZE: 50 * 1024 * 1024, // 50mb
```
barbaracassani: Could it be cautious to have a total max size? E.g. across all the uploaded documents.

app/move/controllers/create/document.js
teneightfive: The problem with doing it in the step config is that we don't have access to `req`. This is trying to say that if a file was uploaded or deleted with the form submission, redirect back to current URL rather than moving to the next step.

Maybe renaming this middleware might help? Something like `isUploadFormSubmission`?

config/index.js
teneightfive: It doesn't seem like that can be set from the [multer limits config](https://github.com/expressjs/multer#limits).

We could set a maximum number of files that can be uploaded?

@jason-sims what are your thoughts on limiting the amount of files a user can upload during this process?

common/services/document.js
teneightfive: Currently, no. But I'm working on another branch to address how the files are stored. And looking to create a separate storage class for uploading the files to the API instead of saving to disk.

app/move/controllers/create/document.js
teneightfive: Good call. Am going to move to a ticket instead.

teneightfive: [P4-976](https://dsdmoj.atlassian.net/browse/P4-976)

config/index.js
teneightfive: @spikeheap might also be one for you ☝️

PR #333, opened 2020-02-05
[P4-978] Ensure current location is persisted on session regeneration

teneightfive:
Currently there is a bug where if the session expires on the frontend,
it will attempted to redirect to the auth service which will make a
call to see if the user is still authenticated.

If they are, it will regenerate the session, persisting some values.

However, the `currentLocation` value is not persisted across this
regeneration which means that the check locations middleware then
redirects the user back to the locations mountpath to choose their
location again even though, in their eyes, they were never logged
out.

To fix this, `req.session.currentLocation` is now persisted across
this session regeneration so that the user remains on the page they
were before the re-authentication.

### DIFF
--- a/app/auth/middleware.js
+++ b/app/auth/middleware.js
@@ -4,7 +4,7 @@ const { decodeAccessToken } = require('../../common/lib/access-token')
 
 function processAuthResponse() {
   return async function middleware(req, res, next) {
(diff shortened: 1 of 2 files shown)


teneightfive:
@starswan I think you have come across this issue before so this change should fix being redirected back to the choose location page without having been signed out.

PR #335, opened 2020-02-06
[P4-935] Save documents without move ID

teneightfive:
This change reworks the document service to allow documents to be saved to the API without needing a move ID.

It also adds a [custom storage engine](https://github.com/expressjs/multer/blob/master/StorageEngine.md) for Multer to save the files directly to the document service rather than to disk.

The previous service also needed an additional library to be created for authorised requests which  has been removed with this change as its no longer needed.

## Todo

- [x] write test for storage engine


common/lib/multer.api-document.storage.js
```diff
@@ -0,0 +1,28 @@
+const concat = require('concat-stream')
+
+const documentService = require('../services/document')
+
+function APIDocumentStorage() {}
+
+APIDocumentStorage.prototype._handleFile = async function _handleFile(
```
barbaracassani: No ES6? :-)

PR #336, opened 2020-02-06
Assessment questions/custom types 2

barbaracassani:
Feb-07-2020 15-48-54

Screenshot 2020-02-07 at 15 52 35

This PR is for P4-961, the special vehicle question.
Although it's a normal assessment question, it needs to be represented with a radio button instead of a checkbox, because it's mandatory. Instead of creating a special field on the template, we have added a configuration option: `explicit: true` that put on a field will transform it in a radio, whilst otherwise treating it as any other assessment question.

I have paired with @teneightfive for most of it.

### Checklist


app/move/controllers/create/assessment.test.js
```diff
@@ -1,52 +1,217 @@
...
-    name: 'first_names',
-  },
-}
-
 describe('Move controllers', function() {
   describe('Assessment controller', function() {
+    let nextSpy
+    let refDataService
+    let req
+    describe('#configure() 2', function() {
```
electrified: Perhaps merge the 2 describe blocks testing configure() and have separate `context`s for the different setup data?

app/move/controllers/create/assessment.test.js
```diff
@@ -1,52 +1,217 @@
...
+        expect(category).to.equal('health')
+        expect(field).to.equal('special_diet_or_allergy')
+        createFields.appendDependent.restore()
+      })
+      it('invokes decorateWithExplicitFields per each field', async function() {
+        sinon.stub(createFields, 'decorateWithExplicitFields')
+        await controller.configure(req, {}, nextSpy)
+        expect(createFields.decorateWithExplicitFields.callCount).to.equal(3)
+        let key
+        ;[, , , key] = createFields.decorateWithExplicitFields.getCall(0).args
```
electrified: Perhaps get the 4th item out of the array normally rather than destructuring?

app/move/controllers/create/assessment.js
```diff
@@ -19,25 +51,22 @@ class AssessmentController extends CreateBaseController {
...
-          }
-        })
-    })
+    const { assessmentCategory } = req.form.options
+    const implicitAnswers = req.form.values[assessmentCategory]
+
+    assessment[assessmentCategory] = req.questions
+      .filter(question => {
+        if (implicitAnswers.includes(question.id)) {
+          return true
```
teneightfive: Could reduce this to:

```js
.filter({ id, key } => {
  return implicitAnswers.includes(id) || req.form.values[`${key}__yesno`] === 'yes'
})
```

app/move/controllers/create/assessment.js
```diff
@@ -1,15 +1,47 @@
...
-      req.form.options.fields = await fieldHelpers.populateAssessmentQuestions(
-        fields
+      const { fields, assessmentCategory } = req.form.options
+      const questions = await referenceDataService.getAssessmentQuestions(
+        assessmentCategory
       )
+
+      const implicitField = createFields.assessmentCategory(assessmentCategory)
+      implicitField.items = questions
+        .filter(question => {
```
teneightfive: Maybe move this to a field helper too. Will make it easier to test in isolation.

It could also then be reduced to this:

```js
...
.filter(fieldHelpers.filterExplicitFields(fields))
...
```

```js
function filterExplicitFields(fields) {
  return { key } => {
    const key = question.key
    const field = fields[key]

    return field && !field.explicit
  }
}
```

teneightfive: I would probably try and extract all of this logic into a field helper too. That way it's not tied specifically to this controller and you can test each method in a simpler way.

It would be nice to reduce the controller to just one call to a helper that returns an unmutated set of fields, like before.

app/move/controllers/create/assessment.js
```diff
@@ -1,15 +1,47 @@
...
+        .map(fieldHelpers.mapReferenceDataToOption)
+
+      req.form.options.fields[assessmentCategory] = implicitField
+
+      req.form.options.fields = mapValues(
+        req.form.options.fields,
+        createFields.appendDependent.bind(null, questions, assessmentCategory)
+      )
+
+      req.form.options.fields = forEach(
```
teneightfive: I wonder if we can also combine this loop with the map. At the moment we're having this controller loop over all fields twice. Could the map call a method within it to append the explicit fields?

app/move/fields/create.js
```diff
@@ -38,6 +40,67 @@ function assessmentCategory(category) {
...
+      },
+      {
+        value: 'no',
+        text: 'No',
+      },
+    ],
+  }
+}
+
+function appendDependent(questions, assessmentCategory, field, key) {
```
teneightfive: This method and `decorateWithExplicitFields` should probably live in the [field helpers](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/helpers/field.js) file.

They're more helper methods than field definitions related specifically to the create journey.

app/move/controllers/create/assessment.js
barbaracassani: Whilst I'd love to get rid of the additional loop, I don't quite see how this could work. In fact, we do mutate `req.form.options.fields` -- this is how we append the additional field. I mean, we rely on a side effect. At least in this way it's fairly isolated...

teneightfive: But we could move the mutation into a separate method that creates a copy of the original fields and appends to that, then returns it.

Trying to apply more of a functional approach where methods return a value rather than updating things too much of their scope.

Granted, we need to mutate the fields object within the controller, but this is a slight limitation of the Form Wizard. What we can try and do is reduce the complexity of these configure methods to call our helpers which are written in a more functional way. That way our helpers are more easily tested as they take an input and we can assert the output.

And our tests for the controllers which test mutation stay as light as possible too.

barbaracassani: Ok, makes sense.

common/helpers/field.js
```diff
@@ -246,14 +212,99 @@ function mapPersonToOption(person) {
...
+  const question = find(questions, { key })
+
+  if (question && field.explicit) {
+    const explicitKey = `${key}__yesno`
+    const explicitField = explicitYesNo(explicitKey)
+    explicitField.items[0].conditional = key
+    collection[explicitKey] = explicitField
+  }
+}
+function explicitYesNo(name) {
```
teneightfive: I think this should still remain in the fields file as this isn't doing any logic. It's a skeleton of the field we want to create.

I think after this, we can get this change in for now. We may need to refactor again to cater for new journeys with Prison -> Court and Prison -> Prison too.

barbaracassani: `explicitYesNo` is called by a func that is in `field.js` -- I reckon it's less awkward to leave it in the same file instead of adding cross-calls.

barbaracassani: Ok, done.

PR #339, opened 2020-02-11
[P4-742] Move button logic to controller

teneightfive:
This change moves the logic to set the correct button text for the
create journey, whether to display continue or schedule move, to
the controller so that this logic doesn't need to live in each step.

It also makes it easier when using the forking via next as the button
text will be dynamically set based on the next step of each current
step.

### DIFF
--- a/app/move/controllers/create/base.js
+++ b/app/move/controllers/create/base.js
@@ -10,9 +10,23 @@ class CreateBaseController extends FormWizardController {
 
   middlewareLocals() {
     super.middlewareLocals()
+    this.use(this.setButtonText)
     this.use(this.setCancelUrl)
     this.use(this.setMoveSummary)
     this.use(this.setJourneyTimer)
+  }
+
+  setButtonText(req, res, next) {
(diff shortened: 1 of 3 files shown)


app/move/controllers/create/base.js
```diff
@@ -10,11 +10,25 @@ class CreateBaseController extends FormWizardController {
...
 
+  setButtonText(req, res, next) {
+    const nextStep = this.getNextStep(req, res)
+    const steps = Object.keys(req.form.options.steps)
+    const lastStep = steps[steps.length - 1]
+    const buttonText = nextStep.includes(lastStep)
+      ? 'actions::schedule_move'
+      : 'actions::continue'
+
+    req.form.options.buttonText = req.form.options.buttonText || buttonText
```
barbaracassani: In which cases does `buttonText` already exist on `options`?

teneightfive: At moment, none in this journey. It was more of a forethought for some of the search steps. We might want the ability to have the button text say `Search` instead of `Continue`.

PR #352, opened 2020-02-18
Only notify slack on CircleCI failures

teneightfive:
This uses the slack Orb in CircleCI config to only alert the channel
on master failures, not branches.


spikeheap:
🙌 ❤️

PR #353, opened 2020-02-18
[P4-1060] Step for the move agreed

barbaracassani:
Not linked from anywhere atm.

Screenshot 2020-02-19 at 11 31 13

### Checklist

### DIFF
--- a/app/move/fields/create.js
+++ b/app/move/fields/create.js
@@ -342,6 +342,45 @@ module.exports = {
       text: 'fields::documents.hint',
     },
   },
+  move_agreed: {
+    validate: 'required',
+    component: 'govukRadios',
+    name: 'move_agreed',
+    fieldset: {
+      legend: {
+        text: 'fields::move_agreed.label',
+        classes: 'govuk-visually-hidden govuk-fieldset__legend--m',
+      },
+    },
+    items: [
+      {
+        id: 'move_agreed',
+        value: true,
+        text: 'Yes',
+        conditional: 'move_agreed_by',
+      },
+      {
+        value: false,
+        text: 'No',
+      },
+    ],
+  },
+  move_agreed_by: {
+    id: 'move_agreed_by',
+    name: 'move_agreed_by',
+    skip: true,
+    component: 'govukInput',
+    classes: 'govuk-input--width-20',
+    label: {
+      text: 'fields::move_agreed_by.label',
+      classes: 'govuk-label--s',
+    },
+    validate: 'required',
+    dependent: {
+      field: 'move_agreed',
+      value: true,
+    },
+  },
   assessmentCategory,
   explicitYesNo,
 }
--- a/app/move/steps/create.js
+++ b/app/move/steps/create.js
@@ -83,6 +83,10 @@ module.exports = {
       'date_custom',
     ],
   },
+  '/agreement-status': {
+    pageTitle: 'moves::agreement_status.heading',
+    fields: ['move_agreed', 'move_agreed_by'],
+  },
   '/court-information': {
     controller: Assessment,
     pageTitle: 'moves::steps.court_information.heading',
--- a/locales/en/fields.json
+++ b/locales/en/fields.json
@@ -137,5 +137,11 @@
     "heading": "Only include documents that will help plan the move.",
     "label": "Upload a file (optional)",
     "hint": "You can upload Word, Excel, PDF and JPEG documents up to 50 megabytes"
+  },
+  "move_agreed": {
+    "label": "Move agreement status"
+  },
+  "move_agreed_by": {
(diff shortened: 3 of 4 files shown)


app/move/fields/create.js
```diff
@@ -336,6 +336,40 @@ module.exports = {
...
+        text: 'Yes',
+        conditional: 'move_who_agreed',
+      },
+      {
+        value: 'no',
+        text: 'No',
+      },
+    ],
+  },
+  move_who_agreed: {
```
teneightfive: Maybe match these to the new API attributes we wrote down - `move_agreed_by`?

app/move/steps/create.js
```diff
@@ -77,6 +77,11 @@ module.exports = {
       'date_custom',
     ],
   },
+  '/move-agreed': {
+    controller: Base,
```
teneightfive: This is already inherited from the main config so no need for it here.

app/move/steps/create.js
```diff
@@ -77,6 +77,11 @@ module.exports = {
...
+    pageTitle: 'moves::details.heading',
```
teneightfive: Worth matching to the step name for consistency. Wonder if `/agreement-status` is a better name for the step too as `/move-agreed` suggests a positive response rather than what we're capturing.

locales/en/moves.json
```diff
@@ -114,5 +114,8 @@
       "content": "Reference number<br><strong>{{moveReference}}</strong>"
     },
     "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled."
+  },
+  "details": {
+    "heading": "Move agreed"
```
teneightfive: We can also do what we do with the assessment question steps and use the question as the heading, then use the visibility class on the fieldset legend.

locales/en/moves.json
```diff
@@ -114,5 +114,8 @@
       "content": "Reference number<br><strong>{{moveReference}}</strong>"
     },
     "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled."
+  },
+  "details": {
```
teneightfive: Keys in the steps object in locales should match the step name so they're easier to find: `agreement-status`

locales/en/fields.json
```diff
@@ -135,5 +135,11 @@
     "heading": "Only include documents that will help plan the move.",
     "label": "Upload a file (optional)",
     "hint": "You can upload Word, Excel, PDF and JPEG documents up to 50 megabytes"
+  },
+  "move_agreed": {
+    "label": "Has this move been agreed with receiving prison?"
```
teneightfive: What about "Move agreement status" as this label is then used for validation and I think that would read better.

So we would have

```
Move agreement status cannot be blank
```

instead of 

```
Has this move been agreed with receiving prison? cannot be blank
```

app/move/fields/create.js
```diff
@@ -336,6 +336,40 @@ module.exports = {
...
+        conditional: 'move_agreed_by',
+      },
+      {
+        value: 'no',
+        text: 'No',
+      },
+    ],
+  },
+  move_agreed_by: {
+    id: 'move_agreed_by',
```
teneightfive: Just spoke to @Andy-Millar and this should be required if they answer `true`.

You'll need to add the `dependent` property too:

```javascript
move_agreed_by: {
  ...
  validate: 'required',
  dependent: {
    field: 'move_agreed',
    value: 'yes',
  }
  ...
}
```

app/move/fields/create.js
```diff
@@ -336,6 +336,40 @@ module.exports = {
...
+    fieldset: {
+      legend: {
+        text: 'fields::move_agreed.label',
+        classes: 'govuk-visually-hidden govuk-fieldset__legend--m',
+      },
+    },
+    items: [
+      {
+        id: 'move_agreed',
+        value: 'yes',
```
teneightfive: This is probably fine for now, but we may want to update the values to `true`/`false` as that value will be sent directly to the API and will likely be how we store it rather than `yes`/`no`.

app/locations/middleware.js
```diff
@@ -9,7 +9,6 @@ function checkLocationsLength(req, res, next) {
   if (req.userLocations.length === 1) {
     return res.redirect(`${req.baseUrl}/${req.userLocations[0].id}`)
   }
-
```
teneightfive: Rogue change crept in?

barbaracassani: Yes -- I had removed one blank line. Now duly restored.

PR #354, opened 2020-02-19
[P4-1061] Step for the move request reason

barbaracassani:
This PR creates the `/prison-transfer-reason` for the OCA flow.

Screenshot 2020-03-02 at 12 57 19

### Checklist

### DIFF
--- a/app/move/controllers/create/prison-transfer-reason.test.js
+++ b/app/move/controllers/create/prison-transfer-reason.test.js
@@ -1,0 +1,68 @@
+const fieldHelpers = require('../../../../common/helpers/field')
+const referenceDataService = require('../../../../common/services/reference-data')
+
+const FormController = require('hmpo-form-wizard').Controller
+
+const PrisonTransferReasonsController = require('./prison-transfer-reason')
+const prisonTransferReasonsController = new PrisonTransferReasonsController({
+  route: '/prison_transfer_reason',
+})
+
+describe('Prison transfer reason controller', function() {
+  let mockRequest
+
+  context('happy path', function() {
+    beforeEach(async function() {
+      mockRequest = {
+        form: {
+          options: {
+            fields: {
+              prison_transfer_reason: {},
+            },
+          },
+        },
+      }
+      sinon.stub(FormController.prototype, 'configure')
+      sinon.stub(fieldHelpers, 'mapReferenceDataToOption').returns({})
+      sinon
+        .stub(referenceDataService, 'getPrisonTransferReasons')
+        .resolves([{}, {}])
+      await prisonTransferReasonsController.configure(mockRequest, {}, () => {})
+    })
+    it('calls reference data service', function() {
+      expect(
+        referenceDataService.getPrisonTransferReasons
+      ).to.have.been.calledOnce
+    })
+    it('creates the correct options', function() {
+      expect(fieldHelpers.mapReferenceDataToOption).to.have.been.calledTwice
+    })
+    it('sets the correct reason items', function() {
+      expect(
+        mockRequest.form.options.fields.prison_transfer_reason.items
+      ).to.deep.equal([{}, {}])
+    })
+    it('invokes the parent configure', function() {
+      expect(FormController.prototype.configure).to.have.been.calledOnce
+    })
+  })
+  context('unhappy path', function() {
+    let stubForNextHandler
+    beforeEach(async function() {
+      sinon
+        .stub(referenceDataService, 'getPrisonTransferReasons')
+        .rejects('referenceDataServiceError')
+      stubForNextHandler = sinon.stub()
+      await prisonTransferReasonsController.configure(
+        {},
+        {},
+        stubForNextHandler
+      )
+    })
+    it('invokes the next() handler', function() {
+      expect(stubForNextHandler).to.have.been.calledWith(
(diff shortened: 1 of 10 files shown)


locales/en/moves.json
```diff
@@ -114,5 +114,8 @@
       "content": "Reference number<br><strong>{{moveReference}}</strong>"
     },
     "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled."
+  },
+  "reason": {
+    "heading": "Request type"
```
teneightfive: Could keep the consistency of using the question as the heading like we have on the other steps?

barbaracassani: In the prototype, the heading has different wording from the question.

app/move/fields/create.js
```diff
@@ -381,6 +381,34 @@ module.exports = {
       text: 'fields::documents.hint',
     },
   },
+  prison_transfer_reason: {
+    validate: 'required',
+    component: 'govukRadios',
+    name: 'reason',
```
teneightfive: This needs to match the key. It's one of the slightly annoying things to keep an eye out for, but its because we use this object to pass to the GOV.UK Design System components.

app/move/controllers/create/prison-transfer-reason.test.js
```diff
@@ -0,0 +1,71 @@
...
+      stubForMapReferenceDataToOption = sinon
+        .stub(fieldHelpers, 'mapReferenceDataToOption')
+        .returns({})
+      stubForGetPrisonTransferReasons = sinon
+        .stub(referenceDataService, 'getPrisonTransferReasons')
+        .resolves([{}, {}])
+      await prisonTransferReasonsController.configure(mockRequest, {}, () => {})
+    })
+    it('calls reference data service', function() {
+      expect(stubForGetPrisonTransferReasons).to.have.been.calledOnce
```
teneightfive: You can actually avoid having to define the stub as a variable and just check the original method:

In beforeEach:
```javascript
...
sinon.stub(fieldHelpers, 'mapReferenceDataToOption').returns({})
...
```

In it block:

```javascript
expect(fieldHelpers.mapReferenceDataToOption).to.have.been.calledOnce
```

I find its a little easier to read too as you're not having to look for what the variables are.

app/move/controllers/create/prison-transfer-reason.test.js
```diff
@@ -0,0 +1,71 @@
...
+    it('appends conditionals to the correct fields', function() {
+      expect(stubForMapReferenceDataToOption).to.have.been.calledTwice
+    })
+    it('sets the correct reason items', function() {
+      expect(
+        mockRequest.form.options.fields.prison_transfer_reason.items
+      ).to.deep.equal([{}, {}])
+    })
+    it('invokes the parent configure', function() {
+      expect(stubForParentConfigure).to.have.been.calledOnce
```
teneightfive: This can also just be:

```javascript
expect(FormController.prototype,.configure).to.have.been.calledOnce
```

Then no need to define an extra variable.

common/lib/api-client/models.js
```diff
@@ -117,6 +117,18 @@ module.exports = {
       collectionPath: 'reference/suppliers',
     },
   },
+  prison_transfer_reason: {
+    attributes: {
+      id: '',
+      type: '',
```
teneightfive: As we discussed we shouldn't include `type` as an attribute as its reserved in JSON:API. Same goes for `id`. They'll get put onto the base object by the [devour client library](https://github.com/twg/devour).

common/helpers/field.js
```diff
@@ -35,7 +35,8 @@ function mapReferenceDataToOption({ id, title, key, conditional, hint }) {
 }
 
 function mapAssessmentQuestionToTranslation(item) {
-  const { category, key, title } = item
+  const { key, title } = item
+  const category = item.category
```
teneightfive: Feels like we don't need this anymore? Can just revert to its original form.

locales/en/moves.json
teneightfive: Might be worth still considering this. But we can review when we look at the journey as a whole afterwards maybe.

app/move/controllers/create/prison-transfer-reason.test.js
barbaracassani: I am aware the variable is not strictly necessary, however I personally find it tidier this way. If that is a problem I am happy to change it.

app/move/steps/create.js
```diff
@@ -107,6 +108,12 @@ module.exports = {
     pageTitle: 'moves::steps.move_date.heading',
     fields: ['date_from', 'date_to'],
   },
+  '/prison-transfer-reason': {
+    controller: PrisonTransferReason,
+    checkJourney: false,
```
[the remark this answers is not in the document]
barbaracassani: It is, but indeed it's better I don't commit it. I'll remove it.

app/move/controllers/create/prison-transfer-reason.test.js
barbaracassani: Well it was quicker to change it than to discuss it, so I changed it 😄

PR #355, opened 2020-02-19
Automate releases

teneightfive:
This  adds a series of scripts and packages to the package manifest to help us automated releases.

It will allow us to run one command to release: `npm run release`. This will then do the following in the background:

- increase version number based on commit history
- update changelog with changes since last release
- create new git tag
- push git tags to remote
- run production builds on CI which are based on git tags
- publish release on GitHub


package.json
```diff
@@ -151,5 +146,62 @@
...
+    "start": "node start.js",
+    "test": "NODE_ENV=test LOG_LEVEL=silent mocha './app/**/*.test.js' './common/**/*.test.js' './config/**/*.test.js' --opts ./test/unit/mocha.opts",
+    "test-e2e:chrome": "testcafe chrome test/e2e/*.test.js",
+    "test-e2e:ci": "npm run test-e2e:chrome -- --color --reporter spec,xunit:reports/testcafe/results-chrome.xml",
+    "test-e2e:local": "E2E_BASE_URL='' npm run test-e2e:chrome -- --debug-on-fail",
+    "update": "ncu -u",
+    "update:check": "ncu -- --error-level 2",
+    "watch:js:client": "webpack --watch --progress",
+    "watch:js:server": "nodemon $NODE_DEBUG_OPTION --inspect",
+    "watch:test": "npm test -- --reporter=min --watch"
```
barbaracassani: We have lots of scripts. Maybe we should group them by theme, something like:
```
release: ...
release:pre: ...
```

teneightfive: I think `prerelease` is actually a npm structure. When running `release` it will automatically run `prerelease` before the script starts and `postrelease` after it runs.

PR #356, opened 2020-02-20
P4 1062

barbaracassani:
Display different options if the user has role Prison or has role Police

### Police

image

### Prison

image

### Checklist

### DIFF
--- a/app/move/controllers/create/move-details.js
+++ b/app/move/controllers/create/move-details.js
@@ -1,4 +1,4 @@
-const { set } = require('lodash')
+const { find, set } = require('lodash')
 const { format, startOfToday, startOfTomorrow, parseISO } = require('date-fns')
 
 const CreateBaseController = require('./base')
@@ -37,10 +37,45 @@ class MoveDetailsController extends CreateBaseController {
         date: filters.formatDateWithDay(res.locals.TOMORROW),
       })
 
+      const itemForPrisonRecall = find(
+        req.form.options.fields.move_type.items,
+        { value: 'prison_recall' }
+      )
+
+      itemForPrisonRecall.conditional = MoveDetailsController.isPolice(req)
+        ? 'additional_information'
+        : 'to_location_prison'
+      itemForPrisonRecall.text = MoveDetailsController.isPolice(req)
+        ? req.t('fields::move_type.items.prison_recall.labelForPrisonRecall')
+        : req.t('fields::move_type.items.prison_recall.label')
+
+      if (MoveDetailsController.isPolice(req)) {
+        delete req.form.options.fields.to_location_prison
+      } else {
+        delete req.form.options.fields.additional_information
+
+        const items = await this.populatePrisonAutocomplete()
+        set(req, 'form.options.fields.to_location_prison.items', items)
+      }
+
       super.configure(req, res, next)
     } catch (error) {
       next(error)
     }
+  }
+
+  static isPolice(req) {
+    return req.session.currentLocation.location_type === 'police'
+  }
+
+  async populatePrisonAutocomplete() {
+    const prisons = await referenceDataService.getLocationsByType('prison')
+    return fieldHelpers.insertInitialOption(
+      prisons
+        .filter(referenceDataHelpers.filterDisabled())
+        .map(fieldHelpers.mapReferenceDataToOption),
+      'prison'
+    )
   }
 
   process(req, res, next) {
--- a/app/move/controllers/create/move-details.test.js
+++ b/app/move/controllers/create/move-details.test.js
@@ -42,10 +42,23 @@ describe('Move controllers', function() {
 
           req = {
             t: sinon.stub().returns('__translated__'),
+            session: {
+              currentLocation: {
+                location_type: 'police',
+              },
+            },
             form: {
               options: {
                 fields: {
                   to_location_court_appearance: {},
+                  move_type: {
+                    items: [
+                      {
+                        value: 'prison_recall',
+                      },
+                    ],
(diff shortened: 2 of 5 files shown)


app/move/controllers/create/move-details.js
```diff
@@ -37,12 +37,35 @@ class MoveDetailsController extends CreateBaseController {
         date: filters.formatDateWithDay(res.locals.TOMORROW),
       })
 
+      const itemForPrisonRecall = find(
+        req.form.options.fields.move_type.items,
+        { value: 'prison_recall' }
+      )
+      if (req.session.user.role === 'Prison') {
```
teneightfive: We should avoid doing any login within this journey based on user role. It opens us up to problems if users have access to different location types.

app/move/controllers/create/move-details.js
```diff
@@ -37,12 +37,35 @@ class MoveDetailsController extends CreateBaseController {
...
+        itemForPrisonRecall.conditional = 'move_to_prison_name'
+        delete req.form.options.fields.additional_information
+        const items = await this.populatePrisonAutocomplete()
+        set(req, 'form.options.fields.move_to_prison_name.items', items)
```
teneightfive: It might be worth taking a look at the previous logic for this:
https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/c49713a0e582ddbaf17d4ea925855f650209f068#diff-14d99f63615581969e8020b943229313

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/bbeb37183ac66517c2725dc1ee97802f9d73894a#diff-4e531508b9cdbccfd41cd6b15225d987

I think currently it won't store the value of the prison as `from_location` which we'll need when sending to the API.

app/move/controllers/create/move-details.js
barbaracassani: Good point, I'll amend accordingly.

app/move/controllers/create/move-details.js
barbaracassani: Now it sends the correct data:
```
{
   "from_location": "e517a984-bb11-4738-aa93-229c22c2d459",
   "to_location": "240f8ea3-f3cf-4598-90b6-4c86c90d6dfc",
   "move_type": "prison_recall",
   "to_location_prison_recall": "240f8ea3-f3cf-4598-90b6-4c86c90d6dfc",
   "to_location_court_appearance": "",
   "date": "2020-02-21",
   "date_type": "today",
   "date_custom": "",
   "can_upload_documents": false
}
```

app/move/controllers/create/move-details.test.js
```diff
@@ -127,6 +140,145 @@ describe('Move controllers', function() {
...
+            }
+            res = {
+              locals: {},
+            }
+          })
+          afterEach(function() {
+            locationsStub.resetHistory()
+          })
+
+          it('when the current location type is prison appends to_location_prison_recall as conditional', async function() {
```
teneightfive: This is where `context` can come in useful for helping break down tests.

There are some examples in the rest of this file that may help. But something like this would be easier to read both in the code and the test outputs:

```javascript
context('when the current location type is prison', function() {
  beforeEach(async function() {
    req.session.currentLocation.location_type = 'prison'
    await controller.configure(req, res, nextSpy)
  })

  it('does thing one', function() {
    ...
  })

  it('does thing two', function() {
    ...
  })
})

context('when the current location type is police', function() {
  beforeEach(async function() {
    req.session.currentLocation.location_type = 'police'
    await controller.configure(req, res, nextSpy)
  })

  it('does thing one', function() {
    ...
  })

  it('does thing two', function() {
    ...
  })
})
```

Keeping `it` blocks to one or two `expect` statements also makes it much easier to debug tests when something fails.

app/move/controllers/create/move-details.test.js
```diff
@@ -197,86 +197,110 @@ describe('Move controllers', function() {
             locationsStub.resetHistory()
           })
 
-          it('when the current location type is prison appends to_location_prison_recall as conditional', async function() {
-            req.session = {
-              currentLocation: {
-                location_type: 'prison',
-              },
+          context(
+            'when the current location type is prison appends to_location_prison as conditional',
```
teneightfive: You could remove the text that now sits in the `if` texts. The context is effectively just saying what the condition is for that group of tests, so in this case: `when the current location type is prison`.

You can also drop the  `async` as there is no `await` used within this function anymore.

app/move/controllers/create/move-details.test.js
```diff
@@ -197,86 +197,110 @@ describe('Move controllers', function() {
...
-            ).to.be.true
-          })
-          it('when the current location type is police appends additional_information as conditional', async function() {
-            req.session = {
-              currentLocation: {
-                location_type: 'police',
-              },
+          )
+          context(
+            'when the current location type is police appends additional_information as conditional',
```
teneightfive: Same as the other context. Purpose is just to set the scene, not describe any outcomes. And can also remove `async` from the function defintion.

app/move/fields/create.js
```diff
@@ -242,7 +242,7 @@ module.exports = {
   to_location_court_appearance: toLocationType('court_appearance', {
     validate: 'required',
   }),
-  to_location_prison_recall: toLocationType('prison_recall', {
+  to_location_prison: toLocationType('prison_recall', {
```
teneightfive: The first argument needs to match  the property name: `to_location_prison: toLocationType('prison', {`. Otherwise the field won't be processed correctly.

PR #357, opened 2020-02-21
[P4-960] Person search logic

teneightfive:
Create logic to determine which person search step a user should see based on the location they are sending someone from.

## Sending someone from Police Custody Suite, Secure Training Centre or Secure Children's Home
police-journey

## Sending someone from Prison
prison-journey

### DIFF
--- a/app/move/controllers/create/base.js
+++ b/app/move/controllers/create/base.js
@@ -67,8 +67,15 @@ class CreateBaseController extends FormWizardController {
   }
 
   saveValues(req, res, next) {
-    req.form.values.from_location_type =
-      req.session.currentLocation.location_type
+    const {
+      id: currentLocationId,
+      location_type: locationType,
+      can_upload_documents: canUploadDocuments,
+    } = req.session.currentLocation
+
+    req.form.values.from_location = currentLocationId
+    req.form.values.from_location_type = locationType
+    req.form.values.can_upload_documents = canUploadDocuments
 
     super.saveValues(req, res, next)
   }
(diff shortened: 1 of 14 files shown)


app/move/controllers/create/base.js
```diff
@@ -65,6 +65,21 @@ class CreateBaseController extends FormWizardController {
...
     next()
   }
+
+  saveValues(req, res, next) {
+    const { currentLocation } = req.session
+    const {
+      id: currentLocationId,
+      location_type: locationType,
+      can_upload_documents: canUploadDocuments,
+    } = currentLocation
```
barbaracassani: You can probably do the destructuring in one go -- is there a reason why there's an interim variable? (not that is very important at all!).

app/move/views/create/no-results.njk
```diff
@@ -0,0 +1,7 @@
+{% extends "form-wizard.njk" %}
+
+{% block contentMain %}
+Can't create new prisoners
```
barbaracassani: Just to be pedantic, we can put this in the locale files.

`{{ t("person::no_results_text") }}`

teneightfive: Doh! These weren't meant to be committed! These are still WIP!

app/move/controllers/create/base.js
teneightfive: Not sure what you mean? So just use `currentLocation.<property>` directly?

teneightfive: Oh I see! `req.session.currentLocation`! Yep, thanks

app/move/views/create/no-results.njk
teneightfive: Also, no pedantic at all. We should try to always use the translations for content.

PR #358, opened 2020-02-24
[P4 1063] - Step for the move range input. Will appear only on journeys whose origin is a prison.

barbaracassani:
Step for the move range input. Will appear only on journeys whose origin is a prison.

Screenshot 2020-02-24 at 13 19 39

Screenshot 2020-02-24 at 13 19 08

### Checklist

### DIFF
--- a/app/move/fields/create.js
+++ b/app/move/fields/create.js
@@ -95,6 +95,29 @@ function explicitYesNo(name) {
   }
 }
 
+const dateField = {
+  validate: ['required', 'date', 'after'],
+  formatter: ['date'],
+  component: 'govukInput',
+  autocomplete: 'off',
+  classes: 'govuk-input--width-10',
+  hint: {
+    text: 'fields::date_custom.hint',
+  },
+}
+
+const dateRangeField = function(id) {
+  return {
+    ...dateField,
+    id: id,
+    name: id,
+    label: {
+      text: `fields::${id}.label`,
+      classes: 'govuk-label--s',
+    },
+  }
+}
+
 module.exports = {
   // Person search
   'filter.police_national_computer': {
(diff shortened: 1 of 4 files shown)


app/move/steps/create.js
```diff
@@ -152,6 +152,10 @@ module.exports = {
     pageTitle: 'moves::steps.document.heading',
     fields: ['documents'],
   },
+  '/date-range': {
```
teneightfive: This might be better called after what is happening on the step `/move-date` for consistency?

All our other step URLs attempt to explain what we're asking so we have a sensible URL structure.

app/move/fields/create.js
```diff
@@ -95,6 +95,29 @@ function explicitYesNo(name) {
...
+const dateRangeField = function(id) {
+  return {
+    ...dateField,
+    id: id,
+    name: id,
+    label: {
+      text: `fields::${id}.label`,
+      classes: 'govuk-label--s',
+    },
+    hint: {
```
teneightfive: If this isn't custom it could be moved to the `dateField` object?

app/move/fields/create.js
```diff
@@ -314,9 +335,9 @@ module.exports = {
     hint: {
```
teneightfive: Could move this to the shared `dateField` object too.

app/move/steps/create.js
teneightfive: Just noticed the position will also need to move. It should come over the `/move-details` step. The other of the properties in the steps object determine the order in the form wizard so its important they sit in the right place.

locales/en/moves.json
```diff
@@ -120,5 +120,8 @@
   },
   "agreement_status": {
     "heading": "Has this move been agreed with receiving prison?"
+  },
+  "date_range": {
```
teneightfive: Sorry, I didn't spot this earlier, but can we move this to the same place as the other step headings please - nested under `steps`. That way it will make them logically grouped and easier to find in the future.

PR #359, opened 2020-02-24
[P4-968] Prison risk logic

teneightfive:
This creates logic to split how different journeys will see the release status question.

Users sending some from Police/STC/SCH will see it asked as part of the risk step.

Prison users will see it asked as a separate step.

### From Police

image

### From Prison

image

### DIFF
--- a/app/move/controllers/create/assessment.js
+++ b/app/move/controllers/create/assessment.js
@@ -39,7 +39,7 @@ class AssessmentController extends CreateBaseController {
     const person = req.sessionModel.get('person') || {}
     const assessment = req.sessionModel.get('assessment') || {}
     const { assessmentCategory } = req.form.options
-    const implicitAnswers = req.form.values[assessmentCategory]
+    const implicitAnswers = req.form.values[assessmentCategory] || []
 
     assessment[assessmentCategory] = req.questions
       .filter(({ id, key }) => {
--- a/app/move/controllers/create/assessment.js
+++ b/app/move/controllers/create/assessment.js
@@ -12,18 +12,26 @@ class AssessmentController extends CreateBaseController {
       const questions = await referenceDataService.getAssessmentQuestions(
         assessmentCategory
       )
+      const implicitQuestions = questions.filter(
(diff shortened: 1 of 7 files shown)


app/move/controllers/create/assessment.js
```diff
@@ -12,18 +12,26 @@ class AssessmentController extends CreateBaseController {
...
 
-      req.form.options.fields[assessmentCategory] = implicitField
+        req.form.options.fields[assessmentCategory] = {
+          ...implicitField,
+          items: implicitQuestions
+            .map(fieldHelpers.mapAssessmentQuestionToConditionalField)
+            .map(fieldHelpers.mapAssessmentQuestionToTranslation)
+            .map(fieldHelpers.mapReferenceDataToOption),
+        }
+      }
```
barbaracassani: Maybe this block of code that populates `assessmentCategory` should go into its own method -- or possibly just the block that builds `items`.

app/move/fields/create.js
```diff
@@ -325,6 +328,10 @@ module.exports = {
   self_harm: assessmentQuestionComments,
   concealed_items: assessmentQuestionComments,
   other_risks: requiredAssessmentQuestionComments,
+  not_for_release: {
+    ...requiredAssessmentQuestionComments,
+    explicit: true,
```
barbaracassani: Glad it gets reused... :-)

app/move/controllers/create/assessment.js
teneightfive: I was kind of thinking of refactoring but didn't want to go down a rabbit hole just yet after we spent a fair bit of time on it last time. 

The main aim was to ensure it didn't create the implicit field if there were no implicit questions for that step.

teneightfive: I did start a refactor to try and remove the need for the `__yesno` suffix being required between a few files so I think we can revisit that later.

barbaracassani: 👍

PR #360, opened 2020-02-24
[P4-961] Add separate special vehicle step

teneightfive:
This change creates a separate step just for capturing the special
vehicle assessment question for journeys where someone is being sent
from Prison.

It uses the same logic for the risk/release status steps.

### Prison journey

image

### Police journey

image

@Andy-Millar I tweaked the label very slightly to match the rest of the page headings/field labels:

I swapped "they" for "this person".


jason-sims:
ok with me

PR #362, opened 2020-02-25
[P4-1086] Prevent prison person creation

teneightfive:
When a user searches for someone with a prison number and there are no
results we need to ensure that they can't create a new person in our
system.

This adds logic to not allow a user to progress to the personal
details step.

It also includes some changes to the display logic of the messages
depending on if there are results or not.

#### Person search

localhost_3000_move_new_person-lookup-pnc (1)

#### Search results

localhost_3000_move_new_person-lookup-results_filter police_national_computer =123 (2)

#### Search no results

localhost_3000_move_new_person-lookup-results_filter police_national_computer =123123123

#### Person search

localhost_3000_move_new_person-lookup-prison-number (2)

#### Search results

localhost_3000_move_new_person-lookup-results_filter police_national_computer =123 (3)

#### Search no results

localhost_3000_move_new_person-lookup-results_filter nomis_offender_no =123 (1)

### DIFF
--- a/app/move/views/create/person-search-results.njk
+++ b/app/move/views/create/person-search-results.njk
@@ -1,9 +1,23 @@
 {% extends "form-wizard.njk" %}
 
+{% set defaultActions %}
+  <a href="{{ backLink }}">{{ t("actions::search_again") }}</a>
+  {{ t("moves::steps.person_search_results.or") }}
+  <a href="?skip">
+    {{ t("moves::steps.person_search_results.move_someone_else") }}
+  </a>
+{% endset %}
+
+{% set prisonActions %}
+  <a href="{{ backLink }}">{{ t("actions::search_again") }}</a>
+{% endset %}
+
 {% block beforeFields %}
-  <h2 class="govuk-heading-l">
+  <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
     {{ t("moves::steps::person_search_results::secondary_heading", {
-      count: resultCount
+      type: "person" | pluralize(resultCount),
+      count: resultCount,
+      searchTerm: "123"
     }) }}
   </h2>
 {% endblock %}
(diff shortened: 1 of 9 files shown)


jason-sims:
Suggested text for Police - I think it may confuse some people
--- 

**Step One**

I do not have this number?
>Book a secure move uses the PNC number to ensure moves are linked to the correct person.

>If you do not have the PNC number you can [move a person without a PNC number]. _**(New person with blank PNC)**_

**Step Two**

3 people found for ` PNC number`

I can't find the person I'm looking for?
> [Search again]  _**(Go back to start of search)**_
> [Use this PNC number anyway]  _**(New person with populated PNC)**_
> [Move a person without a PNC number] _**(New person with blank PNC)**_


Suggested text for Prison
---


**Step One**

I do not have this number?
>Book a secure move uses the prison number to ensure moves are linked to the correct person.

>Ensure the person you are looking for is in NOMIS 

**Step Two**

3 people found for ` Prison number`

I can't find the person I'm looking for?
> [Search again]  _**(Go back to start of search)**_

>Ensure the person you are looking for is in NOMIS

Andy-Millar:
Results page - title should be different, as it's a different page. If we retained the search input on the page would be different. It's just a bit odd with back link navigating to page with same title. It should also replay the search term, as per my screens. 

I've got it as "X matches for A12345"

'People' is more human centered, could also be:
H1: X people found for A12345 (as Jason suggested)

Then the title should be as per my screens with the radio options - I think it's something like 'Select the person you want to move' -bold, lede copy. 

Details copy - shouldn't have a question mark on the one - otherwise I am fine with either. I think giving a benefit for entering the number may help with them skipping it / finding a workaround.

app/move/views/create/person-search.njk
```diff
@@ -3,9 +3,31 @@
...
+    {% endif %}
+  {% endset %}
+
+  {{ govukDetails({
+    summaryText: t("moves::steps.person_search.no_number"),
+    html: detailsHTML
+  }) }}
+
+  {# {% if values.from_location_type == "prison" %}
+    {{ govukInsetText({
```
barbaracassani: Commented code. Is that intentional?

teneightfive: Nope! Good spot :)

teneightfive:
@jason-sims I've update the screenshots and some copy based on the comments.

@Andy-Millar and I sat together again and tweaked some ever so slightly but let's catch up in person as will be easier. 

I think the only main differences are
- not providing a link to create a new person without a PNC after a user has specifically searched with a PNC, we should probably try and encourage them to continue to give us that if creating a new person
- and not including question marks as we didn't feel they're questions, they're the answer to a question

I've also included the search term in the results heading and Andy is happy with it remaining as the subheading after we linked them more visually. 

We've also updated the actions to ensure the user doesn't get left with buttons that won't do anything for that scenario, like no results - you can't just "Continue".

Andy-Millar:
Got a good solution that works well now - much easier chatting through in
person

app/move/views/create/person-search-results.njk
```diff
@@ -1,25 +1,72 @@
...
+{% set prisonActions %}
+  <a href="{{ backLink }}">{{ t("actions::search_again") }}</a>
+{% endset %}
+
 {% block beforeFields %}
-  <h2 class="govuk-heading-l">
+  <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
     {{ t("moves::steps::person_search_results::secondary_heading", {
-      count: resultCount
+      type: "person" | pluralize(resultCount),
```
barbaracassani: 👍 nice!

PR #364, opened 2020-02-26
[P4 1064] Skeleton of the dashboard that will be displaying proposed moves

barbaracassani:
To note:
- this covers the story for the dashboard route only
- displays active moves instead of proposed moves
- displays moves by day instead of by week (will be tackled in P4-1065)
- has no download button

Screenshot 2020-02-26 at 16 40 38
### Checklist

### DIFF
--- a/common/services/move.js
+++ b/common/services/move.js
@@ -42,6 +42,18 @@ const moveService = {
       })
   },
 
+  getProposed({ createdAtRange = [], fromLocationId } = {}) {
+    const [createdAtFrom, createdAtTo] = createdAtRange
+    return moveService.getAll({
+      filter: {
+        'filter[status]': 'proposed',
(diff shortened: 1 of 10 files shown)


common/services/move.js
```diff
@@ -42,6 +42,18 @@ const moveService = {
       })
   },
 
+  getProposed({ createdAtRange = [], fromLocationId } = {}) {
```
teneightfive: Would it be easier to pass in the from and to as separate arguments rather than an array?

barbaracassani: I see your point, but we actually use this pattern elsewhere. I quite like it -- it's semantic, in this instance, and reminiscent of the concept of tuple, that we would use here if we had it in javascript 😃

PR #366, opened 2020-02-26
Remove the need for user roles

teneightfive:
This refactor removes the only current use of a user's role within
the codebase - sending journey times to analytics.

This logic is also no longer what we use to determine different journeys.

It also includes reworking the location of it to be confined to the
actual create journey and its session so we don't need to manipulate
the session within an analytics library and avoids us neeeding to pass
in the request object.

### DIFF
--- a/app/move/controllers/create/save.js
+++ b/app/move/controllers/create/save.js
@@ -1,8 +1,9 @@
-const { omit } = require('lodash')
+const { omit, capitalize } = require('lodash')
 
 const CreateBaseController = require('./base')
 const moveService = require('../../../../common/services/move')
 const personService = require('../../../../common/services/person')
+const analytics = require('../../../../common/lib/analytics')
 
 class SaveController extends CreateBaseController {
   async saveValues(req, res, next) {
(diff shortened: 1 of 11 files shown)


common/lib/user.js
```diff
@@ -4,24 +4,9 @@ function User({ fullname, roles = [], locations = [] } = {}) {
...
-    }
-
-    if (roles.includes('ROLE_PECS_POLICE')) {
-      role = 'Police'
-    }
-
-    return role
-  },
-
   getPermissions(roles = []) {
```
barbaracassani: V. nice. The coexistence of `role` and `roles` was slightly confusing.

app/move/controllers/create/save.js
```diff
@@ -23,13 +24,26 @@ class SaveController extends CreateBaseController {
...
+      new Date().getTime() - req.sessionModel.get('journeyTimestamp')
+    )
 
-    req.journeyModel.reset()
-    req.sessionModel.reset()
-
-    res.redirect(`/move/${id}/confirmation`)
+    try {
+      await analytics.sendJourneyTime({
+        utv: capitalize(req.form.options.name),
```
barbaracassani: I understand we call that `name` already, but we mean something like `journeyName`. Maybe we could think of renaming it at some point to make it clearer (after all, we use names for people and for fields as well, and several other things).

barbaracassani: Oops, scratch that -- it comes from HOF, doesn't it.

teneightfive: Yep, unfortunately it's a HOF thing :(

PR #371, opened 2020-02-28
Update person search

teneightfive:
This includes a couple of small changes to the person search.

- Fix one of the translation keys used
- Make sure the search term is displayed correctly

## Before

image

## After

image

### DIFF
--- a/app/move/views/create/person-search-results.njk
+++ b/app/move/views/create/person-search-results.njk
@@ -29,7 +29,7 @@
     {% set html %}
       {% if values.from_location_type == "prison" %}
         <p class="govuk-body">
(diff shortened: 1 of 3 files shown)


app/move/views/create/person-search-results.njk
```diff
@@ -17,7 +17,7 @@
     {{ t("moves::steps::person_search_results::secondary_heading", {
       type: "person" | pluralize(resultCount),
       count: resultCount,
-      searchTerm: "123"
+      searchTerm: searchTerm
```
barbaracassani: Can use shorthand here

teneightfive: Sadly can't as its nunjucks, not JavaScript :(

PR #373, opened 2020-03-02
[P4-1106] Update assessment question helpers

teneightfive:
This refactors the way the assessment questions are populated. The main thing is to move the logic out of the controller and into one central place.

It also means we don't have to know about the suffix in too many different places. Only the locales files now.

### DIFF
--- a/app/move/fields/create.js
+++ b/app/move/fields/create.js
@@ -30,20 +30,20 @@ const requiredAssessmentQuestionComments = {
   validate: 'required',
 }
 
-function assessmentCategory(category) {
+function implicitAssessmentQuestions(name) {
   return {
+    name,
     component: 'govukCheckboxes',
     multiple: true,
     items: [],
-    name: category,
     fieldset: {
       legend: {
-        text: `fields::${category}.label`,
+        text: `fields::${name}.label`,
         classes: 'govuk-visually-hidden govuk-fieldset__legend--m',
       },
     },
     hint: {
-      text: `fields::${category}.hint`,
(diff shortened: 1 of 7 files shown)


common/helpers/field.js
```diff
@@ -68,7 +63,7 @@ function renderConditionalFields([key, field], index, obj) {
         const fieldName = item.conditional
         const field = fields[fieldName]
 
-        if (!field) {
+        if (!has(field, 'component') || !field.component) {
```
barbaracassani: Do you mean something like
`if (field.component === undefined  || field.component === null)`

Or it is intentional to include every falsey? 🤔

app/move/fields/create.js
```diff
@@ -87,11 +87,12 @@ function explicitYesNo(name) {
...
     items: [
       {
-        value: 'yes',
+        value,
+        conditional,
         text: 'Yes',
       },
       {
-        value: 'no',
+        value: 'false',
```
barbaracassani: False as string?

common/helpers/field.js
```diff
@@ -215,60 +210,56 @@ function mapPersonToOption(person = {}) {
...
-      }
+    if (!field) {
+      return false
     }
 
-    return {
-      ...field,
-      dependent,
-    }
+    return explicit ? field.explicit : !field.explicit
```
barbaracassani: Wow, this is a bit confusing.
If `explicit` is true, then use `field.explicit`
if `explicit` is false, then use the opposite of `field.explicit`?

PR #378, opened 2020-03-05
[P4 1101] Week view in dashboard

barbaracassani:
This PR creates the view week for the dashboard, and also contains some refactoring to use the same url pattern in all the moves routes.

image
image

### Checklist


app/moves/middleware.js
```diff
@@ -11,28 +13,44 @@ const moveService = require('../../common/services/move')
...
     }
 
-    return res.redirect(`${req.baseUrl}/${today}`)
+    return res.redirect(`${req.baseUrl}/day/${today}`)
   },
   saveUrl: (req, res, next) => {
     req.session.movesUrl = req.originalUrl
     next()
   },
+  setPeriod: (req, res, next, period) => {
```
teneightfive: Just had a small thought this morning - are this and status actually being used in the view right now?

If not I'd be tempted to remove the need for these.

app/moves/middleware.js
```diff
@@ -44,6 +62,24 @@ module.exports = {
     res.locals.moveDate = format(parsedDate, moveDateFormat)
```
teneightfive: Is this now even really used anywhere/needed anymore? We're basically replacing the need for this with the range I think.

The only place I can see this being used is in the existing header. But you could replace this with `setDateRange` and use `dateRange | first` again instead of `moveDate`.

app/moves/middleware.js
teneightfive: Ah, I can see its being used for some logic to display the header.

But looking at that logic in the template. If we replace `setMoveDate` with `setDateRange` in  the date param middleware call you could do the template logic based on length. If one item in the array, show first, if two, show the date range in the header?

app/moves/middleware.js
```diff
@@ -11,28 +13,44 @@ const moveService = require('../../common/services/move')
...
   },
   saveUrl: (req, res, next) => {
     req.session.movesUrl = req.originalUrl
     next()
   },
+  setPeriod: (req, res, next, period) => {
+    res.locals.period = period
+    next()
+  },
+  setStatus: (req, res, next, status) => {
```
teneightfive: We can probable remove this middleware. We don't use this value in the view and the subsequent middleware that currently accesses `res.locals.status` could access `req.params.status` to get the same value.

app/moves/index.js
```diff
@@ -20,43 +24,50 @@ const uuidRegex =
...
+  setDateRange,
   setMovesByDateAllLocations,
   setPagination,
   list
 )
 router.get(
-  `/:date/:locationId(${uuidRegex})`,
+  `/:period(week|day)/:date/:locationId(${uuidRegex})`,
   protectRoute('moves:view:by_location'),
+  setDateRange,
```
teneightfive: I spotted below we can probably just call this on the `:date` param middleware seeing as we use it in all routes that contain `:date`. It doesn't look like we have much need for `moveDate` either now so it could be replaced completely with this. And it would reduce the need to duplicate this in all of these route definitions.

app/moves/controllers/download.test.js
```diff
@@ -89,7 +92,7 @@ describe('Moves controllers', function() {
 
       it('should translate filename', function() {
         expect(req.t).to.be.calledOnceWithExactly('moves::download_filename', {
-          date: mockMoveDate,
+          date: '2018-10-10-2018-10-17',
```
teneightfive: Wonder if it might be nicer to use `to` instead of `-`. Could be easier to see the split as currently there are a lot of `-`. Or if we want to keep 2, maybe we remove the dashes splitting the date parts - `20181010-20181017`?

app/moves/middleware.js
```diff
@@ -95,19 +62,20 @@ module.exports = {
     next()
   },
   setPagination: (req, res, next) => {
-    const { moveDate, period = 'week', status = '' } = res.locals
+    const { period = 'week' } = res.locals
+    const { status = '' } = req.params
```
teneightfive: Could also get `period` from `req.params` for consistency. That way, the middleware order isn't as important.

teneightfive: I don't think it needs a default value either to be honest. If it doesn't exist it will be `undefined` which is enough for the falsy checks you currently have.

app/moves/middleware.js
```diff
@@ -136,7 +104,8 @@ module.exports = {
     }
   },
   setMovesByDateRangeAndStatus: async (req, res, next) => {
-    const { dateRange, fromLocationId, status } = res.locals
+    const { dateRange, fromLocationId } = res.locals
+    const { status = '' } = req.params
```
teneightfive: You could also remove the default value here to. don't think it adds any value.

app/moves/middleware.js
teneightfive: Also, you're destructing the req.paras object on line 68 so you can combine those.

app/moves/middleware.js
```diff
@@ -136,7 +104,8 @@ module.exports = {
     }
   },
   setMovesByDateRangeAndStatus: async (req, res, next) => {
-    const { dateRange, fromLocationId, status } = res.locals
+    const { dateRange, fromLocationId } = res.locals
```
teneightfive: You can also get `locationId` from `req.params`. I think its best to use `req.params` within middleware by default, unless that value doesn't exist in the URL, like `dateRange`

app/moves/controllers/download.test.js
barbaracassani: Let's use `to`

app/moves/index.js
```diff
@@ -23,9 +21,7 @@ const uuidRegex =
 
 // Define param middleware
 router.param('locationId', setFromLocation)
-router.param('date', setMoveDate)
```
teneightfive: You could replace this with `setDateRange` now.

It's more efficient using it with [`router.param`](https://expressjs.com/en/api.html#router.param) as I believe express will only load that method in once, rather than each time for each route like we currently have.

PR #380, opened 2020-03-09
[P4-1106] Display NOMIS alerts on assessment pages

teneightfive:
This change includes a way to replay answers that we have for a user in NOMIS around their release status and their need for a special vehicle.

It uses the same pattern for displaying the flags as on the profile page.

## What it looks like

image

image

### DIFF
--- a/app/move/controllers/create/save.js
+++ b/app/move/controllers/create/save.js
@@ -1,9 +1,21 @@
-const { omit, capitalize, flatten, values } = require('lodash')
+const { omit, capitalize, flatten, values, some } = require('lodash')
 
 const CreateBaseController = require('./base')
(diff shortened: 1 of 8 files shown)


app/move/controllers/create/save.js
```diff
@@ -1,10 +1,22 @@
...
 
+function filterAnswer(currentAssessment, searchKey) {
+  return item => {
+    if (
+      !some(currentAssessment, { key: searchKey }) &&
+      item.key === searchKey
+    ) {
+      return false
+    }
+    return true
```
barbaracassani: This can probably be simplified.

teneightfive: Open to suggestions...

PR #382, opened 2020-03-10
[P4-1139] feat: Navigation for multiple types of moves

barbaracassani:
This pr contains the routes and the links for the various types of moves displayed in the dashboard in the OCA workflow. 

image

Small viewport:
image

Very small viewport:
image

### Checklist

### DIFF
--- a/app/moves/middleware.js
+++ b/app/moves/middleware.js
@@ -157,7 +157,7 @@ const movesMiddleware = {
         {
           dateRange,
           status,
-          locationId,
+          fromLocationId: locationId,
         }
       )
 
--- a/app/moves/middleware.js
+++ b/app/moves/middleware.js
@@ -1,5 +1,5 @@
 const { format } = require('date-fns')
-const { find, get, chunk } = require('lodash')
+const { find, get, chunk, cloneDeep } = require('lodash')
 
 const {
   getDateRange,
@@ -7,6 +7,8 @@ const {
   getPeriod,
   dateFormat,
 } = require('../../common/helpers/date-utils')
+
+const presenters = require('../../common/presenters')
 
 const moveService = require('../../common/services/move')
 const { LOCATIONS_BATCH_SIZE } = require('../../config')
@@ -18,8 +20,21 @@ function makeMultipleRequests(service, dateRange, locationIdBatches) {
     )
   )
 }
-
-module.exports = {
+const moveTypeNavigationConfig = [
+  {
+    label: 'moves::dashboard.filter.proposed',
+    filter: 'proposed',
+  },
+  {
+    label: 'moves::dashboard.filter.approved',
+    filter: 'requested,accepted,completed',
+  },
+  {
+    label: 'moves::dashboard.filter.rejected',
+    filter: 'rejected',
+  },
+]
+const movesMiddleware = {
   redirectBaseUrl: (req, res) => {
     const today = format(new Date(), dateFormat)
     const currentLocation = get(req.session, 'currentLocation.id')
@@ -80,6 +95,33 @@ module.exports = {
     }
 
     next()
+  },
+  setMoveTypeNavigation: async (req, res, next) => {
+    const { dateRange } = res.locals
+    const { locationId, period, date } = req.params
+    try {
+      res.locals.moveTypeNavigation = await Promise.all(
+        cloneDeep(moveTypeNavigationConfig).map(moveType => {
+          return moveService
+            .getMovesCount({ dateRange, status: moveType.filter, locationId })
+            .then(count => {
+              const { label } = moveType
+              return {
+                label,
+                value: count,
+                active: moveType.filter === req.params.status,
+                href: `${req.baseUrl}/${period}/${date}${
+                  locationId ? '/' + locationId : ''
+                }/${moveType.filter}`,
+              }
+            })
+            .then(presenters.moveTypesToFilterComponent)
+        })
+      )
+      next()
+    } catch (error) {
+      next(error)
+    }
   },
   setMovesByDateAndLocation: async (req, res, next) => {
     const { dateRange, fromLocationId } = res.locals
@@ -159,3 +201,5 @@ module.exports = {
     }
   },
 }
+
+module.exports = movesMiddleware
--- a/app/moves/views/_layout.njk
+++ b/app/moves/views/_layout.njk
@@ -22,7 +22,7 @@
 
   {% block pageHeader %}
     <header class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
+      <div class="{% block headerGridColumnClass %}govuk-grid-column-two-thirds{% endblock %}">
         <span class="govuk-caption-xl">{{ t(pageTitle) }}</span>
 
         {% block dateTitle %}
--- a/app/moves/views/list-by-status.njk
+++ b/app/moves/views/list-by-status.njk
@@ -1,5 +1,7 @@
 {% extends "./_layout.njk" %}
 
+{% block headerGridColumnClass %}govuk-grid-column-full{% endblock %}
+
 {% block headerActions %}
   {% if canAccess('move:create') %}
     {{ govukButton({
@@ -9,6 +11,9 @@
       classes: "govuk-!-margin-top-2 app-print--hide"
     }) }}
   {% endif %}
+  {{ appFilter({
+    items: moveTypeNavigation
+  }) }}
 {% endblock %}
 
 {% block pageContent %}
@@ -26,4 +31,3 @@
     }) }}
 {% endif %}
 {% endblock %}
-
--- a/common/components/filter/_filter.scss
+++ b/common/components/filter/_filter.scss
@@ -1,0 +1,77 @@
+.app-filter {
+  margin: 0 0 govuk-spacing(3);
+}
+
+.app-filter__list {
+  list-style: none;
+  margin: 0;
+  padding-left: 0;
+  display: flex;
+  justify-content: space-between;
+  align-items: stretch;
+
+  flex-wrap:wrap;
+  flex-direction:row;
+
+  @include govuk-font($size: 19);
+  @include govuk-media-query($until: 370px) {
+    flex-direction:column;
+  }
+  @include govuk-media-query($media-type: print) {
+    display: none;
+  }
+}
+
+.app-filter__list-item {
+  border: 2px solid govuk-colour('blue');
+  flex: 1 1 0;
+  min-height: 70px;
+  display: flex;
+  @include govuk-media-query($until: 370px) {
+    min-height: 90px;
+  }
+}
+
+.app-filter__list-item--active {
+  background: govuk-colour('white');
+  border: 2px solid govuk-colour('black');
+}
+
+.app-filter__link {
+  width: 100%;
+  height: 100%;
+  // cannot remove text-decoration from parts of the link, so we remove it everywhere and re-add it when necessary
+  text-decoration: none;
+  background: govuk-colour('blue');
+  color: govuk-colour('white');
+  display: block;
+  flex-grow: 1;
+
+  &:link,
+  &:active,
+  &:visited {
+    color: govuk-colour('white');
+  }
+
+  &:hover {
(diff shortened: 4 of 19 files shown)


app/moves/middleware.js
```diff
@@ -81,6 +94,42 @@ module.exports = {
 
     next()
   },
+  setMoveTypeNavigationContent: (count, label) => {
+    return `<span class="app-filter__count">${count}</span><span class="app-filter__content">${label}</span>`
```
teneightfive: This feels a little weird to be building HTML in a middleware. 

The idea of having the components is so that the markup is contained just there. If we updated that, for example the element from a `span` to a `p`, or changed the classname this could easily be missed. 

I think it would be cleaner to have the middleware just build objects that can be passed to the views. Or even better, look at how the [presenters](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/master/common/presenters) are used. It's where we transform API data into a structure the component requires.

app/moves/views/list-by-status.njk
```diff
@@ -0,0 +1,71 @@
...
+        {% block dateTitle %}
+          <h1 class="govuk-heading-xl govuk-!-margin-bottom-2">
+            {{ appTime({
+              datetime: datetime,
+              text: displayDate
+            }) }}
+          </h1>
+        {% endblock %}
+
+        {% block dateNavigation %}
```
teneightfive: It feels like there's a few things in this template that could be inherited from a parent layout for the other listing views?

You could add some blocks to that layout to help you inject content, like the filter, where needed rather than having to duplicate the pagination component, header markup, etc.

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,67 @@
...
+
+  &:hover {
+    color: #d5e8f3;
+  }
+
+  &:focus {
+    background-color: govuk-colour('blue');
+  }
+
+  @include govuk-media-query($media-type: print) {
```
teneightfive: Might be worth having a look at the mobile view too:

image

We could stack them, or maybe change the layout.

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,67 @@
...
+    display: none;
+  }
+}
+
+.app-filter__count {
```
teneightfive: It looks like the markup doesn't quite fit:

image

I think the idea of that markup is that `app-filter__value` is the large number, and `app-filter__label` is the text. This makes it less tied to this specific content instance. The _value_ might not always be a "count".

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,67 @@
...
+  @include govuk-font($size: 19, $weight: normal, $line-height: 1.25);
+
+  &:link,
+  &:active,
+  &:visited {
+    color: govuk-colour('white');
+  }
+
+  &:hover {
+    color: #d5e8f3;
```
teneightfive: This doesn't look quite right I don't think:

image

He's the equivalent on Notify which feels a little closer to what we want:
image

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,67 @@
...
+  min-height: 70px;
+  display: flex;
+}
+
+.app-filter__list-item--active {
+  background: govuk-colour('white');
+  border: 2px solid govuk-colour('black');
+  padding: govuk-spacing(4);
+  padding-right: govuk-spacing(6);
+  @include govuk-font($size: 19, $weight: normal, $line-height: 1.25);
```
teneightfive: Do we need to specify the line-height in this mixin? I think `1.25` and `normal` are the defaults.

common/components/filter/_filter.scss
teneightfive: Feels like that hex code may also already be a variable with `govuk-colour`?

common/components/filter/template.njk
```diff
@@ -0,0 +1,17 @@
+<nav class="app-filter">
```
teneightfive: Looks like we're missing some tests for this component. Take a look at some other examples for how we test different scenarios.

common/services/move.js
```diff
@@ -42,22 +42,37 @@ const moveService = {
...
+      'filter[created_at_to]': createdAtTo,
+      'filter[from_location_id]': locationId,
+    }
+    return apiClient
+      .findAll('move', {
+        ...filter,
+        page: 1,
+        per_page: 1,
+      })
+      .then(response => {
```
teneightfive: Could just do:

```javascript
.then(response => response.meta.pagination.total_objects)
```

locales/en/moves.json
```diff
@@ -146,5 +146,18 @@
   },
   "agreement_status": {
     "heading": "Has this move been agreed with receiving prison?"
+  },
+  "proposed_moves_flow": {
```
teneightfive: These could be nested within the object we have for `dashboard` already. Could create an object with that for `filter`. That way we could share that if we use it across other views. Stops it being tied specifically to this _flow_:

```json
...
"dashboard": {
  ...
  "filter": {
    "proposed": "proposed",
    "approved": "approved",
    "rejected": "rejected"
  }
  ...
}
...
```

teneightfive: Don't think we need the label property, that was more for form fields where we'd have multiple properties, one for label, one for hint, etc.

Also, I think they should be lowercased, like the GOV.UK Notify banner as they should read like a sentence: `0 proposed` instead of `0 Proposed`.

app/moves/views/list-by-status.njk
barbaracassani: There's some duplication, indeed -- mostly because I need `govuk-grid-column-full` and the other layouts don't. Is there a way to inject a class that doesn't feel contrived?

common/services/move.js
barbaracassani: Surprising that `prettier` hasn't caught that 🤔

common/components/filter/_filter.scss
barbaracassani: It doesn't -- but has a close-ish match which I am using now.

app/moves/middleware.js
```diff
@@ -81,6 +96,24 @@ module.exports = {
...
   },
+  setMoveTypeNavigation: async (req, res, next) => {
+    const { period, dateRange } = res.locals
+    const { locationId = '' } = req.params
+    try {
+      res.locals.moveTypeNavigation = await Promise.all(
+        cloneDeep(moveTypeNavigationConfig).map(moveType => {
+          return moveService
+            .getMovesCount({ dateRange, status: moveType.filter, locationId })
+            .then(count => {
```
teneightfive: This could be a set of chained `then` which could clean up the number of args you need to send to the presenter:

```javascript
cloneDeep(moveTypeNavigationConfig).map(moveType => {
  return moveService
    .getMovesCount({ dateRange, status: moveType.filter, locationId })
    .then(count => {
      return {
        ...moveType,
        count,
        active: moveType.filter === req.params.status,
        url: ...,
      }
    })
    .then(type => moveTypeForFilter(type))
})
```

barbaracassani: Nice one!!

app/moves/middleware.js
```diff
@@ -81,6 +96,33 @@ module.exports = {
 
     next()
   },
+  setMoveTypeNavigation: async (req, res, next) => {
+    const { period, dateRange } = res.locals
```
teneightfive: You can also get `period` from `req.params`? Removes the dependency on order that way.

app/moves/middleware.js
```diff
@@ -81,6 +96,33 @@ module.exports = {
...
+    const { locationId = '' } = req.params
```
teneightfive: We can drop the default empty string from this I think

app/moves/middleware.js
```diff
@@ -81,6 +96,33 @@ module.exports = {
...
+        cloneDeep(moveTypeNavigationConfig).map(moveType => {
+          return moveService
+            .getMovesCount({ dateRange, status: moveType.filter, locationId })
+            .then(count => {
+              const { label } = moveType
+              return {
+                label,
+                value: count,
+                active: moveType.filter === req.params.status,
+                href: `${req.baseUrl}/${period}/${req.params.date}${
```
teneightfive: Maybe destructure `req.params.date` above for consistency?

app/moves/views/_layout.njk
```diff
@@ -22,7 +22,7 @@
 
   {% block pageHeader %}
     <header class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
+      <div class="{% block mainContainerClass %}govuk-grid-column-two-thirds{% endblock %}">
```
teneightfive: Just thinking about the name, we actually have a div called "main" and this block isn't actually related to that. Maybe something more appropriate to what this actually does might be better - `headerGridColumnClass`?

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,76 @@
...
+  &:active,
+  &:visited {
+    color: govuk-colour('white');
+  }
+
+  &:hover {
+    color: govuk-colour('light-grey');
+  }
+
+  &:focus {
```
teneightfive: image

We might want to set the colour to black in this state. The white on yellow is hard to read.

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,76 @@
...
+  margin: 0;
+  padding-left: 0;
+  display: flex;
+  justify-content: space-between;
+  align-items: stretch;
+
+  flex-wrap:wrap;
+  flex-direction:row;
+
+  @include govuk-font($size: 19, $weight: normal);
```
teneightfive: You can drop the second argument. The default is normal I think?

common/components/filter/_filter.scss
```diff
@@ -0,0 +1,76 @@
...
+  @include govuk-media-query($until: 370px) {
```
teneightfive: Would it be better to use some of the predefined dimensions rather than creating our own?

common/presenters/move-type-for-filter.js
```diff
@@ -0,0 +1,9 @@
+const i18n = require('../../config/i18n')
+const { cloneDeep } = require('lodash')
+
+function moveTypeForFilter(item) {
```
teneightfive: Could we keep the consistency in naming convention for these - `moveTypesToFilterComponent`? It was something that was suggested by a previous developer to try and be verbose and clear about what the presenters are doing and I think it works well.

common/services/move.js
```diff
@@ -42,22 +42,35 @@ const moveService = {
       })
   },
 
-  getMovesByDateRangeAndStatus({
-    dateRange = [],
-    status,
-    fromLocationId,
-  } = {}) {
+  getMovesByDateRangeAndStatus({ dateRange = [], status, locationId } = {}) {
```
teneightfive: Is there a reason this was updated? Feels like the name should still be `fromLocationId` as that is what the service method expects and is using this value for?

common/components/filter/_filter.scss
barbaracassani: This is the dimension that makes it work. The predefined ones always ended up with a stack of 2. We may want to revisit this in the future when we have 4 boxes.

common/services/move.js
barbaracassani: I think the current name is clearer. We get the moves by range and status. Location may not be passed if it's not needed.

common/presenters/move-type-for-filter.js
barbaracassani: ok

common/services/move.js
teneightfive: Sorry, I meant the name of the variable that you renamed. You renamed it from `fromLocationId` to `locationId` - but as far as I can tell the functionality is still the same, ie it is actually the `fromLocationId`, not just an arbitrary `locationId`. It is then specifically passed into the `from_location` filter. So that's why I thought it would make more sense to leave this unchanged.

barbaracassani: Ahhh ... ! Ok. I see what you mean.

PR #385, opened 2020-03-11
[P4-1149] feat: Primary navigation with link to the dashboard

barbaracassani:
This pr removes the beta banner and adds a primary navigation in every page.

N.B. This change affects the whole site! Do not merge unless approved on every level :-)

Screenshot 2020-03-25 at 12 07 15

### Checklist

### DIFF
--- a/app/moves/middleware.js
+++ b/app/moves/middleware.js
@@ -1,6 +1,8 @@
 const { format } = require('date-fns')
 const { find, get, chunk, cloneDeep } = require('lodash')
 
+const permissions = require('../../common/middleware/permissions')
+
 const {
   getDateRange,
   getDateFromParams,
@@ -40,7 +42,16 @@ const movesMiddleware = {
     const currentLocation = get(req.session, 'currentLocation.id')
 
     if (currentLocation) {
-      return res.redirect(`${req.baseUrl}/day/${today}/${currentLocation}`)
+      const userPermissions = get(req.session, 'user.permissions')
+      const canViewProposedMoves = permissions.check(
+        'moves:view:proposed',
+        userPermissions
+      )
+      return res.redirect(
+        `${req.baseUrl}/day/${today}/${currentLocation}/${
+          canViewProposedMoves ? 'proposed' : ''
+        }`
+      )
     }
 
     return res.redirect(`${req.baseUrl}/day/${today}`)
--- a/app/moves/middleware.test.js
+++ b/app/moves/middleware.test.js
@@ -35,37 +35,49 @@ describe('Moves middleware', function() {
     })
 
     context('with current location', function() {
-      context('when location type is police', function() {
+      context(
+        "when user hasn't got permission to see the proposed moves",
+        function() {
+          context('when there is a location', function() {
+            const mockLocation = {
+              id: 'c249ed09-0cd5-4f52-8aee-0506e2dc7579',
+            }
+
+            beforeEach(function() {
+              req.session = {
+                currentLocation: mockLocation,
+                user: {
+                  permissions: [],
+                },
+              }
+
+              middleware.redirectBaseUrl(req, res)
+            })
+
+            it('should redirect to moves by location', function() {
+              expect(res.redirect).to.have.been.calledOnceWithExactly(
+                `/moves/day/${mockMoveDate}/${mockLocation.id}/`
+              )
+            })
+          })
+        }
+      )
+      context('when user has permission to see the proposed moves', function() {
         const mockLocation = {
           id: 'c249ed09-0cd5-4f52-8aee-0506e2dc7579',
-          location_type: 'police',
         }
 
         beforeEach(function() {
-          req.session.currentLocation = mockLocation
+          req.session = {
+            currentLocation: mockLocation,
+            user: {
+              permissions: ['moves:view:proposed'],
+            },
+          }
 
           middleware.redirectBaseUrl(req, res)
         })
-
-        it('should redirect to moves by location', function() {
-          expect(res.redirect).to.have.been.calledOnceWithExactly(
-            `/moves/day/${mockMoveDate}/${mockLocation.id}`
-          )
-        })
-      })
-      context('when location type is prison', function() {
-        const mockLocation = {
-          id: 'c249ed09-0cd5-4f52-8aee-0506e2dc7579',
-          location_type: 'prison',
-        }
-
-        beforeEach(function() {
-          req.session.currentLocation = mockLocation
-
-          middleware.redirectBaseUrl(req, res)
-        })
-
-        it('should redirect to moves by location', function() {
+        it('should redirect to proposed moves by location if the user can see them', function() {
           expect(res.redirect).to.have.been.calledOnceWithExactly(
             `/moves/day/${mockMoveDate}/${mockLocation.id}/proposed`
           )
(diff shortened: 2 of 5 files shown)


app/moves/middleware.js
```diff
@@ -22,13 +22,18 @@ function makeMultipleRequests(service, dateRange, locationIdBatches) {
 module.exports = {
   redirectBaseUrl: (req, res) => {
     const today = format(new Date(), dateFormat)
-    const currentLocation = get(req.session, 'currentLocation.id')
-
-    if (currentLocation) {
-      return res.redirect(`${req.baseUrl}/day/${today}/${currentLocation}`)
+    const currentLocation = get(req.session, 'currentLocation')
+
+    switch (currentLocation && currentLocation.location_type) {
```
teneightfive: I don't think this logic is right at this stage. I think we need to keep the current logic of if a current location is set redirect to a location. 

And maybe in this instance we need to use permissions to determine the next path. So if a user has permission to view proposed moves, they are directed there?

teneightfive: This logic would leave STC/SCH users at a dead end I think.

common/assets/scss/application.scss
```diff
@@ -22,20 +22,23 @@ $govuk-font-family: "Inter", system-ui, sans-serif;
...
 @import "govuk-frontend/govuk/components/panel/panel.scss";
-@import "govuk-frontend/govuk/components/phase-banner/phase-banner.scss";
 @import "govuk-frontend/govuk/components/radios/radios.scss";
 @import "govuk-frontend/govuk/components/select/select.scss";
 @import "govuk-frontend/govuk/components/skip-link/skip-link.scss";
 @import "govuk-frontend/govuk/components/summary-list/summary-list.scss";
 @import "govuk-frontend/govuk/components/textarea/textarea.scss";
 @import "govuk-frontend/govuk/components/warning-text/warning-text.scss";
 
+
```
teneightfive: Any reason for this?

common/templates/layouts/base.njk
```diff
@@ -90,18 +90,17 @@
       }
     ]
   }) }}
+  {{ mojPrimaryNavigation({
+    label: 'Primary navigation',
+    items: [{
+      text: t("default::primary_navigation.dashboard_oca") if CURRENT_LOCATION and CURRENT_LOCATION.location_type === 'prison' else t("default::primary_navigation.dashboard_police"),
```
teneightfive: What about rather than an if/else actually displaying both items based on permissions? No reason OCA shouldn't be able to see this in theory too. Makes this less complicated too.

locales/en/default.json
```diff
@@ -1,6 +1,8 @@
 {
-  "phase_banner": "This is a new service",
-  "phase_banner_with_feedback": "This is a new service – your <a class=\"govuk-link\" href=\"{{url}}\">feedback</a> will help us to improve it.",
   "age": "Age",
-  "age_with_date_of_birth": "{{date_of_birth}} (Age {{age}})"
+  "age_with_date_of_birth": "{{date_of_birth}} (Age {{age}})",
+  "primary_navigation": {
+    "dashboard_police": "Upcoming moves",
```
teneightfive: We should probably use a key that relates to the content of the translation rather than confusing them by using people's roles. "Upcoming moves" isn't also just tied to "Police". Some prison users, STC and SCH users will see this item.

teneightfive:
A small sidenote but this is another PR that I think would have benefit from a couple of smaller commits that would help to explain the changes. Removing the phase banner for example could have been a separate commit with the explanation of having both the phase banner and navigation taking up too much space.

At the moment this is one commit that is doing a number of different things which could have been explained separately and will help us when looking back at those changes.

teneightfive:
@Andy-Millar this change removes the current phase banner (which we talked about a while ago) but it doesn't move the feedback link anyway. I know we discussed moving it to the footer? What are your thoughts on what we can do with it?

@barbaracassani maybe catch up with Andy on this as I don't think we want to just remove the ability for users to provide us with feedback completely.

common/assets/scss/application.scss
barbaracassani: You said we wanted to drop the phase banner because otherwise it would have been too busy with the new navigation...

common/templates/layouts/base.njk
barbaracassani: Both items link to `/moves` that then redirects accordingly. I am not entirely sure how to build the redirect logic if we want OCA to have access to the upcoming moves dashboard. Also, why would they want to? Upcoming moves would contain the active moves visible in the dashboard. That would generate confusion and not add anything.

Andy-Millar:
@teneightfive Let's go with footer (just above it)

app/moves/middleware.js
barbaracassani: so something like   . ?:

     const today = format(new Date(), dateFormat)
    const currentLocation = get(req.session, 'currentLocation.id')

    if (currentLocation) {
      const canViewProposedMoves = permissions.check('move:proposed:view', userPermissions)
      return res.redirect(
        `${req.baseUrl}/day/${today}/${currentLocation}/${canViewProposedMoves ? 'proposed' : ''}`
      )
    }

    return res.redirect(`${req.baseUrl}/day/${today}`)

teneightfive: Yeah, I think that feels a little safe for now 👍

teneightfive: Once we introduce a "dashboard" at this route we can get rid of this anyway

common/assets/scss/application.scss
teneightfive: Sorry, I just meant the addition of another new line.

common/templates/layouts/base.njk
teneightfive: Hmm, yes maybe that's the tricky bit for now.

Ideally these nav items should deep link into the page that they are related too but I suppose our logic of the URLs containing a particular day (by default today or this week) make them trickier to build in a template file.

app/moves/middleware.test.js
```diff
@@ -35,23 +35,71 @@ describe('Moves middleware', function() {
     })
 
     context('with current location', function() {
-      const mockLocationId = 'c249ed09-0cd5-4f52-8aee-0506e2dc7579'
-
-      beforeEach(function() {
-        req.session.currentLocation = {
-          id: mockLocationId,
+      context('when location type is police', function() {
```
teneightfive: Not sure this is the correct context still?

app/moves/middleware.test.js
```diff
@@ -35,23 +35,71 @@ describe('Moves middleware', function() {
...
+          req.session.user.permissions = ['moves:view:proposed']
+          res.redirect.resetHistory()
+          middleware.redirectBaseUrl(req, res)
+          expect(res.redirect).to.have.been.calledOnceWithExactly(
+            `/moves/day/${mockMoveDate}/${mockLocation.id}/proposed`
+          )
+          req.session.user.permissions = []
+        })
       })
+      context('when location type is prison', function() {
```
teneightfive: Might be worth updating this too

common/templates/layouts/base.njk
```diff
@@ -119,6 +119,19 @@
   {% include "includes/messages.njk" %}
 {% endblock %}
 
+{% block footer %}
+    {{ govukFooter({
+      meta: {
+        items: [
+          {
+            href: FEEDBACK_URL,
+            text: t("feedback_link")
```
teneightfive: We should keep the same logic where we only show this if the `FEEDBACK_URL` value is set.

app/moves/middleware.test.js
```diff
@@ -35,32 +35,107 @@ describe('Moves middleware', function() {
     })
 
     context('with current location', function() {
-      const mockLocationId = 'c249ed09-0cd5-4f52-8aee-0506e2dc7579'
+      context(
+        "when user hasn't got permission to see the proposed moves",
+        function() {
+          context('when location type is police', function() {
```
teneightfive: Not sure you fully understood what I meant. I think the tests were fine before, it was just the description in the `context` block.

The current method has zero concern with location type, so there is no need to have tests with different location types, they'll function the same. I was just pointing out the wording was just from the previous implementation.

These two tests don't offer anything different when testing this method.

barbaracassani: Once I had changed the wording of the context, it made more sense to reorganise it this way.

barbaracassani: Following our conversation, I have now removed the redundant test.

PR #386, opened 2020-03-11
[P4-1105] Add image to search results

teneightfive:
This change adds support for images in search results pages.

It uses an API endpoint to gain the image and if at any point this fails, it will fall pack to a placeholder image.

It includes a change to use the same card component from the dashboard to display the person within the search results.

### With image

image

### With fallback image

image

## Todo

- [x] Use correct placeholder image
- [x] Update model test with new `image` model
- [x] Remove singular path override for image model within person resource
- [x] Double check use of card component from dashboard - label style has been updated to match
- [x] Ensure placeholder image works on production with minified assets


app/person/controllers.js
```diff
@@ -0,0 +1,21 @@
...
+      const response = await axios.get(imageUrl, {
+        responseType: 'arraybuffer',
+      })
+
+      res.writeHead(200, {
+        'Content-Type': response.headers['content-type'],
+      })
+      res.end(response.data, 'binary')
+    } catch (error) {
+      res.redirect('/images/person-placeholder.png')
```
teneightfive: I wasn't 100% convinced on this, open to other suggestions, but the alternative was to load the image from file and serve it back.

Also, not sure if we should be storing this in some kind of config? Might be easier that way.

PR #395, opened 2020-03-20
[P4 1082] fix: Prevent garbled message on failed attempt to upload file on STC/SCH move

solidgoldpig:
Files that passed Multer's vaildation but failed to be successfully uploaded to the document store resulted in an error message like this:

Screen Shot 2020-02-24 at 09 20 32

Post fix:

Are_there_any_documents_to_upload__-_Create_a_move

### Checklist

### DIFF
--- a/common/lib/api-client/middleware/post.test.js
+++ b/common/lib/api-client/middleware/post.test.js
@@ -1,7 +1,8 @@
 const FormData = require('form-data')
 
 const postMiddleware = require('./post')
-const { FILE_UPLOADS } = require('../../../../config')
+
+const MAX_FILE_SIZE = 2000
 
 const originalHeaders = {
   original: 'header',
@@ -40,7 +41,7 @@ describe('API Client', function() {
         context('when data is instance of FormData', function() {
           beforeEach(function() {
             payload.req.data = new FormData()
-            response = postMiddleware.req(payload)
+            response = postMiddleware(MAX_FILE_SIZE).req(payload)
           })
 
           it('should update headers', function() {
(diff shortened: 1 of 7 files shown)


common/lib/api-client/middleware/post.test.js
```diff
@@ -1,6 +1,7 @@
 const FormData = require('form-data')
 
 const postMiddleware = require('./post')
+const { FILE_UPLOADS } = require('../../../../config')
```
teneightfive: I find it's best to avoid importing config directly into test files. We've either been using [proxyquire to mock the value](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/app/healthcheck/controllers.test.js#L12) or [injecting config values into the middleware call](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/lib/api-client/index.js#L21). Be good to use one of these approaches to stay consistent and to avoid any issues if the config value changes or becomes dependent on an env var.

locales/en/validation.json
```diff
@@ -10,11 +10,12 @@
...
-  "LIMIT_FILE_SIZE": "must be smaller than 50 MB",
+  "LIMIT_FILE_SIZE": "must be smaller than 50MB",
   "LIMIT_PART_COUNT": "has too many parts",
   "LIMIT_FILE_COUNT": "has too mamy files",
   "LIMIT_FIELD_KEY": "must include a shorter field name",
   "LIMIT_FIELD_VALUE": "must include a shorter field value",
   "LIMIT_FIELD_COUNT": "includes too many fields",
-  "LIMIT_UNEXPECTED_FILE": "must include expected file types"
+  "LIMIT_UNEXPECTED_FILE": "must include expected file types",
+  "API_DOCUMENT_STORAGE_FAILED": "could not be uploaded – try again"
```
teneightfive: Do we need to say "try again"? We don't on any of the other errors is why I ask.

We could even use the same message as the _generic_ validation error above - `something went wrong`?

teneightfive: On second thoughts, `could not be uploaded` is clearer, but maybe we just drop the `try again`?

solidgoldpig: I discussed the message with @Andy-Millar and he wanted us to follow govuk-frontend format as per https://design-system.service.gov.uk/components/file-upload/

common/lib/api-client/middleware/post.test.js
solidgoldpig: Moved loading config var to main middleware file.

PR #397, opened 2020-03-23
[P4-1112] Prison to Court end-to-end tests

teneightfive:
This change includes a further refactor of the end-to-end test suite to allow us to create fixtures more easily and re-use more of the code across different journeys.

It also fixes a few bugs that were spotted within existing libraries whilst running these new scenarios.

If it's easier these can be split out into a separate PR...

### DIFF
--- a/test/e2e/helpers.js
+++ b/test/e2e/helpers.js
@@ -91,25 +91,21 @@ export async function createPersonFixture() {
 /**
  * Select random option from autocomplete menu
  *
- * @param {string} labelText - label text for the option
+ * @param {Selector} selector - An autocomplete field
  * @param {string|number} [optionTextOrIndex] - option text or 0-based index or 'random'.
- * @returns {Selector}
+ * @returns {string} - selected value
  */
-export async function selectAutocompleteOption(labelText, optionTextOrIndex) {
-  const fieldSelector = Selector('.govuk-label').withText(labelText)
+export async function selectAutocompleteOption(selector, optionTextOrIndex) {
+  await t.click(selector)
 
-  await t.click(fieldSelector)
-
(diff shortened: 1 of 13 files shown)


teneightfive:
@solidgoldpig / @barbaracassani / @jcavell I've started to create some e2e tests for the Prison to Court journey...the problem is we don't allow users to create one of these moves without the person existing.

This means that we would need to know a specific prison number and person within that environment. So we will rely on staging/preprod/production (if we test there) having a particular fixture. Have either of you done this in the past and know of the best way to approach it? We probably don't want that prison number/name/any other details we want to check hard coded in our codebase either right?

Would be great to get some thoughts on the best way we can test these journeys

test/e2e/helpers.js
```diff
@@ -50,6 +54,40 @@ async function selectOption(selector, textOrIndex = 'random', cssSelector) {
...
+    }).value,
+    police_national_computer: find(person.identifiers, {
+      identifier_type: 'police_national_computer',
+    }).value,
+  }
+}
+
 /**
  * Select random option from autocomplete menu
  *
```
solidgoldpig: Can't see how to apply the comment to the correct line (101), but suggest changing

`await t.click(fieldSelector)`

to

```
  const autocompleteClicker = fieldSelector.sibling('div').find('input')
  await t.click(autocompleteClicker)
```

which seems robust and does not depend on other events on the page preceding it

test/e2e/helpers.js
```diff
@@ -91,25 +91,21 @@ export async function createPersonFixture() {
...
- * @param {string} labelText - label text for the option
+ * @param {Selector} selector - An autocomplete field
```
solidgoldpig: Yay! Much better

PR #399, opened 2020-03-25
[P4 1152] Split the date fields from the move details steps

barbaracassani:
- This pr separates the date from the move details step
- Two pages for date are possible: one is the normal one (today / tomorrow / other date), the other is the date range
- The date range page has been modified as follow:

Screenshot 2020-03-26 at 15 23 00

### Checklist


app/move/steps/create.js
```diff
@@ -137,9 +150,14 @@ module.exports = {
       'gender_additional_information',
     ],
   },
-  '/move-date': {
-    pageTitle: 'moves::steps.move_date.heading',
+  '/move-date-range': {
     fields: ['date_from', 'date_to'],
+    ...moveDateStep,
```
teneightfive: Probably better to have the spread as the first item, that way you guarantee the properties here will override any defaults set in the shared object.

app/move/steps/create.js
```diff
@@ -137,9 +150,14 @@ module.exports = {
...
+  },
+  '/move-date': {
+    controller: MoveDate,
+    fields: ['date', 'date_type', 'date_custom'],
+    ...moveDateStep,
```
teneightfive: Same here - probably better to have the spread as the first item, that way you guarantee the properties here will override any defaults set in the shared object.

app/move/controllers/create/move-date.js
```diff
@@ -0,0 +1,44 @@
+const { format, startOfToday, startOfTomorrow, parseISO } = require('date-fns')
+
+const CreateBaseController = require('./base')
+const filters = require('../../../../config/nunjucks/filters')
+
+class MoveDetailsController extends CreateBaseController {
```
teneightfive: Should be `MoveDateController`

app/move/controllers/create/move-date.js
```diff
@@ -0,0 +1,44 @@
...
+      moveDate = dateType === 'today' ? startOfToday() : startOfTomorrow()
+    }
+
+    req.form.values.date = format(moveDate, 'yyyy-MM-dd')
+
+    next()
+  }
+}
+
+module.exports = MoveDetailsController
```
teneightfive: Should be `MoveDateController`

app/move/fields/create.js
```diff
@@ -112,7 +112,7 @@ function explicitAssessmentQuestion(name, value, conditional) {
 }
 
 const dateField = {
-  validate: ['date'],
+  validate: ['date', 'required'],
```
teneightfive: This was removed from here as not every date field will necessarily be "required". I think the `required` validation should be done at an individual field level rather than this shared level for future safety.

app/move/fields/create.js
```diff
@@ -344,17 +344,44 @@ module.exports = {
...
+    name: 'has_date_to',
+    fieldset: {
+      legend: {
+        text: 'fields::has_date_to.legend',
+        classes: 'govuk-label--m',
+      },
+    },
+    items: [
+      {
+        value: 'yes',
```
teneightfive: Maybe `true` / `false`? I know we need to use strings for the field values to stay populated but feels nicer to use true/false rather than yes/no as true/false values.

app/move/fields/create.js
barbaracassani: Fair enough, but for now at least they all are required. I am happy to revert it if you want, but we should code for now, not for an hypothetical future 😄

teneightfive: Happy to leave if for now then. 

My thought was more that a shared date configuration should contain date based configuration and field specific configuration (ie whether that field is required or not on the step it is used) should be done at field level, not shared level.

Just felt weird to say every single date field now and forever will always be required. However, felt fine to say every date field should be validated as a date. Is more about logical separating validation and putting them in the more sensible point in the configuration.

barbaracassani:
Thank you @teneightfive , I have now fixed the e-2-e tests (albeit I still see an unrelated failure).

app/move/fields/create.js
solidgoldpig: Personally I find `yes/no` to be more human readable ¯\_(ツ)_/¯

app/move/steps/create.js
teneightfive: Still think the spread should be the first property in this object, then any overrides for this particular step are defined after.

Currently, if `moveDateStep` contained a `controller` property it would take priority over the one defined in this step.

app/move/steps/create.js
teneightfive: Same as the `move-date-range` step, `...moveDateStep` should be the first property in the object.

app/move/fields/create.js
```diff
@@ -344,17 +344,44 @@ module.exports = {
...
+        conditional: 'date_to',
+      },
+      {
+        value: 'no',
+        text: 'No',
+      },
+    ],
+  },
   date_to: {
+    skip: true,
```
teneightfive: Ideally any properties just for this step should be defined after the shared object which is included.

app/move/fields/create.js
barbaracassani: I'll keep yes / no for now if that's ok. 
The reason being: we don't have uniformity at the moment. We use an ample choice of booleans, `'false'` as string, `'no'` etc. 
It would be a nice thing to make it uniform. Maybe on next refactoring Friday?

test/e2e/pages/create-move.js
```diff
@@ -161,11 +161,18 @@ class CreateMovePage extends Page {
...
       ...values,
     }
   }
 
+  /**
+   * Fill in date
+   */
+  async fillInDate() {
+    await t.expect(this.getCurrentUrl()).contains('/move/new/move-date')
+    return selectFieldsetOption('Date', 'Today')
```
teneightfive: Should we still return the inner text?

teneightfive:
Could we also extend the move model to suppress these warnings:

image

Without those defined those values won' be sent to/received from the API either.

app/move/steps/create.js
barbaracassani: This was already addressed. Not sure why it didn't show in the changes.

PR #400, opened 2020-03-26
[P4-1154] Capture hearing time and court case setup

teneightfive:
This PR contains a number of changes related to capturing hearing time for Prison to Court moves.

That change required some extension to existing libraries like formatters, validators and filters.

The major change is currently blocked by some API changes that it requires first so these changes are being split into a separate PR so that they can be merged now.

See #398 for the changes related to the create move journey.

### DIFF
--- a/app/move/validators.js
+++ b/app/move/validators.js
@@ -1,0 +1,13 @@
+const { Controller } = require('hmpo-form-wizard')
+
+module.exports = {
+  time(value) {
+    return (
+      value === '' ||
+      Controller.validators.regex(
+        value,
+        /^([0-9]|0[0-9]|1[0-9]|2[0-3]):([0-5][0-9]) ?([AaPp][Mm])?$/
+      )
+    )
+  },
+}
--- a/app/move/validators.test.js
+++ b/app/move/validators.test.js
@@ -1,0 +1,48 @@
+const validators = require('./validators')
+
+describe('Validators', function() {
+  describe('#time()', function() {
+    describe('invalid values', function() {
+      const inputs = [
+        10,
+        null,
+        '10',
+        '25:00',
+        '10:65',
+        'asdf',
+        '10:00a',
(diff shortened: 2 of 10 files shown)


app/move/validators.test.js
```diff
@@ -0,0 +1,48 @@
...
+        })
+      })
+    })
+
+    describe('valid values', function() {
+      const inputs = [
+        '',
+        '9:00am',
+        '9:00pm',
+        '9:00',
```
solidgoldpig: Aren't 9:00 and 10:00 just as ambiguous as 9 and 10?

app/move/validators.js
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = {
+  time(value) {
+    return (
+      value === '' ||
+      Controller.validators.regex(
+        value,
+        /^([0-9]|0[0-9]|1[0-9]|2[0-3]):([0-5][0-9]) ?([AaPp][Mm])?$/
+      )
+      // Controller.validators.regex(value, /^\d{1,2}:\d{2}([ap]m)?$/)
```
barbaracassani: Maybe remove the commented code; also, you have to read the regex to figure out which format of time this will allow. Maybe add a comment with an example of the desired format?

solidgoldpig: Have we recorded why we're allowing just this format rather than more flexible parsing such as provided by https://sugarjs.com/dates/ ?

teneightfive: So we have a "formatter" and a "validator". The formatter will allow the user to enter a few different formats and then format it in the way the validator will expect.

We can tweak later if we need to.

app/move/validators.test.js
teneightfive: In what sense? As a valid time? `9:00` is at least formatted more like a time than just an integer.

PR #401, opened 2020-03-27
fix: [P4-1041] Extend service timeout

solidgoldpig:
Problem
-------

After successfully reauthenticating a user, the user's session is regenerated. Unfortunately this had the side-effect of losing all the info that HMPO form wizard has stored resulting in the user seeing a session timeout message (or for xhr requests, the cryptic 'Server returned code 0')

Even if the HMPO form wizard was not being lost, the reauthentication redirect would have the effect of causing any posted information to be dropped.

Ideally a new token could just be requested using the refresh token rather than redirecting the user, but the HMPPS OAuth server does not support refresh tokens for the type of grant we require.

Fix
---

- Copy properties from previous session to new one
- Stash POST body (if any) in `req.session.originalRequestBody` when redirecting to /connect/$provider route
- Return 401 for request types that cannot be handled
  - when content-type is `multipart/formdata` (as we can't store the uploaded file)
  - when request is an `xhr` request (since the xhr runs afoul of cross-site concerns)
- Update request's body and method if `req.session.originalRequestBody` is present after successful reauthentication
- Refresh the auth token if the request method is 'GET' and the token is about to expire (see AUTH_EXPIRY_MARGIN env var)

Handling 401 errors
--------------------

Upload failed

upload-auth-expired_mov

Delete failed

delete-auth-expired_mov

Additionally suppresses the delete buttons

Potential enhancements
------------

This should cover the vast majority of cases, but if we find users are still getting stung by it, we could look into

- Automatically refresh the page and display a flash message afterwards
  (I did implement this to show Andy, but we both felt it would be more confusing to the user)
- Add ability to countdown to expiry on the client-side and prompt user to stay "logged in" when expiry is (or is about to be) reached
- Store the uploaded files temporarily (non-trivial)
- Stash and deal with xhr changes (non-trivial)
- Use a refresh token if possible (trivial but dependent on OAuth server changes)

Env vars
--------

Add  `AUTH_EXPIRY_MARGIN` to determine how much time should remain for the auth token to expire before attempting to refresh it. Defaults to 300 seconds (ie 5 minutes)

### Checklist

### DIFF
--- a/app/auth/middleware.js
+++ b/app/auth/middleware.js
@@ -6,7 +6,6 @@ const { decodeAccessToken } = require('../../common/lib/access-token')
 
 function processAuthResponse() {
   return async function middleware(req, res, next) {
-    // const { grant, originalRequestUrl, currentLocation } = req.session
     const accessToken = get(req.session, 'grant.response.access_token')
 
     if (!accessToken) {
@@ -22,7 +21,7 @@ function processAuthResponse() {
         getFullname(accessToken),
       ])
 
-      const previousSession = req.session
+      const previousSession = { ...req.session }
 
       req.session.regenerate(error => {
         if (error) {
--- a/app/auth/middleware.js
+++ b/app/auth/middleware.js
@@ -6,7 +6,7 @@ const { decodeAccessToken } = require('../../common/lib/access-token')
 
 function processAuthResponse() {
   return async function middleware(req, res, next) {
-    const { originalRequestUrl, currentLocation } = req.session
+    // const { grant, originalRequestUrl, currentLocation } = req.session
     const accessToken = get(req.session, 'grant.response.access_token')
 
     if (!accessToken) {
@@ -22,18 +22,28 @@ function processAuthResponse() {
         getFullname(accessToken),
       ])
 
+      const previousSession = req.session
+
       req.session.regenerate(error => {
         if (error) {
           return next(error)
         }
-
         req.session.authExpiry = decodedAccessToken.exp
-        req.session.currentLocation = currentLocation
-        req.session.originalRequestUrl = originalRequestUrl
         req.session.user = new User({
           fullname,
           locations,
           roles: decodedAccessToken.authorities,
+        })
+
+        // copy any previous session properties ignoring grant or any that already exist
+        Object.keys(previousSession).forEach(key => {
+          if (req.session[key]) {
+            return
+          }
+          if (key === 'grant') {
+            return
+          }
+          req.session[key] = previousSession[key]
         })
 
         next()
--- a/app/auth/middleware.test.js
+++ b/app/auth/middleware.test.js
@@ -199,7 +199,7 @@ describe('Authentication middleware', function() {
           })
 
           it('does not copy grant property', function() {
-            expect(req.session.grant).to.equal(undefined)
+            expect(req.session).not.to.contain.property('grant')
           })
 
           it('calls the next action', function() {
--- a/app/auth/middleware.test.js
+++ b/app/auth/middleware.test.js
@@ -41,6 +41,8 @@ describe('Authentication middleware', function() {
           originalRequestUrl: '/test',
           currentLocation: '1234567890',
           regenerate: sinon.stub(),
+          anotherKey: 'abc',
+          grant: 'grantObject',
         },
       }
     })
@@ -190,6 +192,14 @@ describe('Authentication middleware', function() {
 
           it('sets the location in the session', function() {
             expect(req.session.currentLocation).to.equal('1234567890')
+          })
+
+          it('copies additional properties in the session', function() {
+            expect(req.session.anotherKey).to.equal('abc')
+          })
+
+          it('does not copy grant property', function() {
+            expect(req.session.grant).to.equal(undefined)
           })
 
           it('calls the next action', function() {
--- a/common/components/multi-file-upload/multi-file-upload.js
+++ b/common/components/multi-file-upload/multi-file-upload.js
@@ -125,9 +125,10 @@ MultiFileUpload.prototype = {
           row.classList.remove('dz-success')
           row.classList.add('dz-error')
           row.querySelector('[data-dz-errormessage]').innerHTML = errorMessage
-          const buttons = row
-            .closest('.app-multi-file-upload__list')
-            .querySelectorAll('button[name="delete"]')
+
+          const buttons = this.dropzone.previewsContainer.querySelectorAll(
+            'button[data-dz-remove]'
+          )
           const actions = [...buttons]
           actions.forEach(action => {
             action.style.visibility = 'hidden'
--- a/common/components/multi-file-upload/multi-file-upload.js
+++ b/common/components/multi-file-upload/multi-file-upload.js
@@ -116,7 +116,24 @@ MultiFileUpload.prototype = {
           row.parentNode.removeChild(row)
           this.render()
         })
-        .catch(() => false)
+        .catch(error => {
+          let errorMessage = 'could not be deleted'
+          if (error.response && error.response.data) {
+            errorMessage = error.response.data
+          }
+          const row = target.closest('.dz-success')
+          row.classList.remove('dz-success')
+          row.classList.add('dz-error')
+          row.querySelector('[data-dz-errormessage]').innerHTML = errorMessage
+          const buttons = document.querySelectorAll(
+            '.govuk-summary-list__actions button'
+          )
+          const actions = [...buttons]
+          actions.forEach(action => {
+            action.style.visibility = 'hidden'
+          })
+          return false
+        })
     }
   },
 
--- a/common/components/multi-file-upload/multi-file-upload.js
+++ b/common/components/multi-file-upload/multi-file-upload.js
@@ -125,9 +125,9 @@ MultiFileUpload.prototype = {
           row.classList.remove('dz-success')
           row.classList.add('dz-error')
           row.querySelector('[data-dz-errormessage]').innerHTML = errorMessage
-          const buttons = document.querySelectorAll(
-            '.govuk-summary-list__actions button'
-          )
+          const buttons = row
+            .closest('.app-multi-file-upload__list')
+            .querySelectorAll('button[name="delete"]')
           const actions = [...buttons]
           actions.forEach(action => {
             action.style.visibility = 'hidden'
--- a/common/middleware/ensure-authenticated.js
+++ b/common/middleware/ensure-authenticated.js
@@ -9,13 +9,33 @@ function _isExpired(authExpiry) {
 module.exports = function ensureAuthenticated({
   provider,
   whitelist = [],
+  expiryMargin = 5 * 60,
 } = {}) {
   return (req, res, next) => {
-    if (whitelist.includes(req.url) || !_isExpired(req.session.authExpiry)) {
+    let authExpiry = req.session.authExpiry
+    if (req.method === 'GET') {
+      authExpiry -= expiryMargin
+    }
+    if (whitelist.includes(req.url) || !_isExpired(authExpiry)) {
       return next()
     }
 
     req.session.originalRequestUrl = req.originalUrl
+    if (req.method === 'POST') {
+      let errorKey
+      const contentType = req.header('content-type')
+      if (contentType.startsWith('multipart/form-data')) {
+        errorKey = 'MULTIPART_FAILED_AUTH'
+      } else if (req.xhr) {
+        errorKey = 'DELETE_FAILED_AUTH'
+      }
+      if (errorKey) {
+        res.status(401)
+        const errorString = req.t(`validation::${errorKey}`)
+        return res.send(errorString)
+      }
+      req.session.originalRequestBody = req.body
+    }
     res.redirect(`/connect/${provider}`)
   }
 }
--- a/common/middleware/ensure-authenticated.js
+++ b/common/middleware/ensure-authenticated.js
@@ -9,11 +9,11 @@ function _isExpired(authExpiry) {
 module.exports = function ensureAuthenticated({
   provider,
   whitelist = [],
-  expiryMargin = 5 * 60,
+  expiryMargin,
 } = {}) {
   return (req, res, next) => {
     let authExpiry = req.session.authExpiry
-    if (req.method === 'GET') {
+    if (req.method === 'GET' && expiryMargin) {
       authExpiry -= expiryMargin
     }
     if (whitelist.includes(req.url) || !_isExpired(authExpiry)) {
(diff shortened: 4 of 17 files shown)


common/components/multi-file-upload/multi-file-upload.js
```diff
@@ -116,7 +116,24 @@ MultiFileUpload.prototype = {
...
+          let errorMessage = 'could not be deleted'
+          if (error.response && error.response.data) {
+            errorMessage = error.response.data
+          }
+          const row = target.closest('.dz-success')
+          row.classList.remove('dz-success')
+          row.classList.add('dz-error')
+          row.querySelector('[data-dz-errormessage]').innerHTML = errorMessage
+          const buttons = document.querySelectorAll(
+            '.govuk-summary-list__actions button'
```
teneightfive: Wonder if there's a better selector than GOV.UK Design System classnames. 

It's nice to not tie JS modules to classes used for presentation. If there's need for hooks we could use JS prefix classes `js-MultiFileUpload-hookName`?

config/index.js
```diff
@@ -17,6 +17,7 @@ const SESSION = {
   TTL: process.env.SESSION_TTL || 60 * 30 * 1000, // 30 mins
   DB: process.env.SESSION_DB_INDEX || 0,
 }
+const AUTH_EXPIRY_MARGIN = process.env.AUTH_EXPIRY_MARGIN || 5 * 60 // 5 minutes
```
teneightfive: This could be done inline? I think the other process envs defined here are because they're used more than once.

common/middleware/ensure-authenticated.js
```diff
@@ -9,13 +9,33 @@ function _isExpired(authExpiry) {
 module.exports = function ensureAuthenticated({
   provider,
   whitelist = [],
+  expiryMargin = 5 * 60,
```
teneightfive: Do we need to set the default in two places? That could be confusing if we want to update it.

Maybe just set the default here?

common/middleware/ensure-authenticated.test.js
```diff
@@ -7,13 +7,16 @@ describe('Authentication middleware', function() {
...
+      const headerStub = sinon.stub()
+      headerStub.returns('')
       nextSpy = sinon.spy()
       req = {
         url: '/url',
         originalUrl: '/test',
         session: {
           authExpiry: null,
         },
+        header: headerStub,
```
teneightfive: I find it neater to write this as:

```javscript
...
header: sinon.stub().returns(''),
...
```

Then testing assertions like this:

```javascript
expect(req.header).to.be.calledOnceWithExactly(...)
```

common/middleware/ensure-authenticated.test.js
```diff
@@ -54,10 +57,129 @@ describe('Authentication middleware', function() {
...
+
+        it('should set the redirect URL in the session', function() {
+          expect(req.session.originalRequestUrl).to.equal('/test')
+        })
+      })
+
+      context('method is multipart upload', function() {
+        beforeEach(function() {
+          req.method = 'POST'
+          req.header = sinon.stub()
```
teneightfive: You could define the stubs in the parent `beforeEach`. Then in this block where you just want to change what it returns you only have to do:

```javascript
...
req.header.returns('multipart/form-data; boundary=xxxx')
req.t.returns('multipartErrorString')
...
```

common/middleware/ensure-authenticated.js
```diff
@@ -9,13 +9,33 @@ function _isExpired(authExpiry) {
...
+    }
+    if (whitelist.includes(req.url) || !_isExpired(authExpiry)) {
       return next()
     }
 
     req.session.originalRequestUrl = req.originalUrl
+    if (req.method === 'POST') {
+      let errorKey
+      const contentType = req.header('content-type')
+      if (contentType.startsWith('multipart/form-data')) {
```
teneightfive: Minor comment but I find it's a bit easier to read code when there is a bit of breathing room between blocks. Like a new line before/after an `if/else` block or after a block of variables being defined.

common/middleware/ensure-authenticated.js
```diff
@@ -9,13 +9,33 @@ function _isExpired(authExpiry) {
...
+        errorKey = 'MULTIPART_FAILED_AUTH'
+      } else if (req.xhr) {
+        errorKey = 'DELETE_FAILED_AUTH'
+      }
+      if (errorKey) {
+        res.status(401)
+        const errorString = req.t(`validation::${errorKey}`)
+        return res.send(errorString)
+      }
+      req.session.originalRequestBody = req.body
```
teneightfive: Could this cause us any issues if it becomes quite large?

common/middleware/ensure-authenticated.js
```diff
@@ -9,13 +9,33 @@ function _isExpired(authExpiry) {
...
+      const contentType = req.header('content-type')
+      if (contentType.startsWith('multipart/form-data')) {
+        errorKey = 'MULTIPART_FAILED_AUTH'
+      } else if (req.xhr) {
+        errorKey = 'DELETE_FAILED_AUTH'
+      }
+      if (errorKey) {
+        res.status(401)
+        const errorString = req.t(`validation::${errorKey}`)
+        return res.send(errorString)
```
teneightfive: Feels like this middleware is now also doing more than its original intention. 

Maybe this block needs to be its own middleware either before or after. But not the middleware to check someone is authenticated is dealing with validation which feels like its crossing concerns quite a bit.

common/middleware/ensure-authenticated.js
```diff
@@ -9,13 +9,33 @@ function _isExpired(authExpiry) {
...
     }
 
     req.session.originalRequestUrl = req.originalUrl
+    if (req.method === 'POST') {
+      let errorKey
+      const contentType = req.header('content-type')
+      if (contentType.startsWith('multipart/form-data')) {
+        errorKey = 'MULTIPART_FAILED_AUTH'
+      } else if (req.xhr) {
+        errorKey = 'DELETE_FAILED_AUTH'
```
teneightfive: This also feels a bit presumptuous and dangerous to say that an XHR request is always going to be to do with documents failing to delete.

I kind of feel this error handling needs to be kept closer to documents and the form wizard rather than tied to the whole app level authentication.

common/middleware/ensure-body-processed.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function ensureBodyProcessed(authMountpoint = '/auth') {
```
teneightfive: Could this be tied more to the auth processing? Maybe it's just a lib/function we call once when that happens. 

Doesn't feel like its a global level middleware we want to be running on every request (including things like assets). We know the specific point in time and case it needs to happen, so we can run it just the once then.

app/auth/middleware.js
```diff
@@ -6,7 +6,7 @@ const { decodeAccessToken } = require('../../common/lib/access-token')
 
 function processAuthResponse() {
   return async function middleware(req, res, next) {
-    const { originalRequestUrl, currentLocation } = req.session
+    // const { grant, originalRequestUrl, currentLocation } = req.session
```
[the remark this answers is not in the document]
solidgoldpig: Yep.

app/auth/middleware.js
```diff
@@ -22,20 +22,30 @@ function processAuthResponse() {
         getFullname(accessToken),
       ])
 
+      const previousSession = req.session
```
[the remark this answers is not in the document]
solidgoldpig: If the req.session survived the regeneration, we wouldn't need to copy the properties over ;)

Would you be happy with

```
const previousSession = { ...req.session }
```

app/auth/middleware.js
```diff
@@ -22,20 +22,30 @@ function processAuthResponse() {
...
           locations,
           roles: decodedAccessToken.authorities,
         })
 
+        // copy any previous session properties ignoring grant or any that already exist
+        Object.keys(previousSession).forEach(key => {
+          if (req.session[key]) {
+            return
+          }
+          if (key === 'grant') {
```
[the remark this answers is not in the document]
solidgoldpig: I don't think so and we're not overwriting anything that exists. The issue here is that the HMPO form wizard is using arbitrary keys to store so it's not totally predictable.

app/auth/middleware.test.js
```diff
@@ -192,6 +194,14 @@ describe('Authentication middleware', function() {
             expect(req.session.currentLocation).to.equal('1234567890')
           })
 
+          it('copies additional properties in the session', function() {
+            expect(req.session.anotherKey).to.equal('abc')
+          })
+
+          it('does not copy grant property', function() {
+            expect(req.session.grant).to.equal(undefined)
```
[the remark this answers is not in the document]
solidgoldpig: Sure. My rusty mochaness.

common/components/multi-file-upload/multi-file-upload.js
solidgoldpig: How about 

```
          const buttons = row
            .closest('.app-multi-file-upload__list')
            .querySelectorAll('button[name="delete"]')
```

config/index.js
solidgoldpig: Sure

common/middleware/ensure-authenticated.js
solidgoldpig: I think we need the ability to change it without having to make a code commit - hence the env var. I just like having a fallback value in the middleware params (especially since we're not testing what we're sending through to the middleware instantiation from `app/server`)

common/middleware/ensure-authenticated.test.js
solidgoldpig: 👍

common/middleware/ensure-authenticated.js
solidgoldpig: Can we not add that to our prettifier/eslint rules?

common/middleware/ensure-authenticated.js
solidgoldpig: We're never storing the file uploads, so how large are you expecting a standard POST to grow to? Surely if it managed to be post as a regular www/url-encoded form, we won't have issues stashing it in the redis db?

common/middleware/ensure-authenticated.js
solidgoldpig: Is the issue that I've stored the strings in the validation locales file?

I think this definitely the right place to put these checks. Yes, we could turn the auth expiry check into a module, but this code should only be temporary as per Michael Willis's comment that we should soon have auth tokens that actually last for 12 hours.

common/middleware/ensure-authenticated.js
solidgoldpig: It is presumptuous, but until the auth token duration is sorted out and this can be handled properly, I thought it was premature optimisation to do more than this.

common/middleware/ensure-body-processed.js
solidgoldpig: It's not running on every request (certainly not on assets) as it is called directly before the main router is.

I argue that it is the specific time that it needs to run - ie. the request has been authenticated and we're about to have HMPO form wizard (via our router) handle the request. That is the exact point at which we need to update the req values.

common/middleware/ensure-authenticated.test.js
solidgoldpig: OK

common/middleware/ensure-authenticated.js
solidgoldpig: (I've stuffed some new lines in)

common/middleware/ensure-body-processed.js
teneightfive: Sorry, I meant this middleware is being run on every request as it is attached to the global app router.

It feels heavy to have this check run on every request when we know at what point we need to copy across the `originalRequestBody` to `req.body`. Could we do it as part of the auth middleware callback?

app/auth/middleware.js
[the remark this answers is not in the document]
teneightfive: Maybe you're right, it doesn't matter as we're unlikely to change any previous values of `req.session`.

app/auth/middleware.js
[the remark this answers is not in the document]
teneightfive: Yeah I suppose it makes it tricky to whitelist in that case. I can't fully remember our reasoning for needing to regenerate the session but it just felt like copying over the whole session might be invalidating the security decision behind re-generating the session.

I wonder if it was something to do with the possibility of someone having gotten access to it and injecting something into it. With this code we'd be copying across whatever that was again?

common/middleware/ensure-authenticated.js
teneightfive: Yep, totally happy with having an env var to be able to change it per environment. I just meant we're setting a default in the config file **and** here. Maybe if we want a default we should set it in one place - here?

common/middleware/ensure-authenticated.js
teneightfive: Possibly, but not sure if it can be that clever...maybe one to look into

common/middleware/ensure-authenticated.js
teneightfive: I think my concern was that this middleware's purpose was quite straightforward before:

- if session not expired or whitelist URL
  - pass to next middleware
- if expired
  - store current URL
  - redirect to auth URL to handle

Now it's doing lots of things crossing over different boundaries:

- if session not expired or whitelist URL
  - pass to next middleware
- if expired
  - if POST request
     - if multipart form
       - kill process and use `res.send` to send error 
     - if XHR request (assume its a file deletion)
       - kill process and use `res.send` to send error 
  - else
    - store current URL
    - redirect to auth URL to handle

common/middleware/ensure-authenticated.js
solidgoldpig: Defaulting it to undefined here and skipping if so

app/auth/middleware.js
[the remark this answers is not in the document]
solidgoldpig: We're not copying over the grant itself - that's surely the important bit and ensures that the correct code later figures out and sets the correct user permissions. (That will happen anyway, so it is arguable that special-casing the `grant` property is *over-*zealous)

As for injecting anything into it, that means they already have root access or knowledge of the app secret and the ability to alter https-only cookies.

common/middleware/ensure-authenticated.js
solidgoldpig: Updated

common/middleware/ensure-body-processed.js
solidgoldpig: Discussed

common/middleware/ensure-authenticated.js
solidgoldpig: Sorted

common/components/multi-file-upload/multi-file-upload.js
```diff
@@ -116,7 +116,24 @@ MultiFileUpload.prototype = {
...
+          let errorMessage = 'could not be deleted'
+          if (error.response && error.response.data) {
+            errorMessage = error.response.data
+          }
+          const row = target.closest('.dz-success')
+          row.classList.remove('dz-success')
+          row.classList.add('dz-error')
+          row.querySelector('[data-dz-errormessage]').innerHTML = errorMessage
+          const buttons = row
+            .closest('.app-multi-file-upload__list')
```
teneightfive: Just spotted this is using presentation classnames still.

Having a quick look at the element and the module I think this selector might be more appropriate:

```javascript
const buttons = this.dropzone.previewsContainer.querySelectorAll('button[data-dz-remove]')
```

I also noticed there is a `js-` class in that isn't actually being used so we could tidy that element up and just remove that class (`js-upload-delete`) from the template:

image

common/middleware/process-original-request-body.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function processOriginalRequestBody(authMountpoint = '/auth') {
```
teneightfive: It looks like this is missing unit tests. It would be good to add some for this as there is a bit of logic.

solidgoldpig: My local `.gitignore` was hiding it :(

common/components/multi-file-upload/multi-file-upload.js
solidgoldpig: done

PR #402, opened 2020-03-27
[P4-1212]: feat: Dashboard for the OCA journey

barbaracassani:
This is the dashboard for the OCA journey. For now allowances are not included, as agreed with Andrew.
The existing URLs for the moves now acquire the suffix `/outgoing`, in preparation from when we will have `/incoming` as well. The current "default" view per location url is now taken by the dashboard.

Each box in the filter of the dashboard links to the corresponding filtered view; however, `total` links to `proposed`.

Please note before merging this that this changes the behaviour of the existing public website.

No e-2-e tests were harmed in the making of this change.

image

### Default moves view to which the *OCA* user with a location is redirected:
Screenshot 2020-04-01 at 17 14 22

### Default moves view to which the *non-OCA* user with a location is redirected:
Screenshot 2020-04-01 at 17 15 23

### Checklist

### DIFF
--- a/app/moves/controllers/dashboard.js
+++ b/app/moves/controllers/dashboard.js
@@ -1,4 +1,3 @@
-module.exports = function list(req, res) {
-  const template = 'moves/views/dashboard'
-  res.render(template, res.locals)
+module.exports = function dashboard(req, res) {
+  res.render('moves/views/dashboard')
 }
--- a/app/moves/controllers/dashboard.js
+++ b/app/moves/controllers/dashboard.js
@@ -1,5 +1,3 @@
-// const presenters = require('../../../common/presenters')
-
 module.exports = function list(req, res) {
   const template = 'moves/views/dashboard'
   res.render(template, res.locals)
--- a/app/moves/controllers/dashboard.js
+++ b/app/moves/controllers/dashboard.js
@@ -1,0 +1,6 @@
+// const presenters = require('../../../common/presenters')
+
+module.exports = function list(req, res) {
+  const template = 'moves/views/dashboard'
+  res.render(template, res.locals)
+}
--- a/app/moves/index.js
+++ b/app/moves/index.js
@@ -15,8 +15,7 @@ const {
   setMovesByDateAndLocation,
   setMovesByDateRangeAndStatus,
   setMovesByDateAllLocations,
-  setTotalMoves,
-  setTranslatedMoveTypes,
+  setDashboardMoveSummary,
 } = require('./middleware')
 
 const uuidRegex =
@@ -33,7 +32,7 @@ router.use('^([^.]+)$', saveUrl)
 // Define routes
 router.get('/', redirectBaseUrl)
 router.get(
-  '/:period(week|day)/:date',
+  '/:period(week|day)/:date/:view(outgoing)',
   protectRoute('moves:view:all'),
   setMovesByDateAllLocations,
   setPagination,
@@ -42,38 +41,36 @@ router.get(
 router.get(
   `/:period(week|day)/:date/:locationId(${uuidRegex})`,
   protectRoute('moves:view:by_location'),
+  protectRoute('moves:view:proposed'),
+  setMoveTypeNavigation,
+  setDashboardMoveSummary,
+  setPagination,
+  dashboard
+)
+router.get(
+  `/:period(week|day)/:date/:locationId(${uuidRegex})/:view(outgoing)`,
+  protectRoute('moves:view:by_location'),
   setMovesByDateAndLocation,
   setPagination,
   list
-)
-router.get(
-  `/:period(week|day)/:date/:locationId(${uuidRegex})/:view(dashboard)`,
-  protectRoute('moves:view:by_location'),
-  protectRoute('moves:view:proposed'),
-  setMoveTypeNavigation,
-  setTotalMoves,
-  setTranslatedMoveTypes,
-  setPagination,
-  dashboard
 )
 router.get(
   `/:period(week|day)/:date/:locationId(${uuidRegex})/:status(proposed|requested,accepted,completed|rejected)`,
   protectRoute('moves:view:by_location'),
   protectRoute('moves:view:proposed'),
   setMoveTypeNavigation,
-  setTranslatedMoveTypes,
   setMovesByDateRangeAndStatus,
   setPagination,
   listByStatus
 )
 router.get(
-  '/:period(week|day)/:date/download.:extension(csv|json)',
+  '/:period(week|day)/:date/:view(outgoing)/download.:extension(csv|json)',
   protectRoute('moves:download:all'),
   setMovesByDateAllLocations,
   download
 )
 router.get(
-  `/:period(week|day)/:date/:locationId(${uuidRegex})/download.:extension(csv|json)`,
+  `/:period(week|day)/:date/:locationId(${uuidRegex})/:view(outgoing)/download.:extension(csv|json)`,
   protectRoute('moves:download:by_location'),
   setMovesByDateAndLocation,
   download
--- a/app/moves/index.js
+++ b/app/moves/index.js
@@ -3,7 +3,7 @@ const router = require('express').Router()
 
 // Local dependencies
 const { protectRoute } = require('../../common/middleware/permissions')
-const { download, list, listByStatus } = require('./controllers')
+const { dashboard, download, list, listByStatus } = require('./controllers')
 const {
   redirectBaseUrl,
   saveUrl,
@@ -15,6 +15,8 @@ const {
   setMovesByDateAndLocation,
   setMovesByDateRangeAndStatus,
   setMovesByDateAllLocations,
+  setTotalMoves,
+  setTranslatedMoveTypes,
 } = require('./middleware')
 
 const uuidRegex =
@@ -45,10 +47,21 @@ router.get(
   list
 )
 router.get(
+  `/:period(week|day)/:date/:locationId(${uuidRegex})/:view(dashboard)`,
+  protectRoute('moves:view:by_location'),
+  protectRoute('moves:view:proposed'),
+  setMoveTypeNavigation,
+  setTotalMoves,
+  setTranslatedMoveTypes,
+  setPagination,
+  dashboard
+)
+router.get(
   `/:period(week|day)/:date/:locationId(${uuidRegex})/:status(proposed|requested,accepted,completed|rejected)`,
   protectRoute('moves:view:by_location'),
   protectRoute('moves:view:proposed'),
   setMoveTypeNavigation,
+  setTranslatedMoveTypes,
   setMovesByDateRangeAndStatus,
   setPagination,
   listByStatus
--- a/app/moves/middleware.js
+++ b/app/moves/middleware.js
@@ -48,12 +48,12 @@ const movesMiddleware = {
         userPermissions
       )
       const url = canViewProposedMoves
-        ? `${req.baseUrl}/week/${today}/${currentLocation}/dashboard`
-        : `${req.baseUrl}/day/${today}/${currentLocation}/`
+        ? `${req.baseUrl}/week/${today}/${currentLocation}/`
+        : `${req.baseUrl}/day/${today}/${currentLocation}/outgoing`
       return res.redirect(url)
     }
 
-    return res.redirect(`${req.baseUrl}/day/${today}`)
+    return res.redirect(`${req.baseUrl}/day/${today}/outgoing`)
   },
   saveUrl: (req, res, next) => {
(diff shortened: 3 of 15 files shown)


app/moves/controllers/dashboard.js
```diff
@@ -0,0 +1,6 @@
+// const presenters = require('../../../common/presenters')
```
solidgoldpig: Is the comment needed?

barbaracassani: Hmm, no 😄 Shall remove.

app/moves/middleware.js
```diff
@@ -116,24 +115,52 @@ const movesMiddleware = {
...
             })
-            .then(presenters.moveTypesToFilterComponent)
         })
       )
       next()
     } catch (error) {
       next(error)
     }
   },
+  setTotalMoves: (req, res, next) => {
```
teneightfive, line 134: I wonder if this should actually be in that dashboard controller?

The middleware functions are useful for things that are shared across lots of routes, but this is only used by the dashboard right?

app/moves/middleware.js
```diff
@@ -116,24 +115,52 @@ const movesMiddleware = {
...
+    if (!get(res, 'locals.moveTypeNavigation')) {
```
teneightfive, line 135: Also, if its in the controller you can remove logic like this as you can assume it has the correct middleware assigned, otherwise we do want it to blow up and error.

app/moves/middleware.js
```diff
@@ -116,24 +115,52 @@ const movesMiddleware = {
...
+      filter: 'proposed',
+    })
+    res.locals.moveTypeNavigation.unshift(totalMoves)
+    next()
+  },
+  setTranslatedMoveTypes(req, res, next) {
+    if (!get(res, 'locals.moveTypeNavigation')) {
+      return next()
+    }
+    res.locals.moveTypeNavigation = res.locals.moveTypeNavigation.map(
```
teneightfive, line 159: You could also move this to the controller if you wanted to move that logic there?

app/moves/views/dashboard.njk
```diff
@@ -0,0 +1,10 @@
+{% extends "./_layout.njk" %}
+
+{% block headerGridColumnClass %}govuk-grid-column-full view-dashboard{% endblock %}
```
teneightfive, line 3: I'd drop the `view-dashboard` class.

It's much cleaner to apply modifiers to components using BEM rather than do view specific CSS overrides.

common/components/filter/_filter.scss
```diff
@@ -75,3 +75,46 @@
 .app-filter__label {
   text-decoration: underline;
 }
+
+.app-filter--stacked {
+  .app-filter__list {
```
teneightfive, line 80: So another way of doing this, to avoid having to nest all these overrides under the parent modifier is to use [parent selector](https://sass-lang.com/documentation/style-rules/parent-selector):

```scss
.app-filter__list {
  ...

  .app-filter--stacked & {
    flex-direction: column;
  }
}
```

The main benefit of this organisation is that you can see the way a specific element changes based on parent modifiers. So if you change that element, it's clearer that you may affect behaviour for other modifiers.

Not a big deal for now, but one to think about in future when doing similar changes.

common/components/filter/template.njk
```diff
@@ -1,4 +1,4 @@
-<nav class="app-filter">
```
teneightfive, line 1: Maybe add a test for this too. Should be some example you can just copy from other components.

app/moves/controllers/dashboard.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function list(req, res) {
+  const template = 'moves/views/dashboard'
+  res.render(template, res.locals)
```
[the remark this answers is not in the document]
barbaracassani: Ah, good to know. Thank you!

app/moves/index.js
```diff
@@ -31,7 +33,7 @@ router.use('^([^.]+)$', saveUrl)
 // Define routes
 router.get('/', redirectBaseUrl)
 router.get(
-  '/:period(week|day)/:date',
+  '/:period(week|day)/:date/:view(outgoing)',
```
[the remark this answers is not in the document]
barbaracassani: Yup.

app/moves/index.js
```diff
@@ -49,18 +61,19 @@ router.get(
...
   protectRoute('moves:view:proposed'),
   setMoveTypeNavigation,
+  setTranslatedMoveTypes,
   setMovesByDateRangeAndStatus,
   setPagination,
   listByStatus
 )
 router.get(
-  '/:period(week|day)/:date/download.:extension(csv|json)',
+  '/:period(week|day)/:date/:view(outgoing)/download.:extension(csv|json)',
```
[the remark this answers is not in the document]
barbaracassani: Screenshot 2020-04-02 at 14 13 06

We don't.

common/components/filter/_filter.scss
barbaracassani: Ok, thank you for pointing this out.

app/moves/middleware.js
barbaracassani: This is a less clear cut case. `setTranslatedMoveTypes` is actually shared across 2 routes, and I can see it being reused in similar future views.

app/moves/middleware.js
barbaracassani: For now yes, and indeed it's written in such a fashion that it's not really reusable much outside the OCA journey. I'll move it.

app/moves/middleware.test.js
```diff
@@ -656,20 +695,23 @@ describe('Moves middleware', function() {
           moveTypeNavigation: [
             {
               active: false,
+              filter: 'proposed',
+              label: 'moves::dashboard.filter.proposed',
```
teneightfive, line 699: Nice improvement 👍

config/nunjucks/filters.test.js
```diff
@@ -190,6 +191,22 @@ describe('Nunjucks filters', function() {
...
+        const weekOpts = {
+          weekStartsOn: 1,
+        }
+        it('returns "this week"', function() {
+          expect(
+            formatDateRange([
+              startOfWeek(new Date(), weekOpts),
+              endOfWeek(new Date(), weekOpts),
+            ])
+          ).to.equal('This week')
```
teneightfive, line 206: You might want to mock the response of the translation method. If we update that text in the locale this test will fail.

app/moves/controllers/dashboard.js
```diff
@@ -0,0 +1,28 @@
...
+    filter: 'total',
+    href: find(res.locals.moveTypeNavigation, { filter: 'proposed' }).href,
+  }
+  totalMoves.value = res.locals.moveTypeNavigation.reduce(
+    (accumulator, moveType) => {
+      return (accumulator += moveType.value)
+    },
+    0
+  )
+  res.locals.moveTypeNavigation = reject(res.locals.moveTypeNavigation, {
```
teneightfive, line 15: I think it might be cleaner to use a separate variable name here rather than override `moveTypeNavigation`.

The value is used in a separate view and it feels like it will cause us less issues in the future. It's also no longer the "navigation" if you think about it. Maybe it's more a summary of proposed moves?

app/moves/controllers/dashboard.js
```diff
@@ -0,0 +1,28 @@
...
+    filter: 'proposed',
+  })
+  res.locals.moveTypeNavigation.unshift(totalMoves)
```
teneightfive: Could use ES6 spread? I find it can be a little easier to read than `unshift`:

```javascript
res.locals.moveTypeNavigation = [
  ...totalMoves,
  reject(res.locals.moveTypeNavigation, {
    filter: 'proposed',
  }),
]
```

app/moves/middleware.js
```diff
@@ -116,24 +115,46 @@ const movesMiddleware = {
...
+    totalMoves.value = res.locals.moveTypeNavigation.reduce(
+      (accumulator, moveType) => {
+        return (accumulator += moveType.value)
+      },
+      0
+    )
+    res.locals.moveTypeNavigation = reject(res.locals.moveTypeNavigation, {
+      filter: 'proposed',
+    })
+    res.locals.moveTypeNavigation.unshift(totalMoves)
```
teneightfive, line 149: Could use ES6 spread? I find it can be a little easier to read than `unshift`:

```javascript
res.locals.moveTypeNavigation = [
  ...totalMoves,
  reject(res.locals.moveTypeNavigation, {
    filter: 'proposed',
  }),
]
```

app/moves/middleware.js
```diff
@@ -116,24 +115,46 @@ const movesMiddleware = {
...
+      filter: 'total',
+      href: find(res.locals.moveTypeNavigation, { filter: 'proposed' }).href,
+    }
+    totalMoves.value = res.locals.moveTypeNavigation.reduce(
+      (accumulator, moveType) => {
+        return (accumulator += moveType.value)
+      },
+      0
+    )
+    res.locals.moveTypeNavigation = reject(res.locals.moveTypeNavigation, {
```
teneightfive, line 146: I think it might be cleaner to use a separate variable name here rather than override `moveTypeNavigation`.

It just feels less dangerous that overriding the value of something further down the middleware chain. It also feels like for this instance its not really a navigation element anymore anyway. It's more of a summary of proposed moves when being used on the dashboard.

app/moves/middleware.js
barbaracassani: Hmm... but really we want `unshift` there. I do agree that it's not an array method that it's used often, but it fits the case perfectly. I am not sure there's an argument to avoid it -- if anything, let's remind ourselves that it exists, too...

teneightfive: It was more for readability. I think the addition of the spread operator in ES6 just makes what you're trying to do here more readable. But maybe that's more personal preference.

I think it might be more me thinking in a functional programming way to not mutate where possible and to return a new value instead.

app/moves/middleware.js
```diff
@@ -116,24 +115,42 @@ const movesMiddleware = {
...
                 active: moveType.filter === req.params.status,
                 href: `${req.baseUrl}/${period}/${date}${
                   locationId ? '/' + locationId : ''
                 }/${moveType.filter}`,
               }
             })
-            .then(presenters.moveTypesToFilterComponent)
         })
       )
+      res.locals.moveTypeNavigation.map(presenters.moveTypesToFilterComponent)
```
teneightfive, line 129: Any reason why this can't be the same as before?

barbaracassani: You asked for `.map` to be moved outside the presenter. I am not sure I understand what you mean.

barbaracassani: Hmm -- that line makes no sense, you're right.

app/moves/middleware.js
```diff
@@ -116,24 +115,46 @@ const movesMiddleware = {
...
+    }
+    totalMoves.value = res.locals.moveTypeNavigation.reduce(
+      (accumulator, moveType) => {
+        return (accumulator += moveType.value)
+      },
+      0
+    )
+    totalMoves = presenters.moveTypesToFilterComponent(totalMoves)
+    res.locals.dashboardMoveSummary = [
+      { ...totalMoves },
```
teneightfive, line 151: That was my bad in the original suggestion, you can drop the spread and then you don't need to call inside in an object: 

```javascript
res.locals.dashboardMoveSummary = [
  totalMoves,
  ...reject(res.locals.moveTypeNavigation, {
    filter: 'proposed',
  }),
]
```

common/components/filter/filter.yaml
```diff
@@ -25,10 +25,15 @@ params:
...
             description: the content of the small text inside the link
+          - name: classes
+            type: string
+            required: false
+            description: the css class applied to the container
 
 examples:
   - name: default
     data:
+      classes: 'app-filter-stacked'
```
teneightfive, line 36: You might want to split this out into a separate example, as if we start to generate examples using these configurations, the stacked variation will show as the "default" version, which I don't think we want?

In other components we have a [test to pass in the classes attribute](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/components/card/template.test.js#L73-L83) to ensure it works as expected and don't actually worry about creating an "example".

barbaracassani: Ok, I'll keep that in mind. In this case, the attribute doesn't generate any different behaviour, apart from adding the class itself, so I guess it's fine (-ish).

teneightfive: Oh yes, it's missing the double hyphen. Again, it threw me because it had the name of the modifier you created.

PR #406, opened 2020-03-31
[P4-1108] feat: Use supplier name in confirmation message

teneightfive:
This adds the ability to display the supplier that serves that location
on the confirmation page.

If there are no suppliers mapped it uses a fallback.

This change is linked to #404 but has been separated so that it can be merged to avoid conflicts.

### With known supplier

image

### With unknown supplier

image

### DIFF
--- a/app/move/controllers/confirmation.js
+++ b/app/move/controllers/confirmation.js
@@ -1,7 +1,15 @@
+const { get, map } = require('lodash')
+
 function confirmation(req, res) {
   const { move_type: moveType, to_location: toLocation } = res.locals.move
+  const suppliers = get(res.locals, 'move.from_location.suppliers')
+  const supplierName =
+    suppliers && suppliers.length
+      ? map(suppliers, 'name').join(' and ')
+      : req.t('supplier_fallback')
 
   const locals = {
+    supplierName,
     location:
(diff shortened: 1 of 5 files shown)


teneightfive:
@Andy-Millar / @jason-sims I've split this out so that it doesn't get held up by the API changes for the NOMIS sync status message.

app/move/controllers/confirmation.js
```diff
@@ -1,7 +1,15 @@
+const { get, map } = require('lodash')
+
 function confirmation(req, res) {
   const { move_type: moveType, to_location: toLocation } = res.locals.move
+  const suppliers = get(res.locals, 'move.from_location.suppliers')
+  const supplierName =
+    suppliers && suppliers.length
+      ? map(suppliers, 'name').join(' and ')
```
barbaracassani: Is it possible to have more than 2 suppliers? It would result in a chain of `and`. If it's an edge case or otherwise unlikely, ignore this comment.

teneightfive: Yep, currently the API supports the possibility of multiple suppliers as it is returned as an array so this was more just to cover that case rather than just always assume it would be the first. 

That could mislead the user if we only tell them the first item in the array.

PR #407, opened 2020-03-31
fix: Fix broken unit test

barbaracassani:
Fixed unit test which attempted to destructuring an inexistent object.

### Checklist


app/moves/middleware.test.js
```diff
@@ -680,7 +680,13 @@ describe('Moves middleware', function() {
       })
       it('returns predictable results', function() {
         const locals1 = { ...res.locals }
-        middleware.setMoveTypeNavigation({}, res, next)
+        middleware.setMoveTypeNavigation(
+          {
```
teneightfive: You could just pass in `req` that is defined in the `beforeEach`?

PR #410, opened 2020-04-02
fix: Ensure content-type is defined before checking it

solidgoldpig:
Fixes https://sentry.service.dsd.io/mojds/frontend/issues/47341

### Checklist

### DIFF
--- a/common/middleware/ensure-authenticated.js
+++ b/common/middleware/ensure-authenticated.js
@@ -23,7 +23,7 @@ module.exports = function ensureAuthenticated({
     req.session.originalRequestUrl = req.originalUrl
 
     if (req.method === 'POST') {
-      const contentType = req.header('content-type')
+      const contentType = req.get('content-type') || ''
       const isMultipart = contentType.startsWith('multipart/form-data;')
       if (isMultipart || req.xhr) {
         const error = new Error(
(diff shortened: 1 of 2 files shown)


common/middleware/ensure-authenticated.js
```diff
@@ -23,7 +23,7 @@ module.exports = function ensureAuthenticated({
     req.session.originalRequestUrl = req.originalUrl
 
     if (req.method === 'POST') {
-      const contentType = req.header('content-type')
```
teneightfive, line 26: Maybe add a specific test for the scenario where the fallback is used?

common/middleware/ensure-authenticated.test.js
```diff
@@ -14,7 +14,7 @@ describe('Authentication middleware', function() {
         session: {
           authExpiry: null,
         },
-        header: sinon.stub().returns(''),
+        get: sinon.stub().returns(''),
```
teneightfive, line 17: Maybe have this return `undefined` by default in this test suite actually. Then it will force the fallback string.

With this `.returns` it will return a string and the fallback string won't be used.

solidgoldpig: 👍

teneightfive: In fact, just using `.returns()` will return undefined. Although looking at it, a string would be treated as fasly and the empty string would be used.

Might make it easier to read though not returning the same value as the fallback value in the test mock.

common/middleware/ensure-authenticated.js
solidgoldpig: Resolved by returning undefinded from base stub

PR #412, opened 2020-04-03
[P4 1150] Journey for the OCA user

barbaracassani:
This PR provides the journey for the OCA user -- based on the logged in user having the correct role and the current location being a `prison`.

It also includes the tests described in P4-1151.

~For the moment being the tests fail because the api incorrectly validates the date on save.~
The above is now fixed.

### Checklist

### DIFF
--- a/app/move/controllers/create/save.test.js
+++ b/app/move/controllers/create/save.test.js
@@ -554,6 +554,19 @@ describe('Move controllers', function() {
             }
           )
         })
+
+        context('when to location is also Prison', function() {
+          beforeEach(function() {
+            req.form.values.to_location_type = 'prison'
+            controller.process(req, {}, {})
+          })
+          it('will create a status of proposed on the move', function() {
+            expect(req.sessionModel.set).to.have.been.calledWithExactly(
+              'status',
+              'proposed'
+            )
+          })
+        })
       })
     })
 
--- a/app/move/fields/create.js
+++ b/app/move/fields/create.js
@@ -124,6 +124,13 @@ const dateField = {
   },
 }
 
+const moveTypeCourtAppearance = {
+  id: 'move_type',
+  value: 'court_appearance',
+  text: 'fields::move_type.items.court_appearance.label',
+  conditional: 'to_location_court_appearance',
+}
+
 module.exports = {
   // Person search
   'filter.police_national_computer': {
@@ -254,30 +261,23 @@ module.exports = {
   to_location: {},
   move_type: {
     ...cloneDeep(moveType),
+    items: [{ ...moveTypeCourtAppearance }],
+  },
+  move_type__prison_to_prison: {
+    ...cloneDeep(moveType),
     items: [
+      { ...moveTypeCourtAppearance },
       {
-        id: 'move_type',
-        value: 'court_appearance',
-        text: 'fields::move_type.items.court_appearance.label',
-        conditional: 'to_location_court_appearance',
+        value: 'prison',
+        text: 'fields::move_type.items.prison.label',
+        conditional: 'to_location_prison',
       },
-      // TODO: Remove once we want to allow this journey
-      // {
-      //   value: 'prison',
-      //   text: 'fields::move_type.items.prison.label',
-      //   conditional: 'to_location_prison',
-      // },
     ],
   },
   move_type__police: {
     ...cloneDeep(moveType),
     items: [
-      {
-        id: 'move_type',
-        value: 'court_appearance',
-        text: 'fields::move_type.items.court_appearance.label',
-        conditional: 'to_location_court_appearance',
-      },
+      { ...moveTypeCourtAppearance },
       {
         value: 'prison_recall',
         text: 'fields::move_type.items.prison_recall.label',
@@ -469,12 +469,12 @@ module.exports = {
     items: [
       {
         id: 'move_agreed',
-        value: true,
+        value: 'yes',
         text: 'Yes',
         conditional: 'move_agreed_by',
       },
       {
-        value: false,
+        value: 'no',
         text: 'No',
       },
     ],
@@ -492,7 +492,7 @@ module.exports = {
     validate: 'required',
     dependent: {
       field: 'move_agreed',
-      value: true,
+      value: 'yes',
     },
   },
   implicitAssessmentQuestions,
--- a/app/move/steps/create.js
+++ b/app/move/steps/create.js
@@ -55,14 +55,7 @@ const riskStep = {
     {
       field: 'from_location_type',
       value: 'prison',
-      next: [
-        {
-          field: 'to_location_type',
-          value: 'prison',
-          next: 'save',
-        },
-        'health-information',
-      ],
+      next: 'special-vehicle',
     },
     'health-information',
   ],
@@ -80,30 +73,6 @@ const healthStep = {
       next: 'document',
     },
     'save',
-  ],
-}
-
-const moveDateStep = {
-  pageTitle: 'moves::steps.move_date.heading',
-  next: [
-    {
-      field: 'move_type',
-      value: 'court_appearance',
-      next: 'court-information',
-    },
-    {
-      field: 'from_location_type',
-      value: 'prison',
-      next: [
-        {
-          field: 'to_location_type',
-          value: 'prison',
-          next: 'prison-transfer-reason',
-        },
-        'release-status',
-      ],
-    },
-    'risk-information',
   ],
 }
 
@@ -184,7 +153,27 @@ module.exports = {
     next: 'prison-transfer-reason',
   },
   '/move-date': {
-    ...moveDateStep,
+    pageTitle: 'moves::steps.move_date.heading',
+    next: [
+      {
+        field: 'move_type',
+        value: 'court_appearance',
+        next: 'court-information',
+      },
+      {
+        field: 'from_location_type',
+        value: 'prison',
+        next: [
+          {
+            field: 'to_location_type',
+            value: 'prison',
+            next: 'prison-transfer-reason',
+          },
+          'release-status',
+        ],
+      },
+      'risk-information',
+    ],
     controller: MoveDate,
     fields: ['date', 'date_type', 'date_custom'],
   },
--- a/app/move/steps/create.js
+++ b/app/move/steps/create.js
@@ -117,7 +117,9 @@ module.exports = {
       },
       {
         fn: req => {
-          return req.session.user.permissions.includes('moves:view:proposed')
+          return req.session.user.permissions.includes(
+            'move:create:prison_to_prison'
+          )
         },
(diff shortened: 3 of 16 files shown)


app/move/fields/create.js
```diff
@@ -469,12 +468,12 @@ module.exports = {
     items: [
       {
         id: 'move_agreed',
-        value: true,
+        value: 'yes',
```
barbaracassani, line 471: I need to supply some explanation for this. In the model, this is described as a string, so the validation was not working. I could have changed it to boolean in the model, but we use all these as strings so far, so I have changed it here.

app/move/fields/create.js
```diff
@@ -261,12 +261,11 @@ module.exports = {
         text: 'fields::move_type.items.court_appearance.label',
         conditional: 'to_location_court_appearance',
       },
-      // TODO: Remove once we want to allow this journey
-      // {
-      //   value: 'prison',
-      //   text: 'fields::move_type.items.prison.label',
-      //   conditional: 'to_location_prison',
-      // },
+      {
```
teneightfive, line 264: Are we ready to push this whole journey live already?

I'd be concerned that this will then make it available to any existing prison users. I think we still need to create the review workflow for this to be ready, right?

app/move/fields/create.js
teneightfive: When you model, do you mean the API model?

teneightfive: This change also feels like a perfect thing that could be a separate commit, using the commit message to describe why it was changed so when looking back it's clear to understand why.

This GitHub comment will likely get lost during time and somebody may change it back to a boolean, not knowing why it was changed to a string.

app/move/steps/create.js
```diff
@@ -55,7 +55,14 @@ const riskStep = {
...
       field: 'from_location_type',
       value: 'prison',
-      next: 'special-vehicle',
+      next: [
+        {
+          field: 'to_location_type',
+          value: 'prison',
+          next: 'save',
+        },
+        'health-information',
```
teneightfive, line 64: This should remain `special-vehicle`, otherwise it will change the behaviour of the existing prison to court journey.

app/move/steps/create.js
```diff
@@ -55,7 +55,14 @@ const riskStep = {
     {
       field: 'from_location_type',
       value: 'prison',
-      next: 'special-vehicle',
+      next: [
+        {
+          field: 'to_location_type',
+          value: 'prison',
+          next: 'save',
```
teneightfive, line 62: Does the prison to prison journey not collect any information about health affecting transport then?

teneightfive: Ah right. Does the risk step come after the health step in this journey? Is there a specific reason if so? 

If not, I wonder if we keep the ordering of court -> risk (/release status) -> health (/special vehicle) consistent to simplify some of this?

app/move/steps/create.js
```diff
@@ -157,9 +171,11 @@ module.exports = {
       'gender_additional_information',
     ],
   },
+  // OCA journey
   '/move-date-range': {
-    ...moveDateStep,
+    pageTitle: 'moves::steps.move_date.heading',
```
teneightfive, line 176: Any reason to not just keep as it was? This property is the same on the shared step.

app/moves/middleware.js
```diff
@@ -39,17 +39,18 @@ const moveTypeNavigationConfig = [
...
       const userPermissions = get(req.session, 'user.permissions')
       const canViewProposedMoves = permissions.check(
         'moves:view:proposed',
         userPermissions
       )
-      const url = canViewProposedMoves
-        ? `${req.baseUrl}/week/${today}/${currentLocation}/`
-        : `${req.baseUrl}/day/${today}/${currentLocation}/outgoing`
+      const url =
+        canViewProposedMoves && currentLocation.location_type === 'prison'
```
teneightfive, line 51: Could we add a unit test to cover all scenarios.

It looks like we're only covering both, or none. What about either just `canViewProposedMoves` or just `currentLocation.location_type === 'prison'`?

test/e2e/move.new.proposed.test.js
```diff
@@ -0,0 +1,59 @@
...
+import { newMove } from './_routes'
+import { OCAUser } from './roles'
+import { page, moveDetailPage, createMovePage } from './pages'
+import { createPersonFixture } from './helpers'
+
+fixture('New proposed move').beforeEach(async t => {
+  await t.useRole(OCAUser).navigateTo(newMove)
+})
+
+test('Prison to prison as proposed move', async t => {
```
teneightfive, line 12: I think it might be nicer to have these within the existing `move.new.prison.test.js` file to keep things consistent with how we do the Police journeys.

The files are split based on from location, and then the fixtures/tests within those use the different journeys.

test/e2e/roles.js
```diff
@@ -33,3 +33,10 @@ export const prisonUser = Role(home, async t => {
     .typeText('#password', E2E.ROLES.PRISON.password)
     .pressKey('enter')
 })
+
+export const OCAUser = Role(home, async t => {
```
teneightfive, line 37: `ocaUser`? Keeps it consistent with the existing naming convention.

app/move/fields/create.js
barbaracassani: That's correct. I am going to make a change to include an additional step as we discussed this morning.

app/move/steps/create.js
barbaracassani: Basically the whole `moveDateStep` is almost entirely composed of `next` conditions, that we need to override here anyway. All of a sudden it seemed like a booby trap, importing a largish object almost entirely overridden. I think this way is clearer and cleaner.

teneightfive: I was thinking more if something was added to the shared step then it wouldn't be shared any longer. And now we have created a shared step variable which is only used by one step, which voids its "sharedness".

I'd say either:

- remove the `moveDateStep` variable and add `pageTitle` on both steps separately, and add the `next` from `moveDateStep` to the `/move-date` step
- leave it as it is

app/move/steps/create.js
barbaracassani: The sequence is:
- who
- where
- when
- move details (includes the reason for the move, but not the risk information, which is not captured)
- move agreed, and by who
- special vehicle
- save

app/move/steps/create.js
barbaracassani: Well spotted.

app/move/steps/create.js
barbaracassani: It's not actually that much of a complication on the existing journeys. The mistaken change on the `riskStep` might have given the impression that there's more variation than there is.

teneightfive: I see, yeah I thought the risk step was still part of the journey.

app/move/fields/create.js
```diff
@@ -124,6 +124,15 @@ const dateField = {
   },
 }
 
+const items = {
```
teneightfive, line 127: `items` names feels a bit ambiguous given how large this large file now and how many other shared variables we have. Items is a property on many of these fields and this option wouldn't apply to move of them.

What about just creating a variable for the court appearance move type item:

```javascript
moveTypeCourtAppearance: {
   ...
}
```

and inlcuding like this:

```javascript
items: [{ ...moveTypeCourtAppearance }],
```

app/move/fields/create.js
```diff
@@ -253,31 +262,24 @@ module.exports = {
   // move details
   to_location: {},
   move_type: {
+    ...cloneDeep(moveType),
+    items: [{ ...items.moveTypeCourtAppearance }],
+  },
+  move_type__oca: {
```
teneightfive, line 268: Can we remove the associating from to the role from this name. This journey may not always be tied to that specific role.

I know it's difficult to not use the term OCA within the code because that's how we've been referring to this journey in standups, tickets, etc but I think we need to ensure that we keep the role out of logic and naming of things at this level.

app/move/steps/create.js
```diff
@@ -180,6 +192,15 @@ module.exports = {
       'to_location_court_appearance',
     ],
   },
+  '/move-details-oca': {
```
teneightfive, line 195: Same as with the field comment, I think we need to avoid using roles within logic and naming of things.

Is the rule that only people within OCA will be able to propose to move someone from one prison to another then?

app/move/fields/create.js
barbaracassani: Fair enough, but how do we call it then?

app/move/steps/create.js
barbaracassani: As far as I understand, yes -- only OCAs can `propose`. And we haven't got a solid way to identify this journey by location only. I am happy to rename it, but to what?

test/e2e/pages/moves-dashboard.js
```diff
@@ -13,9 +13,10 @@ class MovesDashboardPage extends Page {
       createMoveButton: Selector('.govuk-button').withExactText(
         'Create a move'
       ),
+      filterContainer: Selector('.app-filter'),
       pagination: {
         previousLink: Selector('.app-pagination__list-item--prev a'),
-        todayLink: Selector('.app-pagination__list-item a').withText('Today'),
+        todayLink: Selector('.app-pagination__list-item:nth-of-type(2) a'),
```
teneightfive, line 19: I think it might be safer to use a different selector for "This week", then we will ensure that the relevant users are seeing the correct link, ie in outgoing moves it's done by day therefore they will see "Today" and for your new journeys we are expecting them to see "This week".

So I'd say create a new item and test for that in the smoke test:

```javascript
thisWeekLink: Selector('.app-pagination__list-item a').withText('This week'),
```

app/move/controllers/create/save.test.js
```diff
@@ -554,6 +554,20 @@ describe('Move controllers', function() {
...
           )
         })
+
+        context('when to location is also Prison', function() {
+          beforeEach(function() {
+            req.form.values.to_location_type = 'prison'
+            controller.process(req, {}, {})
+          })
+          it('will create a status of proposed on the move', function() {
+            expect(
```
teneightfive, line 564: This would be easier to read as:

```javascript
expect(req.sessionModel.set).to.be.calledOnceWithExactly(
  'status',
  'proposed'
)
```

And you then don't need to include `isEqual`. Feels unnecessary as mocha already handles assertions.

Or if you need to assert a particular call (where the argument to `getCall` is the call number using zero based index):

```javascript
expect(req.sessionModel.set.getCall(1)).to.be.calledWithExactly(
  'status',
  'proposed'
)
```

barbaracassani: This was done to prevent the test to be unnecessarily bound by order of calls; but fair point re: readability.

PR #413, opened 2020-04-03
fix: Visually hide date fieldset legend

teneightfive:
This change ensures the step now behaves more consistently with other
steps that only have one question. It's also the pattern suggested
on the GOV.UK Design System when asking only one question on a page.

## Before

image

## After

image

### DIFF
--- a/app/move/fields/create.js
+++ b/app/move/fields/create.js
@@ -307,7 +307,7 @@ module.exports = {
     fieldset: {
       legend: {
         text: 'fields::date_type.label',
-        classes: 'govuk-fieldset__legend--m',
+        classes: 'govuk-visually-hidden govuk-fieldset__legend--m',
       },
     },
     items: [


app/move/fields/create.js
```diff
@@ -307,7 +307,7 @@ module.exports = {
     fieldset: {
       legend: {
         text: 'fields::date_type.label',
-        classes: 'govuk-fieldset__legend--m',
+        classes: 'govuk-visually-hidden govuk-fieldset__legend--m',
```
solidgoldpig, line 310: Isn't the govuk design system way actually to set a`isPageHeading` property to `true` on the `legend`?

teneightfive: Yep, good spot. That is the way they suggest doing it now. But that might need some re-working of the current templates.

Maybe one to add to the issues?

PR #415, opened 2020-04-06
Extend end-to-end tests

teneightfive:
## Proposed changes

These set of changes update the way the end-to-end tests are run.

It refactors how forms are filled in, specifically during the move creation steps, to make it more consistent. Forms can now be filled in using a shared method `fillInForm` by passing any field type to it in a defined order. This allows us to to more easily fill in conditional fields.

It also adds page object methods to complete the assessment steps and checks on the move detail page to ensure that the correct flags and comments are displayed.

Form fields are also now defined in the page objects using TestCafe selectors which are passed to the `fillInForm` method. This means forms no longer rely on the fieldset legend of field label to fill in the information meaning end-to-end tests are less likely to break when making content changes.

### Before

<kbd></kbd>

### After

<kbd></kbd>


test/e2e/_helpers.js
```diff
@@ -0,0 +1,230 @@
...
+})
+
+export function generatePerson(overrides = {}) {
+  const firstNames = faker.name.firstName()
+  const lastName = faker.name.lastName()
+  return {
+    lastName,
+    firstNames,
+    fullname: `${lastName}, ${firstNames}`.toUpperCase(),
+    policeNationalComputer: faker
```
solidgoldpig, line 22: I take it we don't need the other identifier_type props ( `criminal_records_office'`, `niche_reference` or `athena_reference`?)

test/e2e/_helpers.js
```diff
@@ -0,0 +1,230 @@
...
+ * Get CSV files downloaded from the app today (glob pattern matched)
+ *
+ * @returns {string[]}
+ */
+export function getCsvDownloadFilePaths() {
+  const dateStamp = format(new Date(), 'yyyy-MM-dd')
+  const globPattern = `${join(
+    homedir(),
+    'Downloads'
+  )}/Moves on*(Downloaded ${dateStamp}*.csv`
```
solidgoldpig, line 198: Pretty funky path/filename there!

test/e2e/_helpers.js
teneightfive: That's a good thought. 

We don't capture those anywhere yet, but I think if a record has them we will display them back in the personal details section so it might be worth adding them to this so we can test they are displayed correctly when present.

test/e2e/_helpers.js
teneightfive: That's true. I think this is the existing code. Haven't really looked at how the download tests work too much yet.

PR #416, opened 2020-04-07
[P4-1237] Ensure court information is only displayed when appropriate

teneightfive:
This change ensures that the court information section is only
displayed on the move detail page when the move is to a court
location.

Previously we would show the court section for all types of moves, but for Police to Prison or Prison to Prison moves for example, we will no longer need to display this information.

## Before

localhost_3000_move_7215c70c-902b-495d-bde1-3973cdd20986 (1)

## After

localhost_3000_move_7215c70c-902b-495d-bde1-3973cdd20986

### DIFF
--- a/app/move/views/_includes/assessment.njk
+++ b/app/move/views/_includes/assessment.njk
@@ -51,18 +51,20 @@
   {% endfor %}
 {% endfor %}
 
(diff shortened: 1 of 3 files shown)


teneightfive:
Has been replaced by #417

PR #418, opened 2020-04-08
[P4-1236] Display hearings on move detail page

teneightfive:
## Proposed changes

This change adds hearings to the display of hearings on the move detail page. It uses the GOV.UK Summary List component to display the information for a hearing and will not display any rows that don't have a value.

It also changes the display slightly based on the number of hearings. If there is only one hearing it condenses the display of numbers and uses a singular heading.

#### NOMIS status

If a court date was not created in NOMIS during the creation of the move it will display a warning message to check manually that the court date has been created.

### With one hearing

<kbd></kbd>

### With multiple hearings

<kbd></kbd>

### With no hearings

<kbd></kbd>

### DIFF
--- a/app/move/controllers/view.js
+++ b/app/move/controllers/view.js
@@ -1,3 +1,5 @@
+const { sortBy } = require('lodash')
+
 const presenters = require('../../../common/presenters')
 
 module.exports = function view(req, res) {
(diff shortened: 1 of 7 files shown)


teneightfive:
Just had a thought - @Andy-Millar do we need to include the court type? We will be getting it from NOMIS and storing it, but don't know if its useful in this screen?

Andy-Millar:
@teneightfive where do we mean display it? Not sure I follow
We need to display it as OMUs prioritise courts in the case of double bookings - ie. Crown Court beats Mag Court

teneightfive:
> @teneightfive where do we mean display it? Not sure I follow
> We need to display it as OMUs prioritise courts in the case of double bookings - ie. Crown Court beats Mag Court

Ah cool. So we need to show it somehow. We're currently not showing the court type on this screen but I'll add it to each hearing section.

teneightfive:
@Andy-Millar have added "Case type" to the hearing deatils

PR #419, opened 2020-04-08
[NOJIRA] Fix end to end test and confirmation page for proposed moves

barbaracassani:
The text on the confirmation page is different between scheduled and proposed moves.

### Proposed
Screenshot 2020-04-08 at 17 23 55

### Scheduled 

Screenshot 2020-04-08 at 17 03 47

Also: fixed broken e2e test.

### Checklist

### DIFF
--- a/locales/en/moves.json
+++ b/locales/en/moves.json
@@ -152,11 +152,14 @@
   },
   "confirmation": {
     "page_title": "Move scheduled for {{name}}",
+    "page_title_proposed": "Move sent for {{name}}",
     "panel": {
       "heading": "Move scheduled",
+      "heading_proposed": "Move sent for review",
       "content": "Reference number<br><strong>{{moveReference}}</strong>"
     },
-    "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled with {{supplier}}."
+    "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled with {{supplier}}.",
+    "detail_proposed": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> has been sent to the Population Management Unit (PMU) for review."
   },
   "agreement_status": {
     "heading": "Has this move been agreed with receiving prison?"
--- a/app/move/views/confirmation.njk
+++ b/app/move/views/confirmation.njk
@@ -5,7 +5,9 @@
 {% endblock %}
 
 {% block pageTitle %}
-  {{ t("moves::confirmation.page_title", { name: move.person.fullname | upper }) }}
+  {{ t("moves::confirmation.page_title", {
+    context: move.status,
+    name: move.person.fullname | upper }) }}
 {% endblock %}
 
 {% block beforeContent %}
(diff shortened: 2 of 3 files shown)


locales/en/moves.json
```diff
@@ -151,12 +151,16 @@
     }
   },
   "confirmation": {
-    "page_title": "Move scheduled for {{name}}",
+    "page_title_requested": "Move scheduled for {{name}}",
```
teneightfive, line 154: I would probably leave this line as is and tweak when the context is sent, that way the default remains the `scheduled` for now and we don't have to add a line here for every status.

locales/en/moves.json
```diff
@@ -151,12 +151,16 @@
...
+    "page_title_proposed": "Move sent for {{name}}",
     "panel": {
-      "heading": "Move scheduled",
-      "content": "Reference number<br><strong>{{moveReference}}</strong>"
+      "heading_requested": "Move scheduled",
+      "heading_proposed": "Move sent",
+      "content_requested": "Reference number<br><strong>{{moveReference}}</strong>",
```
teneightfive, line 159: We probably don't need to use different contexts for the reference number, we'll use the same text here regardless of status.

locales/en/moves.json
teneightfive: In fact, after checking, if it doesn't find the context, it will use the fallback.

So if we leave this line unchanged, if it doesn't find the status it will use that translation. Then we can define a context line for any status where we want to override the default message.

locales/en/moves.json
```diff
@@ -151,12 +151,16 @@
     }
   },
   "confirmation": {
-    "page_title": "Move scheduled for {{name}}",
+    "page_title_requested": "Move scheduled for {{name}}",
+    "page_title_proposed": "Move sent for {{name}}",
     "panel": {
-      "heading": "Move scheduled",
-      "content": "Reference number<br><strong>{{moveReference}}</strong>"
+      "heading_requested": "Move scheduled",
```
teneightfive, line 157: This can be left as `heading` instead of `heading_requested`.

locales/en/moves.json
```diff
@@ -151,12 +151,16 @@
...
+      "heading_proposed": "Move sent",
+      "content_requested": "Reference number<br><strong>{{moveReference}}</strong>",
+      "content_proposed": "Reference number<br><strong>{{moveReference}}</strong>"
     },
-    "detail": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled with {{supplier}}."
+    "detail_requested": "The move for <strong><a href=\"{{href}}\">{{name}}</a></strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong> has been scheduled with {{supplier}}.",
```
teneightfive, line 162: This can be left as `detail` instead of `detail_requested`

locales/en/moves.json
```diff
@@ -151,12 +151,16 @@
...
   },
   "confirmation": {
-    "page_title": "Move scheduled for {{name}}",
+    "page_title_requested": "Move scheduled for {{name}}",
+    "page_title_proposed": "Move sent for {{name}}",
     "panel": {
-      "heading": "Move scheduled",
-      "content": "Reference number<br><strong>{{moveReference}}</strong>"
+      "heading_requested": "Move scheduled",
+      "heading_proposed": "Move sent",
```
teneightfive, line 158: What about "Move sent for review" to be clearer about what is happening next?

PR #420, opened 2020-04-12
Update documents to include human readable filesize

teneightfive:
## Proposed changes

These changes use existing GOV.UK Design system styles to display supporting documents instead of having the need for more internal components.

It also updates the filesize to a more human readable format and tidies up the spacing to be more consistent.

### Before

<kbd></kbd>

### After

<kbd></kbd>


app/move/views/view.njk
```diff
@@ -70,15 +70,31 @@
...
+              class="govuk-link"
+              target="_blank"
+              aria-labelledby="document-{{ loop.index }}">
+              {{- document.filename -}}
+            </a>
+            <span class="govuk-body-s">
+              ({{ document.filesize | filesize }})
+            </span>
+            <div class="govuk-hint govuk-!-font-size-16 govuk-!-margin-bottom-0" id="document-{{ loop.index }}">
+              (Opens in a new window)
```
solidgoldpig, line 93: Add this to the i18n strings?

teneightfive: Good spot

PR #424, opened 2020-04-14
feature: Add ability to edit Police PTR

solidgoldpig:
[P4-1195]

[will add in some screenshots and info when e2e tests written]

### Checklist


app/move/controllers/update/move-details.test.js
```diff
@@ -0,0 +1,166 @@
+const CreateMoveDetails = require('../create/move-details')
+const MixinProto = CreateMoveDetails.prototype
+const UpdateBaseController = require('./base')
+
+const moveService = require('../../../../common/services/move')
+
+// TODO: figure out how to proxyquire lodash without breaking it elsewhere
```
barbaracassani, line 7: I think you are thinking of `require('proxyquire').noPreserveCache();`

solidgoldpig: 👍

app/move/middleware.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
 
     try {
       res.locals.move = await moveService.getById(moveId)
+      res.locals.moveId = moveId
```
teneightfive, line 11: What's the benefit of setting this as a separate variable?

app/move/fields/update.js
```diff
@@ -0,0 +1,23 @@
+const { cloneDeep } = require('lodash')
+const createFields = require('./create')
+
+const updateFields = cloneDeep(createFields)
+
+updateFields.police_national_computer = {
+  ...updateFields.police_national_computer,
+  updateProtect: true,
+  updateComponent: {
```
teneightfive, line 9: Could we make use of the existing `component` property?

app/move/index.js
```diff
@@ -39,13 +54,28 @@ router.use(
...
 
+updateSteps.forEach(updateStep => {
+  const key = Object.keys(updateStep)[0]
+  const updateStepConfig = {
+    ...updateConfig,
+    name: 'update-a-move' + key,
+    journeyName: 'update-a-move' + key,
+  }
+  router.use(
+    '/:moveId',
```
teneightfive, line 73: Wonder if it should include a clear url segment  as well to show a step is being edited: `/:moveId/edit`?

app/move/steps/update.js
```diff
@@ -0,0 +1,43 @@
...
+  '/health-information': Assessment,
+}
+
+const provideStepProps = (key, step) => {
+  const stepProps = {
+    entryPoint: true,
+    checkSession: false,
+    checkJourney: false,
+    update: true,
+    updateBackStep: '/move/:moveId',
```
teneightfive, line 25: Could we make use of the existing properties that exist? Like `backLink`?

app/move/views/view.njk
```diff
@@ -60,12 +60,50 @@
...
+      {% set categoryHeading = t([categoryKey, stepsKey]) %}
+    {% endif %}
+    {% set updateLinkText = t("moves::assessment.categories.update_link", {category: categoryHeading}) %}
+    <p class="update-link"><a href="/move/{{ move.id }}{{ urls[category] }}" data-update-link="{{ category }}">{{ updateLinkText | safe }}</span></a></p>
+  {% endif %}
+{% endmacro %}
+
 {% block contentMain %}
+<style>
+.app-section {
```
teneightfive, line 85: It would be good to move these to the sass rather than having inline CSS

app/move/views/view.njk
```diff
@@ -60,12 +60,50 @@
   </header>
 {% endblock %}
 
+{% set urls = {
+  personal_details: '/personal-details',
```
teneightfive, line 64: This feels like logic that shouldn't live in a view layer. Maybe a middleware/presenter could be used to build links for the sections and use the key as reference in the view.

locales/en/moves.json
```diff
@@ -19,7 +19,8 @@
         "heading": "Supporting documents",
         "heading_plural": "Supporting documents ({{count}})",
         "empty": "No documents uploaded"
-      }
+      },
+      "update_link": "Change <span class=\"govuk-visually-hidden\">{{ category }}</span>"
```
teneightfive, line 23: You could use [i18next contexts](https://www.i18next.com/translation-function/context) to not end up with an empty span if no category is passed:

```
"update_link": "Change",
"update_link_with_description": "Change <span class=\"govuk-visually-hidden\">{{ description }}</span>"
```

and in the view:

```nunjucks
{{ t("update_link", {
  context: "with_description" if category,
  description: category
}) }}
```

app/move/controllers/create/base.js
```diff
@@ -51,16 +51,32 @@ class CreateBaseController extends FormWizardController {
     next()
   }
 
+  getMove(req, res) {
```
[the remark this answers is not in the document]
solidgoldpig: `getMoveModel` ?

the `res` gets used in the update version. And it's keeping the method signatures all aligned (do I need req, res or both - cf. my distaste at passing req and res around in the first place ;-p)

app/move/middleware.js
solidgoldpig: None really. Other than being able to use it as is. I think I probably did this before rewriting elsewhere to have getMove.

app/move/views/view.njk
solidgoldpig: I left it so for discussion. Ideally we could introspect this info from the steps themselves but we don't have a key apart from the final part of the url itself. Should we add a `key` property to the steps or am I missing something?

app/move/views/view.njk
solidgoldpig: Absolutely. But which bit of SASS? Where should the macro in here go? Should it even be a section?

app/move/steps/update.js
solidgoldpig: I'm not sure it serves the same purpose. Ideally this should just be the root route url passed in.

Actually, I know why, I did it - in case we do need a 2-or-more-step update flow. In which case, the lack of explicit `next` would probably suffice.

app/move/fields/update.js
solidgoldpig: It's conditional, so these values are there to be copied on to component if required. Personally, I'd rather have both components in the step set up with conditions that determine whether or not to show them. But that's not really the way HMPO rolls

app/move/formatters.js
```diff
@@ -6,6 +6,9 @@ const formatters = {
     const parsedDate = chrono.en_GB.parseDate(value)
     return parsedDate ? formatDate(parsedDate, format) : value
   },
+  displayDate(value, format = 'd MMM yyyy') {
```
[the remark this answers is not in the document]
solidgoldpig: Fair dos - it seems just as weird to me though to be importing from nunjucks's config. I'd prefer to have core formatters and let HMPO and Nunjucks use/wrap those

app/move/index.js
solidgoldpig: I suppose that we might want to show only parts of a move, so your suggestion is probably a good and sensible one.

app/move/fields/update.js
teneightfive: [The implementation](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/templates/form-wizard.njk#L61-L63) of using components isn't anything to do with HMPO. 

We have created that within our service so we can adapt it if we need to.

It's worth being aware that we're not using the full ["Home Office Forms"](https://github.com/UKHomeOfficeForms) implementation which is quite prescriptive for other things like rendering. We are only using the [Form Wizard](https://github.com/UKHomeOffice/passports-form-wizard) to handle the multi step journeys (and now also edit).

app/move/steps/update.js
teneightfive: But in those cases that can be handled within that specific controller. It would be better to keep it simple for now and reduce the amount of new "update" specific implementations of things.

app/move/controllers/create/base.js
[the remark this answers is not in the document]
teneightfive: Yeah I'm kind of not a fan of passing `req`, `res` around too much but such is the nature of express sometimes, especially when dealing with sessions.

I think a better option here could be to make use of the middleware hooks that the Form Wizard provides in [its lifecycle](https://github.com/UKHomeOffice/passports-form-wizard/wiki/HMPO%20Forms%20Flow.pdf). There are specific hooks for setting things like locals (`middlewareLocals`) if being used in a view layer, like the move summary or setting up things that you may need during the rest of the controller lifecycle (`middlewareSetup`). These all have access to `req` and `res`.

So if we decide we need a consistent way to get a the move or a person we can have a middleware function that sets it on `req` for all controllers. Then whether its a edit or create controller its always on the same value. And we don't need to worry about passing both the `req` and `res` objects around easier.

app/move/fields/update.js
```diff
@@ -0,0 +1,23 @@
+const { cloneDeep } = require('lodash')
+const createFields = require('./create')
+
+const updateFields = cloneDeep(createFields)
+
+updateFields.police_national_computer = {
+  ...updateFields.police_national_computer,
+  updateProtect: true,
```
teneightfive, line 8: What about `readOnly`? Feels just as clear and a more common term for this. It also matches how we use this term in the API.

app/move/middleware.js
teneightfive: Maybe lets remove this then if its not really needed. It's used in various views/controllers already. I'd say unless we're changing it everywhere there's not much point adding it for now.

app/move/views/view.njk
teneightfive: `/common/assets/scss/core` or `/common/assets/scss/utilities` or even [`/common/assets/scss/overrides/_positioning`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/assets/scss/overrides/_positioning.scss)? 

It feels like it would be better to have an override or helper class with what this is doing rather than just a classname of "section". Essentially all you need it to apply `position: relative` to the element so something like `.position-relative` or `.app-!-position-relative` (if using overrides) would be much clearer.

The same could be created for the link too. I'm not really a fan of tying classnames to their content - it makes the style too specific and confusing if the classname is ever re-used for something else. The classname for styles should be more generic description of what it is applying rather than describing what content is inside.

app/move/views/_includes/assessment.njk
```diff
@@ -1,4 +1,5 @@
 {% for assessmentCategory in assessment %}
+<section class="app-section" data-app-section="{{ assessmentCategory.key }}">
```
teneightfive, line 2: It would be good to keep indentation levels for readability.

app/move/views/_includes/assessment.njk
```diff
@@ -110,3 +114,5 @@
     {% endif %}
   {% endif %}
 {% endif %}
+{{ updateLink('court') }}
+</section>
```
teneightfive, line 118: This file seems to be missing a new line, are you using `.editorconfig`? That usually sorts this issue automatically for me when saving files.

app/move/views/view.njk
```diff
@@ -60,12 +60,53 @@
...
+<style>
+.app-section {
+  position: relative;
+}
+.update-link {
+  position: absolute;
+  top: 0.25rem;
+  right: 0;
+}
+.sticky-sidebar .update-link {
```
teneightfive, line 96: It feels like rather than overriding the style for that link when its within a certain element, could we either not apply the class in the first place that sets the element to `position: absolute` or apply a BEM style modifier to position the element statically?

app/move/views/view.njk
```diff
@@ -60,12 +60,53 @@
...
+  position: static;
+  top: auto;
+  margin-top: 20px;
```
teneightfive, line 99: When using spacings we should use the [GOV.UK Design System spacing scale](https://design-system.service.gov.uk/styles/spacing/) if applying within components in the sass, or use the override classes if only applying within a view.

app/move/views/view.njk
```diff
@@ -60,12 +60,53 @@
   </header>
 {% endblock %}
 
+{% set urls = {
```
teneightfive, line 63: I think my concern is still defining maps like this within a view. If we need logic to loop over things and set properties based on certain criteria that kind of logic should live in a piece of middleware/controller where it can be unit tested.

Then you just pass a simple object/array into a component or loop over the items of the object/array in the view.

common/components/display/template.njk
```diff
@@ -0,0 +1,16 @@
+{% from "items/macro.njk" import appItems %}
+
+<div class="app-display govuk-form-group {%- if params.classes %} {{ params.classes }}{% endif %}">
```
teneightfive, line 3: What about `readOnlyField` as the component name? This component is specifically adding a field so it would be good to be clearer about what its doing. "Display" is a bit ambiguous and could be related to anything in the view, rather than specifically displaying a field as uneditable.

app/move/middleware.js
solidgoldpig: Done

app/move/fields/update.js
solidgoldpig: Done

app/move/views/view.njk
solidgoldpig: Added `.app-!-position-relative` and `.app-!-position-top-right`

app/move/views/view.njk
solidgoldpig: Now handled in `view` controller

app/move/views/_includes/assessment.njk
solidgoldpig: Done (I think)

app/move/views/_includes/assessment.njk
solidgoldpig: Done

app/move/views/view.njk
solidgoldpig: Taken care of by addition of explicit positioning classes

app/move/views/view.njk
solidgoldpig: ditto

app/move/views/view.njk
solidgoldpig: Not sure why there are 2 separate comment threads on this line - but, as noted above, moved to the view controller

common/components/display/template.njk
solidgoldpig: Done

app/move/controllers/create/base.js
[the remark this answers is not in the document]
solidgoldpig: Moved methods from controller to request object

app/move/fields/update.js
solidgoldpig: As discussed, I think it's cleaner to determine which component get output in the controller rather than in the template. Especially as all the other update logic has now moved there too.

common/components/items/template.njk
```diff
@@ -0,0 +1,8 @@
+{% from "govuk/components/details/macro.njk" import govukDetails %}
+{% if params.items %}
+<div class="app-items {% if params.classes %}{{ params.classes }}{% endif %}">
```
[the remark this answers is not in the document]
solidgoldpig: OK, removing the items component

app/move/steps/update.js
solidgoldpig: There only 'update' specific thing now is `updateComponent` and that's now handled by the update controller

app/move/controllers/update/base.js
```diff
@@ -0,0 +1,109 @@
+const { get, keys } = require('lodash')
+
+const personService = require('../../../../common/services/person')
+const CreateBaseController = require('../create/base')
+
+class UpdateBaseController extends CreateBaseController {
+  middlewareLocals() {
```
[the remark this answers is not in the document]
solidgoldpig: Drivel

app/move/controllers/cancel.test.js
```diff
@@ -41,6 +41,7 @@ describe('Move controllers', function() {
           redirect: sinon.stub(),
           locals: {
             move: mockMove,
+            moveId: mockMove.id,
```
teneightfive, line 44: Don't think this is needed anymore, right?

app/move/controllers/create/assessment.js
```diff
@@ -64,6 +61,14 @@ class AssessmentController extends CreateBaseController {
           comments: req.form.values[key],
         }
       })
+  }
+
+  saveValues(req, res, next) {
+    const assessment = req.sessionModel.get('assessment') || {}
+    // const assessment = req.sessionModel.get('assessment') || {}
```
teneightfive, line 68: Rogue comment?

app/move/controllers/update/base.js
```diff
@@ -0,0 +1,114 @@
...
+    super.middlewareLocals()
+    this.use(this.setStepUrls)
+    this.use(this.setInitialStep)
+  }
+
+  _setModels(req) {
+    const res = req.res
```
teneightfive, line 14: This feels a little strange? Is it to avoid passing `res` around to?

I think I had something else in mind when I said to make use of the middleware functions that the form wizard provides. Rather than creating extra getters/setters I though it could work to just set the values on `req` instead of having to create get methods on req and create a set of models on req.

app/move/controllers/update/base.js
```diff
@@ -0,0 +1,114 @@
...
+    req.models.person = req.models.move.person
+  }
+
+  getBaseUrl(req) {
+    const moveId = req.getMoveId()
+    return `/move/${moveId}`
+  }
+
+  setButtonText(req, res, next) {
+    req.form.options.buttonText = 'actions::continue'
```
teneightfive, line 25: Could this be set within the wizard options instead?

app/move/controllers/view.js
```diff
@@ -14,6 +15,21 @@ module.exports = function view(req, res) {
     cancellationReason === 'other'
       ? cancellationComments
       : req.t(`fields::cancellation_reason.items.${cancellationReason}.label`)
+
+  const urls = {
```
teneightfive, line 19: Maybe name it something more clear that these are edit URLs?

app/move/views/_includes/assessment.njk
```diff
@@ -1,112 +1,120 @@
...
-      {{ t("moves::assessment.categories.court.heading") }}
-    </h2>
+    {% endfor %}
+    <p class="app-!-position-top-right"><a href="{{ urls.update[assessmentCategory.key] }}" data-update-link="{{ assessmentKey }}">{{ t('moves::update_links.' + assessmentCategory.key, {
+    context: 'information'
+    }) | safe }}</a></p>
+  </section>
+{% endfor %}
+
+<section class="app-section app-!-position-relative" data-app-section="court">
```
teneightfive, line 59: We can drop the `app-section` class right? We're not using it anymore

app/move/views/view.njk
```diff
@@ -61,11 +61,14 @@
 {% endblock %}
 
 {% block contentMain %}
-  <h2 class="govuk-heading-m">
-    {{ t("moves::steps.personal_details.heading") }}
-  </h2>
+  <section class="app-section app-!-position-relative" data-app-section="personal">
```
teneightfive, line 64: We can drop the `app-section` class right? We're not using it anymore

common/assets/scss/overrides/_positioning.scss
```diff
@@ -5,3 +5,15 @@
 .app-\!-float-left {
   float: left !important;
 }
+
+.app-\!-position-relative {
+  position: relative !important;
+}
+
+@media (min-width: 40.0625em) {
```
teneightfive, line 13: Maybe use the media mixins that are provided as part of the GOV.UK Design System for consistency?

common/assets/scss/overrides/_positioning.scss
```diff
@@ -5,3 +5,15 @@
...
+  .app-\!-position-top-right {
+    position: absolute;
+    top: 0;
+    right: 0;
+  }
+}
```
teneightfive, line 19: Seems to be missing a new line again?

teneightfive: Think there's a few files, worth checking and maybe setting up something like .editorconfig to automatically add this.

common/assets/scss/overrides/_visibility.scss
```diff
@@ -13,3 +13,19 @@
     display: block;
   }
 }
+
+@media (min-width: 40.0625em) {
```
teneightfive, line 17: Maybe use the GOV.UK Design System mixins here too?

common/controllers/form-wizard.test.js
```diff
@@ -42,7 +42,7 @@ describe('Form wizard', function() {
         const reqMock = {
           t: sinon.stub().returnsArg(0),
         }
-        errors = controller.getErrors(reqMock, {})
+        errors = controller.getErrors(reqMock, {}) // ?
```
teneightfive, line 45: Was there a reason for this?

common/lib/permissions.js
```diff
@@ -1,4 +1,4 @@
-const policePermissions = [
+const defaultPermissions = [
```
teneightfive, line 1: I'd probably refrain from creating this as we don't necessarily want to start having the idea of a "default" set of permissions for users. It feels safer to define specifically for each group rather than have the idea of a shared permissions for lots of groups.

common/templates/layouts/base.njk
```diff
@@ -2,49 +2,58 @@
 
 {# GOV.UK Design System Components #}
 {% from "govuk/components/back-link/macro.njk"         import govukBackLink %}
-{% from "govuk/components/button/macro.njk"            import govukButton %}
-{% from "govuk/components/checkboxes/macro.njk"        import govukCheckboxes %}
-{% from "govuk/components/details/macro.njk"           import govukDetails %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
```
teneightfive, line 5: This feels like your editor autosaving? Would be good to revert these to remove the noise from this set of changes.

If there are style changes we want to make we can make them separately from feature changes.

locales/en/moves.json
```diff
@@ -185,5 +186,12 @@
   },
   "agreement_status": {
     "heading": "Has this move been agreed with receiving prison?"
+  },
+  "update_links": {
+    "court": "Change <span class=\"app-!-visually-hidden-desktop\">information for the court</span>",
```
teneightfive, line 191: These all seem very similar? What about one key for the link, and passing the translations in separately so we don't have to duplicate the `Change <span>` each time?

teneightfive: You could even use the keys that already exist for the headings of the sections that are being updated for consistency (maybe lowercasing using a nunjucks filter though)

app/move/controllers/update/base.js
teneightfive: I think I see what actually. Maybe we need a separate action actually - `Update` or `Save` instead of `Continue`? Because there is no next step to "continue" to in these instances.

app/move/middleware.js
teneightfive: This still seems in there

common/assets/scss/overrides/_visibility.scss
solidgoldpig: 👍

app/move/controllers/create/assessment.js
solidgoldpig: 👍

app/move/controllers/update/base.js
solidgoldpig: Yep - was going to ask Andy about that

app/move/controllers/view.js
solidgoldpig: But it's an object of all urls. That only currently has update urls in it, but seems cleaner to say `urls.update.foo` than `updateUrls.foo`

common/controllers/form-wizard.test.js
solidgoldpig: https://wallabyjs.com/docs/intro/advanced-logging.html#live-comments

Think I'll just stick to console.logging. At least the linter will spot it

common/lib/permissions.js
solidgoldpig: Refactored so all permissions are explicitly declared and not shared

app/move/controllers/update/base.js
solidgoldpig: (But, yes, it could be set in the wizard options, now you mention it)

common/templates/layouts/base.njk
solidgoldpig: Hopefully sorted

app/move/controllers/update/base.js
solidgoldpig: OK - will revisit in editabilty part deux aka STC/SCH users

locales/en/moves.json
solidgoldpig: Done and reimplemented as a macro (declared in the view template). It seems much cleaner now.

teneightfive:
This PR has been replaced with #450

PR #425, opened 2020-04-14
[NOJIRA]  Date of travel in sidebar

barbaracassani:
Screenshot 2020-04-14 at 17 45 18

The date of travel in the sidebar.

### Checklist

### DIFF
--- a/app/moves/views/list-by-status.njk
+++ b/app/moves/views/list-by-status.njk
@@ -2,7 +2,20 @@
 
 {% block headerGridColumnClass %}govuk-grid-column-full{% endblock %}
 
+{% block dateTitle %}
+  <ul class="govuk-list govuk-!-font-size-16 app-!-float-right app-print--hide">
+    <li>
+      <a href="javascript:window.print()" class="app-icon app-icon--print">
+        {{ t("actions::print_move_list") }}
+      </a>
+    </li>
+  </ul>
+{{ super() }}
+{% endblock %}
+
 {% block headerActions %}
+
+
   {% if canAccess('move:create') %}
     {{ govukButton({
       isStartButton: true,
--- a/app/moves/views/list-by-status.njk
+++ b/app/moves/views/list-by-status.njk
@@ -2,20 +2,7 @@
 
 {% block headerGridColumnClass %}govuk-grid-column-full{% endblock %}
 
-{% block dateTitle %}
-  <ul class="govuk-list govuk-!-font-size-16 app-!-float-right app-print--hide">
-    <li>
-      <a href="javascript:window.print()" class="app-icon app-icon--print">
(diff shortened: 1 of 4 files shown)


app/moves/views/list-by-status.njk
```diff
@@ -2,7 +2,20 @@
 
 {% block headerGridColumnClass %}govuk-grid-column-full{% endblock %}
 
+{% block dateTitle %}
+  <ul class="govuk-list govuk-!-font-size-16 app-!-float-right app-print--hide">
```
teneightfive, line 6: I think it would be good to use a consistent position of these actions like the outgoing view.

With this change the print link now appears between the caption ("Single moves") and the heading ("This 
 week") which doesn't feel correct.

Having taken a quick look at the markup I think it would be better to keep a two-thirds/one-third layout for  the header and the side actions and that the filter could be a separate block outside of this which wouldn't need any of the grid classes at all.

The fact that a block called `dateTitle` now has a print link inside it is also a bit of red flag here.

app/moves/views/list-by-status.njk
```diff
@@ -2,7 +2,20 @@
...
+      <a href="javascript:window.print()" class="app-icon app-icon--print">
+        {{ t("actions::print_move_list") }}
+      </a>
+    </li>
+  </ul>
+{{ super() }}
+{% endblock %}
+
 {% block headerActions %}
+
```
teneightfive, line 17: Can we tidy up these lines.

common/presenters/move-to-meta-list-component.js
```diff
@@ -68,6 +74,14 @@ function moveToMetaListComponent({
         text: filters.formatTime(timeDue),
       },
     },
+    {
```
teneightfive, line 77: It feels like this should actually be instead of the existing date item. Right now we could have the following items all displayed at once:

- Date
- Time due
- Date of travel

It feels like what we need to actually do with this change is allow the existing Date row to support a date range as well as a single value.

And we can check with @Andy-Millar if he wants to update that content to `Date of travel` from what we already have which is just `Date`. Either way I think the label should be consistent between the two scenarios:

image

app/moves/views/list-by-status.njk
barbaracassani: I have reverted this commit, it will be part of a new PR at some point.

common/presenters/move-to-meta-list-component.js
barbaracassani: I have confirmed with @Andy-Millar that date is fine. I was operating under the possibly wrong assumption that the range would go away once the date is fixed. I will change it to prioritise the fixed date if present, and to display the range if not present.

PR #426, opened 2020-04-15
Add primary navigation menu

teneightfive:
## Proposed changes

This introduces a primary navigation to the site. Some of the items are conditional based on permissions so there is a piece of middleware to handle this which passes the final items to the view layer.

### As a Police/Prison/STC/SCH user

<kbd></kdb>

### As an OCA user

<kbd></kdb>

<kbd></kdb>

### As a supplier user

<kbd></kdb>

<kbd></kdb>

### As a user with all permissions

<kbd></kdb>

### DIFF
--- a/app/moves/views/_layout.njk
+++ b/app/moves/views/_layout.njk
@@ -1,7 +1,7 @@
 {% extends "layouts/base.njk" %}
 
 {% block customGtagConfig %}
-  gtag('set', {'page_title': 'Upcoming moves'});
+  gtag('set', {'page_title': '{{ t(pageTitle) }}'});
 {% endblock %}
 
 {% if period === 'day' %}
@@ -23,7 +23,9 @@
   {% block pageHeader %}
     <header class="govuk-grid-row">
       <div class="{% block headerGridColumnClass %}govuk-grid-column-two-thirds{% endblock %}">
-        <span class="govuk-caption-xl">{{ t(pageTitle) }}</span>
+        <span class="govuk-caption-xl">{{ t(pageTitle, {
+          context: "with_date"
+        }) }}</span>
 
         {% block dateTitle %}
           <h1 class="govuk-heading-xl govuk-!-margin-bottom-2">
--- a/common/middleware/set-primary-navigation.js
+++ b/common/middleware/set-primary-navigation.js
@@ -1,0 +1,48 @@
+const { format } = require('date-fns')
+const { get } = require('lodash')
+
+const permissions = require('./permissions')
+
+const baseRegex =
+  '^\\/moves\\/(day|week)(\\/\\d{4}-\\d{2}-\\d{2})(\\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})?'
+const homeRegex = new RegExp(`${baseRegex}$`)
+const proposedRegex = new RegExp(`${baseRegex}\\/proposed$`)
+const outgoingRegex = new RegExp(`${baseRegex}\\/outgoing$`)
+
+function setPrimaryNavigation(req, res, next) {
+  const { TODAY, REQUEST_PATH } = res.locals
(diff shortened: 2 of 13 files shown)


teneightfive:
Make a test Jira comment for P4-1306 #comment Hello world

jason-sims:
As discussed:
- Confirmed if OCA & Prison role have both been added user with see in nav bar "Home, Single moves, Outgoing moves"
Other notes:
- Police/stc/sch role should not have an the OCA or Prison role combination added -  but if they do same behavior will apply. (this might apply to a CDM but never operationally) 
- Suppliers will only have supplier role added.

Cheers

app/moves/views/_layout.njk
```diff
@@ -23,7 +23,9 @@
   {% block pageHeader %}
     <header class="govuk-grid-row">
       <div class="{% block headerGridColumnClass %}govuk-grid-column-two-thirds{% endblock %}">
-        <span class="govuk-caption-xl">{{ t(pageTitle) }}</span>
+        <span class="govuk-caption-xl">{{ t(pageTitle, {
```
solidgoldpig, line 26: For the purposes of this PR this looks great, but is there any reason why we couldn't move the caption inside the h1? eg.  

```
<h1 class="govuk-!-margin-bottom-2">
  <span class="govuk-caption-xl">Upcoming moves for</span>
  <time class="govuk-heading-xl" datetime="2020-04-13">
    Monday 13 Apr 2020
  </time>
</h1>
```

common/middleware/set-primary-navigation.js
```diff
@@ -0,0 +1,48 @@
+const { format } = require('date-fns')
+const { get } = require('lodash')
+
+const permissions = require('./permissions')
+
+const baseRegex =
```
solidgoldpig, line 6: Was going to suggest deferring this to after the running of this middleware but I see that's not possible for all the routes because they have a different structure. Ideally that's what you'd do and just let the relevant controllers mark that a section is active. The form-wizard derived stuff could call such middleware in the locals phase (ie. just before rendering) but the other route controllers would need to do so before calling this.render

app/moves/views/_layout.njk
teneightfive: There isn't. I like that suggestion, particularly the addition of the `time` attribute. The use of `time` is something I wanted to review actually and use more consistently.

We have it in a few places but not in all the places we use dates/times.

I'll create a GitHub issue to address this as a separate change :)

common/middleware/set-primary-navigation.js
teneightfive: Yep, it did feel a bit smelly when writing it 😞 

But I couldn't see another way to effectively highlight the nav items correctly.

teneightfive: It's also made more complicated by our current redirect on `/moves`. But as @barbaracassani and I have discussed this was always more of a temporary thing that once we start actually introducing dashboards at those routes, the redirects can go.

But it was a case of priority in MVP days that we focusing on building the create and list journeys first over the summary dashboards.

app/moves/views/_layout.njk
teneightfive: @solidgoldpig In fact, the reason was because that's how the example is given in the [GOV.UK Design System](https://design-system.service.gov.uk/styles/typography/#headings-with-captions).

But maybe it depends on the content and context. I think you might be right that in this instance that it would read better if the caption was inside and therefore part of the heading as opposed to outside of it. What do you think?

teneightfive: Have created #428 to ensure we use `time` wherever possible.

solidgoldpig: I think it is part of the page heading. And lo... the design system covers this very scenario

> If the caption should be considered part of the page heading, you can also nest the caption within the H1.

```html
<h1 class="govuk-heading-xl">
  <span class="govuk-caption-xl">govuk-caption-xl</span>
  govuk-heading-xl
</h1>
```

PR #427, opened 2020-04-15
[P4-1279] Dashboard for allocations

barbaracassani:
This is the dashboard for allocations. Since we don't have yet the endpoint, the service obviously doesn't work yet. To see it working, uncomment the fake response from the service.

Also, I deliberately ignored the primary navigation since I believe @teneightfive is looking into restoring it -- therefore I will piggyback on his work in a separate PR. If this is not the case, let me know and I will add the primary navigation in the template.

Screenshot 2020-04-16 at 15 31 11

### Checklist


common/middleware/date-utils.js
```diff
@@ -0,0 +1,22 @@
+const {
+  getDateFromParams,
+  getDateRange,
+} = require('../../common/helpers/date-utils')
+
+function setPeriod(req, res, next, period) {
```
teneightfive, line 6: I'd be tempted to split these out actually. I've been looking at the middleware file in moves and wanted to split them out into separate files as I felt like it was getting too large to manage.

The pattern is normally to just use a `middleware.js` file for 1 or 2 middlewares, then if it grows, split into separate files but include a `middleware/index.js` loader file. That way, any previous imports aren't actually broken and it allows for expansion.

teneightfive: So here I'd maybe create `./common/middleware/set-date-period.js` and `./common/middleware/set-date-range.js`. Makes them nice and atomic and testable.

barbaracassani: Yes, makes sense. I'll do it.

app/allocations/middleware.js
```diff
@@ -0,0 +1,51 @@
...
+const allocationService = require('../../common/services/allocation')
+
+const {
+  dateFormat,
+  getDateFromParams,
+  getPeriod,
+} = require('../../common/helpers/date-utils')
+
+const allocationsMiddleware = {
+  // todo: refactor this to be common with the /moves pagination.
```
teneightfive, line 11: I believe if you use capitals as the syntax some code editors will highlight the todo (well, mine does at least) to make them easier to spot (and hopefully address):

image

image

app/allocations/middleware.js
```diff
@@ -0,0 +1,51 @@
...
+    const value = await allocationService.getCount({ dateRange })
+    res.locals.allocationsSummary = [
+      {
+        active: false,
+        label: req.t('allocations::dashboard.label'),
+        href: `/allocations/week/${getDateFromParams(req)}/`,
+        value,
+      },
+    ]
+    /* res.locals.allocationsSummary = [
```
teneightfive, line 40: Rogue comment?

app/allocations/middleware.test.js
```diff
@@ -0,0 +1,132 @@
...
+})
+
+describe('#setPagination', function() {
+  const mockDateRange = ['2020-04-13', '2020-04-19']
+  let req, res, nextSpy
+  beforeEach(function() {
+    this.clock = sinon.useFakeTimers(new Date(mockDateRange[0]).getTime())
+    nextSpy = sinon.spy()
+  })
+
```
teneightfive, line 53: Spotted a few times there's some inconsistency with new line spacing. I find it easier to scan when there is a line between blocks on the same level.

Not a massive deal but might look to see if there's some eslint rules to make it easier when swapping between files.

common/middleware/index.js
```diff
@@ -0,0 +1,7 @@
+const setPeriod = require('./set-period')
+const setDateRange = require('./set-date-range')
+
+module.exports = {
```
teneightfive, line 4: This is a good spot. We should have added an index loader to this folder.

common/middleware/set-period.js
```diff
@@ -0,0 +1,6 @@
+function setPeriod(req, res, next, period) {
```
teneightfive, line 1: I wonder if this should be renamed to `setDatePeriod` now its in a more common place. It feels like "period" on its own isn't as clear now its not contained within the moves app.

app/allocations/middleware.js
barbaracassani: An example of the payload, useful as for now in fact all is returned is a 404. But this might change soon, so I can probably safely remove it.

app/allocations/middleware.js
barbaracassani: Ha! Webstorm does it with both cases 😄 But happy to capitalise it if it's the standard.

app/allocations/middleware.test.js
barbaracassani: I will add the line. We might add https://eslint.org/docs/rules/lines-between-class-members in our eslint config.

PR #429, opened 2020-04-17
Move saving of court hearings to end of create journey

teneightfive:
### Why

Court hearings were previously being saved after the court hearings
step. But this meant they were being sent without a move ID, which is
required to be able to save that hearing to NOMIS.

### What

This change moves that logic until after the creation of a move so that the move ID can
be sent as part of the court hearing. This allows it to be created in NOMIS and associates
that court hearing with the move.

### DIFF
--- a/app/move/controllers/create/court-hearings.js
+++ b/app/move/controllers/create/court-hearings.js
@@ -3,7 +3,6 @@ const { formatISO, parseISO } = require('date-fns')
 
 const CreateBaseController = require('./base')
 const personService = require('../../../../common/services/person')
-const courtHearingService = require('../../../../common/services/court-hearing')
 const componentService = require('../../../../common/services/component')
 const presenters = require('../../../../common/presenters')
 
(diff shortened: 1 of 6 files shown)


common/services/court-hearing.js
```diff
@@ -1,9 +1,22 @@
+const { mapValues, pickBy } = require('lodash')
+
 const apiClient = require('../lib/api-client')()
 
 const courtHearingService = {
+  format(data) {
+    const relationships = ['move']
+
+    return mapValues(pickBy(data), (value, key) => {
```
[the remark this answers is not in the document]
teneightfive: Yes. Will be fixing separately later :/

app/move/controllers/create/court-hearings.js
```diff
@@ -43,43 +42,37 @@ class CourtHearingsController extends CreateBaseController {
...
       court_hearing__comments: comments,
       court_hearing__court_case: courtCaseId,
       court_hearing__start_time: startTime,
     } = req.form.values
-    const courtCase = find(req.courtCases, {
-      id: courtCaseId,
-    })
+    const moveDate = req.sessionModel.get('date')
+    const courtCase = find(req.courtCases, { id: courtCaseId })
+    const hearingDatetime = parseISO(`${moveDate}T${startTime}`)
```
solidgoldpig, line 58: Is this really the only place we do (or will ever do) this? Just seems like a crucial bit of formatting is a bit buried away here

teneightfive: Valid point. It's the only time right now where we ask the user for only a time, and convert it to a datetime. 

It's also quite specific here where we give it the move date (from an earlier step) and use the time entered on this step so felt logical to do it here.

PR #437, opened 2020-04-20
[P4 1309] Generic table for allocations and moves

barbaracassani:
Screenshot 2020-04-21 at 11 51 12

This PR represents a first iteration of the table component, with no pagination nor sorting. I have used the OCA journey to mount it since the PMU dashboard is not ready yet.
A couple of things:
- this PR has unearthed a previously unnoticed bug: the reason for move is not saved correctly. I will investigate the reason
- The default output of the date range filter when passed only one date is different from the format of the single date formatter (you can notice the difference in the second row of the screenshot). This may need to be addressed separately.
- I have kept the table presenter very generic for it to work for both moves and allocations, and I am sure this will cause discussion, since therefore most of the presentational behaviour is handled by the config attached in the controller. If that seems unsavoury, I can move the config elsewhere as an intermediate layer between controller and presenter.

This PR contains also a couple of things that I have done to make the API request for allocations work, since it's now available in a branch.

### Checklist

### DIFF
--- a/app/allocations/middleware.test.js
+++ b/app/allocations/middleware.test.js
@@ -35,7 +35,7 @@ describe('#setAllocationsSummary', function() {
     expect(locals.allocationsSummary).to.deep.equal([
       {
         active: false,
-        label: 'allocations::dashboard.label',
+        label: 'allocations::dashboard.labels.single_requests',
         href: '/allocations/week/2020-01-01/',
         value: 18,
       },
--- a/app/moves/controllers/list-by-status.js
+++ b/app/moves/controllers/list-by-status.js
@@ -5,12 +5,7 @@ module.exports = function list(req, res) {
   const template = 'moves/views/list-by-status'
   const locals = {
     pageTitle: 'moves::dashboard.single_moves',
-    moves: movesByRangeAndStatus.map(
-      presenters.moveToCardComponent({
-        showMeta: false,
-        showTags: false,
-      })
-    ),
+    ...presenters.movesToTable(movesByRangeAndStatus),
   }
 
   res.render(template, locals)
--- a/common/lib/api-client/models.js
+++ b/common/lib/api-client/models.js
@@ -180,4 +180,25 @@ module.exports = {
       filesize: '',
     },
   },
+  allocation: {
+    attributes: {
+      from_location: {
+        jsonApi: 'hasOne',
+        type: 'locations',
+      },
+      to_location: {
+        jsonApi: 'hasOne',
+        type: 'locations',
+      },
+      date: '',
+      prisoner_category: '',
+      sentence_length: '',
+      complex_cases: [],
+      moves_count: 0,
+      complete_in_full: '', // boolean
+      other_criteria: '',
+      created_at: '',
+      updated_at: '',
+    },
+  },
 }
--- a/common/lib/api-client/models.js
+++ b/common/lib/api-client/models.js
@@ -15,6 +15,10 @@ module.exports = {
       date: '',
       date_from: '',
       date_to: '',
+      prison_transfer_reason: {
+        jsonApi: 'hasOne',
+        type: 'prison_transfer_reasons',
+      },
       person: {
         jsonApi: 'hasOne',
         type: 'people',
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -49,6 +49,9 @@ function formatDateRange(dateRange) {
   if (!Array.isArray(dateRange) || dateRange.length !== 2) {
     return dateRange
   }
+  if (!dateRange[1]) {
+    return dateRange[0]
+  }
   const parsedDates = dateRange.map(date => {
     return isDate(date) ? date : parseISO(date)
   })
--- a/locales/en/moves.json
+++ b/locales/en/moves.json
@@ -133,6 +133,8 @@
     "no_results": "No moves requested",
     "single_moves": "Single moves",
     "single_moves_no_results": "No moves proposed",
+    "created_at": "Created at",
+    "move_type": "Move type",
     "filter": {
       "proposed": "proposed",
       "approved": "approved",
--- a/app/allocations/middleware.js
+++ b/app/allocations/middleware.js
@@ -32,7 +32,7 @@ const allocationsMiddleware = {
     res.locals.allocationsSummary = [
       {
         active: false,
-        label: req.t('allocations::dashboard.label'),
+        label: req.t('allocations::dashboard.labels.single_requests'),
         href: `/allocations/week/${getDateFromParams(req)}/`,
         value,
       },
--- a/app/moves/views/list-by-status.njk
+++ b/app/moves/views/list-by-status.njk
@@ -18,9 +18,10 @@
 
 {% block pageContent %}
 {% if moves.length %}
-  {% for move in moves %}
-    {{ appCard(move) }}
-  {% endfor %}
+  {{ govukTable({
+    rows: moves,
+    head: movesHeads
+  }) }}
 {% else %}
     {{ appMessage({
       classes: "app-message--muted govuk-!-margin-top-7",
--- a/common/assets/scss/application.scss
+++ b/common/assets/scss/application.scss
@@ -26,6 +26,7 @@ $govuk-font-family: "Inter", system-ui, sans-serif;
 @import "govuk-frontend/govuk/components/select/select.scss";
 @import "govuk-frontend/govuk/components/skip-link/skip-link.scss";
 @import "govuk-frontend/govuk/components/summary-list/summary-list.scss";
+@import "govuk-frontend/govuk/components/table/table.scss";
 @import "govuk-frontend/govuk/components/textarea/textarea.scss";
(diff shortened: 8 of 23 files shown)


.gitignore
```diff
@@ -72,3 +72,6 @@ typings/
 
 # Test cafe reports
 report*.json
+
+# IDE configs
+.idea
```
teneightfive, line 77: Maybe this could live in your global `.gitconfig`? Feels more specific to the developers chosen environment that related to this application.

app/allocations/middleware.test.js
```diff
@@ -35,7 +35,7 @@ describe('#setAllocationsSummary', function() {
     expect(locals.allocationsSummary).to.deep.equal([
       {
         active: false,
-        label: 'allocations::dashboard.label',
+        label: 'allocations::dashboard.labels.single_requests',
```
teneightfive, line 38: Has this been squashed into the wrong commit? There's no update to the middleware as part of this change?

app/moves/controllers/list-by-status.js
```diff
@@ -1,16 +1,39 @@
 const presenters = require('../../../common/presenters')
+const i18n = require('../../../config/i18n')
+const filters = require('../../../config/nunjucks/filters')
+
+const tableConfig = [
```
teneightfive, line 5: This feels a lot like our [`presenters`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/master/common/presenters) pattern of taking some data from a service or API request and transforming it into how a specific component expects to receive that data. 

Maybe it could be re-worked to continue to use that pattern consistently?

teneightfive: Ah, sorry, hadn't seen you are using presenters. Maybe it could be combined to use them like they're being used to build [other components](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/presenters/person-to-summary-list-component.js)?

common/components/table/template.njk
```diff
@@ -0,0 +1,2 @@
+{% from "govuk/components/table/macro.njk"             import govukTable %}
+{{ govukTable(params) }}
```
teneightfive, line 2: It feels like we could remove this component as its doing the same as the GOV.UK one? Then just call `govukTable()` in your view rather than `appTable()`?

app/moves/controllers/list-by-status.js
barbaracassani: The issue here, as mentioned in the comment PR, is that I am trying to avoid tying the presenters of the rows and table to a particular type of object. Here's moves, but soon those presenters will be used for allocations as well.
If I used two series of presenters there would be a lot of duplication. I could use a mixin-like pattern to have the main presenters methods in common and two different configs, something like
-- allocationToTableRow / allocationToTableHead
-- moveToTableRow / moveToTableHead
if it feels cleaner.

common/components/table/template.njk
barbaracassani: This is the first iteration, I fully expect us to extend (or even fork it) this component soon, hence my creating a place for it.

config/nunjucks/filters.js
```diff
@@ -49,6 +49,9 @@ function formatDateRange(dateRange) {
   if (!Array.isArray(dateRange) || dateRange.length !== 2) {
     return dateRange
   }
+  if (!dateRange[1]) {
```
teneightfive, line 52: Feels like this fix is something separate to adding the table? It could be a separate `fix:` commit rather than being bundled into this "feature".

That way it would be better described in [the changelog](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/CHANGELOG.md) too.

app/allocations/middleware.test.js
barbaracassani: Sorry, yes, this belongs to another commit. Shall amend.

locales/en/moves.json
```diff
@@ -133,6 +133,9 @@
     "no_results": "No moves requested",
     "single_moves": "Single moves",
     "single_moves_no_results": "No moves proposed",
+    "created_at": "Created at",
+    "move_type": "Move type",
+    "date": "Date",
```
teneightfive, line 138: Maybe `date` should live in the default file? Feels like something that has more generic use than being tied to the moves dashboard.

common/lib/api-client/models.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
       move_agreed_by: '',
       cancellation_reason: '',
       cancellation_reason_comment: '',
+      prison_transfer_reason: '',
```
teneightfive, line 11: Didn't know where to put this comment but...it might be worth adding a little piece of text in the commit body as to why this is needed to give some more context to this fix. Maybe what was happening before and what will happen know this has been added.

common/lib/api-client/models.js
```diff
@@ -177,4 +177,19 @@ module.exports = {
       filesize: '',
     },
   },
+  allocation: {
+    attributes: {
+      from_location_id: '',
```
teneightfive, line 182: Are these not a relationship like they are on `move`?

common/lib/api-client/models.js
```diff
@@ -177,4 +177,19 @@ module.exports = {
       filesize: '',
     },
   },
+  allocation: {
```
teneightfive, line 180: It would be good to extend the [models tests](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/lib/api-client/models.test.js) to includes tests for this new model.

common/components/table/template.njk
solidgoldpig: Are we planning to use it anywhere else? Will it incorporate additional pagination elements and/or need to include additional css/js for the added functionality?

Right now though, it seems that the presenter is doing the heavy lifting (and will continue to do so?) and it is currently just a `govukTable`

barbaracassani: It will very likely incorporate other elements; pagination is certainly one.

app/moves/controllers/list-by-status.js
```diff
@@ -1,16 +1,39 @@
...
+  {
+    head: i18n.t('name'),
+    row: data => {
```
solidgoldpig, line 8: See https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/437/files#r412253094 for a suggestion on allowing additional properties to be passed to the individual cells.

app/moves/controllers/list-by-status.js
barbaracassani: However, on second thought, no matter how I change it feels like a complication. I remember you telling me that the presenter should be kept as simple as possible and that any `map` should be invoking it instead of containing it. Therefore, I have kept separated row and heading; if I create a general "moves table presenter" and another "allocation table presenter" that invokes the two presenters with the config, I go against this, not to mention that it becomes clunkier.

What's the main issue with keeping the table config within the controller? It seems acceptable to instruct a generic presenter on what goes where, no?

teneightfive: Yes, I suppose in this instance it's tricky to just call `.map()` on a presenter because it requires all items in the collection.

But I think it feels a little cleaner to have one presenter that transforms a list of moves to a table format. Have a look at what we do with `movesByLocation`. The presenter doesn't necessarily have to transform to a component. 

And with allocations, I'd say cross that bridge when it comes to it. An `allocation` will also have a different structure of a `move` anyway and having config and passing two separate values to build the table could over complicate it.

For this change I'd look at just creating a presenter to take a collection of moves and output in the format the `govukTable` components wants.

That way the controller remains simple too, which is a pattern we've tried to adopt where possible.

common/components/table/template.njk
teneightfive: I'd say we park it for now until that case arises

common/lib/api-client/models.js
barbaracassani: Good point 👍

common/lib/api-client/models.js
```diff
@@ -176,4 +180,25 @@ module.exports = {
...
+      to_location: {
+        jsonApi: 'hasOne',
+        type: 'locations',
+      },
+      date: '',
+      prisoner_category: '',
+      sentence_length: '',
+      complex_cases: [],
+      moves_count: 0,
+      complete_in_full: '', // boolean
```
teneightfive, line 198: This comment feels like a bit of a red flag that maybe we should update the name of the variable? I find a `is_` prefix helps to suggest when something is a Boolean and makes the code much more readable. 

So this could become `is_complete_in_full`. 

Without the “is” this variable could be read as “to” complete in full when used within the app.

smoothcontract: Could this be changed to `complete_in_full: false` instead? I'd prefer not to rename the back end data model attributes at this stage - also I'm not 100% sure that "is complete in full" makes sense - my understanding is that this attribute indicates that the allocation must be completed in full (i.e. if there are 3 prisoners specified in `moves_count` then 3 prisoners must be added to the allocation). It's not meant to be an indication of the move or allocation actually having been completed.

PR #439, opened 2020-04-20
style(linting): Add rule to sort object properties

teneightfive:
### Checklist


solidgoldpig:
Please, please, please... not before #450 is merged in ;)

.eslintrc.js
```diff
@@ -64,8 +69,18 @@ module.exports = {
         'no-unused-expressions': 'off',
         'mocha/no-exclusive-tests': 'error',
         'mocha/no-mocha-arrows': 'error',
+        'sort-keys': 'off',
```
solidgoldpig, line 72: Why is `sort-keys` off for testcafé?

.eslintrc.js
```diff
@@ -64,8 +69,18 @@ module.exports = {
         'no-unused-expressions': 'off',
         'mocha/no-exclusive-tests': 'error',
         'mocha/no-mocha-arrows': 'error',
```
solidgoldpig, line 71: While I'm asking the question about `sort-keys`, hating the `mocha-no-arrows` ;)

solidgoldpig:
My question about `sort-keys` aside, all good 👍

.eslintrc.js
teneightfive: @solidgoldpig There is a reason this is there...[mocha discourages](https://mochajs.org/#arrow-functions) the use of arrow functions ;)

.eslintrc.js
teneightfive: Good question...I'm not sure I fully remember. I think it might have been because there were _a lot_ of warnings to address so I left it off this PR...

.eslintrc.js
solidgoldpig: I understand why - just saying I hates it ;)  Fortunately the auto-fixing sorts it, but I've developed a weird memory muscle now to type `function(){}`

teneightfive: You use VS code too right? I've found the [mocha snippets extension](https://github.com/spoonscen/es6-mocha-snippets-vs-code) a godsend for speeding up writing test blocks :)

You'll need to prefix the shortcuts with `f` to use the non-arrow functions:

`fdes` + <kbd>tab</kbd> = 🥳

solidgoldpig: Giving it a go

solidgoldpig:
```json
{
    "mocha-snippets.function-type": 'arrow' | 'function' | 'both'
}
```

teneightfive:
Closing as feels a little harsh to have this rule in hindsight and doesn't allow for patterns such as ordering shorthand properties within objects first.

PR #443, opened 2020-04-21
[P4-1306] Provide better user feedback when a move conflict occurs

teneightfive:
## Proposed changes

This fix handles `422` errors during move creation where the reason code
is `taken`.

This means that the user is trying to create a duplicate
move (same to and from location, date and person) which the service
doesn't currently allow.

We now provide the user with better feedback that this is what has
happened and a link to the existing move, as well as options to change
the date or location if they still want to create a move for that person.

**Note:** Depends on #442 

### Changing the date

<kbd></kbd>

<kbd></kbd>

### Before

<kbd></kbd>

### After

<kbd></kbd>


jason-sims:
approved from me

PR #450, opened 2020-04-22
feature - Add ability to update moves

solidgoldpig:
Implements P4 1195 - Ability to edit Police PTR

NB. this feature is currently hidden behind the feature flag `FEATURE_FLAG_EDITABILITY`.

Adds the following pages and links to them from the move view

- personal details
- risk information
- health information
- court information

These pages reuse the create move templates, with the only notable differences being

- PNC number is read-only if it has already been set
- uses save and continue rather than continue button

Cancelling or saving returns the user to the move view page.

Move_details_for_GORCZANY__MAZIE

Personal_details_-_update_move

The pages for changing the move details and move date also exist but the links to them have been suppressed until it's confirmed that `from_location` can be changed.

### Checklist

### DIFF
--- a/app/move/controllers/create/base.test.js
+++ b/app/move/controllers/create/base.test.js
@@ -43,30 +43,36 @@ describe('Move controllers', function() {
 
       it('should call set button text method', function() {
         expect(controller.use.getCall(0)).to.have.been.calledWithExactly(
+          controller.setModels
+        )
+      })
+
+      it('should call set button text method', function() {
+        expect(controller.use.getCall(1)).to.have.been.calledWithExactly(
           controller.setButtonText
         )
       })
 
       it('should call set cancel url method', function() {
-        expect(controller.use.getCall(1)).to.have.been.calledWithExactly(
+        expect(controller.use.getCall(2)).to.have.been.calledWithExactly(
           controller.setCancelUrl
         )
       })
 
       it('should call set move summary method', function() {
-        expect(controller.use.getCall(2)).to.have.been.calledWithExactly(
+        expect(controller.use.getCall(3)).to.have.been.calledWithExactly(
           controller.setMoveSummary
         )
       })
 
       it('should call set journey timer method', function() {
-        expect(controller.use.getCall(3)).to.have.been.calledWithExactly(
+        expect(controller.use.getCall(4)).to.have.been.calledWithExactly(
           controller.setJourneyTimer
         )
       })
 
       it('should call correct number of middleware', function() {
-        expect(controller.use).to.be.callCount(4)
+        expect(controller.use).to.be.callCount(5)
       })
     })
 
@@ -224,6 +230,151 @@ describe('Move controllers', function() {
       })
     })
 
+    describe('#setModels()', function() {
+      let res
+      let req
+      let nextSpy
+      beforeEach(function() {
+        res = {
+          locals: {},
+        }
+        req = {}
+        sinon.stub(controller, '_setModels')
+        sinon.stub(controller, '_addModelMethods')
+        nextSpy = sinon.spy()
+        controller.setModels(req, res, nextSpy)
+      })
+
+      it('should set the models', function() {
+        expect(req.models).to.deep.equal({})
+        expect(controller._setModels).to.be.calledOnceWithExactly(req)
+      })
+
+      it('should add model methods', function() {
+        expect(controller._addModelMethods).to.be.calledOnceWithExactly(req)
+      })
+
+      it('invoke next', function() {
+        expect(nextSpy).to.be.calledOnceWithExactly()
+      })
+    })
+
+    describe('#_setModels()', function() {
+      let req
+      const mockSession = {
+        id: '#move',
+        person: {
+          id: '#person',
+        },
+      }
+      beforeEach(function() {
+        req = {
+          models: {},
+          sessionModel: {
+            toJSON: sinon.stub().returns(mockSession),
+            get: sinon
+              .stub()
+              .withArgs('person')
+              .returns(mockSession.person),
+          },
+        }
+        controller._setModels(req)
+      })
+
+      it('should add move model to req', function() {
+        expect(req.models.move).to.deep.equal(mockSession)
+      })
+
+      it('should add person model to req', function() {
+        expect(req.models.person).to.deep.equal(mockSession.person)
+      })
+    })
+
+    describe('#_addModelMethods()', function() {
+      let req
+      beforeEach(function() {
+        req = {}
+        controller._addModelMethods(req)
+      })
+
+      it('should add getMove method to req', function() {
+        expect(typeof req.getMove).to.equal('function')
+      })
+
+      it('should add getMoveId method to req', function() {
+        expect(typeof req.getMoveId).to.equal('function')
+      })
+
+      it('should add getPerson method to req', function() {
+        expect(typeof req.getPerson).to.equal('function')
+      })
+
+      it('should add getPersonId method to req', function() {
+        expect(typeof req.getPersonId).to.equal('function')
+      })
+    })
+
+    describe('#request methods', function() {
+      let req
+
+      context('when models do not exist', function() {
+        beforeEach(function() {
+          req = {
+            models: {},
+          }
+          controller._addModelMethods(req)
+        })
+
+        it('req.getMove should return empty object', function() {
+          expect(req.getMove()).to.deep.equal({})
+        })
+
+        it('req.getMoveId should return', function() {
+          expect(req.getMoveId()).to.be.undefined
+        })
+
+        it('req.getPerson should return empty object', function() {
+          expect(req.getPerson()).to.deep.equal({})
+        })
+
+        it('req.getPersonId should return', function() {
+          expect(req.getPersonId()).to.be.undefined
+        })
+      })
+
+      context('when models do exist', function() {
+        beforeEach(function() {
+          req = {
+            models: {
+              move: {
+                id: '#moveId',
+              },
+              person: {
+                id: '#personId',
+              },
+            },
+          }
+          controller._addModelMethods(req)
+        })
+
+        it('req.getMove should move object', function() {
+          expect(req.getMove()).to.deep.equal(req.models.move)
+        })
+
+        it('req.getMoveId should return move id', function() {
+          expect(req.getMoveId()).to.equal('#moveId')
+        })
+
+        it('req.getPerson should return person object', function() {
+          expect(req.getPerson()).to.deep.equal(req.models.person)
+        })
+
+        it('req.getPersonId should return person id', function() {
+          expect(req.getPersonId()).to.equal('#personId')
+        })
+      })
+    })
+
     describe('#setMoveSummary()', function() {
       let req, res, nextSpy
       const mockSessionModel = {
@@ -244,13 +395,12 @@ describe('Move controllers', function() {
         sinon.stub(presenters, 'moveToMetaListComponent').returnsArg(0)
         nextSpy = sinon.spy()
         req = {
+          getMove: sinon.stub().returns(mockSessionModel),
+          getPerson: sinon.stub().returns(mockSessionModel.person),
           session: {
             currentLocation: {
               title: 'Mock current location',
             },
-          },
-          sessionModel: {
-            toJSON: sinon.stub().returns(mockSessionModel),
           },
         }
         res = {
@@ -262,7 +412,7 @@ describe('Move controllers', function() {
 
       context('with move_type', function() {
         beforeEach(function() {
-          req.sessionModel.toJSON.returns({
+          req.getMove.returns({
             ...mockSessionModel,
             move_type: 'court_appearance',
           })
--- a/app/move/controllers/update/assessment.js
+++ b/app/move/controllers/update/assessment.js
@@ -1,0 +1,71 @@
+const { pick } = require('lodash')
+
+const personService = require('../../../../common/services/person')
+const Assessment = require('../create/assessment')
+
+const UpdateBase = require('./base')
+
+const compare = (a, b) => {
+  const key = 'assessment_question_id'
+  if (a[key] < b[key]) {
+    return -1
+  }
+  if (a[key] > b[key]) {
+    return 1
+  }
+}
+const getAnswerKeys = answers => {
+  return answers
+    .map(x => pick(x, ['key', 'assessment_question_id', 'comments']))
+    .sort(compare)
+}
+
+class UpdateAssessmentController extends UpdateBase {
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setPreviousAssessment)
+  }
+
+  async saveValues(req, res, next) {
+    try {
+      const personId = req.getPersonId()
+      const person = req.getPerson()
+      const assessments = person.assessment_answers || []
+      const fieldKeys = Object.keys(req.form.options.fields)
+
+      const previousAssessments = assessments.filter(
+        ({ key }) => !fieldKeys.includes(key)
+      )
+
+      const newAssessments = this.getAssessments(req, res)
+      const updatedAssessments = [
+        ...previousAssessments,
+        ...newAssessments,
+      ].sort(compare)
+      // TODO: keep all existing NOMIS alerts
+      // answer.nomis_alert_code
+      // TODO: filter out requested answers
+      // 'not_to_be_released', 'special_vehicle'))
+      // TODO: don't set assessments if not from prison?
+      // fromLocationType !== 'prison'
+
+      const originalKeys = getAnswerKeys(assessments)
+      const updatedKeys = getAnswerKeys(updatedAssessments)
+
+      if (JSON.stringify(originalKeys) !== JSON.stringify(updatedKeys)) {
+        await personService.update({
+          id: personId,
+          assessment_answers: updatedAssessments,
+        })
+      }
+
+      next()
+    } catch (error) {
+      next(error)
+    }
+  }
+}
+
+UpdateBase.mixin(UpdateAssessmentController, Assessment)
+
+module.exports = UpdateAssessmentController
--- a/app/move/controllers/update/base.js
+++ b/app/move/controllers/update/base.js
@@ -1,0 +1,108 @@
+const { get, keys } = require('lodash')
+
+const personService = require('../../../../common/services/person')
+const CreateBaseController = require('../create/base')
+
+class UpdateBaseController extends CreateBaseController {
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setNextStep)
+    this.use(this.setInitialStep)
+  }
+
+  _setModels(req) {
+    const res = req.res
+    req.models.move = res.locals.move
+    req.models.person = req.models.move.person
+  }
+
+  getBaseUrl(req) {
+    const moveId = req.getMoveId()
+    return `/move/${moveId}`
+  }
+
+  setCancelUrl(req, res, next) {
+    res.locals.cancelUrl = this.getBaseUrl(req)
+    next()
+  }
+
+  setNextStep(req, res, next) {
+    const nextUrl = this.getBaseUrl(req)
+    req.form.options.next = nextUrl
+    next()
+  }
+
+  setInitialStep(req, res, next) {
+    let initialStep = false
+    const referrer = req.get('referrer')
+    if (!referrer) {
+      initialStep = true
+    } else {
+      const { pathname } = new URL(referrer)
+      if (pathname === this.getBaseUrl(req)) {
+        initialStep = true
+      }
+    }
+    req.initialStep = initialStep
+    next()
+  }
+
+  protectReadOnlyFields(req, values) {
+    const fields = req.form.options.fields
+    Object.keys(fields).forEach(key => {
+      const field = fields[key]
+      if (field.readOnly && values[key] !== undefined && values[key] !== null) {
+        fields[key] = {
+          ...field,
+          ...field.updateComponent,
+          value: values[key],
+        }
+      }
+    })
+  }
+
+  getErrors(req, res) {
+    if (req.initialStep) {
+      return {}
+    }
+    return super.getErrors(req, res)
+  }
+
+  getValues(req, res, callback) {
+    return super.getValues(req, res, (err, values) => {
+      if (err) {
+        return callback(err)
+      }
+
+      try {
+        if (req.initialStep) {
+          values = this.getUpdateValues(req)
+        }
+        this.protectReadOnlyFields(req, values)
+      } catch (error) {
+        return callback(error)
+      }
+
+      callback(null, values)
+    })
+  }
+
+  getUpdateValues(req, res) {
+    const person = req.getPerson()
+    const fields = keys(get(req, 'form.options.fields'))
+    return personService.unformat(person, fields)
+  }
+}
+
+UpdateBaseController.mixin = function mixin(controller, mixinController) {
+  const controllerMethods = Object.getOwnPropertyNames(controller.prototype)
+  Object.getOwnPropertyNames(mixinController.prototype)
+    .filter(key => key !== 'prototype')
+    .forEach(key => {
+      if (!controllerMethods.includes(key)) {
+        controller.prototype[key] = mixinController.prototype[key]
+      }
+    })
+}
+
+module.exports = UpdateBaseController
--- a/app/move/controllers/update/base.test.js
+++ b/app/move/controllers/update/base.test.js
@@ -1,0 +1,570 @@
+const FormController = require('hmpo-form-wizard').Controller
+const { cloneDeep } = require('lodash')
+
+const personService = require('../../../../common/services/person')
+const CreateBaseController = require('../create/base')
+const BaseProto = CreateBaseController.prototype
+
+const UpdateBaseController = require('./base')
+
+const controller = new UpdateBaseController({ route: '/' })
+
+describe('Move controllers', function() {
+  describe('Update base controller', function() {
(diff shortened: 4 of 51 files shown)


app/move/controllers/create/base.test.js
```diff
@@ -224,6 +230,156 @@ describe('Move controllers', function() {
...
+    describe('#setModels()', function() {
+      let res
+      let req
+      let nextSpy
+      beforeEach(function() {
+        res = {
+          locals: {},
+        }
+        req = {
+          // res,
```
teneightfive, line 242: Rogue comments?

common/services/person.js
```diff
@@ -2,6 +2,40 @@ const { mapKeys, mapValues, uniqBy, omitBy, isNil } = require('lodash')
...
+  // health
+  'special_diet_or_allergy',
+  'health_issue',
+  'medication',
+  'wheelchair',
+  'pregnant',
+  'other_health',
+  'special_vehicle',
+]
+const explicitAssessmentKeys = ['special_vehicle']
```
teneightfive, line 37: I think there is another that is explicit - `not_to_be_released`

common/services/person.test.js
```diff
@@ -1,8 +1,18 @@
 const { forEach } = require('lodash')
+const proxyquire = require('proxyquire')
 
+const filters = require('../../config/nunjucks/filters')
 const apiClient = require('../lib/api-client')()
 
-const personService = require('./person')
+const formatDateStub = sinon.stub(filters, 'formatDate').returns('28 Oct 2010')
+
+const unformat = proxyquire('./person-unformat', {
```
teneightfive, line 9: I wonder if its cleaner to test `person.unformat.js` separately in `person.unformat.test.js` then you don't need to worry about nested proxyquires. 

**BUT** not as part of this PR/change. One to think about if you revisit this during STC/SCH/Prison edit changes.

One option could be to use an index loader pattern that is used elsewhere for the person methods. 

So `person.js` becomes:
```
./services
  ...
  /person
    index.js
    format.js
    format.test.js
    unformat.test.js
    unformat.test.js
    create.js
    create.test.js
    ...
  ...
```

app/move/controllers/update/assessment.js
```diff
@@ -0,0 +1,71 @@
...
+    super.middlewareLocals()
+    this.use(this.setPreviousAssessment)
+  }
+
+  async saveValues(req, res, next) {
+    try {
+      const personId = req.getPersonId()
+      const person = req.getPerson()
+      const assessments = person.assessment_answers || []
+      const fieldKeys = keys(get(req, 'form.options.fields'))
```
teneightfive, line 34: Could be `const fieldsKeys = Object.keys(req.form.options.fields)` and do away with needing the lodash methods.

And to be honest, if `req.form.options.fields` isn't there we kind of don't mind this all blowing up as there's likely a bigger issue.

app/move/controllers/update/base.js
```diff
@@ -0,0 +1,109 @@
...
+    next()
+  }
+
+  setNextStep(req, res, next) {
+    const nextUrl = this.getBaseUrl(req)
+    req.form.options.next = nextUrl
+    next()
+  }
+
+  setInitialStep(req, res, next) {
```
teneightfive, line 35: What's the purpose of this? Is it a form-wizard thing or a thing we've added to support edit?

app/move/controllers/update/base.js
```diff
@@ -0,0 +1,109 @@
...
+    return super.getValues(req, res, (err, values) => {
+      if (err) {
+        return callback(err)
+      }
+
+      try {
+        if (req.initialStep) {
+          values = this.getUpdateValues(req)
+        }
+        // values = Object.assign({}, this.getUpdateValues(req), values)
```
teneightfive, line 81: Rogue comment?

app/move/fields/update.js
```diff
@@ -0,0 +1,24 @@
...
+updateFields.police_national_computer = {
+  ...updateFields.police_national_computer,
+  readOnly: true,
+  updateComponent: {
+    component: 'appReadOnlyField',
+    classes: '',
+    items: [
+      {
+        component: 'govukDetails',
+        summaryHtml: 'I need to change the PNC number',
```
teneightfive, line 16: Should these be in the locales files?

app/move/index.js
```diff
@@ -24,6 +33,13 @@ const createConfig = {
   journeyName: 'create-a-move',
   journeyPageTitle: 'actions::create_move',
 }
+const updateConfig = {
+  ...wizardConfig,
+  controller: update.Base,
+  templatePath: 'move/views/create/',
+  template: '../../../form-wizard',
+  journeyPageTitle: 'actions::update_move',
```
teneightfive, line 41: Maybe also add a different `journeyName`. That way it will prevent any clashes on the session namespace.

teneightfive: Ah, it's down down below. Ignore this comment then.

locales/en/actions.json
```diff
@@ -22,5 +22,6 @@
   "print_move_list": "Print moves",
   "print_move_detail": "Print move",
   "sign_out": "Sign out",
-  "switch_location": "Switch location"
+  "switch_location": "Switch location",
+  "save": "Save and continue"
```
teneightfive, line 26: Should this be `save_and_continue`? It's more explicit to the actual content that way.

app/move/views/_includes/assessment.njk
```diff
@@ -1,112 +1,118 @@
...
-
-          {{ answer.comments if answer.comments }}
-
-          {% if answer.created_at %}
-            <div class="govuk-!-margin-top-2 govuk-!-font-size-16">
-              {{ answer.created_at | formatDateWithDay }}
-            </div>
-          {% endif %}
-        {% endset %}
+  <section class="app-!-position-relative" data-app-section="{{ assessmentCategory.key }}">
```
teneightfive, line 2: Is this data attribute used for anything?

app/move/index.js
```diff
@@ -40,13 +56,30 @@ router.use(
...
 router.get('/:moveId', protectRoute('move:view'), view)
+
 router.get('/:moveId/confirmation', protectRoute('move:create'), confirmation)
 router.use(
   '/:moveId/cancel',
   protectRoute('move:cancel'),
   wizard(cancelSteps, cancelFields, cancelConfig)
 )
 
+if (FEATURE_FLAGS.EDITABILITY) {
```
teneightfive, line 67: If its not too much trouble it might be worth seeing if you can move the addition of this feature flag to the other commit. That way when we remove the flag we can reverse that commit and remove any code needed for the feature flag.

test/e2e/_helpers.js
```diff
@@ -5,21 +5,46 @@ import { join } from 'path'
...
+import referenceDataService from '../../common/services/reference-data'
+import { formatDate } from '../../config/nunjucks/filters'
+
+let referenceData
+let genders
+let ethnicities
+let locations
+
+const getReferenceData = async () => {
+  genders = await referenceDataService.getGenders()
```
teneightfive, line 22: These could be done using `Promise.all` to speed things up a little if needed. They don't really need to be run synchronously in this order.

app/move/views/_includes/assessment.njk
solidgoldpig: Yep - it's how I'm targeting the links in the e2e tests

locales/en/actions.json
solidgoldpig: Discussed with @Andy-Millar. 'Save and continue' is the default govuk design system text for a save button.

I would be more inclined to name it `update` and alias it to `save`. But I thought that was overdoing it for now.

app/move/index.js
solidgoldpig: As noted, covered below

app/move/fields/update.js
solidgoldpig: They totally should.

test/e2e/_helpers.js
```diff
@@ -5,21 +5,46 @@ import { join } from 'path'
...
+  genders = genders.filter(gender => gender.title.match(/^(Male|Female)$/i))
+  ethnicities = await referenceDataService.getEthnicities()
+  locations = await referenceDataService.getLocations()
+  referenceData = true
```
teneightfive, line 26: This feels a little flakey. The move fixture uses `locations`, but it relies on the the fact that a person has been created at some point in time for set this value to `true`.

I think it would be safer to have `getReferenceData` check if there is a cached version locally rather than doing that in the fixture. If it exists, return it, if not, make a call to the relevant service and save that value.

There will be a need to create moves without a "person" for the allocations work so it would be good to know that this won't cause us issues when we start testing that.

Something like this:

```javascript
const cachedReferenceData = {}

const getReferenceData = async (key) => {
  if (Object.keys(cachedReferenceData).length) {
    return cachedReferenceData[key]
  }

  cachedReferenceData = {
    genders: await referenceDataService.getGenders().then(genders => genders.filter(gender => gender.title.match(/^(Male|Female)$/i))),
    ethnicities: await referenceDataService.getEthnicities(),
    locations: await referenceDataService.getLocations(),
  }
}


// to get some reference data
getReferenceData('locations')
```

common/services/person.js
solidgoldpig: Added - I think it might be better to add these as properties of the field themselves. It will also make it easier to test. But I'll leave that for a refactor another day...

common/services/person.test.js
solidgoldpig: Sure. I didn't want to add yet more noise at the moment. I really only split it out in the first place to placate the CodeClimate gods

app/move/controllers/create/base.test.js
solidgoldpig: Zapped

test/e2e/pages/move-detail.js
```diff
@@ -79,7 +82,7 @@ class MoveDetailPage extends Page {
             dateOfBirth
           )})`
         : undefined,
-      Gender: gender,
+      // Gender: gender,
```
teneightfive, line 85: Rogue comment?

test/e2e/pages/update-move.js
```diff
@@ -0,0 +1,53 @@
...
+class UpdateMovePage extends CreateMovePage {
+  constructor(id) {
+    super()
+    this.url = `/move/${id}/edit`
+    this.fields = {
+      ...this.fields,
+      policeNationalComputerReadOnly: Selector(
+        '[type=hidden][name=police_national_computer]'
+      ),
+      policeNationalComputerHeading: Selector('.app-read-only-field__heading'),
```
teneightfive, line 17: Maybe these should live under a different property? They're not technically a "field"

app/move/controllers/update/assessment.js
solidgoldpig: Done

app/move/controllers/update/base.js
solidgoldpig: Zapped

test/e2e/pages/update-move.js
```diff
@@ -0,0 +1,53 @@
...
+
+  // /**
+  //  * Fill in personal details
+  //  *
+  //  * @param {Object} options - fields to include or exclude
+  //  * @param {Array} options.include - fields to include
+  //  * @param {Array} options.exclude - fields to exclude
+  //  * @returns {Promise<FormDetails>} - filled in personal details
+  //  */
+  // async fillInPersonalDetails(options = {}) {
```
teneightfive, line 30: Rogue comment?

test/e2e/_helpers.js
solidgoldpig: true dat.

test/e2e/_helpers.js
```diff
@@ -75,6 +112,69 @@ export async function createPersonFixture() {
...
 }
 
+/**
+ * Get a random location
+ *
+ * @param {string} [locationType=police] - type of location
+ *
+ * @returns {string} - location id
+ */
+const getRandomLocation = (locationType = 'police') => {
```
teneightfive, line 122: Maybe we shouldn't have a default? Feels like a safer solution would be to use `all` as the default.

Also, if a `type` is given, it might be better to use `referenceData.getLocationsByType()`. Will be faster to do the filter on the API.

test/e2e/_helpers.js
solidgoldpig: Updated to call the ref data service directly and rely on the api-client's cacheing

test/e2e/_helpers.js
solidgoldpig: Default removed and calls `getLocationsByType` if `locationType` provided, 'getLocations` if not

test/e2e/pages/update-move.js
solidgoldpig: Zapped

test/e2e/pages/update-move.js
solidgoldpig: Do you have a suggestion?

test/e2e/_helpers.js
solidgoldpig: Except we've now moved to calling the ref data service directly so this has gone away

test/e2e/pages/move-detail.js
solidgoldpig: It was failing. But now it's not so reinstated.

app/move/index.js
solidgoldpig: I presume you mean this commit (https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/450/commits/57065c6c64976a2cbafd5a000c44d2f90a59473c). Could you explain what you mean by reverse the commit? Rebase so it was never there or...?

app/move/fields/update.js
solidgoldpig: Moved. And updated `common/helpers/field.js` to handle the new property

app/move/controllers/update/base.js
solidgoldpig: Scenario:
- User goes to move A details
- User edit personal details and attempt to delete the name
- User sees error saying name can't be empty
- User goes move B details (by way of dashboard)
- User sees error again

As the user didn't exit through the gift shop, we haven't been able to delete the session.

If we set `reset` and/or `resetJourney` on these steps then the session data is dropped if there is an error resulting in the user only seeing empty values (or the original data if that's been fetched). Nor can we rely on the `checkJourney` prop because these steps are their own entry points.

Maybe the `journeyName` could be updated so as to be different for each move, but that seems a bit iffy (and I'm not even sure it would work). Even if that was so, it would mean directly going to the page (ie. where moves A and B are the same) would still trigger the error which I think is the wrong UX

Additionally, `req.initialStep` enables not reloading the move's data so it's not without utility.

test/e2e/_helpers.js
```diff
@@ -119,12 +114,14 @@ export async function createPersonFixture(overrides = {}) {
  *
  * @returns {string} - location id
  */
-const getRandomLocation = (locationType = 'police') => {
-  return faker.random.arrayElement(
-    locations
-      .filter(loc => loc.location_type === locationType)
-      .map(({ id }) => id)
-  )
+const getRandomLocation = async (locationType = 'police') => {
```
teneightfive, line 117: I think we should still drop the default. With this the `else` currently won't ever fire.

locales/en/fields.json
```diff
@@ -15,7 +15,11 @@
     "hint": "Select the person you are moving"
   },
   "police_national_computer": {
-    "label": "<abbr title=\"Police National Computer\">PNC</abbr> number"
+    "label": "<abbr title=\"Police National Computer\">PNC</abbr> number",
+    "details": {
+      "summaryHtml": "I need to change the PNC number",
+      "html": "If you have created the movement for the wrong person cancel this request and create a new one"
```
teneightfive, line 21: `movement` -> `move`? We've never referred to it as a "movement" so feels inconsistent. Also, I hate to say it but "movement" reminds of something else which isn't pleasant...

locales/en/actions.json
teneightfive: Oh, sorry I was talking about the key not the content. Totally agree with `Save and continue` as the text. But the key is `save` which would suggest to me the content would be "Save", not "Save and continue". Just means if we ever need to introduce a "Save" action it won't clash.

app/move/index.js
teneightfive: Yep, that commit. And sorry, I meant [`revert`](https://git-scm.com/docs/git-revert), not reverse. Ie, when we're done with the feature flag because the feature is live everywhere, it makes it easy to remove the associated code.

test/e2e/_helpers.js
teneightfive: Yep, much nicer and neater 👍

test/e2e/pages/update-move.js
teneightfive: So there was a pattern of `this.nodes` that had been used elsewhere. Maybe use that for now. But they definitely don't fit into `fields`...

test/e2e/_helpers.js
teneightfive: Looks like the default snuck back in

app/move/controllers/update/personal-details.js
```diff
@@ -0,0 +1,41 @@
...
+    const identifiers = person.identifiers.map(
+      identifier => identifier.identifier_type
+    )
+    const identifiersData = personService.unformat(person, identifiers)
+    req.form.values = Object.assign({}, identifiersData, req.form.values)
+
+    PersonalDetails.prototype.saveValues.apply(this, [req, res, next])
+  }
+
+  // async saveValues(req, res, next) {
```
teneightfive, line 19: Just noticed all these comments. Should they be commented out?

app/move/controllers/update/base.test.js
```diff
@@ -0,0 +1,573 @@
+const FormController = require('hmpo-form-wizard').Controller
+const { cloneDeep } = require('lodash')
+const proxyquire = require('proxyquire')
+
+const personService = require('../../../../common/services/person')
+const CreateBaseController = require('../create/base')
+const BaseProto = CreateBaseController.prototype
+
+const UpdateBaseController = proxyquire('./base', {
```
teneightfive, line 9: You can probably do away with `proxyquire` here. I try to avoid using it unless absolutely necessary.

It doesn't look like this is changing any behaviour as you're loading in the same library?

app/move/controllers/update/move-details.test.js
```diff
@@ -0,0 +1,168 @@
...
+    })
+
+    describe('#getUpdateValues', function() {
+      let req = {}
+      const res = {}
+      beforeEach(function() {
+        req = {
+          getMove: sinon.stub(),
+        }
+        // sinon.stub(controller, 'getMove')
```
teneightfive, line 70: Rogue comment?

app/move/controllers/view.js
```diff
@@ -14,6 +15,21 @@ module.exports = function view(req, res) {
...
+
+  const urls = {
+    update: {},
+  }
+  updateSteps.forEach(step => {
+    const url = Object.keys(step)[0].replace(/^\//, '')
+    const key = url.replace(/-/g, '_')
+    urls.update[key] = `/move/${move.id}/edit/${url}`
+    // some assessment categories aren't suffixed with _information
+    if (key.endsWith('_information')) {
```
teneightfive, line 27: This will likely work for now, but thinking about how the prison journeys work where some make use of the assessment controller on a route that doesn't end with `_information` may start to break this pattern. Something to keep in mind for when you tackle those though, not now.

test/e2e/pages/create-move.js
```diff
@@ -152,11 +155,13 @@ class CreateMovePage extends Page {
    * @param {Object} personalDetails - personal details to fill form in with
    * @returns {Promise<FormDetails>} - filled in personal details
    */
-  async fillInPersonalDetails(personalDetails) {
-    await t.expect(this.getCurrentUrl()).contains('/move/new/personal-details')
+  async fillInPersonalDetails(personalDetails, { include, exclude } = {}) {
```
teneightfive, line 158: Maybe document these new arguments and what they allow you to do

test/e2e/pages/index.js
```diff
@@ -16,4 +17,5 @@ export {
   movesDashboardPage,
   createMovePage,
   cancelMovePage,
+  UpdateMovePage,
```
teneightfive, line 20: I wonder if we should use the same pattern that already exists and instantiate the class above and export an instantiated object here.

test/e2e/pages/move-detail.js
```diff
@@ -189,6 +192,20 @@ class MoveDetailPage extends Page {
...
+  async checkUpdateLink(category, exists = true) {
+    const selector = this.nodes.getUpdateLink(category)
+    await t.expect(selector.exists).eql(exists)
+  }
+
+  async checkNoUpdateLink(category) {
+    await this.checkUpdateLink(category, false)
+  }
+
+  async clickUpdateLink(category, exists = true) {
```
teneightfive, line 205: I wonder if these are actually an abstraction too many. This function doesn't really share anything that isn't already shared. The selector is abstracted to one place using the page nodes and `t.click` feels already clear enough in a test on its own.

test/e2e/pages/move-detail.js
```diff
@@ -189,6 +192,20 @@ class MoveDetailPage extends Page {
       }
     }
   }
+
+  async checkUpdateLink(category, exists = true) {
+    const selector = this.nodes.getUpdateLink(category)
+    await t.expect(selector.exists).eql(exists)
+  }
+
+  async checkNoUpdateLink(category) {
```
teneightfive, line 201: Same here, this doesn't really add any real abstraction. In a test where the link should not appear this feels clear enough:

```javascript
await t.expect(moveDetailPage.nodes.getUpdateLink('risk').exists).not.ok()
```

or 

```javascript
await t.expect(moveDetailPage.nodes.getUpdateLink('risk').exists).ok()
```

test/e2e/_move.js
```diff
@@ -0,0 +1,383 @@
...
+    options
+  )
+}
+
+/**
+ * Check that PNC details are uneditable and displayed corectly on the personal details page
+ *
+ * @returns {undefined}
+ */
+export async function checkPoliceNationalComputerReadOnly() {
```
teneightfive, line 367: I feel some of these should maybe live within the Page Objects? They're doing the same thing the methods there are.

locales/en/actions.json
solidgoldpig: I was talking about the key too ;)

test/e2e/_helpers.js
solidgoldpig: More likely I hallucinated taking it out

locales/en/fields.json
solidgoldpig: That's the copy that @Andy-Millar provided. ¯\_(ツ)_/¯

app/move/controllers/update/personal-details.js
solidgoldpig: Surprised it took you this long to find it - it was my canary in the mine ;)

app/move/controllers/update/move-details.test.js
solidgoldpig: Zapped

app/move/controllers/view.js
solidgoldpig: I think we're better off normalising the keys that we use to refer to these things by.

ie. `risk` or `risk_information` but not both

It is happenstance that it's possible to do this at the moment. And as I've been bleating, a centralised way to get urls pretty please ;)

test/e2e/pages/index.js
solidgoldpig: But it needs to be different for each move id. If the tests always run sequentially, I suppose you could just set the `this.url` value

app/move/index.js
solidgoldpig: OK. But I'll do that last when all the other changes have been made and squashed.

test/e2e/pages/update-move.js
solidgoldpig: Move non-fields to `this.nodes`

test/e2e/_helpers.js
solidgoldpig: I think I hallucinated removing the default

app/move/controllers/update/base.test.js
solidgoldpig: Begone proxyquire

test/e2e/pages/create-move.js
solidgoldpig: Yep. Added.

app/move/index.js
solidgoldpig: Ack - the e2e tests have the flag in too. I'll reverse the order of those commits when the time comes...

locales/en/fields.json
teneightfive: True, but maybe just mention to him and see what he says. I'm pretty sure we had conversations in the early days to make we used `move` instead of `movement` and to try and be consistent.

locales/en/actions.json
teneightfive: Leave if it you're happy. I just felt `save` suggested the text "Save" not "Save and continue".

test/e2e/pages/index.js
teneightfive: It's probably best not to assume they'll run sequentially in that case

locales/en/actions.json
solidgoldpig: Renamed ;)

locales/en/fields.json
solidgoldpig: `move` it is

test/e2e/pages/move-detail.js
solidgoldpig: The conversation will continue...

PR #453, opened 2020-04-23
[P4 1280] List view for allocations

barbaracassani:
This is a partial version of the dashboard for allocations. It doesn't contain yet the progress status, nor the sorting. There's no BE support for either yet, nor there is proper filtering: every allocation is always returned.

One notable awkwardness is that some of our filters are boolean, and when they are in the url, `req.query` returns them as strings. This is due by express depending on qs, where `parseBool` is apparently an hotly debated issue and has never been included.

This has led to reverting the elimination of query-string, and using said module to parse the original query string instead of using `req.query`. It's not that great and we might just decide we don't want boolean filters, although this can be seen as a serious limitation.

Screenshot 2020-04-27 at 11 16 26

### Checklist

### DIFF
--- a/app/allocations/controllers/list.js
+++ b/app/allocations/controllers/list.js
@@ -1,0 +1,12 @@
+const presenters = require('../../../common/presenters')
+
+module.exports = function list(req, res) {
+  const { allocations = [] } = res.locals
+  const template = 'allocations/views/list'
+  const locals = {
+    pageTitle: req.t('allocations::dashboard.heading'),
+    ...presenters.allocationsToTable(allocations),
+  }
+
+  res.render(template, locals)
+}
--- a/app/allocations/middleware.js
+++ b/app/allocations/middleware.js
@@ -1,44 +1,149 @@
 const { format } = require('date-fns')
+const { cloneDeep, isEqual, mapValues } = require('lodash')
+const queryString = require('query-string')
 
 const {
   dateFormat,
   getDateFromParams,
   getPeriod,
 } = require('../../common/helpers/date-utils')
+const presenters = require('../../common/presenters')
 const allocationService = require('../../common/services/allocation')
 
-const allocationsMiddleware = {
-  // TODO: refactor this to be common with the /moves pagination.
-  //  It differs only in the output of the url
-  setPagination: (req, res, next) => {
-    const { period } = req.params
-    const today = format(new Date(), dateFormat)
-    const baseDate = getDateFromParams(req)
-    const interval = period === 'week' ? 7 : 1
-
-    const previousPeriod = getPeriod(baseDate, -interval)
-    const nextPeriod = getPeriod(baseDate, interval)
-
-    res.locals.pagination = {
-      todayUrl: `${req.baseUrl}/${period}/${today}/`,
-      nextUrl: `${req.baseUrl}/${period}/${nextPeriod}/`,
-      prevUrl: `${req.baseUrl}/${period}/${previousPeriod}/`,
-    }
-    next()
+const allocationTypeNavigationConfig = [
+  {
+    label: 'allocations::complete',
+    filter: { complete_in_full: true },
   },
-  setAllocationsSummary: async (req, res, next) => {
-    const { dateRange } = res.locals
-    const value = await allocationService.getCount({ dateRange })
-    res.locals.allocationsSummary = [
-      {
-        active: false,
-        label: req.t('allocations::dashboard.labels.single_requests'),
-        href: `/allocations/week/${getDateFromParams(req)}/`,
-        value,
-      },
-    ]
-    next()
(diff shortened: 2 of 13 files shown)


app/allocations/middleware.js
```diff
@@ -1,44 +1,149 @@
...
+  },
+  {
+    label: 'allocations::incomplete',
+    filter: { complete_in_full: false },
+  },
+]
 
-    const previousPeriod = getPeriod(baseDate, -interval)
-    const nextPeriod = getPeriod(baseDate, interval)
+// TODO: refactor this to be common with the /moves pagination.
```
solidgoldpig, line 24: Is there any reason not to hoist `setPagination` out of /moves and share it?

barbaracassani: They differs in the way they build the urls for prev/next/today, because the urls are different. I toyed with a sort of factory pattern that passed to the factory the schema of the url, but in fact it didn't save any code, and it made it more complicated. 
It remains certainly a todo, but I'd rather not tackle it in this PR.

app/allocations/controllers/list.js
```diff
@@ -0,0 +1,12 @@
+const presenters = require('../../../common/presenters')
+
+module.exports = function list(req, res) {
+  const { allocations = [] } = res.locals
+  const template = 'allocations/views/list'
+  const locals = {
+    pageTitle: req.t('allocations::dashboard.heading'),
+    ...presenters.allocationsToTable(allocations),
```
teneightfive, line 8: It might be worth returning this as a specific variable to the view rather than just `head` and `rows`. I've seen some evolution of some of Andy's designs that display multiple tables on a page.

Feels easier to read the template too if the variable clearly explains the data being loaded into the table.

app/allocations/middleware.test.js
```diff
@@ -82,21 +83,21 @@ describe('#setPagination', function() {
     it('creats correctly todayUrl', function() {
```
teneightfive, line 83: Old typo?

app/allocations/middleware.test.js
```diff
@@ -82,21 +83,21 @@ describe('#setPagination', function() {
...
       expect(res.locals.pagination.todayUrl).to.exist
       expect(res.locals.pagination.todayUrl).to.equal(
-        '/allocations/week/2020-04-13/'
+        '/allocations/week/2020-04-13/outgoing/'
```
teneightfive, line 86: Should our URLs have a trailing slash? I think they all actually get redirected to remove the trailing slash so maybe better to build with that in mind.

app/allocations/views/list.njk
```diff
@@ -0,0 +1,34 @@
...
+      classes: "govuk-!-margin-top-2 app-print--hide"
+    }) }}
+  {% endif %}
+  {{ appFilter({
+    items: allocationTypeNavigation
+  }) }}
+{% endblock %}
+
+{% block pageContent %}
+{% if allocations.length %}
```
teneightfive, line 20: Maybe keep the consistency in nesting and indent this block 2 spaces?

app/allocations/views/list.njk
```diff
@@ -0,0 +1,34 @@
...
+  {{ govukTable({
+    rows: allocations,
+    head: allocationsHeads
+  }) }}
+{% else %}
+    {{ appMessage({
+      classes: "app-message--muted govuk-!-margin-top-7",
+      allowDismiss: false,
+      content: {
+        html: t("moves::dashboard.single_moves_no_results")
```
teneightfive, line 30: Is this using the correct key? Looks like its using content from single requests which might differ.

locales/en/allocations.json
```diff
@@ -2,5 +2,13 @@
   "dashboard": {
     "heading": "Allocations",
     "labels": { "single_requests": "single requests" }
-  }
+  },
+  "complete": "Complete",
```
teneightfive, line 6: Maybe use lowercase for the labels for consistency. 

As before, the label should be read as part of a sentence after the value: `1 complete (allocation)` or `3 incomplete (allocations)`. This doesn't look right if the word is capitalized: `3 Incomplete (allocation)`.

app/allocations/middleware.test.js
barbaracassani: Yup, well spotted.

app/allocations/views/list.njk
barbaracassani: Good point, I shouldn't recycle tokens! Shall amend.

app/allocations/middleware.test.js
barbaracassani: As you say, the trailing slash should be indifferent (interestingly, in hapi it's an option and the default is not to redirect). I have removed them.

PR #456, opened 2020-04-26
[P4-1307] Add image to move detail and create summary

teneightfive:
## Proposed changes

This change adds the person's image to the right hand summary panel
displayed on both the move creation journey and the move detail
page.

## Screenshots

<kbd></kbd>

<kbd></kbd>

### DIFF
--- a/common/assets/scss/core/_elements.scss
+++ b/common/assets/scss/core/_elements.scss
@@ -5,3 +5,8 @@ abbr {
 abbr[title] {
   cursor: help;
 }
+
+img {
+  display: inline-block;
+  max-width: 100%;
+}
--- a/app/move/views/view.njk
+++ b/app/move/views/view.njk
@@ -135,9 +135,15 @@
   {% if moveSummary %}
     <div class="sticky-sidebar">
       <div class="sticky-sidebar__inner">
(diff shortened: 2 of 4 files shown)


common/assets/scss/core/_elements.scss
```diff
@@ -5,3 +5,8 @@ abbr {
 abbr[title] {
   cursor: help;
 }
+
+img {
+  display: inline-block;
+  width: 100%;
```
solidgoldpig, line 11: a) Should this apply to all images?

b) Is this what we actually want? 

In mobile view the image will fill the viewport

Move_details_for_AARDVARKY__ANDREW

vs shrinking in the desktop view

Move_details_for_AARDVARKY__ANDREW

Might `max-width` be more appropriate here?

teneightfive: Good spot. This was a bit of a late night one...

I think we probably do want it to fill the container on smaller layouts. 

Also, the shrinkage on larger screens is a bit of a hack using the GOV.UK grid...probably another one to revisit at some point.

jason-sims:
late comment - good pickup @solidgoldpig. will put something in the backlog to pick up later.

PR #459, opened 2020-04-27
[P4-1155] Display possible conflicts/duplicates from NOMIS

teneightfive:
## Proposed changes

This introduces a new step to show any activities/court dates that a person may already have on that date in NOMIS.

This change is needed because we cannot guarantee if an item is a conflict or not, so we need them to specifically decide. To make that decision we need to give them the information we have.

If the API returns an empty timetable for that person for that day then we skip this step and the user never sees it.

**NOTE:** This PR doesn't yet link the journey up. It creates the skeleton work needed to try and move the code forward and not hold up merging these changes too long.

## Screencapture

These give a rough overview of the two journeys. For latest copy, see the screenshots section.

### When a person has NOMIS activities

<kbd></kbd>

### When a does has no activities

<kbd></kbd>


jason-sims:
Thanks for this - sorry only got to it now.

### Comments

- **Have conflicts path, this is perfect** :apple:

- **No conflicts path, approved** - I do think we are going to have to watch this space research wise. I am interested too see if their may be some other reason (maybe event in NOMIS or something we cannot determine like they don't want it in the diary yet) reason that we might want to show the conflicts page with no conflicts but give them the option to still not add it to NOMIS.

- Also will get a ticket in the backlog for a warning to show where the offender is not currently based at the prison as this will also make it clear that adding something to there diary will not work if they want to proceed (as they are not located there yet)

Thanks for this

app/move/views/_includes/incorrect-nomis-details.njk
```diff
@@ -0,0 +1,23 @@
+{% set html %}
+  <p>
+    {{ t("assessment::retrieved_from_nomis.details.content.lede") }}
+  </p>
+
+  <ul>
+    <li>
+      {{ t("assessment::retrieved_from_nomis.details.content.step_one") }}
```
barbaracassani, line 8: Maybe it's because I am not terribly familiar with the subject, but 1/2/3 is pretty inexpressive to me. Is there a way we can individuate more clearly what's going on in this template?

teneightfive: Hmm. Good point. The content here is split into 3 steps the user must follow to update the content in another system. Those may be tweaked so didn't want to tie the key name to specifically to the content in this sense.

Is there anything you think might be easier to understand?

teneightfive: This is the current content:

image

PR #461, opened 2020-04-27
[P4-1155] Link timetable journey together

teneightfive:
## Proposed

This adds the timetable step to the existing journey that was built in #459.

It now also includes some updates to the way the frontend talks to the API and some extra validation to ensure the start time of the court hearing is in the future.

This work also adds a feature flag so this can be merged and deployed without being immediately released.

### DIFF
--- a/app/move/controllers/create/court-hearings.js
+++ b/app/move/controllers/create/court-hearings.js
@@ -1,9 +1,10 @@
-const { formatISO, parseISO } = require('date-fns')
+const { formatISO, isValid, parseISO } = require('date-fns')
 const { find, pick } = require('lodash')
 
 const presenters = require('../../../../common/presenters')
 const componentService = require('../../../../common/services/component')
 const personService = require('../../../../common/services/person')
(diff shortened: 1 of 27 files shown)


app/move/controllers/create/court-hearings.js
```diff
@@ -43,6 +44,37 @@ class CourtHearingsController extends CreateBaseController {
     }
   }
 
+  process(req, res, next) {
+    const { court_hearing__start_time: startTime } = req.form.values
+    const moveDate = req.sessionModel.get('date')
+    const hearingDate = parseISO(`${moveDate}T${startTime}`)
```
solidgoldpig, line 50: This will give an invalid date if `startTime` is undefined. We can guarantee that `startTIme` will exist and be an empty string at least?

teneightfive: Good point. I've moved that inside the `if` block to only need to do these if `startTime` exists.

If the date is not valid, it will fallback to the original value - for example, during failed validation

PR #462, opened 2020-04-27
fix: Update court cases endpoint to only return active cases

teneightfive:
The API endpoint has been updated to include a filter to show active
court cases rather. This is the corresponding change on the frontend
to support that update.

cc/ @alexdesi


teneightfive:
Merging with Admin privileges as the Codeclimate webhook seems to have been disconnected so isn't fully reporting.

PR #465, opened 2020-04-27
Improve loading of update steps

solidgoldpig:
- Enable multiple steps to be loaded per journey
- Enable different role per journey
- Use key for urls and locales lookup rather than brittle url munging
- Process all update link data up front in move view
- Add ability to moveSummaryList component to output action links
- Enable link to update move date
- Suppress updating of to location route until redirect endpoint in place
- Ensure e2e tests actually change gender

Move_details_for_MCDERMOTT__JAMES

### Checklist

### DIFF
--- a/app/move/controllers/update/base.js
+++ b/app/move/controllers/update/base.js
@@ -76,7 +76,7 @@ class UpdateBaseController extends CreateBaseController {
 
       try {
         if (req.initialStep) {
-          values = this.getUpdateValues(req)
+          values = this.getUpdateValues(req, res)
         }
         this.protectReadOnlyFields(req, values)
       } catch (error) {
--- a/app/move/controllers/view.js
+++ b/app/move/controllers/view.js
@@ -1,7 +1,10 @@
-const { sortBy } = require('lodash')
+const { get, sortBy } = require('lodash')
 
 const presenters = require('../../../common/presenters')
 const updateSteps = require('../steps/update')
+
+const getUpdateLinks = require('./view/view.update.links')
+const getUpdateUrls = require('./view/view.update.urls')
 
 module.exports = function view(req, res) {
   const { move } = res.locals
@@ -16,22 +19,17 @@ module.exports = function view(req, res) {
       ? cancellationComments
       : req.t(`fields::cancellation_reason.items.${cancellationReason}.label`)
   const bannerStatuses = ['cancelled']
+
+  const userPermissions = get(req.session, 'user.permissions')
+  const updateUrls = getUpdateUrls(updateSteps, move.id, userPermissions)
+  const updateActions = getUpdateLinks(updateSteps, updateUrls)
+
   const urls = {
-    update: {},
+    update: updateUrls,
   }
-  updateSteps.forEach(step => {
-    const url = Object.keys(step)[0].replace(/^\//, '')
-    const key = url.replace(/-/g, '_')
-    urls.update[key] = `/move/${move.id}/edit/${url}`
-    // some assessment categories aren't suffixed with _information
-    if (key.endsWith('_information')) {
-      const keyWithoutInformation = key.replace(/_information$/, '')
-      urls.update[keyWithoutInformation] = urls.update[key]
-    }
-  })
 
   const locals = {
-    moveSummary: presenters.moveToMetaListComponent(move),
+    moveSummary: presenters.moveToMetaListComponent(move, updateActions),
     personalDetailsSummary: presenters.personToSummaryListComponent(person),
     tagList: presenters.assessmentToTagList(person.assessment_answers),
     assessment: presenters
@@ -58,6 +56,7 @@ module.exports = function view(req, res) {
       context: status,
       reason,
     }),
+    updateLinks: updateActions,
     urls,
   }
 
--- a/app/move/controllers/view/view.update.links.js
+++ b/app/move/controllers/view/view.update.links.js
@@ -1,0 +1,28 @@
+const i18n = require('../../../../config/i18n')
+
+const getUpdateLinks = (updateSteps, urls) => {
+  const updateLinks = {}
+  updateSteps.forEach(updateJourney => {
+    const category = updateJourney.key
+    if (!urls[category]) {
+      return
+    }
+    const categoryText = i18n.t(`moves::update_link.categories.${category}`)
+    if (categoryText) {
+      updateLinks[category] = {
+        category,
+        attributes: {
+          'data-update-link': category,
+        },
+        href: urls[category],
+        html: i18n.t('moves::update_link.link_text', {
+          context: 'with_category',
+          category: categoryText,
+        }),
+      }
+    }
+  })
+  return updateLinks
+}
+
+module.exports = getUpdateLinks
--- a/app/move/controllers/view/view.update.urls.js
+++ b/app/move/controllers/view/view.update.urls.js
@@ -1,0 +1,19 @@
+const { check } = require('../../../../common/middleware/permissions')
+
+const getUpdateUrls = (updateSteps, moveId, userPermissions) => {
+  const updateUrls = {}
+  updateSteps.forEach(updateJourney => {
+    if (!check(updateJourney.permission, userPermissions)) {
+      return
+    }
+    const steps = updateJourney.steps
+    const entryPointUrl = Object.keys(steps).filter(
+      step => steps[step].entryPoint
+    )[0]
+    const key = updateJourney.key
+    updateUrls[key] = `/move/${moveId}/edit${entryPointUrl}`
+  })
+  return updateUrls
+}
+
+module.exports = getUpdateUrls
--- a/app/move/steps/update.js
+++ b/app/move/steps/update.js
@@ -1,44 +1,88 @@
-const { cloneDeep } = require('lodash')
-
 const {
   Assessment,
   MoveDate,
-  MoveDetails,
+  // TODO: reenable when redirect api available
+  // MoveDetails,
   PersonalDetails,
 } = require('../controllers/update')
 
 const createSteps = require('./create')
 
-const updateControllers = {
-  '/personal-details': PersonalDetails,
-  '/move-details': MoveDetails,
-  '/move-date': MoveDate,
-  '/court-information': Assessment,
-  '/risk-information': Assessment,
-  '/health-information': Assessment,
+const updateStepPropOverrides = {
+  entryPoint: true,
+  backLink: null,
+  next: undefined,
+  buttonText: 'actions::save_and_continue',
 }
 
-const provideStepProps = (key, step) => {
-  const stepProps = {
-    entryPoint: true,
-    backLink: null,
-    next: undefined,
-    buttonText: 'actions::save_and_continue',
-  }
-
-  const updatedStep = {
-    ...cloneDeep(step),
-    ...stepProps,
-  }
-
-  if (updateControllers[key]) {
-    updatedStep.controller = updateControllers[key]
-  }
-  return updatedStep
-}
-
-const updateSteps = Object.keys(createSteps)
-  .filter(key => updateControllers[key])
-  .map(key => ({ [key]: provideStepProps(key, createSteps[key]) }))
+const updateSteps = [
+  {
+    key: 'personal_details',
+    permission: 'move:update',
+    steps: {
+      '/personal-details': {
+        ...createSteps['/personal-details'],
+        ...updateStepPropOverrides,
+        controller: PersonalDetails,
+      },
+    },
+  },
+  // TODO: reenable when redirect api available
+  // {
+  //   key: 'move',
+  //   role: 'move:update',
+  //   steps: {
+  //     '/move-details': {
+  //       ...createSteps['/move-details'],
+  //       ...updateStepPropOverrides,
+  //       controller: MoveDetails,
+  //     },
+  //   },
+  // },
+  {
+    key: 'date',
+    permission: 'move:update',
+    steps: {
+      '/move-date': {
+        ...createSteps['/move-date'],
+        ...updateStepPropOverrides,
+        controller: MoveDate,
+      },
+    },
+  },
+  {
+    key: 'court',
+    permission: 'move:update',
+    steps: {
+      '/court-information': {
+        ...createSteps['/court-information'],
+        ...updateStepPropOverrides,
+        controller: Assessment,
+      },
+    },
+  },
+  {
+    key: 'risk',
+    permission: 'move:update',
+    steps: {
+      '/risk-information': {
+        ...createSteps['/risk-information'],
+        ...updateStepPropOverrides,
+        controller: Assessment,
+      },
+    },
+  },
+  {
+    key: 'health',
+    permission: 'move:update',
+    steps: {
+      '/health-information': {
+        ...createSteps['/health-information'],
+        ...updateStepPropOverrides,
+        controller: Assessment,
+      },
+    },
+  },
+]
 
 module.exports = updateSteps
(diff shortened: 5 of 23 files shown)


app/move/controllers/update/base.js
```diff
@@ -76,7 +76,7 @@ class UpdateBaseController extends CreateBaseController {
 
       try {
         if (req.initialStep) {
-          values = this.getUpdateValues(req)
+          values = this.getUpdateValues(req, res)
```
solidgoldpig, line 79: Missed this in previous PR

app/move/views/_includes/assessment.njk
```diff
@@ -1,5 +1,5 @@
 {% for assessmentCategory in assessment %}
-  <section class="app-!-position-relative govuk-!-margin-top-7" data-app-section="{{ assessmentCategory.key }}">
+  <section class="app-!-position-relative govuk-!-margin-top-7">
```
solidgoldpig, line 2: Not needed any more

test/e2e/pages/create-move.js
```diff
@@ -187,7 +187,7 @@ class CreateMovePage extends Page {
       },
       gender: {
         selector: this.fields.gender,
-        value: faker.random.arrayElement(['Male', 'Female']),
+        value: person.gender,
```
solidgoldpig, line 190: Random selection now handled by `generatePerson`

jason-sims:
tag when you have the build that you want me to review. cannot test the active ff for police user with this deployment. May be easier to review together. Cheers.

solidgoldpig:
> tag when you have the build that you want me to review. cannot test the active ff for police user with this deployment. May be easier to review together. Cheers.

Probably better to wait for the confirmation messages to drop first (which will be along soon - just writing the tests)

app/move/controllers/view/view.update.links.js
```diff
@@ -0,0 +1,23 @@
+const getUpdateLinks = (updateSteps, t, urls) => {
+  const updateLinks = {}
+  updateSteps.forEach(updateJourney => {
+    const category = updateJourney.key
+    if (!urls[category]) {
+      return
+    }
+    const categoryText = t('moves::update_link.categories.' + category)
```
barbaracassani, line 8: I was under the impression that we enforced the use of html template strings for string concatenation, but I might be wrong.

test/e2e/_move.js
```diff
@@ -132,13 +132,7 @@ export async function checkNoUpdateLink(page) {
...
 }
 
-const updatePages = [
-  'personal_details',
-  'risk',
-  'health',
-  'court',
-  'move_details',
-]
+const updatePages = ['personal_details', 'risk', 'health', 'court', 'date']
```
barbaracassani, line 135: The dramatic changes in formatting when you rename / remove an element are always funny.

app/move/controllers/view/view.update.links.js
teneightfive: I think its the preferred pattern but maybe we're missing an eslint rule for it...

app/move/steps/update.js
```diff
@@ -1,44 +1,88 @@
...
+  buttonText: 'actions::save_and_continue',
 }
 
-const updateSteps = Object.keys(createSteps)
-  .filter(key => updateControllers[key])
-  .map(key => ({ [key]: provideStepProps(key, createSteps[key]) }))
+const updateSteps = [
+  {
+    key: 'personal_details',
+    role: 'move:update',
```
teneightfive, line 21: I wonder if this is the wrong term. We've tried really hard to not have "roles" within the app.

When a user signs in, with given "roles", we assign permissions, so maybe if that's what we're trying to do here we should use `permissions` as an array for consistency?

common/components/meta-list/template.njk
```diff
@@ -1,7 +1,7 @@
 <dl class="app-meta-list {% if params.classes -%} {{ params.classes }}{%- endif %}">
   {% for item in params.items %}
     {% if item.value.text or item.value.html %}
-      <div class="app-meta-list__item">
+      <div class="app-meta-list__item app-!-position-relative">
```
teneightfive, line 4: Could this be added using `params.classes` rather than this class applied and style being applied to all areas where this component is used.

common/components/meta-list/template.njk
```diff
@@ -11,6 +11,15 @@
         <dd class="app-meta-list__value">
           {{ item.value.html | safe if item.value.html else item.value.text }}
         </dd>
+
+        {% set updateLink = item.link %}
+        {% if updateLink %}
```
teneightfive, line 16: Also, the addition of this no longer feels content agnostic. The naming of the variables is binding it specifically to the link saying "Update" or "Change", whereas in the terms of a component, it should have more flexibility.

I think it would be cleaner for this component to support a similar pattern to the [GOV.UK Summary list component](https://design-system.service.gov.uk/components/summary-list/) where the component can have "actions" and the presenter sends in the text and href constructed from the URLs object.

teneightfive: I think this pattern of "actions" would be nice to use actually:

```
...
actions: {
  items: [
    {
      href: "#",
      text: "Change",
      visuallyHiddenText: "name"
    }
  ]
}
...
```

app/move/controllers/view/view.update.links.js
solidgoldpig: We don't enforce it - but it is better

app/move/steps/update.js
solidgoldpig: `permission` fine. Why an array though? This gets passed to `protectRoute` which takes a string

teneightfive: Ah yes. That middleware only support a string. And if we need to protect based on multiple permissions we have used that middleware twice.

Maybe you need to cater for that here though. A user may require `move:update` as well as `move:update:documents` for a particular step. Passing just `move:update:documents` wouldn't feel enough as we would require that user to have the permission to update first as well as update at a granular level.

teneightfive: Is this to support STC/SCH where only they can edit/upload documents?

At the moment, that logic is actually tied to the `from_location` rather than a specific permission because of the way it was originally required. We may want the ability to allow other establishments to upload documents, regardless of that user's roles/permissions.

If it's just to support documents at this stage, this may not be the best place for that logic if we need to keep it tied to `from_location`. That logic could live in the [documents controller](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/95be55df32857b54a6a3193cba8c469b524ebf20/app/move/controllers/create/document.js#L16) instead.

common/components/meta-list/template.njk
solidgoldpig: tldr; the meta-list component is already a specialist component with hard-coded fields in it.

That feels like premature optimisation? Do we know that the metalist requires multiple actions?  The following would do what you suggest, but it can only do it by subsuming information about classes and attributes which to me seems even yuckier. 

`common/components/meta-list/template.njk`

```njk
{%- macro _actionLink(action) %}
  <a class="govuk-link {%- if action.classes %} {{ action.classes }}{% endif %}" href="{{ action.href }}" {%- for attribute, value in action.attributes %} {{attribute}}="{{value}}"{% endfor %}>
    {{ action.html | safe if action.html else action.text }}
    {%- if action.visuallyHiddenText -%}
      <span class="govuk-visually-hidden"> {{ action.visuallyHiddenText }}</span>
    {% endif -%}
  </a>
{% endmacro -%}

<dl class="app-meta-list {% if params.classes -%} {{ params.classes }}{%- endif %}">
  {% for item in params.items %}
    {% if item.value.text or item.value.html %}
      <div class="app-meta-list__item {% if item.classes -%} {{ item.classes }}{%- endif %}">
        {% if item.key.html or item.key.text %}
          <dt class="app-meta-list__key">
            {{ item.key.html | safe if item.key.html else item.key.text }}
          </dt>
        {% endif %}

        <dd class="app-meta-list__value">
          {{ item.value.html | safe if item.value.html else item.value.text }}
        </dd>

        {% if item.actions.items.length %}
          <dd class="app-meta-list__actions {%- if item.actions.classes %} {{ item.actions.classes }}{% endif %}">
            {% if item.actions.items.length == 1 %}
              {{ _actionLink(item.actions.items[0]) | indent(12) | trim }}
            {% else %}
              <ul class="app-meta-list__actions-list">
                {% for action in item.actions.items %}
                  <li class="app-meta-list__actions-list-item">
                    {{ _actionLink(action) | indent(18) | trim }}
                  </li>
                {% endfor %}
              </ul>
            {% endif %}
          </dd>
        {% endif %}
      </div>
    {% endif %}
  {% endfor %}
</dl>
```

`common/presenters/move-to-meta-list-component.js`

```js
...
    additional_information: additionalInfo,
  } = {},
  additionalProps = {}
) {
  const destination = get(toLocation, 'title', 'Unknown')
  const destinationLabel =
    moveType === 'prison_recall'
      ? i18n.t('fields::move_type.items.prison_recall.label')
      : destination
  const additionalInformation = additionalInfo ? ` — ${additionalInfo}` : ''
  const items = [
    {
      key: {
        text: i18n.t('fields::from_location.short_label'),
      },
      value: {
        text: get(fromLocation, 'title'),
      },
    },
    {
      ...additionalProps.move,
      key: {
        text: i18n.t('fields::move_type.short_label'),
      },
      value: {
        text: destinationLabel + additionalInformation,
      },
    },
    {
      ...additionalProps.date,
      key: {
        text: i18n.t('fields::date_type.label'),
      },
      value: {
        text: setDateToDisplay({ date, dateFrom, dateTo }),
      },
    },
...
```
`app/move/controllers/view.js`

```js
...
  const updateLinks = getUpdateLinks(updateSteps, req.t, updateUrls)

  const urls = {
    update: updateUrls,
  }

  const getSummaryItemProps = actions => {
    if (!actions) {
      return
    }
    return {
      classes: 'app-!-position-relative',
      actions: {
        classes: 'app-!-position-top-right govuk-!-font-size-16',
        items: [actions],
      },
    }
  }

  const metaListProps = {
    date: getSummaryItemProps(updateLinks.date),
    move: getSummaryItemProps(updateLinks.move),
  }

  const locals = {
    moveSummary: presenters.moveToMetaListComponent(move, metaListProps),
    personalDetailsSummary: presenters.personToSummaryListComponent(person),
...
```

`app/move/controllers/view/view.update.links.js`
```js
...
      updateLinks[category] = {
        category,
        attributes: {
          'data-update-link': category,
        },
        href: urls[category],
        html: t('moves::update_link.link_text', {
          context: 'with_category',
          category: categoryText,
        }),
      }
...
```

Can we leave this for more discussion and not block getting this in for now?

teneightfive: I hear you about we might not need multiple at this stage so that may be overkill. My point was more around this component being more agnostic to the content of the links you're currently adding.

Like with the GOV.UK example, I think the difference between a component and using elements within a template or specific view is that a component should be more generic in its use. The name of things suggests that this would only ever be used to add an "Update" link, but in reality the component doesn't care what the link says.

I also understand we don't want to hold changes up, but I think changes to components just need to be thought of a bit more to not tie them down to single use, otherwise they lost the ability to be as easily re-used.

teneightfive: The change needed could also be handled by using `item.value.html` instead of `item.value.text` and adding the link in the presenter.

And wouldn't require any updates to the component.

app/move/steps/update.js
solidgoldpig: It also allows canAccess to be called within the controller using the value from the config

common/components/meta-list/template.njk
solidgoldpig: Updated as discussed with meta-list component gaining action and not requiring app-!-style overrides

common/components/meta-list/template.njk
solidgoldpig: Made redundant by updating the component styling

app/move/controllers/view/view.update.links.js
```diff
@@ -0,0 +1,26 @@
+const getUpdateLinks = (updateSteps, t, urls) => {
```
teneightfive, line 1: Rather than having to pass in `t` you can directly import the [i18n module](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/5f692a046faea2298687c7a23bc4882734c7a60b/config/i18n.js) and call `i81n.t` on that.

Here's [an example](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/5f692a046faea2298687c7a23bc4882734c7a60b/common/presenters/move-to-card-component.js#L1).

app/move/controllers/view/view.update.urls.js
```diff
@@ -0,0 +1,17 @@
+const getUpdateUrls = (updateSteps, moveId, canAccess) => {
```
teneightfive, line 1: I think rather than passing in the `canAccess` function which is a view helper it would be cleaner to pass in the current user's permissions and use the [permissions module](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/5f692a046faea2298687c7a23bc4882734c7a60b/common/middleware/permissions.js#L3) to do this check.

app/move/controllers/view.js
```diff
@@ -16,22 +19,24 @@ module.exports = function view(req, res) {
...
-    urls.update[key] = `/move/${move.id}/edit/${url}`
-    // some assessment categories aren't suffixed with _information
-    if (key.endsWith('_information')) {
-      const keyWithoutInformation = key.replace(/_information$/, '')
-      urls.update[keyWithoutInformation] = urls.update[key]
+
+  const metaListActions = {}
+  Object.keys(updateLinks).forEach(key => {
+    metaListActions[key] = {
+      classes: 'app-meta-list__action--sidebar',
```
teneightfive, line 33: It feels like all of this should live in the presenter. It's purely related to how we display the actions when they're in that component so feels better to contain them there rather than in this controller.

This controller is starting to contain a lot of logic and we've previously tried (where possible) to keep thin controllers and extract any complicated logic to smaller modules.

common/components/meta-list/template.njk
```diff
@@ -11,6 +11,14 @@
         <dd class="app-meta-list__value">
           {{ item.value.html | safe if item.value.html else item.value.text }}
         </dd>
+
+        {% if item.action.html or item.action.text %}
+          <dd class="app-meta-list__action {% if item.action.classes -%} {{ item.action.classes }}{%- endif %}">
```
teneightfive, line 16: Thanks for these changes. This feels much cleaner from a component perspective and removes the tie to the content of the element.

However, it would be good if this could be split into a separate commit with a little explanation why we changed it (maybe once everything has been rebased).

If you look back on that component with `git blame` embedded in the rest of this PR it will be a bit noisy.

app/move/controllers/view/view.update.links.js
solidgoldpig: Sure, but there is a slight werdness there. I had done that and got bitten by using 

```
const { t } = require('../../../../config/i18n')
```

and getting

```
error: Cannot read property 'translator' of undefined {"stack":"TypeError: Cannot read property 'translator' of undefined\n    at t
```

whereas _obviously_ I should have been doing ;)

```js
const i18n = require('../../../../config/i18n')

module.exports = () => {
...
   const fooStr = i18n.t('foo')
}
```

Maybe it's worth waiting for i18n to actually be properly instantiated before it gets exported. Not a biggy, but principle of least surprise...

app/move/controllers/view/view.update.urls.js
solidgoldpig: Done

app/move/controllers/view.js
solidgoldpig: Done

common/components/meta-list/template.njk
solidgoldpig: Rejigged

app/move/controllers/view/view.update.links.js
teneightfive: I also find the second approach is much testable as you can use sinon to stub the `t` method. Whereas if importing using destructuring you would have to use something lik proxyquire to stub and it gets messy...

PR #467, opened 2020-04-28
[Fix] End-to-end tests for cancelled banner

teneightfive:
## Proposed changes

The banners were recently updated and the identifier used has also
changed.

This caused some of the end-to-end tests to fail.

This fix updates the identifier for that element to the new one.

This PR also contains an update to ensure the status badge is only shown for
cards that require them to be distinguished - cancelled moves are already separated
on the page.

### DIFF
--- a/common/presenters/move-to-card-component.js
+++ b/common/presenters/move-to-card-component.js
@@ -5,6 +5,10 @@ const personToCardComponent = require('./person-to-card-component')
 function moveToCardComponent({ showMeta = true, showTags = true } = {}) {
   return function item({ id, reference, person = {}, status }) {
     const href = `/move/${id}`
+    const excludedBadgeStatuses = ['cancelled']
+    const statusBadge = excludedBadgeStatuses.includes(status)
+      ? undefined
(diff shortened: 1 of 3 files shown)


common/presenters/move-to-card-component.js
```diff
@@ -5,16 +5,18 @@ const personToCardComponent = require('./person-to-card-component')
 function moveToCardComponent({ showMeta = true, showTags = true } = {}) {
   return function item({ id, reference, person = {}, status }) {
     const href = `/move/${id}`
+    const excludedBadgeStatuses = ['cancelled']
+    const statusBadge = excludedBadgeStatuses.includes(status)
```
barbaracassani, line 9: Maybe `const statusBadge = excludedBadgeStatuses.includes(status) &&  { text: i18n.t(`statuses::${status}`) }` is slightly more expressive? I tend to use it instead of ternaries with an undefined. But it's a matter of personal taste, and many people don't like it.

barbaracassani: Ah, forget the above, it becomes `false`, of course, not `undefined`.

PR #471, opened 2020-04-29
Add download and print links to single requests dashboard

teneightfive:
## Proposed changes

This set of changes includes adding back in the download (if a user has permission to download) and print links that appear on the other dashboards for the single requests page.

It also includes some updates to the content to make things more consistency across the different screens.

## Screenshots

*Click screenshots to enlarge*

### "Home"

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

### Single requests

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

### DIFF
--- a/app/moves/controllers/dashboard.js
+++ b/app/moves/controllers/dashboard.js
@@ -1,3 +1,5 @@
 module.exports = function dashboard(req, res) {
-  res.render('moves/views/dashboard')
+  res.render('moves/views/dashboard', {
+    pageTitle: 'moves::dashboard.dashboard',
+  })
 }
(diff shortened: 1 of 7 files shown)


common/presenters/moves-to-table.test.js
```diff
@@ -148,13 +148,13 @@ describe('#movesToTable', function() {
           html: 'Name',
         },
         {
-          html: 'Created at',
+          html: 'Sent on',
```
teneightfive, line 151: @barbaracassani These tests failing when I updated the content in the locales file worried me a bit.

Because this component is now using the other two, I don't think the i18n mocks are really working.

I'm wondering if its cleaner to just have separate transformers for moves/allocations -> table component rather than trying to share the head/rows objects. It feels like its not DRYing it up that much and making it a bit trickier to test. As if we want to mock the deps in this test we have to use proxyquire which always feels a bit annoying. What are your thoughts?

barbaracassani: Addressed in my own pr #476

PR #473, opened 2020-04-29
Use card component within table layout

teneightfive:
This change make use of re-using the card component when displaying
moves within a table layout. This means we can have more consistency between different layouts
when displaying a person's information.

The word "Move" was also removed from the reference label so it could be displayed better in the table layout. This update occurs across existing dashboards and layouts too.

This change also includes an update to the ordering so that they are ordered by the latest created move.

## Screenshots

*Click image to enlarge*

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

### DIFF
--- a/common/presenters/move-to-card-component.js
+++ b/common/presenters/move-to-card-component.js
@@ -2,14 +2,24 @@ const i18n = require('../../config/i18n')
 
 const personToCardComponent = require('./person-to-card-component')
 
-function moveToCardComponent({ showMeta = true, showTags = true } = {}) {
+function moveToCardComponent({
+  isCompact = false,
+  showImage = true,
+  showMeta = true,
+  showTags = true,
+} = {}) {
   return function item({ id, reference, person = {}, status }) {
     const href = `/move/${id}`
     const excludedBadgeStatuses = ['cancelled']
-    const statusBadge = excludedBadgeStatuses.includes(status)
-      ? undefined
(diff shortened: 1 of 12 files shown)


Andy-Millar:
Worth noting - can we update on the 'panel' component on the confirmation page too, along with profile

common/presenters/move-to-card-component.js
```diff
@@ -2,21 +2,32 @@ const i18n = require('../../config/i18n')
...
     const excludedBadgeStatuses = ['cancelled']
-    const statusBadge = excludedBadgeStatuses.includes(status)
-      ? undefined
-      : { text: i18n.t(`statuses::${status}`) }
-    const personCardComponent = personToCardComponent({ showMeta, showTags })({
+    const statusBadge =
+      excludedBadgeStatuses.includes(status) || isCompact
+        ? undefined
+        : { text: i18n.t(`statuses::${status}`) }
+    const personCardComponent = personToCardComponent({
```
solidgoldpig, line 18: Is there a reason for the personToCardComponent to not handle the  `isCompact`ness? Just asking...

teneightfive: At this stage it was only needed for displaying the full "card" so didn't want to complicate the person presenter just yet.

The person presenter is shared between being used in the card and the person search results. In its current singular use, the person doesn't need a "compact" variation.

PR #475, opened 2020-04-30
fix: Update how from and to dates are handled in the summary panel

teneightfive:
Previously we were combining the from and to date into the one date
row on the right hand move summary panel.

We've decided to split this into separate rows to give a clearer picture
of the situation.

This also fixes a bug where the mock date happened to be tomorrow and
we weren't mocking the date formatters which caused the text to be
different and the tests to start failing.

### With only a move date (or all dates)

*No changes - posting for visual reference to other states*

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

### With only a from date

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

### With both from and to date

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

### DIFF
--- a/common/presenters/move-to-meta-list-component.js
+++ b/common/presenters/move-to-meta-list-component.js
@@ -12,24 +12,16 @@ function isRelativeDate(date) {
   )
 }
 
-function formatDateRange(dateFrom, dateTo) {
-  if (dateTo) {
-    return filters.formatDateRange([dateFrom, dateTo])
+function _formatDate(date) {
+  if (!date) {
+    return
   }
-  return `${i18n.t('from')} ${filters.formatDateAsRelativeDay(dateFrom)}`
-}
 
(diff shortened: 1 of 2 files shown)


common/presenters/move-to-meta-list-component.js
```diff
@@ -81,10 +73,26 @@ function moveToMetaListComponent(
...
+      value: {
+        text: !date ? _formatDate(dateFrom) : undefined,
+      },
+    },
+    {
+      key: {
+        text: i18n.t('fields::date_to.label'),
+      },
+      value: {
+        text: !date ? _formatDate(dateTo) : undefined,
```
barbaracassani, line 93: Maybe reverse this, it's usually discouraged to use negative conditions.

Andy-Millar:
Great work

PR #477, opened 2020-04-30
[NOJIRA] chore: Move the translations to the object table cell presenters

barbaracassani:
This PR moves the translations of the table headers to the header presenter itself, in order to make the main method more testable.

### Checklist


common/presenters/object-to-table-presenters.js
```diff
@@ -0,0 +1,7 @@
+const objectToTableHead = require('./object-to-table-head')
+const objectToTableRow = require('./object-to-table-row')
+
+module.exports = {
```
teneightfive, line 4: This feels like this "index loader" is just to make test stubbing easier, right?

Not sure this feels quite right and doesn't follow previous patterns of an index loader being an `index.js` file within a folder. 

I'd either say use `proxyquire` to stub these presenters like we have elsewhere, and has been added in #473 or create these using the same index loader pattern with a folder.

barbaracassani: Fixed 👍

PR #478, opened 2020-04-30
Refactor person service unformat method

solidgoldpig:
- Move all tests to submodule
- Decouple tests from specific fields
- Allow value type keys to be passed through

### Checklist

### DIFF
--- a/common/services/person/person.unformat.test.js
+++ b/common/services/person/person.unformat.test.js
@@ -1,0 +1,278 @@
+const filters = require('../../../config/nunjucks/filters')
+
+const unformat = require('./person.unformat')
+
+describe('Person Service', function() {
+  describe('#unformat()', function() {
+    let person
+    let fields
+    let keys
+    let unformatted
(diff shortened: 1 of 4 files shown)


common/services/person/person.unformat.test.js
```diff
@@ -0,0 +1,271 @@
...
+
+    context('when asking for a date', function() {
+      before(function() {
+        fields = ['dateField']
+        keys = {
+          date: ['dateField'],
+        }
+      })
+      it('should return the value formatted as a date', function() {
+        expect(unformatted).to.deep.equal({ dateField: '1 Jan 2000' })
```
teneightfive, line 148: I think it would be worth stubbing the date format method in this instance. 

If we change the default date format we don't want all these tests to break. We just need to know that it's calling the `formatDate` method with the correct value.

solidgoldpig: Added

PR #479, opened 2020-04-30
Display prison transfer reason

teneightfive:
## Proposed changes

This fixes the display of the prison transfer reason on the list of moves in table format and in the right hand summary panel on the move detail page and during the create a new move journey.

The value of additional comments is now being stored in the `additional_information` attribute in the API. 

This meant some re-work of how we use `additional_information` for the existing prison recall journey.

It ensures they are stored as separate values within the form journey to cater for change of destination but they are stored as `additional_information` when saved to the API.

### Dashboard

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

### Move detail / create journey

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |


common/presenters/move-to-meta-list-component.js
```diff
@@ -101,6 +105,18 @@ function moveToMetaListComponent(
         text: filters.formatTime(timeDue),
       },
     },
+    {
+      key: {
+        text: i18n.t('fields::prison_transfer_type.label'),
+      },
+      value: {
+        text:
```
solidgoldpig, line 113: Finding the legibility of this really hard - it's not the line length it's nested ternaries and the double check plus the concatenation (for the nation). Struggling with the one at line 67 too to be honest. Maybe calculate the value before setting the property?

common/presenters/move-to-meta-list-component.test.js
```diff
@@ -10,7 +10,7 @@ const mockMove = {
   date: '2019-06-09',
   time_due: '2000-01-01T14:00:00Z',
   move_type: 'court_appearance',
-  additional_information: 'Some additional information about this move',
+  // additional_information: 'Some additional information about this move',
```
solidgoldpig, line 13: Is this needed now?

teneightfive: Nope! Good spot

teneightfive:
> Hoping we get to sort out the columns, otherwise great stuff

@Andy-Millar Have potentially found a way and will be coming in a later change...

PR #480, opened 2020-04-30
Show confirmation message after updating move [P4 1353]

solidgoldpig:
If any values on a move are changed, we now display a confirmation message that they have been saved on the move view page.

Move_details_for_SHIELDSY__PHOEBE

### Checklist

### DIFF
--- a/app/move/controllers/update/base.test.js
+++ b/app/move/controllers/update/base.test.js
@@ -574,6 +574,7 @@ describe('Move controllers', function() {
     const res = {}
     let nextSpy
     beforeEach(async function() {
+      sinon.stub(controller, 'setFlash')
       sinon.stub(moveService, 'update').resolves()
       req = {
         getMoveId: sinon.stub().returns('#moveId'),
@@ -607,6 +608,10 @@ describe('Move controllers', function() {
         expect(moveService.update).to.not.be.called
       })
 
+      it('should not set the confirmation message', async function() {
+        expect(controller.setFlash).to.not.be.called
+      })
+
       it('should invoke next with no error', async function() {
         expect(nextSpy).to.be.calledOnceWithExactly()
       })
@@ -625,6 +630,10 @@ describe('Move controllers', function() {
         })
       })
 
+      it('should set the confirmation message', async function() {
+        expect(controller.setFlash).to.be.calledOnceWithExactly(req)
+      })
+
       it('should invoke next with no error', async function() {
         expect(nextSpy).to.be.calledOnceWithExactly()
       })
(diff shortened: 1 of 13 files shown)


app/move/controllers/update/move-date.js
```diff
@@ -1,12 +1,14 @@
...
-const moveService = require('../../../../common/services/move')
 const { formatDate } = require('../../../../config/nunjucks/filters')
 const CreateMoveDateController = require('../create/move-date')
 
 const UpdateBase = require('./base')
 
 class UpdateMoveDetailsController extends UpdateBase {
+  constructor(options) {
+    super(options)
+    this.saveFields = ['date']
```
teneightfive, line 9: Could this be derived from `req.form.options.fields`? I think it only includes fields that are available on this step.

app/move/controllers/update/base.test.js
```diff
@@ -574,6 +574,8 @@ describe('Move controllers', function() {
     const res = {}
     let nextSpy
     beforeEach(async function() {
+      sinon.stub(controller, 'setFlash')
+      // sinon.stub(personService, 'unformat').returns({})
```
teneightfive, line 578: Still needed?

app/move/controllers/update/base.test.js
```diff
@@ -607,6 +609,10 @@ describe('Move controllers', function() {
         expect(moveService.update).to.not.be.called
       })
 
+      it('should set the confirmation message', async function() {
```
teneightfive, line 612: Copy paste jobby? Guessing should be `should not set...`?

app/move/controllers/update/base.test.js
```diff
@@ -625,6 +631,10 @@ describe('Move controllers', function() {
         })
       })
 
+      it('should set the confirmation message', async function() {
```
teneightfive, line 634: Also, you can remove the `async` from all of these functions as there is no `await` inside them.

app/move/controllers/update/move-date.js
solidgoldpig: Unfortunately it includes additional fields (`date_type` and `date_custom`) which would cause this to fail. But come the sooper-dooper model-derived move formatter/unformatter, I think this will be able to be dropped

locales/en/moves.json
```diff
@@ -184,5 +184,34 @@
       "personal_details": "personal details",
       "risk": "risk information"
     }
+  },
+  "update_flash": {
+    "supplier_message": "This move has been updated with the supplier",
```
teneightfive, line 189: I would actually consider dropping this message. It's not _strictly_ true as I'm not sure we notify them of these changes yet. They would have to be subscribed to the webhooks.

Also, we removed that extra copy from the other banners so I feel we could potentially drop this paragraph altogether and just go with the heading. See what @Andy-Millar thinks though.

solidgoldpig: These are the messages that @Andy-Millar and @jason-sims supplied

app/move/controllers/update/base.test.js
solidgoldpig: Zapped

app/move/controllers/update/base.test.js
solidgoldpig: Corrected

app/move/controllers/update/base.test.js
solidgoldpig: You got me - I managed to paste it back in when I split up the commits. Doh!

locales/en/moves.json
jason-sims: For the users benefit we are saying the supplier has received the changes (via email notifications & later webhooks). Can leave copy in for now and will raise separate ticket to change if needed.

PR #482, opened 2020-04-30
Allow custom error labels to be set

teneightfive:
## Proposed changes

This change allows error messages to be customised per field. Previously we built them dynamically using only the field label and validation key.

This new behaviour allows us to either:

- override the message completely
- override the message completely for particular types of errors
- override the field label when it is being used to generate an error
- fallback to default error message using field label

**Note**: This doesn't get _all_ error messages perfect but it more to gives us the ability to customise them properly in the future when we have more time to go through each one individually, possibly with a content designer. It's main purpose was to allow us to fix ones that were very clearly wrong which used the question within an error message.

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |


common/helpers/index.js
```diff
@@ -1,9 +1,13 @@
...
 const referenceData = require('./reference-data')
-
+const foo = 'bar'
+function thing() {
+  return foo
+}
 module.exports = {
   dateUtil,
   field,
+  thing,
```
solidgoldpig, line 11: Um, what's a `thing`?

solidgoldpig:
All good. Just noting that the screenshots use the form `Enter ...` (per the design system guidelines) whereas what's actually output is `... cannot be blank`.

common/helpers/index.js
teneightfive: Bah! Good  spot!

teneightfive:
> All good. Just noting that the screenshots use the form `Enter ...` (per the design system guidelines) whereas what's actually output is `... cannot be blank`.

@solidgoldpig doh! That was a case of me taking screenshots then not actually pushing the latest changes. The locales updates should now be there.

PR #484, opened 2020-04-30
[P4 1281] Complete journey for creating an allocation

barbaracassani:
The complete journey for creating an allocation.

Screenshot 2020-05-04 at 18 54 44

Screenshot 2020-05-04 at 18 55 34

Screenshot 2020-05-04 at 18 55 15

### Checklist


common/middleware/locals.js
```diff
@@ -16,6 +17,7 @@ module.exports = function setLocals(req, res, next) {
     USER: req.session.user,
     CURRENT_LOCATION: req.session.currentLocation,
     MOVES_URL: req.session.movesUrl || movesApp.mountpath,
+    ALLOCATION_URL: allocationsApp.mountpath,
```
teneightfive, line 20: I don't think you necessarily need to copy this pattern here. The purpose of the moves value is so that it will remember any current day/location/filter you have when you go back rather than taking you always to "today".

Allocations sounds like it will sit outside of that style so you can just link to the allocations base path from the controller.

locales/en/messages.json
```diff
@@ -5,5 +5,11 @@
       "heading_proposed": "Move sent for review",
       "content": "Move reference<br><strong>{{moveReference}}</strong>"
     }
+  },
+  "create_allocation": {
+    "success": {
+      "heading": "Allocation requested",
+      "heading_proposed": "Move sent for review"
```
teneightfive, line 12: You don't need this context. This was to support the different states when creating a move. It uses a context of `proposed` to display a different heading.

common/services/allocation.js
```diff
@@ -8,6 +8,27 @@ function urlKeyToFilter(additionalFilters) {
   })
 }
 const allocationService = {
+  format(data) {
+    const booleansAndNulls = ['complete_in_full', 'sentence_length']
```
solidgoldpig, line 12: I'm not suggesting that we do anything about it as part of this PR - you are stepping in the footsteps of those who went before, but maybe we should consider figuring out what kind of formatting the field needs from its definition. Of course, there is a bit of a disjoint between the model and fields so hmmm...

Andy-Millar:
Small tweak for me please - for the freetext box (under other criteria question) ca we have this as 'Give details' please (to follow the pattern we've used everywhere). Also this freetext box in mandatory.

common/services/allocation.js
barbaracassani: Hmm... please explain. Are you suggesting we use some kind of Hungarian notation? Or just document the format required in the FE?

barbaracassani:
@Andy-Millar I have made the change as requested 👍

common/helpers/field/index.js
```diff
@@ -30,6 +37,10 @@ function mapReferenceDataToOption({ id, title, key, conditional, hint }) {
     option.key = key
   }
 
+  if (checked) {
```
teneightfive, line 40: I don't think we need to extend this helper for this functionality. This isn't something related to the item coming back from the reference data service, this is you needing to default an item to checked in a specific controller.

I would move that logic outside of this helper and keep it only where you need it.

common/services/allocation.js
solidgoldpig: Not hungarian notation. Not just documenting. Adding a property to the fields that enables it to derived automatically and not have to update the format methods manually

```js
fields: {
  complete_in_full: {
    format: 'boolean',
    ...
  },
  an_other: {
    format: 'identifier'
  ...
  }
...
```

common/helpers/field/index.js
barbaracassani: I think it's no really different from `hint` or `conditional` in the strict sense that we map attributes from a configuration object to the dom.

barbaracassani:
I have reordered the commits as requested.

common/services/allocation.js
barbaracassani: It's a good idea @solidgoldpig .

PR #490, opened 2020-05-06
[P4-1467] Filter single requests correctly

teneightfive:
## Proposed changes

Single requests were previously making use of existing moves services and URLs and passing a list of statuses directly from the URL to the API.

The problem with this is that to correctly filter down the moves to only display single requests we actually need to pass other filters to the API, like `move_type`. It also wasn't using the correct statuses within the existing URLs.

This work abstracts that work to a single requests service so that if statuses/filters change within the API they are contained within one place in this app.

That allows us to display moves based on filters in the UI and keep that abstraction away from URLs/controllers/middleware.

There should be no major difference in the screens for this change apart from the URL changes.

This is also a start to try and split the single requests work into its own sub-app as the way of displaying them feels separate enough to moves to warrant a new area within this app.

### DIFF
--- a/common/services/single-request.js
+++ b/common/services/single-request.js
@@ -1,0 +1,53 @@
+const { pickBy } = require('lodash')
+
+const moveService = require('./move')
+
+const singleRequestService = {
+  getAll({
+    status,
+    moveDate = [],
+    createdAtDate = [],
+    fromLocationId,
+    toLocationId,
+    isAggregation = false,
+  } = {}) {
+    const [moveDateFrom, moveDateTo] = moveDate
+    const [createdAtFrom, createdAtTo] = createdAtDate
+
+    let statusFilter
+    switch (status) {
+      case 'pending':
+        statusFilter = {
+          'filter[status]': 'proposed',
+        }
+        break
+      case 'accepted':
+        statusFilter = {
+          'filter[status]': 'requested,accepted,completed',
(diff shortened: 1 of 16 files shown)


common/services/single-request.js
```diff
@@ -0,0 +1,53 @@
...
+
+    let statusFilter
+    switch (status) {
+      case 'pending':
+        statusFilter = {
+          'filter[status]': 'proposed',
+        }
+        break
+      case 'accepted':
+        statusFilter = {
```
solidgoldpig, line 25: Is there a story here as to why these 3 terms are agglomerated?

teneightfive: There is. We are making re-use of the statuses that we know exist in our API. This combines all statuses that signify a request has been "approved" so we want to ensure they will continue to be displayed in the "approved" filter after a move progressing through more states.

solidgoldpig: (but I can see it makes the urls much cleaner ;)

teneightfive: In the future we will want to capture changes to state and may be able to derive the "approved" state another way, which I why I wanted to create this service to abstract that.

Currently if a new status was added that suggests a move has been approved we would have had to add it across various places. And also, deriving these particular moves required more filters than just status so also felt it was appropriate to abstract that to this service.

PR #491, opened 2020-05-06
chore: Allow locations to be set by env var during development

teneightfive:
This allows the locations to be set using an environment variable
populated with NOMIS Agency IDs.

This is useful during development when working offline or when the auth
service is not working.


teneightfive:
> Utter filth^w genius

🤭 🤫

PR #493, opened 2020-05-06
[P4 1440] Provide ability to update to location frontend

solidgoldpig:
Adds ability to change where a move is going to or, if a prison recall, amend the move’s additional information.

Adds link to edit move details from move view page.

Move_details_for_MELLIE__ROGER

Prison recall move
Where_is_this_person_moving__-_update_move

Other move types
Where_is_this_person_moving__-_update_move

### Checklist

### DIFF
--- a/app/move/controllers/update/move-details.js
+++ b/app/move/controllers/update/move-details.js
@@ -1,29 +1,76 @@
 const { get, pick } = require('lodash')
 
+const moveService = require('../../../../common/services/move')
 const CreateMoveDetailsController = require('../create/move-details')
 
 const UpdateBase = require('./base')
 
 class UpdateMoveDetailsController extends UpdateBase {
+  middlewareSetup() {
+    CreateMoveDetailsController.prototype.middlewareSetup.apply(this)
+    this.use(this.filterMoveTypes)
+  }
+
   getUpdateValues(req, res) {
     const move = req.getMove()
     if (!move) {
       return {}
     }
 
-    const values = pick(move, ['move_type'])
-    if (values.move_type) {
-      const toLocation = get(move, 'to_location.id')
-      if (toLocation) {
-        values[`to_location_${values.move_type}`] = toLocation
-      }
+    const values = pick(move, ['move_type', 'additional_information'])
+
+    const moveType = move.move_type
+    if (values.additional_information) {
+      values[`${moveType}_comments`] = values.additional_information
+    }
+    const toLocation = get(move, 'to_location.id')
+    if (toLocation) {
+      values[`to_location_${moveType}`] = toLocation
     }
 
     return values
   }
 
-  saveValues(req, res, next) {
-    this.saveMove(req, res, next)
+  filterMoveTypes(req, res, next) {
+    const move = req.getMove()
+    const moveType = move.move_type
+    const moveTypeField = get(req, 'form.options.fields.move_type')
+    moveTypeField.items = moveTypeField.items.filter(
+      item => item.value === moveType
+    )
+    next()
+  }
+
+  async saveValues(req, res, next) {
+    try {
+      const moveId = req.getMoveId()
+      const { values } = req.form
+      const move = req.getMove()
+      const moveType = move.move_type
+      const toLocation = values[`to_location_${moveType}`]
+      const additionalInformation = values[`${moveType}_comments`]
+
+      if (toLocation !== get(move, 'to_location.id')) {
+        const notes = req.t('moves::redirect_notes', req.session.user)
+
+        await moveService.redirect({
+          id: moveId,
+          notes,
+          to_location: {
+            id: toLocation,
+          },
+        })
+      } else if (additionalInformation !== move.additional_information) {
+        await moveService.update({
+          id: moveId,
+          additional_information: additionalInformation,
+        })
+      }
+
+      next()
+    } catch (err) {
+      next(err)
+    }
   }
 }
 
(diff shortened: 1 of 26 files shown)


common/lib/api-client/middleware/post.js
```diff
@@ -11,9 +11,9 @@ module.exports = function post(maxFileSize) {
       })
 
       if (req.method === 'POST' && req.data instanceof FormData) {
-        payload.req.maxContentLength = maxFileSize
-        payload.req.maxBodyLength = maxFileSize
-        payload.req.headers = {
+        req.maxContentLength = maxFileSize
```
teneightfive, line 14: This doesn't have any knock-on side effects to the downloads, right?

common/lib/api-client/models.test.js
```diff
@@ -161,4 +161,25 @@ describe('API client models', function() {
       })
     })
   })
+
+  context('Move events model', function() {
```
teneightfive, line 165: Have a look at how previous models have been tested. There is a configuration object which will load the fixture and test against the model.

It would be good to keep a consistent test format for now. This was done with the vision to move the fixtures out to a shared contracts location so that we can properly test this relationship between the API and the frontend.

common/services/move.test.js
```diff
@@ -1,8 +1,15 @@
+const proxyquire = require('proxyquire')
+
 const apiClient = require('../lib/api-client')()
 
-const moveService = require('./move')
 const personService = require('./person')
 
+const moveService = proxyquire('./move', {
+  'date-fns': {
+    formatISO: sinon.stub().returns('#timestamp'),
```
teneightfive, line 9: We can probably not worry about mocking this method for the sake of not having to introduce proxyquire to this test.

app/move/controllers/update/move-details.js
```diff
@@ -1,29 +1,76 @@
 const { get, pick } = require('lodash')
 
+const moveService = require('../../../../common/services/move')
 const CreateMoveDetailsController = require('../create/move-details')
 
 const UpdateBase = require('./base')
 
 class UpdateMoveDetailsController extends UpdateBase {
+  middlewareLocals() {
```
teneightfive, line 9: This would probably be better as `middlewareSetup`. More consistent with where we've done this on other controllers and it's more related to setting up the fields than setting "locals" for the view layer.

locales/en/moves.json
```diff
@@ -219,5 +219,6 @@
         "message": "$t(moves::update_flash.supplier_message)"
       }
     }
-  }
+  },
+  "redirect_notes": "Requested by {{fullname}} (user_name:{{username}} / user_id:{{userId}})"
```
teneightfive, line 223: Is this something we play back to a user ever or just for our records?

I'm wondering if we _really_ need to send the person's username/user ID along with their name.

There will be a separate piece of work coming soon which will look at moving the auth to the API so that the API will know who does what.

And I wonder if we leave trying to send audit traces as part of "notes" until then and for now we just send the name which is the information we already have.

common/lib/api-client/middleware/post.js
solidgoldpig: It's the same object, innit. e2e tests all continue to work.

app/move/controllers/update/move-details.js
solidgoldpig: You're right. And that's what I did

```js
  middlewareSetup() {
    CreateMoveDetailsController.prototype.middlewareSetup()
    this.use(this.filterMoveTypes)
  }
```

which led to this error

```
/hmpps-book-secure-move-frontend/node_modules/hmpo-form-wizard/lib/controller/index.js:59
        if (!this.router) throw new Error('Cannot use middleware outside of middleware mixins');
                          ^

Error: Cannot use middleware outside of middleware mixins
    at CreateBaseController.use
```

but your comment made me revisit the path of least resistance I'd taken and realised that of course the create version's `middlewareSetup` method has a `super` in it, so the solution is to just apply the this

```js
  middlewareSetup() {
    CreateMoveDetailsController.prototype.middlewareSetup.apply(this)
    this.use(this.filterMoveTypes)
  }
```

Good call. Obviously now I have to rewrite the tests all over again ;)

solidgoldpig: Sorted

common/lib/api-client/middleware/post.js
teneightfive: >It's the same object, innit

👍 

>e2e tests all continue to work

Not sure they would pick up if this changed though as the fixture files are small and is this not related to when things go wrong with larger files?

locales/en/moves.json
solidgoldpig: Yes, the auth logging would indeed be better, but I've checked with Jason and he's happy with this.

common/services/move.test.js
solidgoldpig: That seems...  back to front.

solidgoldpig: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/495

common/lib/api-client/models.test.js
solidgoldpig: Sorted and added to the fixtures' readme

common/services/move.test.js
solidgoldpig: Added a test t ensure that `formatISO` is called with expected args

PR #496, opened 2020-05-07
[P4-1364] [P4-1362] Review a single request (approve/reject)

teneightfive:
## Proposed changes

This introduces the ability for a user to approve or reject a single request (a `proposed` move). It uses the form wizard to create a simple one step journey and creates extra service methods to interact with the API.

**NOTE**: This requires two changes to the API to be fully supported. One to extend the cancellation reasons to include `rejected` and another to remove date validation when sending a `PATCH` with `proposed` moves.

## Screenshots

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

### DIFF
--- a/common/lib/user.test.js
+++ b/common/lib/user.test.js
@@ -219,6 +219,20 @@ describe('User class', function() {
       })
     })
 
+    context('when user has ROLE_PECS_PMU', function() {
+      beforeEach(function() {
+        permissions = user.getPermissions(['ROLE_PECS_PMU'])
+      })
+
+      it('should contain correct permission', function() {
+        expect(permissions).to.deep.equal([
+          'allocations:view',
+          'allocation:create',
+          'move:review',
+        ])
+      })
+    })
+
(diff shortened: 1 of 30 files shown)


common/lib/user.test.js
```diff
@@ -219,6 +219,20 @@ describe('User class', function() {
       })
     })
 
+    context('when user has ROLE_PECS_PMU', function() {
+      beforeEach(function() {
+        permissions = user.getPermissions(['ROLE_PECS_PMU'])
+      })
+
+      it('should contain correct permission', function() {
+        expect(permissions).to.deep.equal([
```
solidgoldpig, line 228: Maybe it's an idea to make a helper function that sorts the arrays so that we don't have to worry about the order...

teneightfive: In the test or in the library?

I realised we were missing this test so have added it as part of this change.

PR #498, opened 2020-05-07
fix: Prevent update court link when move not to court

solidgoldpig:
### Checklist

### DIFF
--- a/app/move/controllers/update/court.js
+++ b/app/move/controllers/update/court.js
@@ -1,0 +1,24 @@
+const { get } = require('lodash')
+
+const AssessmentController = require('./assessment')
+
+class UpdateCourtController extends AssessmentController {
+  // TODO: replace this with a centralised more form-wizard-based protection
+  configure(req, res, next) {
+    const isCourt =
+      get(res, 'locals.move.to_location.location_type') === 'court'
+
+    if (!isCourt) {
+      const error = new Error(
+        'Updating court information is not possible for this move'
+      )
+      error.statusCode = 404
+
+      return next(error)
+    }
+
+    super.configure(req, res, next)
+  }
+}
+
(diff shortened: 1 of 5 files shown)


app/move/views/_includes/assessment.njk
```diff
@@ -92,7 +92,7 @@
         }) }}
       {% endif %}
     {% endif %}
-  {% endif %}
 
-  {{ updateLink(updateLinks.court) }}
+    {{ updateLink(updateLinks.court) }}
+  {% endif %}
 </section>
```
teneightfive, line 98: I think it might be better to move this whole `section` within that block. Otherwise in some instances we'd be left with just an empty `<section>` tab on the page.

app/move/controllers/update/court.js
```diff
@@ -0,0 +1,24 @@
...
+  // TODO: replace this with a centralised more form-wizard-based protection
+  configure(req, res, next) {
+    const isCourt =
+      get(res, 'locals.move.to_location.location_type') === 'court'
+
+    if (!isCourt) {
+      const error = new Error(
+        'Forbidden. Updating court information is not allowed for this move'
+      )
+      error.statusCode = 403
```
teneightfive, line 15: I wonder if this should `404` instead. We're saying that this information doesn't exist for this move type so maybe a `404` is a better error code?

Or we do what the form wizard does and just redirect past it.

app/move/views/_includes/assessment.njk
solidgoldpig: 👍

app/move/controllers/update/court.js
solidgoldpig: Now `404`s

PR #499, opened 2020-05-07
[P4 1287] Single allocation view

barbaracassani:
The links cancel and remove from allocations are dummies; so is the allocation history tab.

It has been tried only with fake data, which may explain why there is no picture and no meta data returned from the backend.

With empty slots:
Screenshot 2020-05-12 at 14 42 18
With filled slots:
Screenshot 2020-05-12 at 14 42 29

### Checklist

### DIFF
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -1,0 +1,25 @@
+const presenters = require('../../../common/presenters')
+
+module.exports = function view(req, res) {
+  const { allocation } = res.locals
+  const { moves } = allocation
+  const locals = {
+    dashboardUrl: '/allocations',
+    allocationDetails: presenters.allocationToMetaListComponent(allocation),
+    allocationSummary: presenters.allocationToSummaryListComponent(allocation),
+    allocationPeople: {
+      emptySlots: moves.filter(move => !move.person).length,
+      filledSlots: moves
+        .filter(move => move.person)
+        .map(move => {
+          const { person } = move
+          if (!person.fullname) {
+            person.fullname = `${person.first_names} ${person.last_name}`
+          }
+          return person
+        })
+        .map(presenters.personToCardComponent()),
+    },
+  }
+  res.render('allocation/views/view', locals)
+}
--- a/app/allocation/controllers/view.test.js
+++ b/app/allocation/controllers/view.test.js
@@ -1,0 +1,104 @@
+const presenters = require('../../../common/presenters')
+
+const handler = require('./view')
+
+describe('view allocation', function() {
+  let locals
+  let render
+  let output
+  beforeEach(function() {
+    sinon.stub(presenters, 'allocationToMetaListComponent').returns({})
+    sinon.stub(presenters, 'allocationToSummaryListComponent').returns({})
+    render = sinon.stub()
+    locals = {
+      allocation: {
+        id: '06464fbd-b78a-4e47-b65c-8ab3c3867196',
+        type: 'allocations',
+        moves_count: 22,
+        date: '2020-05-06',
+        prisoner_category: 'c',
+        sentence_length: 'long',
+        complex_cases: [
+          {
+            title: 'Segregated prisoners',
+            answer: true,
+          },
+          {
+            title: 'Self harm / prisoners on ACCT',
+            answer: true,
+          },
+          {
+            title: 'Mental health issues',
+            answer: false,
+          },
+          {
+            title: 'Integrated Drug Treatment System',
+            answer: false,
+          },
+        ],
+        complete_in_full: false,
+        other_criteria: 'no',
+        created_at: '2020-05-01T10:44:00+01:00',
+        updated_at: '2020-05-01T10:44:00+01:00',
+        from_location: {
+          title: 'BERWYN (HMP)',
+          location_type: 'prison',
+        },
+        to_location: {
+          title: 'GARTREE (HMP)',
+          location_type: 'prison',
+        },
+        moves: [
+          {},
+          {},
+          {
+            person: {
+              first_names: 'John',
+              last_name: 'Doe',
+            },
+          },
+        ],
+      },
+    }
+    handler(
+      {},
+      {
+        locals,
+        render,
+      }
+    )
+    output = render.firstCall.lastArg
+  })
+  it('creates allocationDetails on locals with the result of the correct presenter', function() {
+    expect(output.allocationDetails).to.exist
+    expect(
+      presenters.allocationToMetaListComponent
+    ).to.have.been.calledOnceWithExactly(locals.allocation)
+  })
+  it('creates allocationSummary on locals with the result of the correct presenter', function() {
+    expect(output.allocationSummary).to.exist
+    expect(
+      presenters.allocationToSummaryListComponent
+    ).to.have.been.calledOnceWithExactly(locals.allocation)
+  })
+  it('calls render with the template', function() {
+    expect(render).to.have.been.calledWithExactly('allocation/views/view', {
+      allocationDetails: {},
+      allocationSummary: {},
+      allocationPeople: {
+        emptySlots: 2,
+        filledSlots: [
+          {
+            href: undefined,
+            image_alt: 'JOHN DOE',
+            image_path: undefined,
+            meta: { items: [] },
+            tags: { items: [] },
+            title: { text: 'JOHN DOE' },
+          },
+        ],
+      },
+      dashboardUrl: '/allocations',
+    })
+  })
+})
--- a/app/allocation/index.js
+++ b/app/allocation/index.js
@@ -5,6 +5,7 @@ const FormWizardController = require('../../common/controllers/form-wizard')
 const { protectRoute } = require('../../common/middleware/permissions')
 
 const confirmation = require('./controllers/create/confirmation')
+const view = require('./controllers/view')
 const { createFields } = require('./fields')
 const { setAllocation } = require('./middleware')
 const { createSteps } = require('./steps')
@@ -33,7 +34,7 @@ router.get(
   protectRoute('allocation:create'),
   confirmation
 )
-
+router.get('/:allocationId', protectRoute('allocations:view'), view)
 // Export
 module.exports = {
   router,
--- a/app/allocation/views/view.njk
+++ b/app/allocation/views/view.njk
@@ -1,0 +1,93 @@
+{% extends "layouts/two-column.njk" %}
+
+{% set additionalContainerClass = "sticky-sidebar-container" %}
+
+{% block pageTitle %}
+  {{ t("allocations::view.page_title") }}
+{% endblock %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    text: t("actions::back_to_dashboard"),
+    classes: "app-print--hide",
+    href: dashboardUrl
+  }) }}
+
+  <a href="javascript:window.print()" class="app-!-float-right app-print--hide govuk-!-margin-top-3 app-icon app-icon--print">
+    {{ t("actions::print_move_detail") }}
+  </a>
+
+  {{ super() }}
+{% endblock %}
+
+{% block contentHeader %}
+  <header class="govuk-!-margin-bottom-8 govuk-!-padding-left-0 govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-1">
+      {{ t("allocations::view.heading", {
+        count: allocation.moves_count,
+        from_location: allocation.from_location.title,
+        to_location: allocation.to_location.title,
+        date: allocation.date | formatDate
+      })}}
+    </h1>
+  </header>
+{% endblock %}
+
+{% block contentMain %}
+  <section>
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-1">{{ t("allocations::view.criteria_for_allocation") }}</h2>
+    <span class="govuk-caption-m govuk-!-margin-bottom-4">
+      {{ t("allocations::view.created_at") }}
+      {{ allocation.created_at | formatDateAsRelativeDay }}
+      {{ allocation.created_at | formatTime }}
+    </span>
+
+    {{ govukSummaryList(allocationDetails) }}
+
+    {% if allocation.complete_in_full and allocationPeople.emptySlots %}
+      {{ govukWarningText({
+        text: t("allocations::view.complete_in_full"),
+        iconFallbackText: "Warning"
+      }) }}
+    {% endif %}
+  </section>
+
+  <section class="govuk-!-margin-top-8">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-1">{{ t("allocations::view.people.heading") }}</h2>
+    <h3 class="govuk-heading-s">{{ t("allocations::view.people.slots_filled", {
+      slots_filled: allocationPeople.filledSlots.length,
+      total_slots: allocation.moves_count,
+      person_or_people: t('person', {
+        count: allocation.moves_count
+      })
+    }) }}</h3>
+  </section>
+
+  {% for i in range(0, allocationPeople.emptySlots) -%}
+    <div class="app-border-top-1">
+      {{ appMessage({
+        classes: "app-message--muted govuk-!-margin-top-3 govuk-!-margin-bottom-3 govuk-!-padding-right-8",
+        allowDismiss: false,
+        content: {
+          html: t("allocations::view.people.empty_slot")
+        }
+      }) }}
+    </div>
+  {%- endfor %}
+
+  {% for slot in allocationPeople.filledSlots %}
+    {{ appCard(slot) }}
+  {% endfor %}
+
+{% endblock %}
+
+{% block contentSidebar %}
+  <div class="sticky-sidebar">
+    <div class="sticky-sidebar__inner">
+      <h3 class="govuk-heading-m app-border-top-2 app-border--blue govuk-!-padding-top-4 govuk-!-margin-bottom-3">
+        {{ t("allocations::view.allocation_details") }}
+      </h3>
+      {{ appMetaList(allocationSummary) }}
+    </div>
+  </div>
+{% endblock %}
--- a/common/presenters/allocation-to-summary-list-component.js
+++ b/common/presenters/allocation-to-summary-list-component.js
@@ -1,0 +1,40 @@
+const i18n = require('../../config/i18n')
+const filters = require('../../config/nunjucks/filters')
+
+function allocationToSummaryListComponent(allocation) {
+  const {
+    moves,
+    from_location: fromLocation,
+    to_location: toLocation,
+    date,
+  } = allocation
+  return {
+    items: [
+      {
+        key: {
+          text: i18n.t('allocations::view.sidebar.number_of_prisoners'),
+        },
+        value: { text: moves.length },
+      },
+      {
+        key: {
+          text: i18n.t('allocations::view.sidebar.move_from'),
+        },
+        value: { text: fromLocation.title },
+      },
+      {
+        key: {
+          text: i18n.t('allocations::view.sidebar.move_to'),
+        },
+        value: { text: toLocation.title },
+      },
+      {
+        key: {
+          text: i18n.t('allocations::view.sidebar.date'),
+        },
+        value: { text: filters.formatDateAsRelativeDay(date) },
+      },
+    ],
+  }
+}
+module.exports = allocationToSummaryListComponent
--- a/locales/en/actions.json
+++ b/locales/en/actions.json
@@ -16,6 +16,7 @@
   "continue": "Continue",
   "restart": "Restart",
   "cancel": "Cancel",
+  "cancel_allocation": "Cancel allocation",
   "cancel_move": "Cancel this move",
   "cancel_move_confirmation": "Cancel move",
   "download": "Download moves",
--- a/locales/en/allocations.json
+++ b/locales/en/allocations.json
@@ -38,10 +38,7 @@
       "none_provided": "None provided",
       "other_criteria": "Other criteria",
       "sentence_length": {
-        "label": "Time left to serve",
-        "value": "Any length of sentence",
-        "value_short": "16 months or less to serve",
-        "value_long": "16 months or longer to serve"
+        "label": "Time left to serve"
       }
     },
     "complete_in_full": "This allocation should be filled in full - consider prisoners that are the next best for this allocation",
--- a/locales/en/allocations.json
+++ b/locales/en/allocations.json
@@ -24,5 +24,39 @@
   },
   "allocation_criteria": {
     "page_title": "Allocation criteria"
+  },
+  "view": {
+    "page_title": "View allocation",
+    "heading": "{{count}} person from {{from_location}} to {{to_location}}, {{date}}",
+    "heading_plural": "{{count}} people from {{from_location}} to {{to_location}}, {{date}}",
+    "criteria_for_allocation": "Criteria for allocation",
+    "created_at": "Created {{datetime}}",
+    "fields": {
+      "estate_type": "Estate type",
+      "category": "Category",
+      "complex_cases": "Complex cases for prisons to agree",
+      "none_provided": "None provided",
+      "other_criteria": "Other criteria",
+      "sentence_length": {
+        "label": "Time left to serve",
+        "value": "Any length of sentence",
+        "value_short": "16 months or less to serve",
+        "value_long": "16 months or longer to serve"
+      }
+    },
+    "complete_in_full": "This allocation should be filled in full - consider prisoners that are the next best for this allocation",
+    "people": {
+      "slots_filled": "{{slots_filled}} of {{total_slots}} {{person_or_people}} allocated",
+      "heading": "People for allocation",
+      "empty_slot": "Person to be allocated"
+    },
+    "cancel_allocation": "Cancel allocation",
+    "allocation_details": "Allocation details",
+    "sidebar": {
+      "number_of_prisoners": "Number of people",
+      "move_from": "Move from",
+      "move_to": "Move to",
+      "date": "Date"
+    }
   }
 }
--- a/locales/en/fields.json
+++ b/locales/en/fields.json
@@ -260,9 +260,9 @@
   "sentence_length": {
     "label": "Sentence length",
     "items": {
-      "long": "Over 16 months only",
-      "short": "Under 16 months only",
-      "any": "Any sentence"
+      "length_long": "Over 16 months",
+      "length_short": "16 months or less",
+      "length": "Any time to serve"
     }
   },
   "complete_in_full": {
--- a/app/allocation/fields/sentence-length.js
+++ b/app/allocation/fields/sentence-length.js
@@ -14,15 +14,15 @@ const sentenceLength = {
   items: [
     {
       value: 'null',
-      text: 'fields::sentence_length.items.any',
+      text: 'fields::sentence_length.items.length',
     },
     {
       value: 'short',
-      text: 'fields::sentence_length.items.short',
+      text: 'fields::sentence_length.items.length_short',
     },
     {
       value: 'long',
-      text: 'fields::sentence_length.items.long',
+      text: 'fields::sentence_length.items.length_long',
     },
   ],
 }
--- a/common/lib/api-client/models.js
+++ b/common/lib/api-client/models.js
@@ -218,6 +218,10 @@ module.exports = {
       other_criteria: '',
       created_at: '',
       updated_at: '',
+      moves: {
+        jsonApi: 'hasMany',
+        type: 'moves',
+      },
     },
   },
   allocation_complex_case: {
--- a/common/presenters/allocation-to-meta-list-component.js
+++ b/common/presenters/allocation-to-meta-list-component.js
@@ -1,0 +1,69 @@
+const i18n = require('../../config/i18n')
+const filters = require('../../config/nunjucks/filters')
+
+function allocationToMetaListComponent(allocation) {
+  const {
+    to_location: toLocation,
+    prisoner_category: prisonerCategory,
+    complex_cases: complexCases,
+    other_criteria: otherCriteria,
+    sentence_length: sentenceLength,
+  } = allocation
+  return {
+    rows: [
+      {
+        key: {
+          text: i18n.t('allocations::view.fields.estate_type'),
+        },
+        value: {
+          // TODO replace with region when we have it
+          text: filters.startCase(toLocation.location_type),
+        },
+      },
+      {
+        key: {
+          text: i18n.t('allocations::view.fields.category'),
+        },
+        value: {
+          text: filters.startCase(prisonerCategory),
+        },
+      },
+      {
+        key: {
+          text: i18n.t('allocations::view.fields.sentence_length.label'),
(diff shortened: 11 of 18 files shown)


app/allocation/views/view.njk
```diff
@@ -0,0 +1,89 @@
...
+    items: [
+      {
+        label: t("allocations::view.allocation_tab"),
+        id: "allocation-tab",
+        panel: {
+          html: allocationTab
+        }
+      },
+      {
+        label: t("allocations::view.allocation_history_tab"),
```
teneightfive, line 69: I'm not sure any of this is actually in scope initially?

app/allocation/views/view.njk
```diff
@@ -0,0 +1,119 @@
...
+            to_location: allocation.to_location.title,
+            date: allocation.date | formatDate
+          }) | safe }}
+    </h1>
+  </header>
+{% endblock %}
+
+{% block contentMain %}
+  {% set allocationTab %}
+    <section class="app-!-position-relative">
```
solidgoldpig, line 39: Do you need the `app-!-position-relative`s ?

common/assets/scss/application.scss
```diff
@@ -27,6 +27,7 @@ $govuk-font-family: "Inter", system-ui, sans-serif;
 @import "govuk-frontend/govuk/components/skip-link/skip-link.scss";
 @import "govuk-frontend/govuk/components/summary-list/summary-list.scss";
 @import "govuk-frontend/govuk/components/table/table.scss";
+@import "govuk-frontend/govuk/components/tabs/tabs.scss";
```
teneightfive, line 30: We can remove the need for these at this stage. This change should be MVP of an allocation.

locales/en/allocations.json
```diff
@@ -24,5 +24,40 @@
...
     "page_title": "Allocation criteria"
+  },
+  "view": {
+    "page_title": "View allocation",
+    "heading": "{{moves_count}} {{person_or_people}} from {{from_location}} to {{to_location}}, {{date}}",
+    "criteria_for_allocation": "Criteria for allocation",
+    "created_at": "Created {{datetime}}",
+    "allocation_tab": "Allocation",
+    "allocation_history_tab": "Allocation history",
+    "fields": {
```
teneightfive, line 35: Could we re-use the labels from the fields translation file?

app/allocation/controllers/view.js
```diff
@@ -0,0 +1,43 @@
...
+      other_criteria: otherCriteria,
+      sentence_length: sentenceLength,
+      to_location: toLocation,
+      moves,
+    },
+  } = res.locals
+  const locals = {
+    dashboardUrl: '/allocations',
+    allocationDetails: presenters.allocationToMetaListComponent({
+      t: req.t,
```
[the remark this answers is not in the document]
barbaracassani: I am happy to do so, but this is easier to mock in tests.

app/allocation/views/view.njk
```diff
@@ -0,0 +1,119 @@
...
+    {% endfor %}
+
+    <p class="govuk-!-margin-top-9 govuk-!-margin-bottom-0">
+      <a href="{{ allocation.id }}/cancel" class="app-link--destructive">
+        {{ t("actions::cancel_allocation") }}
+      </a>
+    </p>
+  {% endset -%}
+
+  {{ govukTabs({
```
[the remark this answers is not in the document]
barbaracassani: Ok shall remove.

app/allocation/index.js
```diff
@@ -33,7 +34,7 @@ router.get(
   protectRoute('allocation:create'),
   confirmation
 )
-
+router.get('/:allocationId/', protectRoute('allocations:view'), view)
```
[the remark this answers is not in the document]
barbaracassani: Do we need to have both `allocations:view` (plural, pre-existing) and `allocation:view` (singular) ?

app/allocation/controllers/view.js
```diff
@@ -0,0 +1,43 @@
...
+      toLocation,
+      date,
+    }),
+    allocationPeople: presenters.allocationPeopleToCards({
+      movesCount,
+      moves,
+    }),
+  }
+  res.render('allocation/views/view', {
+    ...res.locals,
```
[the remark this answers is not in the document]
barbaracassani: Well, for it to exist in the view I need to pass it to `render`, and I need to mix the new properties in. So yes, I do need it -- unless I am misunderstanding your comment.

app/allocation/controllers/view.js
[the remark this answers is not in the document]
teneightfive: Surely it's just as easy using `sinon.stub`?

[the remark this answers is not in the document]
teneightfive: Can you not use `sinon.stub` the same way just mocking the imported library?

I think it's cleaner to not have to pass a translation library into the presenter. And is more consistent with how we have done it in other presenters.

app/allocation/index.js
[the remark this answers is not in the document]
teneightfive: Maybe we can just use the same permissions for this part of the app

app/allocation/controllers/view.js
```diff
@@ -0,0 +1,43 @@
...
+      sentenceLength,
+    }),
+    allocationSummary: presenters.allocationToSummary({
+      t: req.t,
+      movesCount,
+      fromLocation,
+      toLocation,
+      date,
+    }),
+    allocationPeople: presenters.allocationPeopleToCards({
```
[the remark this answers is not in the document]
barbaracassani: I am reusing `personToCardComponent` which is called with `.map` on this intermediate presenter, which also creates the empty slots.

locales/en/allocations.json
barbaracassani: The wording is different -- I believe, intentionally.

app/allocation/views/view.njk
```diff
@@ -0,0 +1,119 @@
+{% extends "layouts/two-column.njk" %}
+
+{% set additionalContainerClass = "sticky-sidebar-container" %}
+
+{% block pageTitle %}
+  {{ t("allocations::view.page_title") }}
```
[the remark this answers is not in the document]
barbaracassani: I did consider it, but it's difficult to display something that it's both meaningful and short. Allocations are not necessarily associated to people, and locations risk to make it verbose.

app/allocation/views/view.njk
barbaracassani: No! Well spotted :-)

app/allocation/controllers/view.js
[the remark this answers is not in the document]
teneightfive: I think you are. Unless you are overriding properties from `res.locals`, which probably isn't a good idea here anyway, you don't need to include it in the second argument to `res.render`.

Anything previously assigned to `res.locals` will already be available as a variable within the view layer. You can just pass `locals` as the second argument to the render method:

```javascript
res.render('allocation/views/view', locals)
```

app/allocation/controllers/view.js
```diff
@@ -0,0 +1,38 @@
...
+      moves_count: movesCount,
+      other_criteria: otherCriteria,
+      sentence_length: sentenceLength,
+      to_location: toLocation,
+      moves,
+    },
+  } = res.locals
+  const locals = {
+    dashboardUrl: '/allocations',
+    allocationDetails: presenters.allocationToMetaListComponent({
```
teneightfive, line 19: In this instance you can just pass the allocation in, save you having to destructure it and re-build it in the argument.

The nice thing about destructing in the presenter is it will then ignore any properties it doesn't need.

It will allow you to reduce the size of this controller a lot too.

app/allocation/controllers/view.test.js
```diff
@@ -0,0 +1,105 @@
+const presenters = require('../../../common/presenters')
+const filters = require('../../../config/nunjucks/filters')
+
+const handler = require('./view')
+
+describe('view allocation', function() {
+  let locals
+  let t
```
teneightfive, line 8: Is this still needed now we don't pass it in?

app/allocation/controllers/view.test.js
```diff
@@ -0,0 +1,105 @@
...
+  let render
+  let output
+  beforeEach(function() {
+    sinon.stub(filters, 'formatDate').returns('__date__')
+    sinon.stub(filters, 'formatTime').returns('__time__')
+    sinon.stub(filters, 'formatDateAsRelativeDay').returns('__relative_date__')
```
teneightfive, line 14: The filters don't appear to be part of the view controller, do we need to import and stub these methods?

app/allocation/views/view.njk
```diff
@@ -0,0 +1,97 @@
...
+  </a>
+
+  {{ super() }}
+{% endblock %}
+
+{% block contentHeader %}
+  <header class="govuk-!-margin-bottom-8 govuk-!-padding-left-0 govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-1">
+      {{ t("allocations::view.heading", {
+        person_or_people: "person" | pluralize(allocation.resultCount),
```
teneightfive, line 27: It might be worth looking at i18next [in-built pluralization](https://www.i18next.com/translation-function/plurals) to handle this.

app/allocation/views/view.njk
```diff
@@ -0,0 +1,97 @@
...
+        moves_count: allocation.moves_count,
+        from_location: allocation.from_location.title,
+        to_location: allocation.to_location.title,
+        date: allocation.date | formatDate
+      }) | safe }}
```
teneightfive, line 32: Does this need to be unescaped using `safe`? Doesn't look like there's any HTML in the translation

app/allocation/views/view.njk
```diff
@@ -0,0 +1,97 @@
...
+      }) }}
+    {% endif %}
+  </section>
+
+  <section class="govuk-!-margin-top-8">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-1">{{ t("allocations::view.people.heading") }}</h2>
+    <h3 class="govuk-heading-s">{{ t("allocations::view.people.slots_filled", {
+      slots_filled: allocationPeople.filledSlots.length,
+      total_slots: allocation.moves_count,
+      person_or_people: "person" | pluralize(allocation.moves_count)
```
teneightfive, line 61: Again, could also look at i18next's [in-built pluralization](https://www.i18next.com/translation-function/plurals).

app/allocation/views/view.njk
```diff
@@ -0,0 +1,97 @@
...
+        }
+      }) }}
+    </div>
+  {%- endfor %}
+
+  {% for slot in allocationPeople.filledSlots %}
+    {{ appCard(slot) }}
+  {% endfor %}
+
+  <p class="govuk-!-margin-top-9 govuk-!-margin-bottom-0">
```
teneightfive, line 81: Might be worth leaving this out until it's actually working and hooked up to a journey. Otherwise it's just a `404` at this stage?

common/assets/scss/application.scss
teneightfive: This can be removed for now. I noticed bringing in lots of the GOV.UK Design System is bloating the css file so we should do it sparingly.

locales/en/allocations.json
```diff
@@ -24,5 +24,40 @@
   },
   "allocation_criteria": {
     "page_title": "Allocation criteria"
+  },
+  "view": {
+    "page_title": "View allocation",
+    "heading": "{{moves_count}} {{person_or_people}} from {{from_location}} to {{to_location}}, {{date}}",
+    "criteria_for_allocation": "Criteria for allocation",
+    "created_at": "Created {{datetime}}",
+    "allocation_tab": "Allocation",
```
teneightfive, line 33: Maybe we can tidy up some of the keys that are no longer needed like these ones?

locales/en/allocations.json
teneightfive: I would check with @Andy-Millar but I feel like it might be better to have consistency in labelling between capture and replay if we can.

locales/en/allocations.json
```diff
@@ -24,5 +24,40 @@
...
+        "value": "Any length of sentence",
+        "value_short": "16 months or less to serve",
+        "value_long": "16 months or longer to serve"
+      }
+    },
+    "complete_in_full": "This allocation should be filled in full - consider prisoners that are the next best for this allocation",
+    "people": {
+      "slots_filled": "{{slots_filled}} of {{total_slots}} {{person_or_people}} allocated",
+      "heading": "People for allocation",
+      "empty_slot": "Prisoner to be allocated"
```
teneightfive, line 52: We're using a combination of `People` and `Prisoner` between `heading` and `empty_slot`. We should probably stick to one term.

locales/en/allocations.json
```diff
@@ -24,5 +24,40 @@
...
+    },
+    "cancel_allocation": "Cancel allocation",
+    "allocation_details": "Allocation details",
+    "sidebar": {
+      "number_of_prisoners": "Number of prisoners",
```
teneightfive, line 57: Same here. We should use one of `People` or `Prisoners` consistently.

locales/en/allocations.json
Andy-Millar: I've gone backwards and forwards on this..

Allocations are ONLY for prisoners (not detainees). 

Problem with people is it is a bit ambiguous what is meant is meant by people - ie. is it prisoners.... But it's more humanising than the term prisoners.

Problem with prisoners is it's too long and causes problems in things like titles. But it provides far more clarity.

I've deliberately added the term prisoner in detail, and kept the term people in the title to try and get the best of both worlds. ie. Clearer when we are asking specific questions (prisoner), shorter and more humanising in titles (people). 

If we go for one, it should be people, but I'm concerned it might cause a bit of confusion. We do refer to prisoners as people elsewhere in other journeys though.

locales/en/allocations.json
barbaracassani: @Andy-Millar we have two sets of wordings for the sentence length. In the radio buttons, we have
```
      "long": "Over 16 months only",
      "short": "Under 16 months only",
      "any": "Any sentence"
```
Whilst in the detail view of the allocation, we have:
```
        "any": "Any length of sentence",
        "short": "16 months or less to serve",
        "long": "16 months or longer to serve"
```
Would it be possible to stick with one set for both?

locales/en/allocations.json
barbaracassani: Ok -- so what is my action here? Is change needed?

locales/en/allocations.json
Andy-Millar: @barbaracassani 

Worth noting this copy was updated a while ago in the clickable prototype - I'm guessing you built this from a screenshot which is out of date. Going forward we have been working on links to prototype rather than screenshot, so hopefully this will happen less and less going forward. 

Here's the page in prototype: https://invis.io/YET7ANHWV9K#/412212287_IPTC-MVP-03 

Copy in prototype for PMU creating a move is:

- "any": "Any time to serve"
- "short": "16 months or less",
- "long": "Over 16 months",

For the allocation page we should have the same - though the invision prototype does have the words "to serve" at the end, we can remove them - just to clarify it should be:

- "any": "Any time to serve"
- "short": "16 months or less",
- "long": "Over 16 months"

locales/en/allocations.json
Andy-Millar: Just go for people everywhere. But if there is problems, may need to change later

locales/en/allocations.json
barbaracassani: Ok, thank you for the clarification @Andy-Millar -- shall amend.

locales/en/allocations.json
barbaracassani: 👍

app/allocation/controllers/view.js
```diff
@@ -0,0 +1,18 @@
+const presenters = require('../../../common/presenters')
+
+module.exports = function view(req, res) {
+  const { allocation } = res.locals
+  const {
+    allocation: { moves_count: movesCount, moves },
+  } = res.locals
```
teneightfive, line 7: It looks like `res.locals` is being destructured twice, could them be combined?

common/presenters/allocation-people-to-cards.js
```diff
@@ -0,0 +1,22 @@
+const personToCardComponent = require('./person-to-card-component')
+
+function allocationPeopleToCards({ movesCount, moves = [] }) {
+  const emptySlots = movesCount - moves.length
+  const mapper = personToCardComponent({
+    showImage: true,
```
teneightfive, line 6: These are actually the defaults so they don't need to be passed here.

common/presenters/allocation-people-to-cards.js
```diff
@@ -0,0 +1,22 @@
...
+    showMeta: true,
+    showTags: true,
+  })
+  const filledSlots = moves.map(move => {
+    const { person } = move
+    if (!person.fullname) {
+      person.fullname = `${person.first_names} ${person.last_name}`
```
teneightfive, line 13: Maybe have a look at how moves are formatted in the moves service. This is done as part of the response on a move to keep it in one place rather than have to do this in the presenter.

common/presenters/allocation-to-summary-list-component.js
```diff
@@ -0,0 +1,40 @@
+const i18n = require('../../config/i18n')
+const filters = require('../../config/nunjucks/filters')
+
+function allocationToSummaryListComponent(allocation) {
+  const {
+    moves_count: movesCount,
```
teneightfive, line 6: I think we should be careful using this. Abeer and I discussed removing this as it's not really necessary. The information can be inferred by the number of moves on an allocation.

We were also discussing passing a "count" as part of the request rather than an attribute within an actual allocation. So to future proof that we should use `allocation.moves.length` instead.

locales/en/actions.json
```diff
@@ -16,6 +16,7 @@
   "continue": "Continue",
   "restart": "Restart",
   "cancel": "Cancel",
+  "cancel_allocation": "Cancel allocation",
```
teneightfive, line 19: `Cancel this allocation`? Remains consistent with the copy used below.

locales/en/allocations.json
```diff
@@ -24,5 +24,36 @@
...
+    "criteria_for_allocation": "Criteria for allocation",
+    "created_at": "Created {{datetime}}",
+    "fields": {
+      "estate_type": "Estate type",
+      "category": "Category",
+      "complex_cases": "Complex cases for prisons to agree",
+      "none_provided": "None provided",
+      "other_criteria": "Other criteria",
+      "sentence_length": {
+        "label": "Time left to serve"
```
teneightfive, line 41: I think we still need to address the difference here. 

The label in the form journey is `Sentence length` but in the display we are saying `Time left to serve`. I think @Andy-Millar wanted to move more to this language, but either way we should be consistent with the label in the data capture and the replay of the data.

locales/en/allocations.json
```diff
@@ -24,5 +24,36 @@
...
+      }
+    },
+    "complete_in_full": "This allocation should be filled in full - consider prisoners that are the next best for this allocation",
+    "people": {
+      "slots_filled": "{{slots_filled}} of {{total_slots}} {{person_or_people}} allocated",
+      "heading": "People for allocation",
+      "empty_slot": "People to be allocated"
+    },
+    "cancel_allocation": "Cancel allocation",
```
teneightfive, line 50: This same text seems to also be in `actions.json`. Do we need both?

locales/en/allocations.json
```diff
@@ -24,5 +24,36 @@
...
+    "allocation_details": "Allocation details",
+    "sidebar": {
+      "number_of_prisoners": "Number of people",
+      "move_from": "Move from",
```
teneightfive, line 54: This content already exists within the existing move journeys. We should try and re-use those if possible to keep consistency between the journeys.

locales/en/allocations.json
```diff
@@ -24,5 +24,36 @@
...
+      "move_to": "Move to",
+      "date": "Date"
```
teneightfive, line 56: Same here. This label already exists within moves. We should re-use where possible for consistency.

locales/en/allocations.json
Andy-Millar: @barbaracassani Worth noting the empty slot should be "Person to be allocated" not people

locales/en/allocations.json
Andy-Millar: It's not sentence length - we originally misunderstood and have updated to 'Time left to serve' for a good while now

locales/en/actions.json
Andy-Millar: Worth noting @teneightfive, on the cancel page the red button is 'Cancel move'. Probably should also be consistent in that case.

Or the text link should set the expectation that you are not cancelling by clicking the link (ie. "I need to cancel this move / allocation"

Worth catching up to talk through.

locales/en/allocations.json
barbaracassani: Sentence length is the name of the field described by the label. I think it's best to stick to that for the field names, as copy may change over time.

locales/en/allocations.json
barbaracassani: So, in the past, when I recycled labels you told me not to because the copy might diverge over time. Do we recycle or not?

locales/en/allocations.json
barbaracassani: See above. It's not clear to me if our policy is to reuse or to err on the side of caution. Personally I'd rather not recycle as it could lead to unintended consequences down the line, if for example the copy is radically changed for one journey in a way that doesn't fit the other.

common/presenters/allocation-to-summary-list-component.js
barbaracassani: Wait -- `moves_count` pre-exists to the allocations. How I understand it, is that `moves_count` represents the number of slots available, but the moves associated are only the already filled slots. Otherwise, how could we display the empty slots?
@cwrw can you confirm this?

app/allocation/controllers/view.js
barbaracassani: As far as I am aware, you can't destructure an object at several levels in one pass.

locales/en/allocations.json
Andy-Millar: No - because it's not sentence length - it's 'Time left to serve' (or sentence remaining). Sentence length and sentence remaining are completely different. Label and field needs to be 'Time left to serve' in all instances - allocation page and when PMU create an allocation

app/allocation/controllers/view.js
barbaracassani: I have replaced the deep destructuring now.

locales/en/allocations.json
Andy-Millar: I've marked this as unresolved as the label needs to be updated

common/presenters/allocation-to-summary-list-component.js
teneightfive: My understanding was the moves are always there, as we need to know them to be able to "fill" them at the later stages.

My understanding was that we would return the number of moves, and we can determine "unfilled" on the frontend as those that don't contain a `person` relationship. The `moves_count` is a superfluous piece of information that we can already determine.

locales/en/actions.json
teneightfive: Yep. I think we use a slight difference between button and link text. One to suggest you might go somewhere, one to suggest you'll do the final action.

common/presenters/allocation-people-to-cards.js
barbaracassani: The sole reason why this line of code exist, it's to adapt the object to the existing presenter. The `Person` service adds `fullname` to the person object, but in this case we return the data as supplied by the nested object in the api. If I do it in the service, I have to add a deep parsing of the results to add an arbitrary property that isn't supplied by the api and strictly speaking wouldn't be needed if not for this particular presenter. I am not sure it would be an improvement. If anything, we ought to remove `fullname` from usage in the presenter.

locales/en/actions.json
barbaracassani: Hmm -- ok, so shall I change it it "I need to cancel this allocation"?

Andy-Millar: Hold off for a bit on this one @barbaracassani - need to think it through as it affects other journeys.

@teneightfive Cancel this XXX and Cancel XXX doesn't do enough to suggest difference in my opinion. Let's catch up and chat through with other journeys

locales/en/allocations.json
teneightfive: @barbaracassani I think you have misunderstood the issue. The issue is not with the `name` of the field, that's irrelevant and should match to the API for ease in sending that data.

My comments were around the mis-match in content we are displaying in two parts of the app. The label during creating a new allocation and when we display the answer to that information in the allocation detail should be the same. Which was my original suggestion to use the translation keys from the fields to ensure they will remain consistent. This is what we do in the move detail and something I think we should do here.

It's our responsibility to understand the codebase and the problem and to ensure consistency and if there are slight discrepancies that we see mock-ups we should have a quick conversation around them with @Andy-Millar and get them resolved.

The main goal is to create a consistent and simple to use UI rather than build exactly from screens. Sometimes content can be updated between tickets/changes so we just need to keep on eye on things to keep them inline.

locales/en/allocations.json
teneightfive: That would have been a different case. In this instance the label is being used to refer to the same thing as the other dashboard. And in those cases we should remain consistent.

common/presenters/allocation-to-summary-list-component.js
barbaracassani: Discussed with @cwrw and we are keeping `moves_count` for the moment being. Eventually we will instead parse the moves and count which ones have a person assigned.

locales/en/allocations.json
teneightfive: Again, it depends on the context. In the case here where we want to create unity between how we refer to a from and to location I think it makes sense to re-use.

And also in the case of labels. Unless they are drastically different we should use the same labels when capturing the data and playing it back so a user is aware what it refers to.

Often it's a judgement call though, there isn't a one-size-fits-all policy for this.

barbaracassani: The lack of policy in general, is very disempowering. It seems to me that whatever choice I make, the other way will be argued for, which leads to non-uniformity and general waste of time / energies. I think that a blanket policy may leave some edge cases out, but should nonetheless be decided for, and exceptions evaluated if they are such, e.g. exceptions.

common/presenters/allocation-to-summary-list-component.js
barbaracassani: Discussed again with @cwrw and @teneightfive and we may as well change it now, since `moves_count` will be deprecated at some point. Shall remove it.

locales/en/allocations.json
barbaracassani: I have now updated the token `Sentence length` to read `Time left to serve`.

locales/en/allocations.json
teneightfive: The default policy has always been to re-use code and translation to ensure consistency. Especially when we are describing related resources. 

When we are talking about a different resource, it is better to use a separate translation key.

app/allocation/controllers/view.test.js
```diff
@@ -0,0 +1,83 @@
...
+        render,
+      }
+    )
+    output = render.firstCall.lastArg
+  })
+  it('creates allocationDetails on locals with the result of the correct presenter', function() {
+    expect(output.allocationDetails).to.exist
+    expect(
+      presenters.allocationToMetaListComponent
+    ).to.have.been.calledWithExactly(locals.allocation)
```
teneightfive, line 67: I can be useful to use `calledOnceWithExactly` in instances like this to ensure a method isn't called multiple times by mistake when changing code in the future.

app/allocation/index.js
teneightfive, line 37: This doesn't need the trailing slash. Trailing slashes haven't been defined on any other routes in the application.

app/allocation/views/view.njk
teneightfive: Following up on this - do we need to use `| safe`?

app/allocation/views/view.njk
teneightfive: Following up on this - maybe use the built-in pluralization support.

app/allocation/views/view.njk
teneightfive: Following up on this too - can we please remove this and add this as part of the work to introduce cancellation an allocation. 

Otherwise this just introduces a `404`.

There will be further work around support if a person has the permission to cancel too so it would be cleaner to group that work together.

common/presenters/allocation-people-to-cards.js
teneightfive: So the reason I suggested it is that it's not just `fullname` you'll likely need.

I think there are other things like `image` that get added when formatting a person as part of a move in the move service.

I think this page will also require the use of the person's image so you'll find yourself repeating that code here. And we add support for any other properties we will also have to update it here.

Ideally it would be good to have that done as part of devour but I don't think there are available hooks to do that.

common/presenters/allocation-people-to-cards.js
```diff
@@ -0,0 +1,26 @@
...
+  const emptySlots = slotsByFilledStatus[1].length
+  const filledSlots = slotsByFilledStatus[0].map(move => {
+    const { person } = move
+    if (!person.fullname) {
+      person.fullname = `${person.first_names} ${person.last_name}`
+    }
+    return mapper(person)
+  })
+  return {
+    emptySlots,
```
teneightfive, line 22: This is a little hard to read as part of the "presenter". Returning a count does't really fit in with transforming an object into something that needs to be used by a component or another entity. 

I feel this presenter can be simplified a lot but just returning an array of cards for filled slots, and the count of empty slots can be sent to the view via the controller using a filter on the collection:

```javascript
const emptySlots = moves.filter(move => !move.person).length
```

At the moment the name of the presenter is saying it will return cards but it actually returns an object with a count and then a list of cards.

If you remove the count you could even remove the need for this presenter altogether and re-use the existing person-to-card-component:

```javascript
const filledMoves = moves
  .filter(move => move.person)
  .map(move => move.person)
  .map(personToCardComponent())

const locals = {
  ...
  emptySlots,
  filledMoves,
  ...
}
```

This greatly improves its readability and maintainability. If we make changes to how the card displays those changes will be consistently carried across the application.

And if you don't feel moving the addition of `fullname` to use the shared formatter is necessary:

```javascript
const filledMoves = moves
  .filter(move => move.person)
  .map(move => {
    return {
      ...move.person,
      person.fullname: `${person.first_names} ${person.last_name}`,
    }
  })
  .map(personToCardComponent())
```

common/presenters/allocation-to-summary-list-component.js
teneightfive, line 6: Can we use `moves.length` here too please so that we don't introduce more debt around `moves_count`.

locales/en/fields.json
```diff
@@ -259,11 +259,9 @@
   },
   "sentence_length": {
     "label": "Sentence length",
-    "items": {
```
teneightfive, line 262: Any reason to remove the `items` key? It's something we've used consistently elsewhere to show the keys for items of a field.

PR #505, opened 2020-05-12
chore: Add e2e tests for updating move date

solidgoldpig:
Adds e2e tests for checking that move date can be updated.

Extends create-move's `fillInDate` method to allow an optional date type or value to be passed to it.

### Checklist

### DIFF
--- a/test/e2e/_move.js
+++ b/test/e2e/_move.js
@@ -410,6 +410,21 @@ export async function checkUpdateMoveDetails() {
 }
 
 /**
+ * Change move date and confirm changes on move view page
+ *
+ * @returns {undefined}
+ */
+export async function checkUpdateMoveDate(date = 'Tomorrow') {
+  const updateMovePage = await clickUpdateLink('date')
+
+  const updatedMove = await updateMovePage.fillInDate(date)
+
+  await updateMovePage.submitForm()
+
+  await moveDetailPage.checkMoveDetails(updatedMove)
+}
+
+/**
  * Check that PNC details are uneditable and displayed corectly on the personal details page
  *
  * @returns {undefined}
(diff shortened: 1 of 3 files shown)


test/e2e/_move.js
```diff
@@ -409,6 +409,21 @@ export async function checkUpdateMoveDetails() {
   await moveDetailPage.checkMoveDetails(updatedMove)
 }
 
+/**
+ * Change move date and confirm changes on move view page
+ *
+ * @returns {undefined}
+ */
+export async function checkUpdateMoveDate(date = 'Tomorrow') {
+  const updateMovePage = await clickUpdateLink('date')
```
teneightfive, line 418: Does `clickUpdateLink` return a page object then?

test/e2e/move.update.police.test.js
```diff
@@ -92,4 +93,12 @@ if (FEATURE_FLAGS.EDITABILITY) {
       await checkUpdateMoveDetails()
     }
   )
+
+  test('User should be able to change move date to tomorrow', async () => {
+    await checkUpdateMoveDate()
+  })
+
+  test('User should be able to change move date to another date', async () => {
+    await checkUpdateMoveDate(3)
```
teneightfive, line 102: This felt a little hard to read just looking at this call. It wasn't easy to tell `3` was going to set a custom date, 3 days from now.

I wonder if it might be a bit simpler to just use a faker date in the future in the `fillInDate` method and not worry about needing to define how many days away it is being changed to.

test/e2e/_move.js
solidgoldpig: Yep.

test/e2e/move.update.police.test.js
solidgoldpig: Personally, I'd like to use something like 'moment` and be able to pass stuff like `+3d`. (But didn't want to introduce or reinvent it here). Will move the number stuff out.

PR #506, opened 2020-05-12
Link dashboard directly to single request review page

teneightfive:
## Proposed changes

This links the single requests dashboard directly to the review
page for a move if they have permission to access it.

It also includes a column width change to the table to ensure dates will always have enough room.

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |


common/presenters/allocations-to-table.test.js
```diff
@@ -92,30 +92,30 @@ describe('#allocationsToTable', function() {
...
+        text: 'HMP Ashfield',
       })
     })
     it('returns move date on the fifth cell', function() {
       expect(output.rowsForAllocationTable[0][4]).to.deep.equal({
-        html: '3 May 2020',
+        text: '3 May 2020',
       })
     })
     xit('returns progress on the sixth cell', function() {
```
solidgoldpig, line 121: Should this still be skipped?

common/presenters/allocations-to-table.js
```diff
@@ -5,33 +5,43 @@ const tablePresenters = require('./table')
...
-    row: data => filters.formatDate(data.created_at),
+    row: {
+      text: data => filters.formatDate(data.created_at),
+    },
   },
   {
     head: 'allocations::move_from',
-    row: 'from_location.title',
+    row: {
+      text: 'from_location.title',
```
solidgoldpig, line 24: Why all the `htmls`s to 'text`s?

common/presenters/allocations-to-table.test.js
teneightfive: Not sure. It was an existing test that @barbaracassani I think added as not sure what is going to go in that column for allocations.

common/presenters/allocations-to-table.js
teneightfive: They were all made `html` originally but that shouldn't be the default. Using `html` in the [Table component](https://design-system.service.gov.uk/components/table/) will escape the input. We should only use `html` if we need to and can ensure there is no unescaped user inputted data in there.

PR #509, opened 2020-05-13
[P4 1196] Provide ability for STC user to update move documents

solidgoldpig:
Users with `move:update` permission can now update documents for moves when the current location allows documents.
 
Adds link to move view
Move_details_for_TREMBLAY__AUSTEN

Adds update document view
Are_there_any_documents_to_upload__-_update_move
(Looks and functions identically to create's document view)

### Checklist

### DIFF
--- a/app/move/controllers/update/document.js
+++ b/app/move/controllers/update/document.js
@@ -1,0 +1,59 @@
+const { get } = require('lodash')
+
+const moveService = require('../../../../common/services/move')
+const DocumentUploadController = require('../create/document')
+
+const UpdateBase = require('./base')
+
+class UpdateDocumentUploadController extends UpdateBase {
+  configure(req, res, next) {
+    const canUploadDocuments = get(
+      req,
+      'session.currentLocation.can_upload_documents'
+    )
+    if (!canUploadDocuments) {
+      const error = new Error(
+        'Document upload is not possible for this location'
+      )
+      error.statusCode = 404
+      return next(error)
+    }
+    DocumentUploadController.prototype.configure.apply(this, [req, res, next])
+  }
+
+  getUpdateValues(req, res) {
+    const move = req.getMove()
+    if (!move) {
+      return {}
+    }
+    const values = {}
+    if (req.initialStep) {
+      req.sessionModel.set('documents', move.documents)
+    }
+    values.documents = req.sessionModel.get('documents')
+    return values
+  }
+
+  async successHandler(req, res, next) {
+    if (!req.xhr) {
+      try {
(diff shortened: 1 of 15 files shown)


app/move/controllers/update/document.js
```diff
@@ -0,0 +1,59 @@
...
+    const values = {}
+    if (req.initialStep) {
+      req.sessionModel.set('documents', move.documents)
+    }
+    values.documents = req.sessionModel.get('documents')
+    return values
+  }
+
+  async successHandler(req, res, next) {
+    if (!req.xhr) {
```
teneightfive, line 38: This might read a little nicer the other way around. If `xhr`, call parent, otherwise save using the service.

app/move/controllers/update/document.js
```diff
@@ -0,0 +1,59 @@
...
+  getUpdateValues(req, res) {
+    const move = req.getMove()
+    if (!move) {
+      return {}
+    }
+    const values = {}
+    if (req.initialStep) {
+      req.sessionModel.set('documents', move.documents)
+    }
+    values.documents = req.sessionModel.get('documents')
```
teneightfive, line 33: Doesn't the parent getValues do something like retrieving this? Wonder if we can re-use that across all the update controllers.

app/move/controllers/update/document.test.js
```diff
@@ -0,0 +1,267 @@
...
+      })
+
+      context('When current location does allow uploads', function() {
+        beforeEach(function() {
+          req.session.currentLocation.can_upload_documents = true
+          controller.configure(req, res, nextSpy)
+        })
+
+        it('should not call next', function() {
+          expect(nextSpy).to.not.be.called
```
teneightfive, line 73: I wasn't aware `to.not.be` existed too. I've always used `not.to.be`..._to be_ or _not to be_...

solidgoldpig: I hate mocha

app/move/controllers/update/document.js
solidgoldpig: Yes, now it's all in place, we can probably streamline this. As part of the move to sticking everything on the `req` object

app/move/controllers/update/document.test.js
solidgoldpig: Wonder if there's an eslint rule to enforce a particular variant? Manually flipped for now

app/move/controllers/update/document.js
solidgoldpig: Top Gun-style double-inverted

PR #514, opened 2020-05-14
refactor: Handle redirecting of different move types

teneightfive:
## Proposed changes

This changes the way default URLs are handled within the moves app. Now each "view" has a default time period which is handled at the base router level.

This simplifies the main navigation and any other areas where we will need to link to the default
view for a particular list of moves.

### DIFF
--- a/app/moves/constants.js
+++ b/app/moves/constants.js
@@ -1,0 +1,10 @@
+const CONSTANTS = {
+  DEFAULTS: {
+    QUERY: {
+      requested: { status: 'pending' },
+      outgoing: {},
+    },
+  },
+}
+
+module.exports = CONSTANTS
--- a/app/moves/constants.js
+++ b/app/moves/constants.js
@@ -4,6 +4,10 @@ const CONSTANTS = {
       requested: { status: 'pending' },
       outgoing: {},
     },
+    TIME_PERIOD: {
+      requested: 'week',
+      outgoing: 'day',
+    },
   },
 }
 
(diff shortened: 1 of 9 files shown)


app/moves/constants.js
```diff
@@ -0,0 +1,10 @@
+const CONSTANTS = {
```
solidgoldpig, line 1: Should we have a noddy tautological test to ensure that if anyone changes these values, they at the very least have to stop and fix the test and maybe think whether there are any conserquences of changing it?

app/moves/middleware/redirect-default-query.js
```diff
@@ -0,0 +1,16 @@
+const { isEmpty } = require('lodash')
+const querystring = require('qs')
+
+function redirectDefaultQuery(defaults = {}) {
+  return function handleQueryRedirect(req, res, next, view) {
+    const query = defaults[view]
+
+    if (!isEmpty(query) && isEmpty(req.query)) {
+      return res.redirect(`${req.originalUrl}?${querystring.stringify(query)}`)
```
solidgoldpig, line 9: Personally I'd prefer to have the url construction in another method. a) it makes testing both the construction and the redirection simpler and  b) it could be reused.

app/moves/constants.js
teneightfive: So in theory we don't actually care in terms of the unit test I don't think.

That's probably something that we should do with the e2e tests. Ie, if I click the tab in the navigation, by default I see View X.

But unfortunately those tests are still WIP... :/

app/moves/middleware/redirect-default-query.js
teneightfive: I'm inclined to agree for a more complex URL, which there will be some coming but this one felt simple enough to leave within this method at this stage.

PR #517, opened 2020-05-15
[P4 1276] cancel allocation without specifying a reason

barbaracassani:
This PR contains the redirection of the PMU users to a (temporary?) dashboard, and add the cancellation journey to the allocation view, without specifying a reason yet.

Upon logging in, the PMU user is presented with this dashboard:

Screenshot 2020-05-18 at 12 48 58

Clicking on "Allocations" (the single request links does not go anywhere for now):
Screenshot 2020-05-18 at 12 49 20

The single view has now the cancel link:
Screenshot 2020-05-18 at 12 49 35

Intermediate page that will contain the reason, when it's supported by the BE:
Screenshot 2020-05-18 at 12 49 44

Confirmation:
Screenshot 2020-05-18 at 12 53 37

### Checklist

### DIFF
--- a/app/allocation/controllers/cancel.js
+++ b/app/allocation/controllers/cancel.js
@@ -1,0 +1,21 @@
+const FormWizardController = require('../../../common/controllers/form-wizard')
+const allocationService = require('../../../common/services/allocation')
+
+class CancelController extends FormWizardController {
+  async successHandler(req, res, next) {
+    const { id: allocationId } = res.locals.allocation
+
+    try {
+      await allocationService.cancel(allocationId)
+
+      req.journeyModel.reset()
+      req.sessionModel.reset()
+
+      res.redirect(`/allocation/${allocationId}`)
+    } catch (error) {
+      next(error)
+    }
+  }
+}
+
+module.exports = CancelController
--- a/app/allocations/middleware.js
+++ b/app/allocations/middleware.js
@@ -1,5 +1,5 @@
 const { format } = require('date-fns')
-const { cloneDeep, isEqual, mapValues } = require('lodash')
+const { mapValues } = require('lodash')
 const queryString = require('query-string')
 
 const {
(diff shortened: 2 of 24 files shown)


app/home/controllers.js
```diff
@@ -5,6 +5,10 @@ const permissions = require('../../common/middleware/permissions')
 function dashboard(req, res) {
   const userPermissions = get(req.session, 'user.permissions')
 
+  if (permissions.check('allocations:view', userPermissions)) {
```
teneightfive, line 8: I'm just working on supporting this on the dashboard so you can probably remove this commit from this PR. Maybe keep a forked branch in case we want to merge them still afterwards.

app/allocations/middleware.js
```diff
@@ -49,14 +50,38 @@ function setPagination(req, res, next) {
   next()
 }
 async function setAllocationsSummary(req, res, next) {
-  const { dateRange } = req.params
-  const value = await allocationService.getCount({ dateRange })
+  const { dateRange } = res.locals
+  const allocationsCount = await allocationService.getCount({ dateRange })
```
teneightfive, line 43: We can remove this commit too as it's replicating what the main dashboard now does. I'm changing allocations to use the shared logic around listing results and enabling them to see the shared dashbard.

That will avoid us adding more to this complicated area for now. So I'd say pull this commit into a separate branch too for now and focus this change on just supporting cancelling an allocation.

common/lib/api-client/models.js
```diff
@@ -43,6 +43,10 @@ module.exports = {
         jsonApi: 'hasMany',
         type: 'events',
       },
+      allocations: {
```
teneightfive: I think this should be singular: `allocation`

app/allocation/controllers/cancel.js
```diff
@@ -0,0 +1,21 @@
+const FormWizardController = require('../../../common/controllers/form-wizard')
```
teneightfive, line 1: Can we add a unit test for this file please.

barbaracassani: Done 👍

PR #518, opened 2020-05-15
Allow PMU users to view all single requests

teneightfive:
## Proposed changes

This set of changes refactors and updates the permissions and routing for moves.

It allows OCA users to see single requests and makes the middleware more re-usable by other
parts of the application but using a configuration approach.

It also adds the correct permissions for Population Management Unit (PMU) users to see all
single requests for this MVP release.

### DIFF
--- a/app/move/index.js
+++ b/app/move/index.js
@@ -63,7 +63,11 @@ router.use(
   wizard(createSteps, createFields, createConfig)
 )
 router.get('/:moveId', protectRoute('move:view'), view)
-router.get('/:moveId/confirmation', protectRoute('move:create'), confirmation)
+router.get(
+  '/:moveId/confirmation',
+  protectRoute(['move:create', 'move:review']),
+  confirmation
+)
 router.use(
   '/:moveId/cancel',
   protectRoute('move:cancel'),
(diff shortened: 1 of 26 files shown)


app/move/index.js
```diff
@@ -63,7 +63,11 @@ router.use(
   wizard(createSteps, createFields, createConfig)
 )
 router.get('/:moveId', protectRoute('move:view'), view)
-router.get('/:moveId/confirmation', protectRoute('move:create'), confirmation)
+router.get(
+  '/:moveId/confirmation',
+  protectRoute(['move:create', 'move:review']),
```
barbaracassani, line 68: Why do we need a separate permission for `move:review`? Surely it's akin to create, in practice?

teneightfive: Nope, they are different. This permission is for PMU to be allowed to accept or reject a proposed move but we are re-using the confirmation page in that journey. But the permission to create shouldn't allow a user to review a proposed move, likewise, being able to create a move doesn't mean you should be able to review a proposed move.

tl;dr we don't want to give PMU the ability to create single moves, just allocation moves

barbaracassani: 👍

PR #520, opened 2020-05-18
test: Increase the speed of the editability tests

teneightfive:
This tries to reduce the need to create a new move for each editability
test in order to speed them up.

It also moves away from moves being created based on a user and their role
as this is a concept we have tried to stay away from. We can keep some
checks linked to roles based around permissions but when creating those
fixtures they shouldn't be tied to a person's role.

### DIFF
--- a/test/e2e/_move.js
+++ b/test/e2e/_move.js
@@ -2,14 +2,7 @@ import faker from 'faker'
 import { Selector, t } from 'testcafe'
 
 import { createMoveFixture, fillInForm, expectStatusCode } from './_helpers'
-import {
-  ocaUser,
-  policeUser,
-  prisonUser,
-  stcUser,
-  supplierUser,
-} from './_roles'
-import { home, getMove, getUpdateMove } from './_routes'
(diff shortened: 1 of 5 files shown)


test/e2e/_move.js
```diff
@@ -27,24 +20,19 @@ import UpdateMovePage from './pages/update-move'
...
-  await t.useRole(options.user)
-
   if (
     !options.moveOverrides ||
     (options.moveOverrides && !options.moveOverrides.from_location)
   ) {
-    await t.navigateTo(home)
-
-    const currentLocation = await page.nodes.locationMeta.getAttribute(
+    const currentLocationId = await page.nodes.locationMeta.getAttribute(
```
solidgoldpig, line 30: Is it an idea to blow up if the meta location element doesn't have a value?

teneightfive: Good point. What do you think is the best way? Should we throw an error? Or do you think we could get away with a TestCafe assertion to check that it's not undefined?

PR #523, opened 2020-05-18
Re-use collection structure for allocations

teneightfive:
## Proposed changes

These changes re-use the methods used within the moves app to display a set of results. It's starts to extract some of the shared middleware to a common place as well as addressing some issues with filtering allocations correctly.

It introduces support for displaying only allocations for a particular location as well as all locations and adds in the foundations for filtering and support to re-use the filter for the unified dashboard.


barbaracassani:
Ran the code, and it's 👍

PR #525, opened 2020-05-19
chore(single requests): Add from location to table

teneightfive:
## Proposed changes

Currently some users so a table with all locations. In the table view, it's hard to see where the different moves are coming from.

This change adds a column to display the location of the sending location.

### DIFF
--- a/locales/en/moves.json
+++ b/locales/en/moves.json
@@ -109,16 +109,11 @@
     "single_moves_no_results_pending": "No single requests pending review",
     "single_moves_no_results_approved": "No single requests have been approved",
     "single_moves_no_results_rejected": "No single requests have been rejected",
+    "move_from": "Move from",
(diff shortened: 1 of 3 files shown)


locales/en/moves.json
```diff
@@ -109,16 +109,11 @@
...
     "move_to": "Move to",
     "earliest_move_date": "Earliest date of travel",
     "created_at": "Sent on",
-    "move_type": "Move type",
-    "filter": {
-      "proposed": "pending review",
-      "approved": "approved",
-      "rejected": "rejected",
-      "total": "sent for review"
-    }
```
barbaracassani, line 121: Are we sure those are now unused? Just checkin'.

teneightfive: Nope, the unified dashboard at the root of the application uses the existing translations for these.

PR #530, opened 2020-05-19
[P4-1464] Listing allocations on the dashboard

teneightfive:
## Proposed changes

This set of changes adds the allocations panel to the dashboard as well as tidying up some of the language used across the allocations section.

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |


common/middleware/set-primary-navigation.js
```diff
@@ -18,15 +18,24 @@ function setPrimaryNavigation(req, res, next) {
...
+      active: req.path.startsWith('/moves') && req.path.endsWith('/requested'),
+    })
+  }
+
+  if (permissions.check('allocations:view', userPermissions)) {
+    items.push({
+      text: req.t('primary_navigation.allocations'),
+      href: '/allocations',
+      active:
+        req.path.startsWith('/allocations') && req.path.endsWith('/outgoing'),
```
solidgoldpig, line 30: Is this active check correct?

teneightfive: Yep, so the allocations URLs contain the same route structure as moves to share the code around collections.

The both use: `/:base/:date/:locationId?/:view`. This is looking at the base and the last path.

For allocations, we could potentially drop the second check as we only list one type at the moment.

solidgoldpig: but for super clarity, it was the second part of the check I was asking about

 `&& req.path.endsWith('/outgoing')`

since surely `/allocations` itself wouldn't be matched - presumably we redirect away from that so it doesn't arise

PR #534, opened 2020-05-20
[P4 1276] e2e tests for cancel and create allocation

barbaracassani:
These are the tests for cancel and create allocation. They pass locally (most of the time), but need a test user to be added to this PR for it to run.

### Checklist

### DIFF
--- a/test/e2e/_roles.js
+++ b/test/e2e/_roles.js
@@ -26,3 +26,8 @@ export const prisonUser = Role(home, async t => {
 export const ocaUser = Role(home, async t => {
   await page.signIn(E2E.ROLES.OCA)
 })
+
+export const pmuUser = Role(home, async t => {
+  await page.signIn(E2E.ROLES.PMU)
+  await page.chooseAllLocations()
+})
--- a/test/e2e/_routes.js
+++ b/test/e2e/_routes.js
@@ -19,3 +19,5 @@ export const newMove = `${E2E.BASE_URL}/move/new`
 export const getMove = id => `${E2E.BASE_URL}/move/${id}`
 export const getUpdateMove = (id, page) =>
   `${E2E.BASE_URL}/move/${id}/edit/${getUpdatePageStub(page)}`
+export const newAllocation = `${E2E.BASE_URL}/allocation/new`
+export const allocations = `${E2E.BASE_URL}/allocations`
--- a/test/e2e/allocation.cancel.test.js
+++ b/test/e2e/allocation.cancel.test.js
@@ -1,0 +1,26 @@
+import { pmuUser } from './_roles'
+import { allocations, newAllocation } from './_routes'
+import { allocationJourney } from './pages/'
+
+fixture('Cancel allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+  await allocationJourney.createAllocation()
+  await t.navigateTo(allocations)
+})
+test('Cancel allocation', async t => {
+  await t.click(
+    allocationJourney.allocationCancelPage.nodes.linkToFirstAllocation
+  )
+  await allocationJourney.scrollToBottom()
+  await t.click(allocationJourney.allocationCancelPage.nodes.cancelLink)
+  await t
+    .expect(allocationJourney.getCurrentUrl())
+    .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}\/cancel\/reason$/)
+  await allocationJourney.allocationCancelPage.submitForm()
+  await t
+    .expect(allocationJourney.getCurrentUrl())
+    .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}$/)
+  await t
+    .expect(allocationJourney.allocationCancelPage.nodes.statusHeading.exists)
+    .ok()
+})
--- a/test/e2e/allocation.new.test.js
+++ b/test/e2e/allocation.new.test.js
@@ -1,0 +1,36 @@
+import { pmuUser } from './_roles'
+import { newAllocation } from './_routes'
+import { allocationJourney } from './pages/'
+
+fixture('New PMU allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+})
+test('Create allocation and verify the result', async t => {
+  const allocationId = await allocationJourney.createAllocation()
+  await t.navigateTo(`/allocation/${allocationId}`)
+  ;['summary', 'meta'].forEach(async section => {
+    for (const o of allocationJourney.allocationViewPage.nodes[section].keys) {
+      const textExists = allocationJourney.allocationViewPage.getDlDefinitionByKey(
+        allocationJourney.allocationViewPage.nodes[section].selector,
+        o
+      )
+      await t.expect(textExists).ok()
+    }
+  })
+})
+test('Check validation errors on allocation details page', async t => {
+  await allocationJourney.submitForm()
+
+  for (const item of allocationJourney.allocationDetailsPage.errorLinks) {
+    const error = allocationJourney.findErrorInList(item)
+    await t.expect(error).ok()
+  }
+})
+test('Check validation errors on allocation criteria page', async t => {
+  await allocationJourney.triggerValidationOnAllocationCriteriaPage()
+
+  for (const item of allocationJourney.allocationCriteriaPage.errorLinks) {
+    const error = allocationJourney.findErrorInList(item)
+    await t.expect(error).ok()
+  }
+})
--- a/test/e2e/pages/allocation-cancel.js
+++ b/test/e2e/pages/allocation-cancel.js
@@ -1,0 +1,18 @@
+import { Selector } from 'testcafe'
+
+import Page from './page'
+
+class AllocationCancelPage extends Page {
+  constructor() {
+    super()
+    this.nodes = {
+      ...this.nodes,
+      linkToFirstAllocation: Selector('.govuk-table td:first-of-type a'),
+      cancelLink: Selector('a').withText('Cancel allocation'),
+      statusHeading: Selector(
+        '.app-message.app-message--temporary .app-message__heading'
+      ).withText('Allocation cancelled'),
+    }
+  }
+}
+export default AllocationCancelPage
--- a/test/e2e/pages/allocation-criteria.js
+++ b/test/e2e/pages/allocation-criteria.js
@@ -1,0 +1,62 @@
+import faker from 'faker'
+import { Selector } from 'testcafe'
+
+import { fillInForm } from '../_helpers'
+
+import Page from './page'
+
+class AllocationCriteriaPage extends Page {
+  constructor() {
+    super()
+    this.url = '/allocation/new/allocation-criteria'
+    this.fields = {
+      prisonerCategory: Selector('#prisoner_category'),
+      sentenceLength: Selector('[name="sentence_length"]'),
+      complexCases: Selector('#complex_cases'),
+      completeInFull: Selector('#complete_in_full'),
+      hasOtherCriteria: Selector('#has_other_criteria'),
+      otherCriteria: Selector('#other_criteria'),
+      errorSummary: Selector('.govuk-error-summary'),
+    }
+    this.errorSummary = Selector('.govuk-error-summary ul')
+    this.errorLinks = [
+      '#prisoner_category',
+      '#sentence_length',
+      '#complete_in_full',
+      '#has_other_criteria',
+    ]
+  }
+
+  fill() {
+    const fieldsToFill = [
+      {
+        selector: this.fields.prisonerCategory,
+        value: 'C',
+        type: 'radio',
+      },
+      {
+        selector: this.fields.sentenceLength,
+        type: 'radio',
+      },
+      {
+        selector: this.fields.complexCases,
+        type: 'checkbox',
+      },
+      {
+        selector: this.fields.completeInFull,
+        type: 'radio',
+      },
+      {
+        selector: this.fields.hasOtherCriteria,
+        type: 'radio',
+        value: 'Yes',
+      },
+      {
+        selector: this.fields.otherCriteria,
+        value: faker.lorem.sentence(6),
+      },
+    ]
+    return fillInForm(fieldsToFill)
+  }
+}
+export default AllocationCriteriaPage
--- a/test/e2e/pages/allocation-details.js
+++ b/test/e2e/pages/allocation-details.js
@@ -9,7 +9,7 @@ class AllocationDetailsPage extends Page {
   constructor() {
     super()
     this.url = '/allocation/new/allocation-details'
-    this.fields = {
+    this.nodes = {
       movesCount: Selector('#moves_count'),
       fromLocation: Selector('#from_location'),
       toLocation: Selector('#to_location'),
@@ -27,24 +27,24 @@ class AllocationDetailsPage extends Page {
   fill() {
     const fieldsToFill = [
       {
-        selector: this.fields.movesCount,
+        selector: this.nodes.movesCount,
         value: '3',
       },
       // TODO it would be better to let the value blank here, so a random one would be used.
       // However, it currently makes the test fail when the chosen option is not in the visible
       // portion of the list. This clearly needs to be addressed.
       {
-        selector: this.fields.fromLocation,
+        selector: this.nodes.fromLocation,
         type: 'autocomplete',
         value: 0,
       },
       {
-        selector: this.fields.toLocation,
+        selector: this.nodes.toLocation,
         type: 'autocomplete',
         value: 1,
       },
       {
-        selector: this.fields.date,
+        selector: this.nodes.date,
         value: format(endOfWeek(new Date()), 'dd/MM/yyyy'),
       },
     ]
--- a/test/e2e/pages/allocation-details.js
+++ b/test/e2e/pages/allocation-details.js
@@ -1,0 +1,54 @@
+import { endOfWeek, format } from 'date-fns'
+import { Selector } from 'testcafe'
+
+import { fillInForm } from '../_helpers'
+
+import Page from './page'
+
+class AllocationDetailsPage extends Page {
+  constructor() {
+    super()
+    this.url = '/allocation/new/allocation-details'
+    this.fields = {
+      movesCount: Selector('#moves_count'),
+      fromLocation: Selector('#from_location'),
+      toLocation: Selector('#to_location'),
+      date: Selector('#date'),
+    }
+    this.errorSummary = Selector('.govuk-error-summary__list')
+    this.errorLinks = [
+      '#moves_count',
+      '#to_location',
+      '#from_location',
+      '#date',
+    ]
+  }
+
+  fill() {
+    const fieldsToFill = [
+      {
+        selector: this.fields.movesCount,
+        value: '3',
+      },
+      // TODO it would be better to let the value blank here, so a random one would be used.
+      // However, it currently makes the test fail when the chosen option is not in the visible
+      // portion of the list. This clearly needs to be addressed.
+      {
+        selector: this.fields.fromLocation,
+        type: 'autocomplete',
+        value: 0,
+      },
+      {
+        selector: this.fields.toLocation,
+        type: 'autocomplete',
+        value: 1,
+      },
+      {
+        selector: this.fields.date,
+        value: format(endOfWeek(new Date()), 'dd/MM/yyyy'),
+      },
+    ]
+    return fillInForm(fieldsToFill)
+  }
+}
+export default AllocationDetailsPage
--- a/test/e2e/pages/allocation-journey.js
+++ b/test/e2e/pages/allocation-journey.js
@@ -1,0 +1,51 @@
+import { t } from 'testcafe'
+
+import AllocationCancelPage from './allocation-cancel'
+import AllocationCriteriaPage from './allocation-criteria'
+import AllocationDetailsPage from './allocation-details'
+import AllocationViewPage from './allocation-view'
+import Page from './page'
+
+import { allocationJourney } from './index'
+
+const allocationDetailsPage = new AllocationDetailsPage()
+const allocationCriteriaPage = new AllocationCriteriaPage()
+const allocationCancelPage = new AllocationCancelPage()
+const allocationViewPage = new AllocationViewPage()
+
+class AllocationJourney extends Page {
+  constructor() {
+    super()
+    this.allocationCancelPage = allocationCancelPage
+    this.allocationCriteriaPage = allocationCriteriaPage
+    this.allocationDetailsPage = allocationDetailsPage
+    this.allocationViewPage = allocationViewPage
+  }
+
+  async createAllocation() {
+    await allocationDetailsPage.fill()
+    await this.submitForm()
+    await allocationCriteriaPage.fill()
+    await this.submitForm()
+    const currentUrl = await allocationJourney.getCurrentUrl()
+    await t
+      .expect(currentUrl)
+      .match(
+        new RegExp('/allocation/' + this.uuidRegex.source + '/confirmation')
+      )
+    return currentUrl.match(new RegExp(allocationJourney.uuidRegex))[0]
+  }
+
+  async triggerValidationOnAllocationCriteriaPage() {
+    await allocationDetailsPage.fill()
+    await this.submitForm()
+    await this.submitForm()
+  }
+
+  findErrorInList(hrefAttribute) {
+    return allocationDetailsPage.errorSummary
+      .child('a')
+      .withAttribute('href', hrefAttribute)
+  }
+}
+export default AllocationJourney
--- a/test/e2e/pages/allocation-view.js
+++ b/test/e2e/pages/allocation-view.js
@@ -1,0 +1,28 @@
+import { Selector } from 'testcafe'
+
+import Page from './page'
+
+class AllocationViewPage extends Page {
+  constructor() {
+    super()
+    this.nodes = {
+      heading: Selector('h1.govuk-heading-l'),
+      listElements: {
+        category: 'Category',
+      },
+      summary: {
+        selector: Selector('.govuk-summary-list'),
+        keys: [
+          'Time left to serve',
+          'Complex cases for prisons to agree',
+          'Other criteria',
+        ],
+      },
+      meta: {
+        selector: Selector('.app-meta-list'),
+        keys: ['Number of people', 'Move from', 'Move to', 'Date'],
+      },
+    }
+  }
+}
+export default AllocationViewPage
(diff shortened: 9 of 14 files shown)


test/e2e/_routes.js
```diff
@@ -19,3 +19,6 @@ export const newMove = `${E2E.BASE_URL}/move/new`
 export const getMove = id => `${E2E.BASE_URL}/move/${id}`
 export const getUpdateMove = (id, page) =>
   `${E2E.BASE_URL}/move/${id}/edit/${getUpdatePageStub(page)}`
+export const newAllocation = `${E2E.BASE_URL}/allocation/new`
+// export const allocations = `${E2E.BASE_URL}/allocations/week/2020-05-19/outgoing`
```
teneightfive, line 23: Rogue comment?

test/e2e/_roles.js
```diff
@@ -26,3 +26,11 @@ export const prisonUser = Role(home, async t => {
 export const ocaUser = Role(home, async t => {
   await page.signIn(E2E.ROLES.OCA)
 })
+
+export const pmuUser = Role(home, async t => {
+  await t
```
teneightfive, line 31: Can we use the `page.signIn` method for consistency please.

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,52 @@
+import { Selector, t } from 'testcafe'
+
+import { pmuUser } from './_roles'
+import { allocations, newAllocation } from './_routes'
+import { allocationDetails, allocationCriteria, page } from './pages/'
+
+const selectors = {
```
teneightfive, line 7: Can we use [Page Objects](https://devexpress.github.io/testcafe/documentation/guides/concepts/page-model.html) for defining selectors please. This keeps inline with the existing pattern and tries to keep selectors to one place so if they change we don't have to update them through test files.

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,52 @@
...
+    cancelLink: Selector('.app-link--destructive').withText(
+      'Cancel allocation'
+    ),
+    statusHeading: Selector(
+      '.app-message.app-message--temporary .app-message__heading'
+    ).withText('Allocation cancelled'),
+  },
+}
+
+async function createAllocation() {
```
teneightfive, line 19: We can also move this to a page object.

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,52 @@
...
+  await page.submitForm()
+  return t
+    .expect(page.getCurrentUrl())
+    .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}\/confirmation$/)
+}
+
+fixture('New PMU allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+})
+test('by default', t => {
```
teneightfive, line 32: This test doesn't really feel very clear and descriptive. In the test output on CI it would just read:

```
New PMU allocation
  ✓ by default
```

Maybe we can follow the pattern that has been used elsewhere:

```
New allocation
  ✓ with all allocation criteria
```

It might also be good to test variations of the criteria screen within this section.

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,52 @@
...
+  return createAllocation()
+})
+
+fixture('Cancel allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+  await createAllocation()
+  await t.useRole(pmuUser).navigateTo(allocations)
+})
+test('Cancel allocation', async t => {
+  await t.click(selectors.cancelJourney.linkToFirstAllocation)
```
teneightfive, line 42: Can we also move these selectors to a page object. We can have a separate cancel-allocation page object like we have for [cancel-move](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/3a64f5b2d6669722908c86c41cd87e9ca7b78e9d/test/e2e/pages/cancel-move.js#L1)

test/e2e/pages/allocation-criteria.js
```diff
@@ -0,0 +1,53 @@
...
+        selector: this.fields.hasOtherCriteria,
+        type: 'radio',
+        value: 'Yes',
+      },
+      {
+        selector: this.fields.otherCriteria,
+        value: 'lorem ipsum',
+      },
+    ]
+    return await fillInForm(fieldsToFill)
```
teneightfive, line 50: You don't need to `await` on a `return`. You can then also remove the `async` from the function call.

test/e2e/pages/allocation-criteria.js
```diff
@@ -0,0 +1,53 @@
+import { Selector } from 'testcafe'
+
+import { fillInForm } from '../_helpers'
+
+import Page from './page'
+
+class AllocationCriteria extends Page {
```
teneightfive, line 7: Can we continue to use the consistent naming of pages and include `Page` in the name: `AllocationCriteriaPage`.

test/e2e/pages/allocation-details.js
```diff
@@ -0,0 +1,44 @@
+import { endOfWeek, format } from 'date-fns'
+import { Selector } from 'testcafe'
+
+import { fillInForm } from '../_helpers'
+
+import Page from './page'
+
+class AllocationDetails extends Page {
```
teneightfive, line 8: Same for this page name please.

test/e2e/pages/allocation-details.js
```diff
@@ -0,0 +1,44 @@
...
+  async fill() {
+    const fieldsToFill = [
+      {
+        selector: this.fields.movesCount,
+        value: '3',
+      },
+      {
+        selector: this.fields.fromLocation,
+        type: 'autocomplete',
+        value: 0,
```
teneightfive, line 29: I think it don't pass a value it will use a random value which I think would be better to test with.

test/e2e/pages/page.js
```diff
@@ -70,6 +70,19 @@ export default class Page {
       .notContains('/auth/login')
   }
 
+  /**
+   * Select "all locations"
+   *
+   * @returns {Promise}
+   */
+  async chooseAllLocations() {
+    const allLocationsLink = Selector('main p a').withAttribute(
```
teneightfive, line 79: Maybe just use `Selector('a')` here. It means it's less likely to break if we move the position of that element in the DOM.

test/e2e/_routes.js
barbaracassani: oops, yes.

test/e2e/allocation.new.test.js
barbaracassani: 👍

test/e2e/pages/allocation-details.js
barbaracassani: This is something that I wanted to discuss.  For some reason, the test fails intermittently if the chosen option is not immediately visible. This makes the test reliable, however is clearly not a permanent solution.

test/e2e/_allocation.js
```diff
@@ -0,0 +1,28 @@
...
+    .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}\/confirmation$/)
+}
+async function triggerValidationOnAllocationDetailsPage() {
+  await page.submitForm()
+}
+
+async function triggerValidationOnAllocationCriteriaPage() {
+  await allocationDetailsPage.fill()
+  await page.submitForm()
+  await page.submitForm()
```
teneightfive, line 21: Is this double submit intentional?

test/e2e/allocation.cancel.test.js
```diff
@@ -0,0 +1,22 @@
+import createAllocation from './_allocation'
+import { pmuUser } from './_roles'
+import { allocations, newAllocation } from './_routes'
+import { allocationCancelPage, page } from './pages/'
+
+fixture('Cancel allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+  await createAllocation()
+  await t.useRole(pmuUser).navigateTo(allocations)
```
teneightfive, line 9: You shouldn't need to use `useRole` again here as you're already signed in using that role.

test/e2e/allocation.cancel.test.js
```diff
@@ -0,0 +1,22 @@
+import createAllocation from './_allocation'
+import { pmuUser } from './_roles'
+import { allocations, newAllocation } from './_routes'
+import { allocationCancelPage, page } from './pages/'
+
+fixture('Cancel allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+  await createAllocation()
```
teneightfive, line 8: I think it might be nice to use a similar approach to the moves fixtures used with the edit tests to create an allocation fixture using the API service rather than needing to go through all the steps each time.

test/e2e/_allocation.js
```diff
@@ -0,0 +1,28 @@
+import { t } from 'testcafe'
+
+import { allocationDetailsPage, allocationCriteriaPage, page } from './pages/'
+
+async function createAllocation() {
```
teneightfive, line 5: We actually wanted to move away from this pattern of `_allocation`/`_move` file. The move example is a bit of a cross between helpers and a page object right now and I think we need to split it up a little moving forward. 

It would be good not to use this pattern for allocations as it will just introduce more tech debt. These methods look like they would fit into the page object model so if we need them to live somewhere, they should live within the page objects.

test/e2e/pages/index.js
```diff
@@ -15,6 +21,9 @@ const dashboardPage = new DashboardPage()
 
 export {
   page,
+  allocationCancelPage,
+  allocationCriteriaPage,
```
teneightfive, line 25: These could be combined to follow the same pattern for moves: `createAllocationPage`?

test/e2e/_allocation.js
barbaracassani: @teneightfive In this case yes, but in general, I had pushed a bunch of incomplete stuff this morning because I needed to change branch, and I sometimes forget to unstash. It wasn't meant to be reviewed, sorry. I appreciate the eagerness, but maybe next time wait till I re-request the review, just not to waste time.

test/e2e/pages/index.js
barbaracassani: But why exactly? The split between files helps keep them small and tidy, and the common methods are in `_allocation`. What are in your opinion the drawbacks of such approach?

teneightfive: It was just a suggestion to follow the pattern for creating a move. If we ended up with more steps we might up with lots of pages that was the only reason I was suggesting that.

My comments aren't always things we must do, they're talking points. You know more about this piece of code so it's just to create a discussion around changes and get a shared goal moving forward.

The pattern of `_allocation` isn't something I think we want to maintain though. I feel the things added to `_allocation` could easily fit within a page object model so rather than maintain two patterns (or continue to add to the two existing patterns) we should strive to use one approach as it will be easier to maintain long term. @solidgoldpig and I discussed moving some things out of `_move` to a more page object approach but as is often the case we had other delivery pressures.

So in summary, I think it would be better to keep these changes to using one approach rather than two.

test/e2e/pages/allocation-details.js
teneightfive: Ah okay. This feels like the issue we have seen elsewhere. Can we add a `TODO` around this and a little description of why we're always selecting the first item. It will help our future selves and future developers if somebody else picks up this codebase.

test/e2e/allocation.cancel.test.js
barbaracassani: Whilst I love the approach you mention, I am not entirely sure it makes sense here. I would make helpers that will be used literally only once, as there are only 2 pages in the create allocation. It would be nice, but does it represent business / technical value?

test/e2e/pages/index.js
barbaracassani: Ok, I acknowledge it was a talking point -- however you formally requested changes, so it wasn't immediately evident what was requested.
I am ok to reincorporate the three files (the two pages files, plus `_allocation`) as long as the pattern makes sense. However, whilst I appreciate that we want to move away from the status quo, I need to be able to appreciate its drawbacks. Currently, I see none, unless we want to genericise most actions across the board (e.g. `moves`,  `allocation` and whatever new comes up). Which might be a good idea, but way outside the scope of this PR, which meant to simply complement the allocations feature.

In short -- what do you suggest I do?

teneightfive: Sorry for the confusion. The requested changes were to things that weren’t needed or I felt could be updated to keep inline with the rest of the codebase so that it remains maintainable and consistent. 

This particular line isn’t a big issue as it keeps to the concept of using page objects to structure tests and was added as a comment, not a change request.  

My concern is with the addition of the `_allocation` file as this doesn’t follow the status quo, this in fact starts to move away from it and will it harder for us to maintain the page object pattern moving forward. The desired goal is to use page objects which are a widely understood concept within end to end testing and come with benefits such as isolating where the tests interact with elements of the UI instead of having those interactions in multiple places. 

So in short, can we move the methods from `_allocation` to a page object.

test/e2e/allocation.cancel.test.js
teneightfive: I think the long term technical benefit would be efficiency and re-usability. I imagine we will need to create an allocation to test further parts of that section of the app? For example, testing the lists with allocations in different states and filters applied. It will be more efficient to not have to run through those pages each time. 

The business value comes from the codebase being easier to work on as we continue. If we do each ticket in isolation and never consider the wider codebase it becomes difficult to maintain or handover in the future. 

if there is a solution that is more efficient and can be achieved in a similar amount of time we should aim for that.

test/e2e/pages/index.js
barbaracassani: Ok, have a look now, if it seems better.

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,25 @@
+import { pmuUser } from './_roles'
+import { newAllocation } from './_routes'
+import { allocationJourney } from './pages/'
+
+fixture('New PMU allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+})
+test('Create allocation with all the options', t =>
```
teneightfive, line 8: I think it would be good to actually test the result of the creation as part of this test.

Maybe the `createAllocation` method should just create the allocation rather than be concerned with assertions, the current name is a bit misleading.

If you have a look at the creating of moves, the test files call page object methods to do particular interactions and end by calling methods that assert particular behaviour.

test/e2e/pages/allocation-cancel.js
```diff
@@ -0,0 +1,18 @@
+import { Selector } from 'testcafe'
+
+import Page from './page'
+
+class AllocationCancelPage extends Page {
+  constructor() {
+    super()
+    this.fields = {
+      linkToFirstAllocation: Selector('.govuk-table td:first-of-type a'),
```
teneightfive, line 9: This feels like a copy/paste from another file but now doesn't really fit. These elements aren't _fields_. This object was used in other page objects to contain the fields on a form page.

Elsewhere we have use `this.nodes` to contain other elements on a page.

test/e2e/pages/allocation-cancel.js
```diff
@@ -0,0 +1,18 @@
+import { Selector } from 'testcafe'
+
+import Page from './page'
+
+class AllocationCancelPage extends Page {
+  constructor() {
+    super()
+    this.fields = {
```
teneightfive, line 8: This (`this.fields`) feels like a copy/paste from another file but now doesn't really fit. These elements aren't _fields_. This object was used in other page objects to contain the fields on a form page.

Elsewhere we have use `this.nodes` to contain other elements on a page.

test/e2e/pages/allocation-criteria.js
```diff
@@ -0,0 +1,61 @@
...
+        type: 'radio',
+      },
+      {
+        selector: this.fields.hasOtherCriteria,
+        type: 'radio',
+        value: 'Yes',
+      },
+      {
+        selector: this.fields.otherCriteria,
+        value: 'lorem ipsum',
```
teneightfive, line 55: You could use `faker.lorem.sentence(6)` to generate the data here like has been done previously.

test/e2e/pages/allocation-journey.js
```diff
@@ -0,0 +1,52 @@
...
+    await this.submitForm()
+  }
+
+  findErrorInList(hrefAttribute) {
+    return allocationDetailsPage.fields.errorSummaryList
+      .child('a')
+      .withAttribute('href', hrefAttribute)
+  }
+
+  scrollToBottom() {
```
teneightfive, line 46: This method (`scrollToBottom`) and the methods around validation `findErrorInList`/`triggerValidationOnAllocationDetailsPage` feel like they would be better placed in a more generic place, like the main `Page` page object.

test/e2e/pages/allocation-journey.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  async triggerValidationOnAllocationDetailsPage() {
+    await this.submitForm()
+  }
+
+  async triggerValidationOnAllocationCriteriaPage() {
+    await allocationDetailsPage.fill()
+    await this.submitForm()
+    await this.submitForm()
```
teneightfive, line 37: What's the reason for needing to submit the form twice?

test/e2e/pages/allocation-journey.js
```diff
@@ -0,0 +1,52 @@
...
+    await allocationCriteriaPage.fill()
+    await this.submitForm()
+    return t
+      .expect(this.getCurrentUrl())
+      .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}\/confirmation$/)
+  }
+
+  async triggerValidationOnAllocationDetailsPage() {
+    await this.submitForm()
```
teneightfive, line 31: Doesn't feel like we need to have this method as it's calling a method that already exists on the base Page object to submit the form.

Instead of 

```javascript
await allocationJourney.triggerValidationOnAllocationDetailsPage()
```

you can just use

```javascript
await page.submitForm()
```

test/e2e/pages/allocation-journey.js
barbaracassani: To trigger the errors on the second page, as it's submitted without being filled.

test/e2e/allocation.new.test.js
barbaracassani: I can't quite see how it's misleading... it's what it does! However, it's a good idea to test its result, and I have now added it.

test/e2e/pages/allocation-view.js
```diff
@@ -0,0 +1,29 @@
+import { Selector } from 'testcafe'
+
+import Page from './page'
+
+class AllocationViewPage extends Page {
+  getValueElementByKeyText(text) {
```
teneightfive, line 6: This looks very similar to `page.getDlDefinitionByKey`. Can the existing method not be used to determine the same outcome?

test/e2e/allocation.cancel.test.js
```diff
@@ -0,0 +1,26 @@
...
+fixture('Cancel allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+  await allocationJourney.createAllocation()
+  await t.navigateTo(allocations)
+})
+test('Cancel allocation', async t => {
+  await t.click(
+    allocationJourney.allocationCancelPage.nodes.linkToFirstAllocation
+  )
+  await allocationJourney.scrollToBottom()
```
teneightfive, line 14: Maybe this and `getCurrentUrl` could be called using `page.scrollToBottom()` and `page.getCurrentUrl()` to make it more clear which page object those methods actually exist on.

test/e2e/pages/allocation-view.js
barbaracassani: I hadn't seen that one. Will amend.

test/e2e/allocation.cancel.test.js
barbaracassani: I understand your argument. The reason why I haven't done it is because I am not too keen on importing both parent and child objects in order to consume methods that exist on the parent and are not overridden in the child. It's like throwing a tiny curveball to those who read the codebase, for I act there like there was a specific reason (overridden method? no inheritance at all?) to not consume everything from the child. Makes sense?

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,40 @@
+import { pmuUser } from './_roles'
+import { newAllocation } from './_routes'
+import { allocationJourney } from './pages/'
+
+fixture('New PMU allocation').beforeEach(async t => {
+  await t.useRole(pmuUser).navigateTo(newAllocation)
+})
+test('Create allocation and verify the result', async t => {
+  await allocationJourney.createAllocation()
+  const currentUrl = await allocationJourney.getCurrentUrl()
```
teneightfive, line 10: I have a feeling this is the kind of thing that caused some flakiness with tests previously.

I think it's safer to call `getCurrentUrl()` within the expect call:

```javascript
await t
    .expect(allocationJourney.getCurrentUrl())
    .match(...)
```

test/e2e/allocation.new.test.js
```diff
@@ -0,0 +1,40 @@
...
+        allocationJourney.allocationViewPage.nodes[section].selector,
+        o
+      )
+      await t.expect(textExists).ok()
+    }
+  })
+})
+test('Check validation errors on allocation details page', async t => {
+  await allocationJourney.submitForm()
+  for (const item of allocationJourney.allocationDetailsPage.nodes
```
teneightfive, line 27: I think this is the kind of thing that would've been nice to extract to the main page object. To have a shared method to check the error messages on a form page. 

You're already doing it twice in this file and it's something we can definitely use elsewhere to improve our test suite.

test/e2e/allocation.cancel.test.js
```diff
@@ -0,0 +1,26 @@
...
+  await t.click(allocationJourney.allocationCancelPage.nodes.cancelLink)
+  await t
+    .expect(allocationJourney.getCurrentUrl())
+    .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}\/cancel\/reason$/)
+  await t.click(allocationJourney.allocationCancelPage.nodes.submitButton)
```
teneightfive, line 19: Can we use `page.submitForm()` for consistency please. Then you don't need to create `allocationJourney.allocationCancelPage.nodes.submitButton` either.

test/e2e/pages/allocation-criteria.js
```diff
@@ -0,0 +1,62 @@
...
+
+import { fillInForm } from '../_helpers'
+
+import Page from './page'
+
+class AllocationCriteriaPage extends Page {
+  constructor() {
+    super()
+    this.url = '/allocation/new/allocation-criteria'
+    this.nodes = {
```
teneightfive, line 12: I think you may have misunderstood a previous comment. The name of this property now doesn't fit what it contains. This property does contain _fields_ so it would be better to use `this.fields = {}`. 

The other comment was when `this.fields` contained non-field elements of the page.

A clearer way to define this would probably be:

```javascript
this.fields = {
  prisonerCategory: Selector('#prisoner_category'),
  ...
}
this.errorMessages = [
  Selector('a').withAttribute('href','#prisoner_category'),
  ...
]
```

test/e2e/pages/allocation-cancel.js
```diff
@@ -0,0 +1,18 @@
...
+class AllocationCancelPage extends Page {
+  constructor() {
+    super()
+    this.nodes = {
+      linkToFirstAllocation: Selector('.govuk-table td:first-of-type a'),
+      cancelLink: Selector('a').withText('Cancel allocation'),
+      statusHeading: Selector(
+        '.app-message.app-message--temporary .app-message__heading'
+      ).withText('Allocation cancelled'),
+      submitButton: Selector('.govuk-button').withAttribute('type', 'submit'),
```
teneightfive, line 14: This can be removed and instead just call `page.submitForm()`

test/e2e/allocation.cancel.test.js
teneightfive: Let's leave this for a future improvement actually.

test/e2e/allocation.new.test.js
teneightfive: Let's leave this for a future improvement actually.

test/e2e/pages/allocation-details.js
```diff
@@ -0,0 +1,54 @@
...
+
+import { fillInForm } from '../_helpers'
+
+import Page from './page'
+
+class AllocationDetailsPage extends Page {
+  constructor() {
+    super()
+    this.url = '/allocation/new/allocation-details'
+    this.nodes = {
```
teneightfive, line 12: Could we also rename this to `this.fields` and `this.nodes` please. I think it would continue to upload the previous pattern and make it clear what the fields on the page are, compared to other elements.

test/e2e/allocation.new.test.js
barbaracassani: This is not a match as in, expect something to match. It's match as in, extract a piece of a string. How that would work if I put it in an expect block?

barbaracassani: Btw, so far this particular step has caused no flakiness whatsoever. The only thing that has created problems is the interaction with elements off the viewport.

teneightfive: What's the goal? To navigate to the actual allocation using an ID?

It feels like it would be nicer and cleaner if `allocationJourney.createAllocation()` returned you the allocation, along with the ID. But I get that's its not as easy as the other fixtures because it's not hitting the API and therefore returning the ID.

How about placing the ID somewhere in the DOM? We are [doing something](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/23eb7e810f18c5117dc1923c10dfcd4480f23713) similar to know which current location a user has selected and it would be a nicer way to get that information than having to get it from the URL via a regex. I think the `getCurrentUrl` hasn't felt robust enough so the more we can avoid using it, I think, the better.

You could also add it to the template using a `data-` attribute to something like the confirmation banner. That way, in `allocationJourney.createAllocation()` you can retrieve it from there and return it as part of that method.

test/e2e/pages/allocation-details.js
barbaracassani: Ok

test/e2e/pages/allocation-criteria.js
barbaracassani: Ok, I have separated the error bits.

test/e2e/allocation.new.test.js
teneightfive: After that chat on slack I think just moving this logic to the `createAllocation` would be enough here and would tidy this up a little.

So that method then returns the ID and we have it contained in one place.

PR #536, opened 2020-05-21
[P4 1310] Display cancelled allocations

barbaracassani:
Screenshot 2020-05-21 at 11 06 43

This change aims to display the cancelled allocations in the dashboard. Currently not supported by the api service -- the implementation should be merged soon.

### Checklist

### DIFF
--- a/app/allocations/middleware/set-results.allocations.js
+++ b/app/allocations/middleware/set-results.allocations.js
@@ -3,17 +3,20 @@ const allocationService = require('../../../common/services/allocation')
 
 async function setResultsAllocations(req, res, next) {
   try {
-    const allocations = await allocationService.getByDateAndLocation(
-      req.body.allocations
-    )
+    const [activeAllocations, cancelledAllocations] = (
+      await Promise.all([
+        allocationService.getActiveAllocations(req.body.allocations),
+        allocationService.getCancelledAllocations(req.body.allocations),
+      ])
+    ).map(response => response.flat())
 
     req.results = {
-      active: allocations,
-      cancelled: [],
+      active: activeAllocations,
+      cancelled: cancelledAllocations,
     }
     req.resultsAsTable = {
-      active: presenters.allocationsToTable(allocations),
-      cancelled: [],
+      active: presenters.allocationsToTable(activeAllocations),
+      cancelled: presenters.allocationsToTable(cancelledAllocations),
     }
 
     next()
--- a/app/allocations/middleware/set-results.allocations.test.js
+++ b/app/allocations/middleware/set-results.allocations.test.js
@@ -9,6 +9,10 @@ const mockActiveMoves = [
   { id: '3', foo: 'bar', status: 'completed' },
   { id: '4', fizz: 'buzz', status: 'completed' },
 ]
+const mockCancelledMoves = [
+  { id: '5', foo: 'bar', status: 'cancelled' },
+  { id: '6', fizz: 'buzz', status: 'cancelled' },
+]
 
 describe('Allocations middleware', function() {
   describe('#setResultsAllocations()', function() {
@@ -17,7 +21,8 @@ describe('Allocations middleware', function() {
     let next
 
     beforeEach(function() {
-      sinon.stub(allocationService, 'getByDateAndLocation')
+      sinon.stub(allocationService, 'getActiveAllocations')
+      sinon.stub(allocationService, 'getCancelledAllocations')
       sinon.stub(presenters, 'allocationsToTable').returnsArg(0)
       next = sinon.stub()
       res = {}
@@ -34,17 +39,25 @@ describe('Allocations middleware', function() {
 
     context('when service resolves', function() {
       beforeEach(async function() {
-        allocationService.getByDateAndLocation.resolves(mockActiveMoves)
+        allocationService.getActiveAllocations.resolves(mockActiveMoves)
+        allocationService.getCancelledAllocations.resolves(mockCancelledMoves)
         await middleware(req, res, next)
       })
 
       it('should call the data service with request body', function() {
         expect(
-          allocationService.getByDateAndLocation
+          allocationService.getActiveAllocations
         ).to.have.been.calledOnceWithExactly({
-          status: 'proposed',
-          moveDate: ['2019-01-01', '2019-01-07'],
           fromLocationId: '123',
+          moveDate: ['2019-01-01', '2019-01-07'],
+          status: 'proposed',
+        })
+        expect(
+          allocationService.getCancelledAllocations
+        ).to.have.been.calledOnceWithExactly({
+          fromLocationId: '123',
+          moveDate: ['2019-01-01', '2019-01-07'],
+          status: 'proposed',
         })
       })
 
@@ -52,7 +65,7 @@ describe('Allocations middleware', function() {
         expect(req).to.have.property('results')
         expect(req.results).to.deep.equal({
           active: mockActiveMoves,
-          cancelled: [],
+          cancelled: mockCancelledMoves,
         })
       })
 
@@ -60,13 +73,21 @@ describe('Allocations middleware', function() {
         expect(req).to.have.property('resultsAsTable')
         expect(req.resultsAsTable).to.deep.equal({
           active: mockActiveMoves,
-          cancelled: [],
+          cancelled: mockCancelledMoves,
         })
       })
 
       it('should call allocationsToTable presenter', function() {
-        expect(presenters.allocationsToTable).to.be.calledOnceWithExactly(
+        expect(presenters.allocationsToTable).to.be.calledTwice
+      })
+      it('should call allocationsToTable presenter for active moves', function() {
+        expect(presenters.allocationsToTable.firstCall.firstArg).to.deep.equal(
           mockActiveMoves
+        )
+      })
+      it('should call allocationsToTable presenter for cancelled moves', function() {
+        expect(presenters.allocationsToTable.secondCall.firstArg).to.deep.equal(
+          mockCancelledMoves
         )
       })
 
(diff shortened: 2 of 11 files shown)


app/allocations/middleware/set-results.allocations.js
```diff
@@ -3,17 +3,27 @@ const allocationService = require('../../../common/services/allocation')
 
 async function setResultsAllocations(req, res, next) {
   try {
-    const allocations = await allocationService.getByDateAndLocation(
-      req.body.allocations
-    )
+    const [activeAllocations, cancelledAllocations] = (
+      await Promise.all([
+        allocationService.getByDateAndLocation({
```
teneightfive, line 8: Can we extract this to the service level. Any hardcoded statuses that relate to the API should live in the service so that we contain that logic to one place.

This file should make one call to something like `allocationService.getActive(req.body.allocations)` and `allocationService.getCancelled(req.body.allocations)`.

This layer should be super simple and just call the relevant service method.

common/templates/collection-as-table.njk
```diff
@@ -30,4 +30,24 @@
...
     }) }}
   {% endif %}
+
+  {% if resultsAsTable.cancelled.rows.length %}
+    {% set html %}
+      <div>
+        {{ govukTable({
+          rows: resultsAsTable.cancelled.rows,
+          head: resultsAsTable.cancelled.head,
+          classes: "govuk-secondary-text-colour"
```
teneightfive, line 40: Is there a reason to apply this knocked back colour to the whole table?

I think the results being nested within the details is enough to show they are linked to the cancelled action.

common/presenters/table/object-to-table-head.js
```diff
@@ -1,8 +1,11 @@
 const i18n = require('../../../config/i18n')
 
 function objectToTableHead(schemaEl) {
+  const { head } = schemaEl
+  const prop = head.text ? 'text' : 'html'
```
teneightfive, line 5: Minor detail but the default should really be `text`. Using the `html` property on GOV.UK Design System components essentially uses the `| safe` filter in nunjucks which will output the value unescaped.

It is safer to default to using `text` with these components and only use `html` when we need to.

common/templates/collection-as-table.njk
barbaracassani: This is something @Andy-Millar explicitly requested.

common/presenters/table/object-to-table-head.js
barbaracassani: Actually it's a good point, shall amend.

app/allocations/middleware/set-results.allocations.test.js
```diff
@@ -34,40 +38,61 @@ describe('Allocations middleware', function() {
 
     context('when service resolves', function() {
       beforeEach(async function() {
-        allocationService.getByDateAndLocation.resolves(mockActiveMoves)
+        allocationService.getByDateAndLocation
```
teneightfive, line 41: These stubs can be updated, you're not longer actually calling this method within this middleware so they are no longer feel like the right thing to stub.

That way you can continue to use `calledOnceWithExactly` on each call too.

app/allocations/middleware/set-results.allocations.test.js
```diff
@@ -34,40 +38,61 @@ describe('Allocations middleware', function() {
...
           active: mockActiveMoves,
-          cancelled: [],
+          cancelled: mockCancelledMoves,
         })
       })
 
       it('should call allocationsToTable presenter', function() {
-        expect(presenters.allocationsToTable).to.be.calledOnceWithExactly(
+        expect(presenters.allocationsToTable).to.be.calledTwice
+        expect(presenters.allocationsToTable.firstCall.firstArg).to.deep.equal(
```
teneightfive, line 90: It might actually be nice to split these tests out. It's a bit easier to debug with trying to keep an approach of one assertion to an `it` block. So you could have:

```javascript
it('should call allocationsToTable presenter correct number of times', function() {
  ...
})

it('should call allocationsToTable presenter for active moves', function() {
  ...
})

it('should call allocationsToTable presenter for cancelled moves', function() {
  ...
})
```

common/presenters/table/object-to-table-head.test.js
```diff
@@ -18,7 +25,10 @@ describe('table head presenter', function() {
     const output = schemaExample.map(objectToTableHead)
     expect(output).to.deep.equal([
       {
-        html: 'move_type::label',
+        text: 'move_type::label',
```
teneightfive, line 28: It doesn't really feel like the update to this test file explains well the new behaviour. 

The intended behaviour is a little hidden. It would be good to have separate tests using `context` to describe the two scenarios:

```javascript
context('with text content', function() {
  ...
})

context('with html content', function() {
  ...
})

context('with attributes', function() {
  ...
})
```

common/services/allocation.js
```diff
@@ -41,19 +41,34 @@ const allocationService = {
...
         'filter[date_from]': moveDateFrom,
         'filter[date_to]': moveDateTo,
       }),
     })
   },
+  getActiveAllocations(allocationsParams) {
+    return allocationService.getByDateAndLocation({
+      ...allocationsParams,
+      // TODO the status may or may not be "proposed"
+      status: 'proposed',
```
teneightfive, line 63: I believe this could likely be `['filled', 'unfilled']`?

common/templates/collection-as-table.njk
teneightfive: Then maybe it could be something that lives in the presenter. Their purpose is to transform something into the expected output for a component. 

See the example above how the output of the presenter for active moves is passed straight into the component.

teneightfive: Actually, let's just leave this for now to get this in.

common/assets/scss/helpers/_typography.scss
```diff
@@ -4,3 +4,7 @@
 
 // Ensure font override is loaded
 @include _inter-ui-font-face();
+
+.govuk-secondary-text-colour {
```
teneightfive, line 8: I'm not sure this is the correct place for this class. The Design System readme [suggests that helpers](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/README.md#helpers) "should not output any CSS".

Maybe this class would live better within [utilities](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/README.md#utilities) - "Utility classes – for example clearing floats, visually hiding content"

common/presenters/table/object-to-table-head.js
```diff
@@ -1,8 +1,11 @@
 const i18n = require('../../../config/i18n')
 
 function objectToTableHead(schemaEl) {
+  const { head } = schemaEl
+  const prop = head.html ? 'html' : 'text'
```
teneightfive, line 5: I think this can be improved a little to match the [`object-to-table-row`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/53987ba7e66e6ef1e8b57065a7c62e8cd872ae15/common/presenters/table/object-to-table-row.js#L17-L21) solution as currently if neither `html` or `text` are actually passed to this function it will still output an empty `text` property.

common/assets/scss/utilities/_typography.scss
```diff
@@ -0,0 +1,3 @@
+.govuk-secondary-text-colour {
```
teneightfive, line 1: The only issue with using the `.govuk-` prefix here is that it suggests that the definition is coming from the GOV.UK Design System. 

The namespace pattern allows somebody to see which design system that definition exists in. I'm wondering if it's better to use the `.app-` prefix so its clear it is a definition within our local design system, not the GOV.UK one.

It could also lead to unintended side effects if, in later version they introduce the same class with more style definitions.

common/presenters/table/object-to-table-head.js
teneightfive: This change still produces the same error where if neither `html` or `text` properties are passed, a `text` property is still output with an empty string.

common/presenters/table/object-to-table-head.test.js
```diff
@@ -1,28 +1,84 @@
...
 const objectToTableHead = require('./object-to-table-head')
 
 describe('table head presenter', function() {
-  it('returns the heading specified', function() {
+  beforeEach(function() {
     sinon.stub(i18n, 't').returnsArg(0)
-    const output = schemaExample.map(objectToTableHead)
-    expect(output).to.deep.equal([
+  })
+  context('with html', function() {
```
teneightfive, line 9: I think you're missing two scenarios from this test:

`with both html and text`

and 

`with neither html or text`

common/templates/collection-as-table.njk
```diff
@@ -30,4 +30,24 @@
...
+        }) }}
+      </div>
+    {% endset %}
+    {{ govukDetails({
```
teneightfive, line 44: Maybe add a bit of breathing room here. It can help with scanning to use a new line between blocks when they are at the same indentation.

common/templates/collection-as-table.njk
```diff
@@ -30,4 +30,24 @@
       }
     }) }}
   {% endif %}
+
+  {% if resultsAsTable.cancelled.rows.length %}
+    {% set html %}
+      <div>
```
teneightfive, line 36: Do we need this extra `div` or can we just use the `table` element that is output from `govukTable`?

common/presenters/table/object-to-table-head.js
barbaracassani: I will add `pickBy` but to be honest, I don't quite see a scenario in which this matters. Maybe you can help me figure out one?

common/presenters/table/object-to-table-head.test.js
barbaracassani: Again, I will add the tests, but how is this a realistic scenario?

teneightfive: Fine, let's leave this scenario for now. My point was more around the tests not describing what the current intended behaviour is in those scenarios.

It doesn't show that if you pass neither or both what actually happens. Which means that if the behaviour were to change in the future you wouldn't see the side effects based on the original intended behaviour. 

I think the tests can be more descriptive about what this presenter does in certain scenarios rather than just creating one high level test that ticks the code coverage box by running over each line at least once.

PR #537, opened 2020-05-21
Add progress column to allocations table

teneightfive:
## Proposed changes

This set of changes adds the progress column to the allocations table so that different users can see an overview of the state of the allocations in that time period.

### All locations (as a Population Management Unit user)

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

### Single locations (as a Operational Capacity Allocation user)

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |


Andy-Millar:
We should update the CTA to be "Create a move >" rather than create an allocation. Unifies the content, and we specifically pressed this language during testing and had no issues with the term 'move'

Andy-Millar:
Otherwise all good - I've just realised this PR is specifically for progress column!

common/presenters/table/object-to-table-head.js
```diff
@@ -2,12 +2,16 @@ const { pickBy } = require('lodash')
...
-  const prop = head.html ? 'html' : 'text'
+function objectToTableHead({ head } = {}) {
+  if (!head) {
+    return undefined
+  }
+
   return pickBy({
     ...head,
-    [prop]: i18n.t(schemaEl.head[prop]),
+    html: head.html ? i18n.t(head.html) : undefined,
```
solidgoldpig, line 12: Not sure why this change was necessary - the dynamic prop seems more readable to me. Is there any reason the `pickBy` is actually necessary?

app/allocations/middleware/set-results.allocations.test.js
```diff
@@ -19,14 +20,24 @@ describe('Allocations middleware', function() {
...
       sinon.stub(allocationService, 'getCancelledAllocations')
-      sinon.stub(presenters, 'allocationsToTable').returnsArg(0)
+      sinon
+        .stub(permissions, 'check')
+        .returns(false)
+        .withArgs('allocation:person:assign', ['allocation:person:assign'])
+        .returns(true)
+      sinon
+        .stub(presenters, 'allocationsToTableComponent')
+        .callsFake(() => allocationsToTableStub)
```
solidgoldpig, line 36: Any reason to not use `returns` rather than `callsFake`?

common/presenters/allocations-to-table-component.js
```diff
@@ -0,0 +1,129 @@
...
+const filters = require('../../config/nunjucks/filters')
+const componentService = require('../services/component')
+
+const tablePresenters = require('./table')
+
+function tagByRemaining(totalSlots, unfilledSlots) {
+  const isComplete = unfilledSlots === 0
+  const inProgress = unfilledSlots !== totalSlots
+  const text = i18n.t('collections::labels.progress_status', {
+    context: inProgress ? (isComplete ? 'complete' : 'remaining') : '',
```
solidgoldpig, line 11: `const context = inProgress ? (isComplete ? 'complete' : 'remaining') : ''`

(Still find that a bit of an overloaded double ternary)

and then this just provide an object to map the classes

```
const contextClasses = {
  complete: 'govuk-tag--green',
  remaining: 'govuk-tag-yellow',
  default: 'govuk-tag--red'
}
...
const classes = contextClasses[context] || contextClasses.default
```
(or similar)

common/presenters/table/object-to-table-head.js
barbaracassani: The `pickBy` is meant to whittle away the empty props. However, I agree that the dynamic prop seems more readable.

app/allocations/middleware/set-results.allocations.test.js
teneightfive: Ooo, didn't think of that...I'll give it a whirl!

common/presenters/table/object-to-table-head.js
teneightfive: Yep, `pickBy` is being used to ensure we don't end up with something like this:

```
{
  html: undefined,
  text: '',
}
```

The thing I was trying to prevent was an empty `text` property being output if provided with neither `html` or `text`. Ie, it will only translate `html` and `text` if you actually provide them and it will clean out any falsy properties provided.

common/presenters/allocations-to-table-component.js
teneightfive: That does read much nicer. And then can actually move that out of those methods to avoid duplication.

common/presenters/table/object-to-table-head.js
barbaracassani: But why should one add both `html` and `text`? Wouldn't that be a self-evident dev error?

teneightfive: It's not the both scenario this prevents, its the neither. It's preventing a misleading situation where you provide nothing, but you get a `text` property with an empty string.

As you say, if you don't provide either, it's likely an error, so this shouldn't silently provide you with a `text` property. It should return you back nothing because you haven't provided the correct properties.

barbaracassani: If we decide we want to prevent dev errors, maybe we should introduce type checks.

PR #548, opened 2020-05-27
fix(allocations): Copy changes on the allocation(s) app

barbaracassani:
Changes made:

* allocations dashboards: Column title from 'Date' ‘Date of travel’ 
* allocations dashboards: filter name from 'Total' to ‘Total allocations’
* Individual allocations summary record page ‘Estate type’ prison hidden from view
* OCA adds; Page title for the first page becomes ‘Who do you want to add to this allocation?’
* OCA adds: added right hand column to the search person and search results page

Also: renamed / moved a few things for better readability.

Screenshot 2020-06-01 at 09 43 57

### Checklist

### DIFF
--- a/app/move/controllers/assign/base.js
+++ b/app/move/controllers/assign/base.js
@@ -1,6 +1,9 @@
+const { pick } = require('lodash')
+
+const presenters = require('../../../../common/presenters')
 const CreateBaseController = require('../create/base')
 
-class CreateAllocationBaseController extends CreateBaseController {
+class AssignBaseController extends CreateBaseController {
   middlewareLocals() {
     this.use(this.setMove)
     CreateBaseController.prototype.middlewareLocals.apply(this)
@@ -10,6 +13,18 @@ class CreateAllocationBaseController extends CreateBaseController {
   setCancelUrl(req, res, next) {
     const allocationId = res.locals.move.allocation.id
     res.locals.cancelUrl = `/allocation/${allocationId}`
+    next()
+  }
+
+  setMoveSummary(req, res, next) {
+    const currentLocation = req.session.currentLocation
+    // TODO remove pick when the api return move_agreed as null
+    // unless of course further unwanted properties have been added in the meantime
+    const moveSubset = pick(res.locals.move, ['to_location', 'date'])
+    res.locals.moveSummary = presenters.moveToMetaListComponent({
+      ...moveSubset,
+      from_location: currentLocation,
+    })
     next()
   }
 
@@ -29,10 +44,7 @@ class CreateAllocationBaseController extends CreateBaseController {
   }
 }
 
-CreateAllocationBaseController.mixin = function mixin(
-  controller,
-  mixinController
-) {
+AssignBaseController.mixin = function mixin(controller, mixinController) {
   const controllerMethods = Object.getOwnPropertyNames(controller.prototype)
   Object.getOwnPropertyNames(mixinController.prototype)
     .filter(key => key !== 'prototype')
@@ -43,4 +55,4 @@ CreateAllocationBaseController.mixin = function mixin(
     })
 }
 
-module.exports = CreateAllocationBaseController
+module.exports = AssignBaseController
--- a/app/move/controllers/assign/base.test.js
+++ b/app/move/controllers/assign/base.test.js
@@ -1,4 +1,6 @@
+const presenters = require('../../../../common/presenters')
 const CreateBaseController = require('../../../move/controllers/create/base')
+
 const BaseProto = CreateBaseController.prototype
 
 const UpdateBaseController = require('./base')
@@ -32,10 +34,42 @@ describe('Assign controllers', function() {
     })
 
     describe('#setMoveSummary()', function() {
-      it('should inherit setMoveSummary from CreateBaseController', function() {
-        expect(controller.setMoveSummary).to.exist.and.equal(
-          BaseProto.setMoveSummary
+      let locals
+      let next
+      beforeEach(function() {
+        next = sinon.stub()
+        sinon
+          .stub(presenters, 'moveToMetaListComponent')
+          .returns({ summary: {} })
+        locals = {
+          move: {
+            to_location: 'b',
+            other_prop: 'c',
+          },
+        }
+        controller.setMoveSummary(
+          {
+            session: {
+              currentLocation: 'a',
+            },
+          },
+          {
+            locals,
+          },
+          next
         )
+      })
+      it('creates moveSummary on the locals', function() {
+        expect(locals.moveSummary).to.exist
+        expect(locals.moveSummary).to.deep.equal({ summary: {} })
+      })
+      it('invokes moveToMetaListComponent with a subset of properties', function() {
+        expect(
+          presenters.moveToMetaListComponent
+        ).to.have.been.calledWithExactly({
+          from_location: 'a',
+          to_location: 'b',
+        })
       })
     })
 
@@ -81,7 +115,7 @@ describe('Assign controllers', function() {
     })
 
     it('should call correct number of additional middleware', function() {
-      expect(controller.use).to.be.callCount(2)
+      expect(controller.use).to.have.been.calledTwice
     })
   })
 
--- a/app/move/controllers/assign/person-search-results.js
+++ b/app/move/controllers/assign/person-search-results.js
@@ -1,8 +1,8 @@
 const MoveCreatePersonSearchResultsController = require('../../../move/controllers/create/person-search-results')
 
-const PersonAssignBase = require('./base')
+const AssignBaseController = require('./base')
 
-class PersonSearchResultsController extends PersonAssignBase {
+class PersonSearchResultsController extends AssignBaseController {
   middlewareSetup() {
     super.middlewareSetup()
     this.use(this.setPeople)
@@ -15,7 +15,7 @@ class PersonSearchResultsController extends PersonAssignBase {
   }
 }
 
-PersonAssignBase.mixin(
+AssignBaseController.mixin(
   PersonSearchResultsController,
   MoveCreatePersonSearchResultsController
 )
--- a/app/move/steps/create.js
+++ b/app/move/steps/create.js
@@ -15,7 +15,7 @@ const {
 const personSearchStep = {
   controller: PersonSearch,
   buttonText: 'actions::search',
-  template: 'person-search',
+  template: '../person-search',
   pageTitle: 'moves::steps.person_search.heading',
   next: [
     {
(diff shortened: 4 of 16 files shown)


app/move/controllers/assign/base.js
```diff
@@ -13,6 +15,21 @@ class CreateAllocationBaseController extends CreateBaseController {
     next()
   }
 
+  setAllocationBlurb(req, res, next) {
```
teneightfive, line 18: I don't think this is the most efficient place for this method. This is run on every step, but in reality it will only be needed on the step that uses this field.

It would be better in the two controllers related to people. But as discussed on slack I think a better approach would be to use more generic hint text like we have elsewhere and to provide the dynamic information via the right hand panel.

app/move/steps/create.js
```diff
@@ -16,7 +16,7 @@ const {
 const personSearchStep = {
   controller: PersonSearch,
   buttonText: 'actions::search',
-  template: 'person-search',
+  template: '../person-search',
```
teneightfive, line 19: Don't think there is a reason to create custom templates for these steps. It can be updated via the field/step configuration.

locales/en/moves.json
```diff
@@ -5,13 +5,16 @@
   "steps": {
     "person_search": {
       "heading": "Who is being moved?",
+      "heading_add_to_allocation": "Who do you want to add to this allocation?",
```
teneightfive, line 8: The language and naming of these things feels a little disjointed. We're now referring to "add_to_allocation" where previously we've used the term "assign" and related it to the move, not the allocation.

Might be worth keeping a consistent context of `assign` when talking about how we add a person to a move. That's what is really happening underneath, regardless of how we suggest to a user that it is part of the allocation.

app/move/steps/create.js
barbaracassani: I haven't created templates -- I have _moved_ them. They are shared between `create` and `assign` but incongruously, they were in the `create` template folder.

app/move/controllers/assign/base.js
barbaracassani: Indeed. I have run the matter past @Andy-Millar , awaiting feedback.

Andy-Millar:
Quick comment for me - @barbaracassani what happened to the 'Prisoner to be allocated' title for the side panel? We should be displaying that instead 'Allocation details' - as we discussed, it makes the page after this less weird when it displays the person being moved.

barbaracassani:
@Andy-Millar good point -- amended now.

barbaracassani:
@Andy-Millar good point, now addressed.

locales/en/allocation.json
```diff
@@ -14,6 +14,11 @@
     "content": "<p>As this person needs a special vehicle they cannot be part of this allocation.</p>\n<p>If you would still like to move this person you can create a single request for them to move to another prison.",
     "return_link_text": "Return to allocation"
   },
+  "assign": {
+    "sidebar": {
+      "heading": "Prisoner to be allocated"
```
teneightfive, line 19: There is [already a key](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/locales/en/default.json#L20) for this value that we should use to be consistent.

I think we may tweak that content in a future PR to use `prisoner` instead of `person` but for this  change we should use the same key so we avoid having to maintain that copy in two places.

locales/en/allocations.json
```diff
@@ -1,7 +1,8 @@
 {
   "complete": "complete",
   "incomplete": "incomplete",
-  "total": "total",
+  "total": "total allocations",
+  "date": "Date of travel",
```
teneightfive, line 5: I'm sure this content already exists. Would be good to make use  of existing content instead of adding more to maintain.

common/templates/form-wizard.njk
```diff
@@ -87,12 +87,16 @@
 {% endblock %}
 
 {% block contentSidebar %}
-  {% if person %}
+  {% if person or showAllocationDetails %}
```
teneightfive, line 90: It would be good to have this logic not be specific to "allocation details". If you think about what we're showing it's no longer really "allocation" specific so this variable name feels little misleading too.

common/templates/form-wizard.njk
```diff
@@ -87,12 +87,16 @@
...
     <div class="sticky-sidebar">
       <div class="sticky-sidebar__inner">
         {% block summaryPanelContent %}
           <h3 class="govuk-heading-m app-border-top-2 app-border--blue govuk-!-padding-top-4 govuk-!-margin-bottom-3">
-            {{ person.fullname | upper }}
+            {% if person and person.fullname %}
```
teneightfive, line 95: This whole block could be:

```nunjucks
{{ person.fullname | upper or t("awaiting_person") }}
```

feels easier to read and neater.

common/templates/form-wizard.njk
barbaracassani: What would you suggest instead? The situation where we have a move but not a person is completely specific to allocation, so `showMoveDetails` wouldn't be less misleading...

teneightfive: I played around and realised this logic isn't actually working correctly either. This is the new journey now:

image

I don't think we need to tie this to a boolean that we need to specifically set on two particular steps. I got this working correctly using this logic:

```nunjucks
{% if person.id or move.id %}
  ...
{% endif %}
```

teneightfive: Means we don't need to do all the changes to the controllers.

locales/en/moves.json
```diff
@@ -5,6 +5,7 @@
   "steps": {
     "person_search": {
       "heading": "Who is being moved?",
+      "heading_assign_to_allocation": "Who do you want to add to this allocation?",
```
teneightfive, line 8: Could we maybe use consistent terms for these variables. When @solidgoldpig did the original work we are using the term `assign` rather than `assign_to_allocation`. This also fits better the mental model of what is _actually_ happening. 

A person is not actually being assigned to an allocation, rather to the move that was previously created _by_ the allocation. This is why we moved the logic to assign a person to the move part of the application and away from allocation.

Think of the allocation of just a shortcut way to manage a group of moves expected to happen to/from the same place and on the same day. The further actions around adding/removing a person are done on specific moves.

common/templates/form-wizard.njk
barbaracassani: Damn! good spot.

barbaracassani: Damn! Good spot.

barbaracassani: This also solves the issue of the naming...

teneightfive: Exactly :)

app/move/controllers/assign/base.test.js
```diff
@@ -184,4 +220,54 @@ describe('Assign controllers', function() {
...
+      let req
+      let res
+      beforeEach(function() {
+        req = { ...mockReq }
+        res = { locals: { ...mockLocals } }
+        req.form.options.showAllocationDetails = true
+        controller.setAllocationDetails(req, res, next)
+      })
+      afterEach(function() {
+        next.resetHistory()
```
teneightfive, line 244: If you assign next as `sinon.stub()` within the `beforeEach` block it will use the global `afterEach` to reset the stub after each test case.

app/move/controllers/assign/base.js
```diff
@@ -13,6 +16,16 @@ class CreateAllocationBaseController extends CreateBaseController {
     next()
   }
 
+  setMoveSummary(req, res, next) {
+    const currentLocation = req.session.currentLocation
+    const moveSubset = pick(res.locals.move, ['to_location', 'date'])
```
teneightfive, line 21: I realised that this is to strip out the `move_agreed` field, right? As this is set to `false` even though technically is hasn't been answered yet. 

This is covering up that that is infact more of a bug with the API. We started to look to address this as part of prison to court as it should really be set to `null` if it wasn't provided with `POST /moves`. 

This fix also now prevents the answer showing up when the user actually answers that step in this journey. 

I think we should maybe wrap this in a TODO so that we know it needs to be addressed once we solve it in the API.

barbaracassani: You have inferred it correctly. I will add a TODO (and amend the commit wording)

teneightfive: I wonder if actually we should be using `omit` instead of `pick`? That way any other future attributes from the move that the presenter needs to handle display it will still receive?

app/move/controllers/assign/base.js
```diff
@@ -1,6 +1,9 @@
+const { pick } = require('lodash')
+
+const presenters = require('../../../../common/presenters')
 const CreateBaseController = require('../create/base')
 
-class CreateAllocationBaseController extends CreateBaseController {
+class AssignToAllocationBaseController extends CreateBaseController {
```
teneightfive, line 6: Could we do the same with all these naming changes too please and remove the reference to the allocation. I'd either suggest `AssignPersonToMoveBaseController` or `AssignBaseController`. I think the second fits better with the existing naming convention but happy if you want to be more explicit.

app/move/controllers/assign/person-search-results.js
```diff
@@ -1,12 +1,13 @@
...
-const PersonAssignBase = require('./base')
+const AssignToAllocationBaseController = require('./base')
 
-class PersonSearchResultsController extends PersonAssignBase {
+class PersonSearchResultsController extends AssignToAllocationBaseController {
   middlewareSetup() {
     super.middlewareSetup()
     this.use(this.setPeople)
     this.use(this.setPeopleItems)
+    this.use(this.setMoveSummary)
```
teneightfive, line 10: Is this not already being called as part of `super.middlewareLocals`? I'd say that's probably the more logical place to actually make this call.

barbaracassani: Yup, good point.

app/move/controllers/assign/base.js
barbaracassani: To be honest, I intentionally went for `pick` to avoid unintentionally displaying other things that may be added in the future. The requirements here call for a specific set of properties displayed, not for `everything but`.

common/templates/form-wizard.njk
barbaracassani: Done 👍

app/move/controllers/assign/base.js
teneightfive: But that's more the responsibility of the presenter than this method I'd say. That's why we only destructure particular properties within the presenter so that we're not displaying everything.

Let's leave this for now. I've raised [a ticket](https://dsdmoj.atlassian.net/browse/P4-1608) to address this issue in the correct place.

We will need to ensure this is reverted to correctly display the answer to the move agreed after it has been answered during this journey to consistently follow the pattern we have used elsewhere in the service.

Issue #495, opened 2020-05-07
Use rewire instead of proxyquire


solidgoldpig:
`rewire` is much better than `proxyquire` (says its author)

https://dev.to/thekashey/please-stop-playing-with-proxyquire-11j4

But it is really, so we should probably use it.

PR #549, opened 2020-05-28
P4 1349 remove person from allocation

solidgoldpig:
Adds unassign route to move

Remove_person_from_allocation_-_Cancel_allocation

Adds unassign link to allocation view and suppresses status badge

| Before | After |
| ------ | ----- |
|View_allocation|View_allocation|

### Checklist

### DIFF
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -4,7 +4,12 @@ module.exports = function view(req, res) {
   const { allocation } = res.locals
   const { moves } = allocation
   const bannerStatuses = ['cancelled']
+
   const movesWithoutPerson = moves.filter(move => !move.person)
+  const moveToCardComponent = presenters.moveToCardComponent({
+    showStatus: false,
+  })
+
   const locals = {
     dashboardUrl: '/allocations',
     /* eslint-disable indent */
@@ -22,7 +27,7 @@ module.exports = function view(req, res) {
       emptySlots: movesWithoutPerson.length,
       filledSlots: moves
         .filter(move => move.person)
-        .map(presenters.moveToCardComponent()),
+        .map(move => ({ ...move, ...moveToCardComponent(move) })),
     },
   }
   res.render('allocation/views/view', locals)
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -27,7 +27,11 @@ module.exports = function view(req, res) {
       emptySlots: movesWithoutPerson.length,
       filledSlots: moves
         .filter(move => move.person)
-        .map(move => ({ ...move, ...moveToCardComponent(move) })),
+        .map(move => ({
+          id: move.id,
+          fullname: move.person.fullname,
+          card: moveToCardComponent(move),
+        })),
     },
   }
   res.render('allocation/views/view', locals)
--- a/app/allocation/controllers/view.test.js
+++ b/app/allocation/controllers/view.test.js
@@ -1,3 +1,5 @@
+const { expect } = require('chai')
+
 const presenters = require('../../../common/presenters')
 
 const handler = require('./view')
@@ -118,9 +120,14 @@ describe('view allocation', function() {
     })
 
     it('calls render with the template', function() {
-      expect(mockRes.render).to.have.been.calledWithExactly(
+      expect(
+        JSON.parse(JSON.stringify(mockRes.render.getCall(0).args))
+      ).to.deep.equal([
         'allocation/views/view',
         {
+          dashboardUrl: '/allocations',
+          messageContent: 'allocations::statuses.description',
+          unassignedMoveId: '123',
           allocationDetails: {},
           allocationSummary: {},
           allocationPeople: {
@@ -128,26 +135,16 @@ describe('view allocation', function() {
             filledSlots: [
               {
                 id: '456',
-                person: {
-                  first_names: 'John',
-                  last_name: 'Doe',
-                },
+                person: { first_names: 'John', last_name: 'Doe' },
               },
               {
                 id: '011',
-                person: {
-                  first_names: 'Phil',
-                  last_name: 'Jones',
-                },
+                person: { first_names: 'Phil', last_name: 'Jones' },
               },
             ],
           },
-          unassignedMoveId: '123',
-          messageContent: 'allocations::statuses.description',
-          messageTitle: undefined,
-          dashboardUrl: '/allocations',
-        }
-      )
+        },
+      ])
     })
   })
 
@@ -169,6 +166,33 @@ describe('view allocation', function() {
     it('does create a message banner content', function() {
       expect(locals.messageContent).to.equal(
         'allocations::statuses.description'
+      )
+    })
+
+    it('calls render with the template', function() {
+      expect(mockRes.render).to.be.calledOnceWithExactly(
+        'allocation/views/view',
+        {
+          dashboardUrl: '/allocations',
+          messageTitle: 'allocations::statuses.cancelled',
+          messageContent: 'allocations::statuses.description',
+          unassignedMoveId: '123',
+          allocationDetails: {},
+          allocationSummary: {},
+          allocationPeople: {
+            emptySlots: 2,
+            filledSlots: [
+              {
+                id: '456',
+                person: { first_names: 'John', last_name: 'Doe' },
+              },
+              {
+                id: '011',
+                person: { first_names: 'Phil', last_name: 'Jones' },
+              },
+            ],
+          },
+        }
       )
     })
   })
--- a/app/allocation/controllers/view.test.js
+++ b/app/allocation/controllers/view.test.js
@@ -1,5 +1,3 @@
-const { expect } = require('chai')
-
 const presenters = require('../../../common/presenters')
 
 const handler = require('./view')
--- a/app/allocation/controllers/view.test.js
+++ b/app/allocation/controllers/view.test.js
@@ -49,6 +49,7 @@ describe('view allocation', function() {
         person: {
           first_names: 'John',
           last_name: 'Doe',
+          fullname: 'John Doe',
         },
       },
       { id: '789' },
@@ -57,6 +58,7 @@ describe('view allocation', function() {
         person: {
           first_names: 'Phil',
           last_name: 'Jones',
+          fullname: 'Phil Jones',
         },
       },
     ],
@@ -135,11 +137,27 @@ describe('view allocation', function() {
             filledSlots: [
               {
                 id: '456',
-                person: { first_names: 'John', last_name: 'Doe' },
+                fullname: 'John Doe',
+                card: {
+                  id: '456',
+                  person: {
+                    first_names: 'John',
+                    last_name: 'Doe',
+                    fullname: 'John Doe',
+                  },
+                },
               },
               {
                 id: '011',
-                person: { first_names: 'Phil', last_name: 'Jones' },
+                fullname: 'Phil Jones',
+                card: {
+                  id: '011',
+                  person: {
+                    first_names: 'Phil',
+                    last_name: 'Jones',
+                    fullname: 'Phil Jones',
+                  },
+                },
               },
             ],
           },
@@ -184,11 +202,27 @@ describe('view allocation', function() {
             filledSlots: [
               {
                 id: '456',
-                person: { first_names: 'John', last_name: 'Doe' },
+                fullname: 'John Doe',
+                card: {
+                  id: '456',
+                  person: {
+                    first_names: 'John',
+                    last_name: 'Doe',
+                    fullname: 'John Doe',
+                  },
+                },
               },
               {
                 id: '011',
-                person: { first_names: 'Phil', last_name: 'Jones' },
+                fullname: 'Phil Jones',
+                card: {
+                  id: '011',
+                  person: {
+                    first_names: 'Phil',
+                    last_name: 'Jones',
+                    fullname: 'Phil Jones',
+                  },
+                },
               },
             ],
           },
--- a/app/allocation/views/view.njk
+++ b/app/allocation/views/view.njk
@@ -119,18 +119,20 @@
   {% endif %}
 
   {% for slot in allocationPeople.filledSlots %}
-  <section class="app-!-position-relative">
-    {{ appCard(slot) }}
-    {% if canAccess('allocation:person:assign') %}
-      <p class="app-!-position-top-right govuk-!-margin-top-3">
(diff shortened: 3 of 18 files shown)


app/allocation/views/view.njk
```diff
@@ -119,7 +119,18 @@
   {% endif %}
 
   {% for slot in allocationPeople.filledSlots %}
+  <section class="app-!-position-relative">
     {{ appCard(slot) }}
+    {% if canAccess('allocation:person:assign') %}
+      <p class="app-!-position-top-right govuk-!-margin-top-3">
```
teneightfive, line 125: My concern with this would be if a person's name stretches to the edge of the card, does that mean this button will be positioned over their name and over the link to the move?

app/allocation/views/view.njk
```diff
@@ -119,7 +119,18 @@
   {% endif %}
 
   {% for slot in allocationPeople.filledSlots %}
+  <section class="app-!-position-relative">
```
teneightfive, line 122: Maybe keep consistent indentation with this change. 

I know it comes with the catch of git thinking more has changed than actually has, but I feel it's better to maintain longterm scanability and readability of the code in this instance.

app/allocation/controllers/view.js
```diff
@@ -22,7 +27,7 @@ module.exports = function view(req, res) {
       emptySlots: movesWithoutPerson.length,
       filledSlots: moves
         .filter(move => move.person)
-        .map(presenters.moveToCardComponent()),
+        .map(move => ({ ...move, ...moveToCardComponent(move) })),
```
teneightfive, line 30: This feels a little messy combining the the two properties back together. And could lead to unintended circumstances in the future of properties being overridden.

If, for this solution we need access to the original move I think it would be better to assign the result of the presenter to a separate property. 

Or even better split the original move and the presenter to separate properties for this collection that you send to the view.

Or even, even better. Send an object with the card result and only the properties you need:

```javascript
.map(move => {
  return {
    id: move.id,
    fullname: move.person.fullname,
    card: moveToCardComponent(move),
  }
})
```

locales/en/allocations.json
```diff
@@ -37,10 +37,13 @@
       "empty_slot": "Person to be allocated"
     },
     "people_assign": {
+      "special_vehicle_instruction": "You can’t add people who require a special vehicle to an allocation - you need to <a href=\"{{href}}\">create a single request</a> instead",
+      "no_slots_filled": "No prisoners added to allocation",
```
teneightfive, line 41: What was the outcome of `person` vs `prisoner`? It feels inconsistent to say "Add **person**" on the button but have the text here say "No **prisoners** added...".

Feels like we need to stick to one or the other?

solidgoldpig: `prisoner` was confirmed. Leaving until we raise the issue of content design…

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,56 @@
+const FormWizardController = require('../../../common/controllers/form-wizard')
+const allocationService = require('../../../common/services/allocation')
+const moveService = require('../../../common/services/move')
+
+class UnassignController extends FormWizardController {
+  middlewareLocals() {
```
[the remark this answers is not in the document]
solidgoldpig: 👍

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,56 @@
...
+    const { move } = res.locals
+
+    if (!move.allocation) {
+      return res.redirect(`/move/${move.id}`)
+    }
+    if (!move.person) {
+      return res.redirect(`/allocation/${move.allocation.id}`)
+    }
+
+    const allocation = await allocationService.getById(move.allocation.id)
```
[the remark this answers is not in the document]
solidgoldpig: 👍

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,56 @@
...
+
+    res.locals.person = move.person
+
+    next()
+  }
+
+  async saveValues(req, res, next) {
+    try {
+      const { id } = res.locals.move
+      await moveService.update({
```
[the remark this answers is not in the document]
solidgoldpig: 👍

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,56 @@
...
+        move_agreed: false,
+        move_agreed_by: '',
+      })
+    } catch (err) {
+      return next(err)
+    }
+    super.saveValues(req, res, next)
+  }
+
+  async successHandler(req, res, next) {
```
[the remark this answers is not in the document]
solidgoldpig: 👍

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,56 @@
...
+        person: {
+          id: null,
+        },
+        move_agreed: false,
+        move_agreed_by: '',
+      })
+    } catch (err) {
+      return next(err)
+    }
+    super.saveValues(req, res, next)
```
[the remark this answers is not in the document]
solidgoldpig: 👍

app/move/views/unassign.njk
```diff
@@ -0,0 +1,26 @@
...
+    text: t("actions::back_to_allocation"),
+    href: "/allocation/" + allocation.id
+  }) }}
+
+  {{ super() }}
+{% endblock %}
+
+{% block contentMain %}
+  <p class="govuk-lede">
+    {{ t("allocation::unassign.detail", {
```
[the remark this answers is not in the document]
solidgoldpig: 👍

app/allocation/views/view.njk
solidgoldpig: 👍

app/allocation/controllers/view.js
solidgoldpig: 👍

app/allocation/controllers/view.test.js
```diff
@@ -1,3 +1,5 @@
+const { expect } = require('chai')
```
teneightfive, line 1: Isn't this already available globally?

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,60 @@
...
+
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setMoveRelationships)
+  }
+
+  async setMoveRelationships(req, res, next) {
+    const { move } = res.locals
+
+    const allocation = await allocationService.getById(move.allocation.id)
```
teneightfive, line 32: Just spotted you'll need a `try/catch` around this to catch any promise failures.

That also then led me to ask whether we need to make this call at all as `allocation` is available as a relationship on move: `move.allocation`?

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,60 @@
...
+      const { id } = res.locals.move
+      await moveService.unassign(id)
+      super.saveValues(req, res, next)
+    } catch (err) {
+      next(err)
+    }
+  }
+
+  successHandler(req, res) {
+    const { id: allocationId } = res.locals.allocation
```
teneightfive, line 51: Again, think about the fact an allocation actually exists on the move this could be:

```javascript
const { id: allocationId } = res.locals.move.allocation
```

Then does that mean you don't need to set `res.locals.allocation` in `setMoveRelationships`?

app/move/controllers/unassign.test.js
```diff
@@ -0,0 +1,277 @@
+const proxyquire = require('proxyquire')
+
+const FormWizardController = require('../../../common/controllers/form-wizard')
+
+const allocationGetByIdStub = sinon.stub()
+allocationGetByIdStub.resolves({ id: '__allocation__' })
+const moveUnassignStub = sinon.stub()
+moveUnassignStub.resolves({})
+
+const UnassignController = proxyquire('./unassign', {
```
teneightfive, line 10: You could drop the need for proxyquire altogether here and stub the services using:

```javascript
const allocationService = require('../../../common/services/allocation')

...

beforeEach(function() {
  sinon.stub(allocationService, getById)
})
```

If find this method much cleaner to use where possible. Then you can also drop the need for having to do things like `. resetHistory()` as it's already being done globally with sinon.

app/move/controllers/unassign.test.js
```diff
@@ -0,0 +1,277 @@
...
+          expect(res.redirect).to.not.be.called
+        })
+
+        it('should call next', function() {
+          expect(next).to.be.calledOnceWithExactly()
+        })
+      })
+    })
+
+    describe('#setMoveRelationships()', function() {
```
teneightfive, line 149: This method is missing the failure test if something goes wrong with the promise.

However, you may be able to avoid this altogether if you don't need to make this extra API call - that would be the much preferred solution in my opinion.

app/move/views/unassign.njk
```diff
@@ -0,0 +1,25 @@
...
+      name: person.fullname,
+      move: move,
```
teneightfive, line 16: This isn't being used by the looks of it, so can be dropped?

common/services/move.test.js
```diff
@@ -892,6 +892,47 @@ describe('Move Service', function() {
     })
   })
 
+  describe('#unassign()', function() {
+    context('without move ID', function() {
+      it('should reject with error', function() {
```
teneightfive, line 897: Does this pass? The method doesn't seem to include this like the other methods do.

Would be good to add it though and ensure these tests are functioning as expected.

app/allocation/controllers/view.test.js
solidgoldpig: I have no idea how that got there - I certainly didn't do it manually

app/move/controllers/unassign.js
solidgoldpig: 👍 


As for the necessity of the call, we do need it as we don't have the moves on the move's allocation relationship object. Maybe once we've properly implemented json:api includes we can return to this (and many other bits)

app/move/controllers/unassign.js
solidgoldpig: Have updated to get the id from move's allocation rather than, er, the allocation

See above why we do need the actual allocation

app/move/controllers/unassign.test.js
solidgoldpig: It literally wasn't working when I did it the way you're suggesting previously. But now it works.

I've made the change but I do disagree. It's only because of mocha's lack of parallelisation that this works and also it works for some things and not for others.  (*cough* https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/495)

app/move/views/unassign.njk
solidgoldpig: It is being used.

common/services/move.test.js
solidgoldpig: It passes. It's just called the its own `update` method rather than the api-client's. I don't understand what you mean by this or by adding it?

app/move/controllers/unassign.test.js
solidgoldpig: 👍 to the failure test

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,64 @@
...
+const allocationService = require('../../../common/services/allocation')
+const moveService = require('../../../common/services/move')
+
+class UnassignController extends FormWizardController {
+  middlewareChecks() {
+    this.use(this.checkAllocation)
+    super.middlewareChecks()
+  }
+
+  async checkAllocation(req, res, next) {
```
teneightfive, line 11: Sorry, missed this last time but you can drop the `async` as you're not doing an `await` anymore.

common/services/move.test.js
teneightfive: Ah yes, sorry I missed that. That's probably a nicer way to handle it actually.

app/move/views/unassign.njk
teneightfive: Sorry, I meant `move: move`. That value isn't being used in the translation so we can remove it from being passed into this `t()` call.

app/move/controllers/unassign.js
```diff
@@ -0,0 +1,64 @@
...
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setMoveRelationships)
+  }
+
+  async setMoveRelationships(req, res, next) {
+    try {
+      const { move } = res.locals
+
+      const allocation = await allocationService.getById(move.allocation.id)
```
teneightfive, line 33: Taking another look I don't think we use the rest of the details in this journey so we can remove the need to make this service call and also remove the `async` and `try/catch` from the function.

app/move/controllers/unassign.js
solidgoldpig: ^^ Doh! No, you're right, now that we're not passing the ids for move, allocation and person to double-check what we're removing, there is zero need  for the allocation call.

Zapped.

app/move/controllers/unassign.test.js
```diff
@@ -0,0 +1,263 @@
...
+        it('should redirect to allocation', function() {
+          expect(res.redirect).to.be.calledOnceWithExactly('/allocation/6789')
+        })
+
+        it('should not call next', function() {
+          expect(next).to.not.be.called
+        })
+      })
+
+      context('with allocation and person', async function() {
```
teneightfive, line 106: You can also remove the `async` here as it's no longer an `await` on the function call.

app/move/controllers/unassign.js
solidgoldpig: ^^ See above for why I'm wrong

app/move/views/unassign.njk
solidgoldpig: 👍

app/move/controllers/unassign.test.js
teneightfive: Yeah I'd actually be keen to explore rewire. I'm not really a fan of proxyquire and that's why I try and avoid using it where we can and use `sinon.stub()`. It makes it easier to change the behaviour in nested tests of a stub, otherwise you have to proxyquire within a `beforeEach` if sending the stub in via proxyquire. And that gets...well, messy very quickly.

app/move/controllers/unassign.js
solidgoldpig: 👍

app/move/controllers/unassign.js
solidgoldpig: 👍

app/move/controllers/unassign.test.js
solidgoldpig: 👍

PR #552, opened 2020-05-28
P4 1336 confirm person added to allocation

solidgoldpig:
Adds a warning to confirmation page that user should add the move to NOMIS

Person_assigned

Also reinstates the link to assign another person if allocation still has unfilled moves

### Checklist

### DIFF
--- a/app/move/controllers/assign/confirmation.js
+++ b/app/move/controllers/assign/confirmation.js
@@ -13,12 +13,13 @@ class ConfirmationController extends PersonAssignBase {
       const allocationId = res.locals.move.allocation.id
 
       const allocation = await allocationService.getById(allocationId)
-      const unassignedMove =
-        allocation.moves.filter(
-          move => !move.person || move.person === null
-        )[0] || {}
 
-      res.locals.unassignedMoveId = unassignedMove.id
+      const unassignedMoves = allocation.moves.filter(move => !move.person)
+      const unassignedMoveId = unassignedMoves.length
+        ? unassignedMoves[0].id
+        : undefined
+
+      res.locals.unassignedMoveId = unassignedMoveId
       next()
     } catch (err) {
       next(err)
--- a/app/move/controllers/assign/confirmation.js
+++ b/app/move/controllers/assign/confirmation.js
@@ -1,0 +1,25 @@
+const allocationService = require('../../../../common/services/allocation')
+
+const PersonAssignBase = require('./base')
+
+class ConfirmationController extends PersonAssignBase {
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setAssignNext)
+  }
+
+  async setAssignNext(req, res, next) {
+    const allocationId = res.locals.move.allocation.id
+
+    const allocation = await allocationService.getById(allocationId)
+    const unfilledMove =
+      allocation.moves.filter(
+        move => !move.person || move.person === null
+      )[0] || {}
+
+    res.locals.assignNext = unfilledMove.id
+    next()
+  }
+}
+
+module.exports = ConfirmationController
--- a/app/move/controllers/assign/confirmation.js
+++ b/app/move/controllers/assign/confirmation.js
@@ -5,20 +5,20 @@ const PersonAssignBase = require('./base')
 class ConfirmationController extends PersonAssignBase {
   middlewareLocals() {
     super.middlewareLocals()
-    this.use(this.setAssignNext)
+    this.use(this.setUnassignedMove)
   }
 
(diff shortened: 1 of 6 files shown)


app/move/controllers/assign/confirmation.js
```diff
@@ -0,0 +1,25 @@
...
+class ConfirmationController extends PersonAssignBase {
+  middlewareLocals() {
+    super.middlewareLocals()
+    this.use(this.setAssignNext)
+  }
+
+  async setAssignNext(req, res, next) {
+    const allocationId = res.locals.move.allocation.id
+
+    const allocation = await allocationService.getById(allocationId)
```
teneightfive, line 14: This will need to be wrapped in a `try/catch` in case the promise doesn't resolve.

app/move/controllers/assign/confirmation.js
```diff
@@ -0,0 +1,25 @@
...
+    const unfilledMove =
+      allocation.moves.filter(
+        move => !move.person || move.person === null
+      )[0] || {}
+
+    res.locals.assignNext = unfilledMove.id
```
teneightfive, line 20: What if we use [`unassignedMoveId`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/608bcfb1f1223fd1da25cadc21b8a1da3694c9d3/app/allocation/views/view.njk#L97) which is the same variable name we use in the allocation section of the app to refer to the same thing. Be nice to make it clear it's referring to the same scenario.

app/move/controllers/assign/confirmation.test.js
```diff
@@ -0,0 +1,111 @@
...
+        })
+
+        it('should not set a move to assign next', function() {
+          expect(res.locals.assignNext).to.be.undefined
+        })
+      })
+
+      describe('When unfilled moves remain', function() {
+        beforeEach(async function() {
+          allocationGetByIdStub.resolves({
```
teneightfive, line 88: Looks like we're missing a test for when this promise rejects.

solidgoldpig: ???   `describe('When allocation service returns an error'...` is the next block

app/move/controllers/assign/confirmation.js
solidgoldpig: 👍

app/move/controllers/assign/confirmation.js
solidgoldpig: 👍

app/move/controllers/assign/confirmation.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  async setUnassignedMove(req, res, next) {
+    try {
+      const allocationId = res.locals.move.allocation.id
+
+      const allocation = await allocationService.getById(allocationId)
+      const unassignedMove =
+        allocation.moves.filter(
+          move => !move.person || move.person === null
```
teneightfive, line 18: Won't `!move.person` already return true for `null` values?

ie, this can become:

```javascript
const unassignedMove = allocation.moves.filter(move => !move.person)[0] || undefined
```

It might also read a little easier as something like this too:

```javascript
const unassignedMoves = allocation.moves.filter(move => !move.person)
const unassignedMoveId = unassignedMoves.length ? unassignedMoves[0].id : undefined
```

app/move/controllers/assign/confirmation.test.js
```diff
@@ -0,0 +1,123 @@
+const proxyquire = require('proxyquire')
+
+const AssignBaseController = require('./base')
+
+const allocationGetByIdStub = sinon.stub().resolves({ moves: [] })
+
+const ConfirmationController = proxyquire('./confirmation', {
```
teneightfive, line 7: I think we could strip `proxyquire` from this test too. Will give us less areas to replace if we move to require...

app/move/controllers/assign/confirmation.test.js
teneightfive: My bad...

app/move/controllers/assign/confirmation.js
solidgoldpig: Yes. Think I got confuzzled because we have to pass `{id: null}` to the api. Personally, I prefer not having to check for the length, but switching it

app/move/controllers/assign/confirmation.test.js
solidgoldpig: Again this wasn't working. Now it is. ¯\_(ツ)_/¯

👍

PR #553, opened 2020-05-28
[P4 1445] Display additional info when no allocation move agreement

solidgoldpig:
### What changed

Updates `agreement` step of move `assign` journey to use an extended version of the `move_agreed` field. Adds `move_not_agreed_instruction` as conditional field when `move_agreed` answer is`false`.

### Why did it change

Adds text with instructions of how to proceed if the move of a person to be assigned to an allocation has not yet been agreed.

### Issue tracking

- [P4-1445](https://dsdmoj.atlassian.net/browse/P4-1445)

## Screenshots

| Before | After |
| ------ | ----- |
|Has_this_move_been_agreed_with_receiving_prison__-_person_assign|Has_this_move_been_agreed_with_receiving_prison__-_person_assign |

#### Manual testing

- [x] Chrome
- [x] Firefox

### Environment variables

### Other considerations

### DIFF
--- a/app/move/fields/move-not-agreed-instruction.js
+++ b/app/move/fields/move-not-agreed-instruction.js
@@ -1,0 +1,12 @@
+const moveNotAgreedInstruction = {
+  skip: true,
+  component: 'raw',
+  html:
+    '<p class="govuk-body">Complex cases must be discussed and agreed with the receiving prison.</p><p class="govuk-body">This includes:<p><ul class="govuk-list govuk-list--bullet"><li>Segregated prisoners</li><li>Self harm/prisoners on ACCT</li><li>Mental health issues</li><li>Integrated Drug Treatment System</li></ul>',
+  dependent: {
+    field: 'move_agreed',
+    value: 'false',
+  },
+}
+
+module.exports = moveNotAgreedInstruction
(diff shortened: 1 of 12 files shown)


app/move/fields/move-not-agreed-instruction.js
```diff
@@ -0,0 +1,12 @@
+const moveNotAgreedInstruction = {
+  skip: true,
+  component: 'raw',
+  html:
+    '<p class="govuk-body">Complex cases must be discussed and agreed with the receiving prison.</p><p class="govuk-body">This includes:<p><ul class="govuk-list govuk-list--bullet"><li>Segregated prisoners</li><li>Self harm/prisoners on ACCT</li><li>Mental health issues</li><li>Integrated Drug Treatment System</li></ul>',
```
teneightfive, line 5: Not particularly a big fan of having all this markup in a field config but if that's what we need to build...

teneightfive:
Just making note of the chat we had on slack...I wonder if a potential improvement in the future is to have this as a separate step. It seems we want to encourage the user to do a certain action (by using the "must") but at the same time, not disrupt their existing workflow too much.

Maybe using a separate step if the user selects "No" to replay this information will encourage them to agree with a prison for future transactions. It also simplifies the way we need to maintain that code which feels like a nice bonus :)

PR #555, opened 2020-05-29
chore: Call /moves/id/redirects endpoint

solidgoldpig:
## Proposed changes

Switches endpoint used by `move.redirect` from `/moves/:moveId/events` to `/moves/:moveId/redirects`

Updates models test to handle empty `204` responses

### Why did it change

We are moving away from generic `../events` endpoints to specific ones for each ttype

### Issue tracking

- [P4-1588](https://dsdmoj.atlassian.net/browse/P4-1588)

#### Automated testing

### Environment variables

### Documentation

- [x] Documented in the [models README](ministryofjustice/hmpps-book-secure-move-frontend/blob/master/test/fixtures/api-client/README.md)

### Other considerations


common/lib/api-client/models.test.js
```diff
@@ -52,78 +56,91 @@ const testCases = {
...
       method: 'findAll',
       httpMock: 'get',
+      statusCode: 200,
     },
     {
       method: 'find',
       httpMock: 'get',
       mockPath: '/1',
       args: '1',
+      statusCode: 200,
```
teneightfive, line 128: Maybe have the status code default to `200` unless specificed?

common/lib/api-client/models.test.js
```diff
@@ -136,36 +153,60 @@ const client = proxyquire('./', {
...
-            )
-
             flatten([response.data]).forEach(item =>
               checkProperties(item, model.attributes)
             )
           })
         })
       })
+      forEach(testCases[modelName], testCase => {
+        describe(`#${testCase.method}() with 204 response`, function() {
```
teneightfive, line 194: Maybe combine these tests and include the response code in the test description. It feels like this approach could become a pain if we add further response codes. We'd have to add another loop and describe going over all tests cases and return for tests that don't match the response code for that describe block.

You could then include the logic in the `it` block that for `204` it would expect no output. All other response codes then expect an output.

That also feels better because in reality we should update the POST tests to have a response code of `201` instead of `200`.

common/lib/api-client/models.test.js
```diff
@@ -25,17 +25,20 @@ const testCases = {
...
       httpMock: 'get',
+      statusCode: 200,
     },
   ],
   person: [
     {
       method: 'create',
       httpMock: 'post',
       args: {},
+      statusCode: 200,
```
teneightfive, line 41: Now we have this support we should actually look to use the correct status code for created resources. POSTs should return a `201` status.

common/lib/api-client/models.js
```diff
@@ -244,9 +240,8 @@ module.exports = {
       collectionPath: 'reference/allocation_complex_cases',
     },
   },
-  event: {
```
teneightfive, line 247: You may need to double check no other service methods are using this model. I seem to remember something to do with cancelling allocations temporarily sharing this model.

common/lib/api-client/models.test.js
solidgoldpig: I had it like that originally, but thought it was probably better to be explicit. Happy to snip it out though - your call

common/lib/api-client/models.test.js
solidgoldpig: Good point

common/lib/api-client/models.js
teneightfive: Yep, seems we're using it to [cancel allocations](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/4c888e03116954df6cc85a941d9c50276a10ef57/common/services/allocation.js#L12).

So we may have to leave this existing model for now and add a new one for `redirect`.

PR #557, opened 2020-06-01
Incoming moves

teneightfive:
### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/moves/middleware/set-body.outgoing.js
```diff
@@ -0,0 +1,19 @@
+const { get, set } = require('lodash')
+
+function setBodyOutgoing(req, res, next) {
+  const { dateRange, locationId } = req.params
```
barbaracassani, line 4: This is quite literally the same as `set incoming`. Can you incorporate them together, maybe through a function that accepts the type of move as parameters and returns the middleware function?

common/middleware/set-primary-navigation.test.js
```diff
@@ -199,6 +221,48 @@ describe('#setPrimaryNavigation()', function() {
...
+            {
+              active: false,
+              href: '/moves/incoming',
+              text: 'primary_navigation.incoming',
+            },
+          ])
+        })
+      })
+
+      context('on incoming page', function() {
```
barbaracassani, line 233: This is becoming a big file. Maybe split it between the lines of outgoing and incoming?

common/middleware/set-primary-navigation.js
```diff
@@ -39,6 +39,14 @@ function setPrimaryNavigation(req, res, next) {
...
   }
 
+  if (permissions.check('moves:view:incoming', userPermissions)) {
+    items.push({
+      text: req.t('primary_navigation.incoming'),
+      href: '/moves/incoming',
+      active: req.path.startsWith('/moves') && req.path.endsWith('/incoming'),
+    })
+  }
+
```
barbaracassani, line 49: Not in scope for this PR, but maybe we could have an object keyed by permission and with the `item` as value, then iterate it to retrieve the wanted matches.

common/services/move.js
```diff
@@ -34,6 +49,43 @@ const moveService = {
     page = 1,
     isAggregation = false,
   } = {}) {
+    // TODO: This is more of a temporary solution to solve the problem where
+    // the API doesn't have a concept of what locations a user has access to
+    //
+    // Once Auth is moved to the API we would be able to remove this as the API
+    // would know to only return moves that a user has access to
+    if (
```
solidgoldpig, line 57: You might find this verbose and/or indirect

```js
    const getMultipleRequestsPromise = (direction) => {
      if (
        (filter[`filter[${direction}_location_id]`] || '').split(',').length >
        LOCATIONS_BATCH_SIZE
      ) {
        return makeMultipleRequests(
          moveService.getAll,
          {
            filter,
            combinedData,
            page,
            isAggregation,
          },
          `filter["filter[${direction}_location_id]"]`
        )
      }
    }

    const fromPromise = getMultipleRequestsPromise('from')
    if (fromPromise) { return fromPromise }

    const toPromise = getMultipleRequestsPromise('to')
    if (toPromise) { return toPromise }
``

File: common/services/move.js [hunk 5]
```diff
@@ -34,6 +49,43 @@ const moveService = {
...
+        {
+          filter,
+          combinedData,
+          page,
+          isAggregation,
+        },
+        'filter["filter[to_location_id]"]'
+      )
+    }
+
```
solidgoldpig, line 88: Might be useful adding a comment after this pointing out that this is where the real meat of the method begins

app/moves/middleware/set-body.outgoing.js
teneightfive: You're right. I did toy with that idea too. The two major differences are with the property that is used to set it on the body (which is currently to aid re-using multiple of these middleware on the dashboard and I wonder if there's a better way to actually do that too) and the other is that one uses `fromLocationId` and the other `toLocationId`. 

Both could be passed into a function which returns the middleware but I didn't know if that would add too much logic and make it harder to read. What do you think?

It would look something like this:

```javascript
// middleware
function setBodyMoves(bodyKey, locationKey) {
  return function handleBody(req, res, next) {
    const { dateRange, locationId } = req.params
    const locations =
      locationId ||
      get(req.session, 'user.locations', [])
        .map(location => location.id)
        .join(',')

    set(req, `body.${bodyKey}`, {
      dateRange,
      [locationKey]: locations,
    })

    next()
  }
}


// moves/index.js
setBodyMoves('outgoing', 'fromLocationId')
...
setBodyMoves('incoming', 'toLocationId')
```

If that still feels okay to read I'll update them to use this shared middleware.

common/middleware/set-primary-navigation.test.js
teneightfive: Not sure what you mean by splitting it, could you explain a little further or post an example?

It also felt a little repetitive to me testing the highlight tab, I wonder if I could look at checking the number of `active = false` items and then just check that the `active = true` item matches the one expected and if that would be enough. What do you think?

common/middleware/set-primary-navigation.js
teneightfive: I may actually just add this here as it feels like this is the point it could use a small refactor to reduce repetitiveness.

common/services/move.js
teneightfive: Hmm, actually there might be a nice middleground using that approach to move the check to `makeMultipleRequests` instead...let me have a try...

common/services/move.js
teneightfive: I think your suggestion of moving the above chunks out into a separate method too might make that a bit clearer. Let me see if there's a tweak that can improve this.

app/moves/middleware/set-body.outgoing.js
solidgoldpig: Seems good to me

common/middleware/set-primary-navigation.test.js
teneightfive: Have pushed a fixup, let me know what you think

app/moves/middleware/set-body.outgoing.js
teneightfive: @barbaracassani if you're happy too I'll do that as a fixup

common/middleware/set-primary-navigation.js
solidgoldpig: something like

```js
const navPermissions = {
  'moves:view:outgoing': {
    text: 'primary_navigation',
    href: '/moves/outgoing',
    active: {
      startsWith: '/moves',
      endsWith: '/outgoing'
    }
  },
  'allocations': {
    text: 'primary_navigation',
    href: '/allocations',
    active: {
      startsWith: '/allocations',
      endsWith: '/outgoing'
    }
  }
  ...
}

Object.keys(navPermissions).forEach(permission => {
  if (permissions.check(permission, userPermissions)) {
    const { text, href, active = {} } = navPermissions[permission]
    const { startsWith, endsWith } = active
    items.push({
      text: req.t(text),
      href,
      active: (!startsWith || req.path.startsWith(startsWith)) && (!endsWith || req.path.endsWith(endsWith)),
    })
  }
})
```

Could a permission require more than one item to be added to the nav?

solidgoldpig: Or indeed, what you've done in the !fixup.

common/middleware/set-primary-navigation.test.js
solidgoldpig:
```
set-primary-navigation.test.js 

=>

set-primary-navigation.test.js
set-primary-navigation.ougoing.test.js
set-primary-navigation.incoming.test.js
```
I'm assuming that's what you meant B? (And if so, I'm all in favour of that)

common/services/move.js
teneightfive: Have updated it a bit so the initial function is a little easier to read. Let me know what you think.

common/middleware/set-primary-navigation.test.js
teneightfive: Hmm, I'm not sure I like the idea of that because it breaks the pattern of having a test file per file. Making it easy to see when scanning directories when test files are missing.

That's the main reason I prefer the pattern of side-by-side test files over a `/test` directory. 

I think that pattern would make sense if the main file was also split but just splitting a test file feels a little odd.

common/middleware/set-primary-navigation.js
teneightfive: >Could a permission require more than one item to be added to the nav?

Not at this stage. Could also extend if it we need that in the future. Tried to go with the simplest cleanup for now.

common/middleware/set-primary-navigation.test.js
barbaracassani: @solidgoldpig yes, that's what I meant :-)

app/moves/middleware/set-body.moves.js
```diff
@@ -0,0 +1,21 @@
...
+      get(req.session, 'user.locations', [])
+        .map(location => location.id)
+        .join(',')
+
+    set(req, `body.${property}`, {
+      dateRange,
+      [locationProperty]: locations,
+    })
+
+    next()
```
barbaracassani, line 17: Nice 👍

PR #561, opened 2020-06-02
[P4-1447] feat(allocations): Added sorting to allocation tables

barbaracassani:
## Proposed changes

This PR adds the ability to sort (server-side) to the allocation table (only the active allocations one. The cancelled one currently inherits the same filters but has no sorting capabilities of its own).

### What changed

Headers of the columns were changed to accomodate links that refresh the page with the various sorting options. If a column is sortable, the first click triggers a sort ascending. Further clicks will toggle the order. Clicking on another column triggers a sort ascending on that one, and so on.

### Why did it change

We want the users to be able to find allocations quickly.

### Issue tracking

- [P4-1447]()

## Screenshots
Screenshot 2020-06-04 at 11 41 35

Detail of active current sort:
Screenshot 2020-06-04 at 12 01 54

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/allocations/middleware/set-body.allocations.js
+++ b/app/allocations/middleware/set-body.allocations.js
@@ -3,11 +3,13 @@ const { set } = require('lodash')
 const dateHelpers = require('../../../common/helpers/date-utils')
 
 function setBodyAllocations(req, res, next) {
-  const { status } = req.query
+  const { status, sortBy, sortDirection } = req.query
   const { dateRange, locationId } = req.params
 
   set(req, 'body.allocations', {
     status,
+    sortBy,
+    sortDirection,
     moveDate: dateRange || dateHelpers.getCurrentWeekAsRange(),
     fromLocationId: locationId,
   })
--- a/app/allocations/middleware/set-results.allocations.js
+++ b/app/allocations/middleware/set-results.allocations.js
@@ -7,12 +7,14 @@ const allocationService = require('../../../common/services/allocation')
 async function setResultsAllocations(req, res, next) {
   const currentLocationId = get(req.session, 'currentLocation.id')
   const userPermissions = get(req.session, 'user.permissions')
+  const query = req.query
   const displayConfig = {
     showRemaining: permissions.check(
       'allocation:person:assign',
       userPermissions
     ),
     showFromLocation: !currentLocationId,
+    query,
   }
 
   try {
@@ -31,9 +33,10 @@ async function setResultsAllocations(req, res, next) {
       active: presenters.allocationsToTableComponent(displayConfig)(
         activeAllocations
       ),
-      cancelled: presenters.allocationsToTableComponent(displayConfig)(
-        cancelledAllocations
-      ),
+      cancelled: presenters.allocationsToTableComponent({
+        ...displayConfig,
+        isSortable: false,
+      })(cancelledAllocations),
     }
 
     next()
--- a/common/presenters/allocations-to-table-component.js
+++ b/common/presenters/allocations-to-table-component.js
@@ -27,11 +27,15 @@ function _byAdded(totalSlots, unfilledSlots) {
 function allocationsToTableComponent({
   showFromLocation = false,
   showRemaining = false,
+  query,
+  isSortable = true,
 } = {}) {
   const tableConfig = [
     {
       head: {
-        text: 'collections::labels.move_size',
+        isSortable,
+        sortKey: 'moves_count',
+        html: 'collections::labels.move_size',
         attributes: {
           width: '120',
         },
@@ -79,7 +83,11 @@ function allocationsToTableComponent({
     {
       /* eslint-disable indent */
       head: showFromLocation
-        ? { text: 'collections::labels.from_location' }
+        ? {
+            sortKey: 'from_location',
+            html: 'collections::labels.from_location',
+            isSortable,
+          }
         : undefined,
       /* eslint-enable indent */
       row: {
@@ -88,7 +96,9 @@ function allocationsToTableComponent({
     },
     {
       head: {
-        text: 'collections::labels.to_location',
+        isSortable,
+        sortKey: 'to_location',
+        html: 'collections::labels.to_location',
       },
       row: {
         text: 'to_location.title',
@@ -96,7 +106,9 @@ function allocationsToTableComponent({
     },
     {
       head: {
-        text: 'fields::date_custom.label',
+        isSortable,
+        sortKey: 'date',
+        html: 'fields::date_custom.label',
         attributes: {
           width: '135',
         },
@@ -110,7 +122,7 @@ function allocationsToTableComponent({
   return function buildTable(allocations) {
     return {
       head: tableConfig
-        .map(tablePresenters.objectToTableHead)
+        .map(tablePresenters.objectToTableHead(query))
         .filter(column => column),
       rows: allocations.map(tablePresenters.objectToTableRow(tableConfig)),
     }
--- a/common/presenters/table/object-to-table-head.js
+++ b/common/presenters/table/object-to-table-head.js
@@ -1,17 +1,67 @@
 const { pickBy } = require('lodash')
+const queryString = require('query-string')
 
 const i18n = require('../../../config/i18n')
 
-function objectToTableHead({ head } = {}) {
-  if (!head) {
-    return undefined
+function buildUrl({ sortKey }, query) {
+  return queryString.stringify({
+    ...query,
+    sortBy: sortKey,
+    sortDirection: query.sortDirection === 'asc' ? 'desc' : 'asc',
+  })
+}
+
+function buildAriaLabel(head, { sortDirection }) {
+  // the label indicates the result of the sorting action triggered by the link,
+  // which is the opposite of the current state
+  return i18n.t('sort', {
+    context: sortDirection === 'asc' ? 'descending' : 'ascending',
+    label: i18n.t(head.html).toLowerCase(),
+  })
+}
+
+function buildWrapper(head, query) {
+  const label = i18n.t(head.html)
+
+  if (!head.isSortable) {
+    return label
   }
 
-  return pickBy({
-    ...head,
-    html: head.html ? i18n.t(head.html) : undefined,
-    text: head.text ? i18n.t(head.text) : undefined,
-  })
+  const url = buildUrl(head, query)
+  const ariaLabel = buildAriaLabel(head, query)
+  return `<a aria-label="${ariaLabel}" class="sortable-table__button" role="button" href="?${url}">${label}</a>`
+}
+
+function assignSortAttributes(
+  { attributes, isSortable, sortKey },
+  { sortBy, sortDirection }
+) {
+  if (!isSortable) {
+    return attributes
+  }
+
+  const isCurrentSortColumn = sortKey === sortBy
+  const ariaSortDirection = sortDirection === 'asc' ? 'ascending' : 'descending'
+
+  return {
+    ...attributes,
+    'aria-sort': isCurrentSortColumn ? ariaSortDirection : 'none',
+  }
+}
+
+function objectToTableHead(query = {}) {
+  return function objectToTableHead({ head } = {}) {
+    if (!head) {
+      return undefined
+    }
+
+    return pickBy({
+      ...head,
+      attributes: assignSortAttributes(head, query),
+      html: head.html ? buildWrapper(head, query) : undefined,
+      text: head.text ? i18n.t(head.text) : undefined,
+    })
+  }
 }
 
 module.exports = objectToTableHead
--- a/common/presenters/table/object-to-table-head.test.js
+++ b/common/presenters/table/object-to-table-head.test.js
@@ -6,12 +6,14 @@ describe('table head presenter', function() {
   let output
 
   beforeEach(function() {
-    sinon.stub(i18n, 't').returnsArg(0)
+    sinon.stub(i18n, 't').callsFake(key => {
+      return `${key} (translated)`
+    })
   })
 
   context('with no args', function() {
     beforeEach(function() {
-      output = objectToTableHead()
+      output = objectToTableHead({})()
     })
 
     it('should return undefined', function() {
@@ -25,7 +27,7 @@ describe('table head presenter', function() {
 
   context('with html', function() {
     beforeEach(function() {
-      output = objectToTableHead({
+      output = objectToTableHead()({
         head: {
           html: 'move_type::label',
         },
@@ -34,7 +36,7 @@ describe('table head presenter', function() {
 
     it('should return correct header', function() {
       expect(output).to.deep.equal({
-        html: 'move_type::label',
+        html: 'move_type::label (translated)',
       })
     })
 
@@ -45,7 +47,7 @@ describe('table head presenter', function() {
 
   context('with text', function() {
     beforeEach(function() {
-      output = objectToTableHead({
+      output = objectToTableHead()({
         head: {
           text: 'move_type::label',
         },
@@ -54,7 +56,7 @@ describe('table head presenter', function() {
 
     it('should return correct header', function() {
       expect(output).to.deep.equal({
-        text: 'move_type::label',
+        text: 'move_type::label (translated)',
       })
     })
 
@@ -65,7 +67,7 @@ describe('table head presenter', function() {
 
   context('with attributes', function() {
     beforeEach(function() {
-      output = objectToTableHead({
+      output = objectToTableHead()({
         head: {
           html: 'move_type::label',
           attributes: {
@@ -77,7 +79,7 @@ describe('table head presenter', function() {
 
     it('should return correct header', function() {
       expect(output).to.deep.equal({
-        html: 'move_type::label',
+        html: 'move_type::label (translated)',
         attributes: {
           width: 1,
         },
@@ -91,7 +93,7 @@ describe('table head presenter', function() {
 
   context('with neither html nor test', function() {
     beforeEach(function() {
-      output = objectToTableHead({})
+      output = objectToTableHead()({})
     })
 
     it('should return undefined', function() {
@@ -105,7 +107,7 @@ describe('table head presenter', function() {
 
   context('with both html and text', function() {
     beforeEach(function() {
-      output = objectToTableHead({
+      output = objectToTableHead()({
         head: {
           html: 'move_type::label-html',
           text: 'move_type::label-text',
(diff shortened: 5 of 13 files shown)


common/presenters/allocations-to-table-component.js
```diff
@@ -88,6 +98,8 @@ function allocationsToTableComponent({
     },
     {
       head: {
+        sortable: true,
+        html: 'to_location',
```
teneightfive, line 102: Should this be `key`? Looks like it may have been a rogue copy/paste.

common/presenters/table/object-to-table-head.js
```diff
@@ -1,15 +1,45 @@
...
+      sortBy: key,
+      sortDirection: currentSort,
+    },
+  })
+}
+
+function buildWrapper(head) {
+  const label = i18n.t(head.html)
+  const url = buildUrl(head)
+  let sort = 'none'
```
teneightfive, line 25: Having a quick read up on [`aria-sort`](https://www.digitala11y.com/aria-sort-properties/) it seems it should be applied to the `th` element rather than the anchor element.

It also looks like if no attribute is provided the default value is `none` so we can probably just drop this attribute on all other headers and just apply it with the relevant sort to the column that the table is sorted by.

common/presenters/table/object-to-table-head.js
```diff
@@ -1,15 +1,45 @@
 const { pickBy } = require('lodash')
+const queryString = require('query-string')
 
 const i18n = require('../../../config/i18n')
 
+function buildUrl(head) {
+  const {
+    currentUrl: { pathname, query },
```
teneightfive, line 8: I wonder if it would be neater to not need to pass the entire URL into this presenter, and just pass in the existing querystring. You could then use the `query-string` library to override the sort keys as necessary for each sortable column.

The url for the link can then just be the querystring: `?foo=bar&sortBy=field&sortDirection=ascending`

common/presenters/table/object-to-table-head.js
```diff
@@ -1,15 +1,45 @@
...
+    },
+  })
+}
+
+function buildWrapper(head) {
+  const label = i18n.t(head.html)
+  const url = buildUrl(head)
+  let sort = 'none'
+  if (head.currentSort) {
+    sort = head.currentSort === 'asc' ? 'descending' : 'ascending'
```
teneightfive, line 27: This logic seems the wrong way around?

app/allocations/middleware/set-body.allocations.js
```diff
@@ -6,6 +6,8 @@ function setBodyAllocations(req, res, next) {
 
   set(req, 'body.allocations', {
     status,
+    sortBy: req.query.sortBy,
```
teneightfive, line 9: These values could be destructured above like `status`

app/allocations/middleware/set-results.allocations.js
```diff
@@ -7,12 +7,20 @@ const allocationService = require('../../../common/services/allocation')
...
+  }
   const displayConfig = {
     showRemaining: permissions.check(
       'allocation:person:assign',
       userPermissions
     ),
     showFromLocation: !currentLocationId,
+    sortDirection,
+    sortBy,
+    currentUrl,
```
teneightfive, line 23: As mentioned, this could just take the current querystring instead of the whole URL, and replace just the sort portion of the querstring

app/allocations/middleware/set-body.allocations.js
teneightfive: I wonder if using the sort as an object may also make it neater to work with and override in the allocations to table presenter too?

So the URL would be:

```javascript
?foo=bar&sort[by]=field&sort[direction]=asc
```

and you just pass the sort object on the body:

```javascript
const { status, sort } = req.query

set(req, 'body.allocations', {
  status,
  sort,
  ...
})
```

common/presenters/table/object-to-table-head.js
barbaracassani: Yeah, the html is all over the place. I wanted it to work with the `sortable-table` styles which are very different: it's all client-side so there are no links to speak of. I just wanted to see the arrows, so I hammered it in. This is not meant to be the solution, just a stepping stone. It's likely that `aria-sort` must be applied to a `th`, which is not possible afaik without overriding the govuk-table.

One thing I would like to discuss is the possibility of just copying / adapting the `sortable-table` styles instead of importing them. This would make things much simpler.

app/allocations/middleware/set-body.allocations.js
barbaracassani: Yeah, I toyed with the idea of a tuple-like object. Or just an object. It's a possibility.

common/presenters/table/object-to-table-head.js
barbaracassani: Good point.

common/presenters/table/object-to-table-head.js
barbaracassani: On second thought -- I actually reverse it twice. Before passing it, to build the url, and then again to set the arrow.
It should be enough to reverse it once at the end.

common/assets/scss/application.scss
```diff
@@ -40,6 +40,7 @@ $govuk-font-family: "Inter", system-ui, sans-serif;
 @import "@ministryofjustice/frontend/moj/components/organisation-switcher/organisation-switcher";
 @import "@ministryofjustice/frontend/moj/components/primary-navigation/primary-navigation";
 @import "@ministryofjustice/frontend/moj/components/badge/badge";
+@import "@ministryofjustice/frontend/moj/components/sortable-table/_sortable-table";
```
teneightfive, line 43: You don't need to include the underscore in this definition. The [underscore pattern](https://sass-lang.com/guide#topic-4) is used to define partials in sass and to tell it not to turn that file in a `.css` file when you run the sass command on a folder or set of files.

When including partials using `@import` you don't need to specifically add the underscore. It would be neater to keep the same pattern we have in this file and also if the underscore is ever removed from the Design System it wouldn't break these references.

common/presenters/table/object-to-table-head.js
barbaracassani: I have realised I can pass `aria-sort` to the th through attributes. Still doesn't solve the `button` issue, but it's a step in the right direction.

common/assets/scss/application.scss
barbaracassani: It may be that you're right but when I tried without `_` I got a compilation error. The compiler was trying to include the js instead (!!). I am not entirely sure why.

common/presenters/allocations-to-table-component.js
```diff
@@ -27,11 +27,15 @@ function _byAdded(totalSlots, unfilledSlots) {
...
   showFromLocation = false,
   showRemaining = false,
+  query,
+  enableSorting = true,
 } = {}) {
   const tableConfig = [
     {
       head: {
-        text: 'collections::labels.move_size',
+        sortable: enableSorting,
```
teneightfive, line 36: I find the `is` prefix pattern quite nice for things like this. Makes it easy to see what properties are a boolean. You could then actually use that same name for the argument as it fits both scenarios and pass it directly in here. And I think it might help to give a clearer name to `key` to make it more obvious that it is referring to the sort key:

```javascript
function allocationsToTableComponent({
  ...
  isSortable = true,
})

...

head: {
  isSortable,
  sortKey: 'moves_count',
  ...
}
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,49 @@
...
+  const { query, key } = head
+  return queryString.stringify({
+    ...query,
+    sortBy: key,
+    sortDirection: (query && query.sortDirection) === 'asc' ? 'desc' : 'asc',
+  })
+}
+
+function buildWrapper(head) {
+  const label = i18n.t(head.html)
```
teneightfive, line 16: This might read easier if you return the non-sortable option first. It also saves having to call `buildUrl` on columns that aren't sortable. Something like this:

```javascript
function buildSortableLink({ sortable, html, query, key }) {
  const label = i18n.t(head.html)

  if (!sortable) {
    return label
  }

  return `<a class="sortable-table__button" role="button" href="?${buildUrl(query, key)}">${label}</a>`
}
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,49 @@
 const { pickBy } = require('lodash')
+const queryString = require('query-string')
 
 const i18n = require('../../../config/i18n')
 
-function objectToTableHead({ head } = {}) {
-  if (!head) {
-    return undefined
+function buildUrl(head) {
+  const { query, key } = head
```
teneightfive, line 7: This could be destructured directly in the arguments:

```javascript
function buildUrl({ query = {}, key }) {
  ...
}
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,49 @@
...
+  return queryString.stringify({
+    ...query,
+    sortBy: key,
+    sortDirection: (query && query.sortDirection) === 'asc' ? 'desc' : 'asc',
```
teneightfive, line 11: Using a default for `query` in the arguments destructuring would allow you to drop the check for `query` here and simplify this line a little:

```javascript
sortDirection: query.sortDirection === 'asc' ? 'desc' : 'asc',
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,49 @@
...
-    html: head.html ? i18n.t(head.html) : undefined,
-    text: head.text ? i18n.t(head.text) : undefined,
-  })
+function objectToTableHead(query) {
+  return function objectToTableHead({ head } = {}) {
+    if (!head) {
+      return undefined
+    }
+    if (head.sortable) {
+      head.query = query
```
teneightfive, line 30: This feels a little heavy and difficult to read to set the query on each head element. You have access to the function that you call that builds the element and URL so maybe just pass it in once rather than having to attach it to the head object:

```javascript
html: head.html ? buildWrapper(head, query) : undefined,
```

instead of:

```javascript
html: head.html ? buildWrapper(head) : undefined,
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,49 @@
...
+      head.attributes = {
+        ...head.attributes,
```
teneightfive, line 32: What about just assigning this property in the case a column is sortable instead of re-assigning the attributes object:

```javascript
const { isSortable, sortKey } = head

if (isSortable) {
  const isCurrentSortColumn = sortKey === query.sortBy
  head.attributes['aria-sort'] = isCurrentSortColumn
    ? query.sortDirection === 'asc'
      ? 'ascending'
      : 'descending'
    : 'none',
}
```

I also find creating a variable (`isCurrentSortColumn`) for some booleans in a ternary can make the code a bit easier to read.

barbaracassani: The current version not only assigns the property, but also initialises `attributes` if it's `undefined`.

barbaracassani: The variable is a good point, and I have added it.

app/allocations/middleware/set-body.allocations.js
```diff
@@ -6,6 +6,8 @@ function setBodyAllocations(req, res, next) {
...
+    sortDirection: req.query.sortDirection,
```
teneightfive, line 10: Think the initial comment here got missed so adding to this line instead. It would be good to destructure these from `req.query` like `status` is so that we don't have two methods for getting values from `req.query`.

common/presenters/allocations-to-table-component.js
```diff
@@ -27,11 +27,15 @@ function _byAdded(totalSlots, unfilledSlots) {
...
   showRemaining = false,
+  query,
+  isSortable = true,
 } = {}) {
   const tableConfig = [
     {
       head: {
-        text: 'collections::labels.move_size',
+        isSortable,
+        key: 'moves_count',
```
teneightfive, line 37: Might be good to give this a clearer name like `sortKey`? 

Reading this back without the context of the change in 2 months it wouldn't be clear that this value is only used for sorting.

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,61 @@
...
+    ...query,
+    sortBy: key,
+    sortDirection: query.sortDirection === 'asc' ? 'desc' : 'asc',
+  })
+}
+
+function buildAriaLabel(head, query) {
+  // the label indicates the result of the sorting action triggered by the link,
+  // which is the opposite of the current state
+  return `${i18n.t('sort')} ${i18n.t(head.html).toLowerCase()} ${
```
teneightfive, line 17: It might be neater to have this as one translation key where you pass in the label and use `context` for the direction:

```json
// translation
"sort_ascending": "Sort {{label}} ascending",
"sort_descending": "Sort {{label}} descending"
```

```javascript
return i18n.t('sort', {
  context: query.sortDirection === 'asc' ? i18n.t('descending') : i18n.t('ascending'),
  label: i18n.t(head.html),
})
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,61 @@
...
+  }
+  const isCurrentSortColumn = head.key === query.sortBy
+  return {
+    ...head.attributes,
+    'aria-sort': isCurrentSortColumn
+      ? query.sortDirection === 'asc'
+        ? 'ascending'
+        : 'descending'
+      : 'none',
   }
```
solidgoldpig, line 44: I do find nested ternaries a bit head-melting. Could we not do something like

```js
  let ariaSort = 'none'
  if (head.key === query.sortBy) {
    const ariaSortLabels = {
      asc: 'ascending',
      desc: 'descending'
    }
    ariaSort = ariaSortLabels[query.sortDirection]
  }
  return {
    ...head.attributes,
    'aria-sort': ariaSort
  }
```

solidgoldpig: actually, having hoisted that out of the return statement, there's probably little to be gained from the object map now...

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,61 @@
...
-    ...head,
-    html: head.html ? i18n.t(head.html) : undefined,
-    text: head.text ? i18n.t(head.text) : undefined,
-  })
+function objectToTableHead(query) {
+  return function objectToTableHead({ head } = {}) {
+    if (!head) {
+      return undefined
+    }
+    head.attributes = assignSortAttributes(head, query)
```
teneightfive, line 52: This could be combined within the return to make it a little easier to read:

```javascript
    return pickBy({
      ...head,
      attributes: assignSortAttributes(head, query),
      html: head.html ? buildWrapper(head, query) : undefined,
      text: head.text ? i18n.t(head.text) : undefined,
    })
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,61 @@
...
+    return label
+  }
+  const url = buildUrl(head, query)
+  const ariaLabel = buildAriaLabel(head, query)
+  return `<a aria-label="${ariaLabel}" class="sortable-table__button" role="button" href="?${url}">${label}</a>`
+}
+
+function assignSortAttributes(head, query) {
+  if (!head.isSortable) {
+    return head.attributes ? { ...head.attributes } : undefined
```
teneightfive, line 34: I wonder if this could be neatened up using defaults. There's a lot of ternaries within this file now, and as Alex said, I'm similar in that I find them harder to read when there are lots of them.

teneightfive: You could destructure the argument and possible use a separate variable for the second ternary to make it a bit easier to read I think:

```javascript
function assignSortAttributes({ attributes = {}, key, isSortable }, { sortBy, sortDirection }) {
  if(!isSortable) {
    return attributes
  }

  const isCurrentSortColumn = key === sortBy
  const ariaSortDirection = sortDirection === 'asc' ? 'ascending' : 'descending'

  return {
    ...attributes,
    'aria-sort': isCurrentSortColumn 
      ? ariaSortDirection
      : 'none',
  }
}
```

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,61 @@
 const { pickBy } = require('lodash')
+const queryString = require('query-string')
 
 const i18n = require('../../../config/i18n')
 
-function objectToTableHead({ head } = {}) {
-  if (!head) {
-    return undefined
+function buildUrl({ key }, query = {}) {
```
teneightfive, line 6: Why not set the default value for query at the first point of entry for this file: 

```javascript
function objectToTableHead(query = {}) {
  ...
}
```

Then you can destructure here to only include what you need:

```javascript
function buildUrl({ key }, { sortDirection }) {
  ...
}
```

common/presenters/table/object-to-table-head.test.js
```diff
@@ -129,4 +131,64 @@ describe('table head presenter', function() {
       expect(i18n.t.callCount).to.equal(2)
     })
   })
+  context(
+    'with isSortable, and current active sorting on the column',
```
teneightfive, line 135: I think it would be nicer to use nexted contexts here. In both contexts you're saying "with isSortable" but it think it would be clearer to use something like this instead:

```javascript
context('with isSortable', function() {
  context('with current column sorted', function() {

  })

  context('with another column sorted', function() {

  })
})
```

For me, things like _and_ in a `context` description and _when_ or _if_ in an `it` description are a bit of a flag that things should be re-worked a little.

teneightfive:
@barbaracassani another question - should there be a default sort order?

If so, you can add it to the [default query constant](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/13cd3d6d2a270b09f5daed0efbfe7b5c2ec5c32f/app/allocations/constants.js#L17) in the allocations app like so:

```javascript
const DEFAULTS = {
  QUERY: {
    outgoing: { status: '', sortBy: 'date', sortDirection: 'asc' },
  },
  ...
}
```

common/presenters/table/object-to-table-head.test.js
barbaracassani: 👍

common/presenters/table/object-to-table-head.js
barbaracassani: I have adopted @teneightfive suggestion (see below) of using an intermediate variable. Does it look better now?

common/presenters/table/object-to-table-head.js
barbaracassani: 👍

app/allocations/middleware/set-body.allocations.js
barbaracassani: 👍

common/presenters/allocations-to-table-component.js
```diff
@@ -27,11 +27,15 @@ function _byAdded(totalSlots, unfilledSlots) {
 function allocationsToTableComponent({
   showFromLocation = false,
   showRemaining = false,
+  query,
+  isSortable = true,
```
teneightfive, line 31: Just realised it looks like there is no test to test the default scenario and the scenario where this is overridden to `false`. Could we add some tests for this flag please.

common/presenters/allocations-to-table-component.js
```diff
@@ -27,11 +27,15 @@ function _byAdded(totalSlots, unfilledSlots) {
 function allocationsToTableComponent({
   showFromLocation = false,
   showRemaining = false,
+  query,
```
teneightfive, line 30: There seems to be no test to ensure that this value is passed correctly to the `objectToTableHead` presenter. It would be good to add something around that so if this property is removed the tests would fail.

common/presenters/table/object-to-table-head.js
```diff
@@ -1,17 +1,60 @@
...
+    return label
+  }
+  const url = buildUrl(head, query)
+  const ariaLabel = buildAriaLabel(head, query)
+  return `<a aria-label="${ariaLabel}" class="sortable-table__button" role="button" href="?${url}">${label}</a>`
+}
+
+function assignSortAttributes(head, { sortBy, sortDirection }) {
+  if (!head.isSortable) {
+    return head.attributes ? { ...head.attributes } : undefined
```
teneightfive, line 35: I don't think you need to use spread if the attributes already exist here?

Wouldn't this be enough:

```javascript
return head.attributes ? head.attributes : undefined
```

And I think if you destructure head in the argument this could read even nicer:

```javascript
function assignSortAttributes({ attributes, isSortable, sortKey }, { sortBy, sortDirection }) {
  if (!isSortable) {
    return attributes
  }
  
  const isCurrentSortColumn = sortKey === sortBy
  const ariaSortDirection = sortDirection === 'asc' ? 'ascending' : 'descending'

  return {
    ...attributes,
    'aria-sort': isCurrentSortColumn ? ariaSortDirection : 'none',
  }
}
```

I think it also helps to be more consistent with the destructuring of the arguments rather than have a mix of patterns (unless you need to return `head` in its entirety, which is doesn't seem like you do in this case).

PR #564, opened 2020-06-05
Tweak allocation detail page

teneightfive:
### What changed

- allocation detail page (see screenshots)
  - Date shows date as well as relative date
  - Updated labels on criteria to be consistent with creation journey
  - Included allocation information into page title to make it more useful and to help distinguish different pages in browser history
  - Updated button/added language to be consistent with listing page
  - Re-used same card component for outgoing moves without a person
- allocation criteria step (during creation)
  - Updated labels to be consistent with criteria table

### Why did it change

When this section of the app was created there were time constraints that reduced the time available to ensure we were being as consistent and re-usable as possible.

The majority of these changes ensure we are re-using code and content were possible to ensure that we are providing a consistent experience across the service.

## Screenshots

*Click images to enlarge*

### Without permission to assign (PMU user)

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

### With permission to assign (OCA user)

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/allocation/controllers/view.test.js
```diff
@@ -96,17 +103,17 @@ describe('view allocation', function() {
         locals = mockRes.render.firstCall.lastArg
       })
 
-      it('creates allocationDetails with result of the presenter', function() {
-        expect(locals.allocationDetails).to.exist
+      it('creates `criteria` with result of the presenter', function() {
```
barbaracassani, line 106: Why the backtick?

app/allocation/controllers/view.test.js
```diff
@@ -117,6 +124,19 @@ describe('view allocation', function() {
       it('should create unassigned move ID', function() {
         expect(locals.unassignedMoveId).to.equal('123')
       })
+
+      it('sould order moves without person first', function() {
```
barbaracassani, line 128: typo :-)

app/allocation/controllers/view.test.js
```diff
@@ -265,4 +272,52 @@ describe('view allocation', function() {
...
+    let locals
+
+    beforeEach(function() {
+      mockReq.session.user.permissions = ['allocation:person:assign']
+
+      handler(mockReq, mockRes)
+      locals = mockRes.render.firstCall.lastArg
+    })
+
+    it('sould filter out moves without a person', function() {
```
barbaracassani, line 286: Same typo as above...

app/allocation/views/view.njk
```diff
@@ -31,111 +40,100 @@
...
+
+    {% if canAccess('allocation:person:assign') and remainingCount %}
+      <p>
+        {{ t("allocation::view.people.special_vehicle_instruction", {
+          href:"/move/new"
+        }) | safe }}
+      </p>
+
+      {{ govukButton({
+        href: "/move/" + unassignedMoveId + "/assign",
```
barbaracassani, line 101: Really unimportant, but maybe use template strings.

app/allocation/views/view.njk
```diff
@@ -31,111 +40,100 @@
...
-      {{ appCard(slot.card) }}
       {# TODO: consider handling actions within card component #}
       {# TODO: consider whether long names may cause overlap problems #}
-      {% if canAccess('allocation:person:assign') %}
+      {% if move.person and canAccess('allocation:person:assign') %}
         <p class="app-!-position-top-right govuk-!-margin-top-3">
           {{ govukButton({
-            href: '/move/' + slot.id + "/unassign",
-            html: t("allocations::view.person_unassign", {name: slot.fullname}),
+            href: '/move/' + move.id + "/unassign",
```
barbaracassani, line 118: As above, maybe use template strings.

common/presenters/allocation-to-meta-list-component.test.js
```diff
@@ -5,99 +5,90 @@ const presenter = require('./allocation-to-meta-list-component')
...
+    expect(output.items).to.be.an('array')
+    expect(output.items.length).to.equal(4)
   })
-  it('has the prisoner category as first item', function() {
-    expect(output.rows[0]).to.deep.equal({
+  it('has the number of prisoners as first item', function() {
+    expect(output.items[0]).to.deep.equal({
       key: {
-        text: 'allocations::view.fields.category',
+        text: 'fields::moves_count.label',
```
barbaracassani, line 56: I have found useful to stop using `returnsArg` and instead using `callsFake`, so to append something to the string that effectively verifies it has gone through the `i18n.t` function.

common/presenters/move-to-meta-list-component.test.js
```diff
@@ -164,105 +161,16 @@ describe('Presenters', function() {
...
-        sinon.stub(filters, 'formatDateWithDay').returnsArg(0)
         transformedResponse = moveToMetaListComponent(mockMove)
       })
 
-      afterEach(function() {
-        this.clock.restore()
-      })
-
       describe('response', function() {
         it('should render `date from` row', function() {
```
barbaracassani, line 175: Oh, we are using the backticks to highlight the name of a property? Where does it work?

app/allocation/views/view.njk
teneightfive: Unfortunately it's not possible as this is a nunjucks file

app/allocation/views/view.njk
teneightfive: Same as above. I'm not aware it's possible to use template strings in a nunjucks file.

Although this has highlighted the mix of single/double quotes which I'll fix...

common/presenters/move-to-meta-list-component.test.js
teneightfive: I think it's actually just something I've used when I've remembered to try and show that the assertion is checking for a particular thing. Not sure it has any actual effect in mocha though.

common/presenters/allocation-to-meta-list-component.test.js
teneightfive: I do agree that simply checking for the inputted i18n key isn't often enough of a robust test though.

solidgoldpig: ???

Why isn't checking the inputted i18n key (and any context) not enough? Why is it good enough for other methods but not the `t`?

app/allocation/views/view.njk
solidgoldpig: ++1 to being consistent with our quotes in the nunjucks templates.

Should we consider linting them?

But at a deeper level, I think we should be computing these values outside of the templates if we can (especially urls). As little logic and assignment of values in the templates as possible.

app/allocation/controllers/view.js
```diff
@@ -1,38 +1,51 @@
...
   })
 
+  const removeUnassignedMoves = move => {
+    return !(
+      permissions.check('allocation:person:assign', userPermissions) &&
+      !move.person
+    )
+  }
+
   const locals = {
```
solidgoldpig, line 24: I know some of us value terseness (and ternaries, ternaries, ternaries...) but I find it really hard to see what's actually being exported to locals with all the inline assignments. Not just here (and in fact this is hardly the most egregious example) but I just thought I'd mention it

teneightfive: Yeah, I do see what you mean. Especially with files like this where there are a number of things being passed down to the view.

I wonder if we try to adopt a pattern where  the object passed to the render method always uses shorthand variable names and they are defined using variables within the controller to aid readability. What do you think?

Although, of course, there could be exception to the rule if they are simple assignments such as `query: req.query`.

common/presenters/allocation-to-meta-list-component.test.js
teneightfive: Sorry, to clarify, I meant stubbing `t` to return the first argument, and then simply checking that the end object contains the translation key isn’t enough, which I think is what Barbara was also saying.

Checking specifically that `t` is called with the key and any other expected arguments, like context, is what I was suggesting we should be doing and I feel is sufficient.

teneightfive: Those cases were we do not check that `t` was actually called using sinon and `calledWithExactly` could lead to us missing regressions where the second argument of `t` is changed, so I think we should always attempt to check it is called correctly where possible.

teneightfive:
> A general consideration, is that it would help slightly the reviewer to have a list of the intended changes, instead of having to play spot the difference on the -- certainly very useful! -- screenshots. In any case nothing that warrants change requests.

Yep, valid point as there were a number of screenshots for these changes. I updated the description with a short summary of the key differences.

app/allocation/views/view.njk
teneightfive: I'd be keen to lint, but I wasn't aware there was a way to easily lint nunjucks files?

So up until now it's unfortunately been a consistency I've just tried to preserve via code reviews.

I do agree it would also be good to tidy up some of the large view files. They're containing quite a lot of logic that isn't easily as easy to test with just end-to-end tests.

PR #569, opened 2020-06-05
Prevent e2e fixtures using disabled reference data

solidgoldpig:
### What changed

Ethnicities and genders have disabled values removed.

### Why did it change

Test fixtures could occasionally generate data with a value that was disabled which led to the test failing.

This brings the fixtures in line with what the frontend expects.

### Issue tracking
n/a

### Environment variables

### Other considerations

### DIFF
--- a/test/e2e/_helpers.js
+++ b/test/e2e/_helpers.js
@@ -8,6 +8,7 @@ import glob from 'glob'
 import { find, get, isArray, isNil } from 'lodash'
 import { ClientFunction, RequestLogger, Selector, t } from 'testcafe'
 
+import referenceDataHelpers from '../../common/helpers/reference-data'
 import moveService from '../../common/services/move'
 import personService from '../../common/services/person'
 import referenceDataService from '../../common/services/reference-data'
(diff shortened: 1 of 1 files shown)


test/e2e/_helpers.js
```diff
@@ -17,14 +18,19 @@ export const scrollToTop = ClientFunction(() => {
...
   const firstNames = faker.name.firstName()
   const lastName = faker.name.lastName()
 
   let genders = await referenceDataService.getGenders()
-  // TODO: implement proper test to render this filter unnecessary
-  genders = genders.filter(gender => gender.title.match(/^(Male|Female)$/i))
-  const ethnicities = await referenceDataService.getEthnicities()
+  // TODO: implement proper test to render male/female filter unnecessary
+  genders = genders
+    .filter(filterDisabled)
```
teneightfive, line 30: I think it would be good to also add this filter to all other calls to reference data to prevent this happening  with other datasets. 

From a quick scan it looks like we make a similar call in [`createPersonFixture`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/3c5538c7c9d8a66c90d3e2fbba93c5263e24b614/test/e2e/_helpers.js#L66-L67).

So might be worth having a search for calls to the `referenceDataService` as I'm sure we're doing calls to locations, etc too?

We designed all the reference data endpoints in the API to return the same format and include a `disabled_at` attribute so that they could be consistently filtered out in the way to support us changing the data over time.

solidgoldpig: For the call in createPersonFIxture you reference, we then immediately filter it down by the value that's been passed in so it wouldn't be an issue there, but the getLocations call could indeed be susceptible to the same issue, so I've added filtered versions of the methods in the _helpers file to call rather than the refData ones directly.

I have a niggling feeling this should be the default behaviour of the refData methods, and that we should ask for the unfiltered list explicitly...

teneightfive: I was going to suggest the same after I saw your changes 😂 

Now that we're calling it in multiple places it should probably be something rolled into the service rather than where the service is called to prevent unfiltered results being returned again in the future.

Up to you if you want to do as part of this or add a GH issue/TODO to track it.

test/e2e/allocation.new.test.js
```diff
@@ -10,7 +10,7 @@ test('Create allocation and verify the result', async t => {
   await t.navigateTo(`/allocation/${allocationId}`)
   ;['summary', 'meta'].forEach(async section => {
     for (const o of allocationJourney.allocationViewPage.nodes[section].keys) {
-      const textExists = allocationJourney.allocationViewPage.getDlDefinitionByKey(
+      const textExists = await allocationJourney.allocationViewPage.getDlDefinitionByKey(
```
teneightfive, line 13: I pushed a similar change to fix this issue and just made it a bit more readable:
https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/564/commits/bc59ae218ea425f0ff0cc8e333a7783523c7545e

`o` in the loop seemed a bit ambiguous so updated it to be `key` and passed the promise directly into the expect call which is what we've done elsewhere.

PR #577, opened 2020-06-08
Update allocation confirmation page

teneightfive:
### What changed

The confirmation page for the assign journey of a move has been updated to use the existing shared confirmation page.

### Why did it change

The main reason is to create consistency between journeys. It re-uses a lot of the existing logic around moves and brings the display of content alongside other confirmation pages to be more consistent.

### Confirmation pages

- Updated use of bold to be consistent with existing use
- Updated return text to be consistent with existing use
- Updated success message to be more accurate of action that just occurred
- Updated page title

Added some existing confirmation page screenshots for consistency reference.

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | NA |
| <kbd></kbd> | NA |
| <kbd></kbd> | <kbd></kbd> |

### Button on assign journey

- Updated button to match action

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/move/views/confirmation.njk
```diff
@@ -1,13 +1,14 @@
 {% extends "layouts/base.njk" %}
 
 {% block customGtagConfig %}
-  gtag('set', {'page_title': 'Move requested'});
+  gtag('set', {'page_title': '{{ "Person added to allocation" if move.allocation else "Move " + move.status }}' });
```
solidgoldpig, line 4: Just a thought. Could we actually use different templates for move and assign? As long as we never need another variant, I suppose this is cool

```
{% block customGtagConfig %}
  gtag('set', {'page_title': '{% block CustomGtagContent %}{% endblock %}'});
{% endblock %}
...
{% block CustomGtagContent %} Person added to allocation{% endblock %}
...
{% block CustomGtagContent %} Person added to allocation{% endblock %}
```

teneightfive: I suppose I was trying to keep them combined so that we keep consistency between confirmation pages.

So any of the actions that occur on a move use this same confirmation so felt better to maintain having one approach. The display and layout is all the same but because they were maintained separately we ended up with them looking differently.

solidgoldpig: Sorry, to be clear. I meant have the templates share the same base confirmation template and just override the bits as needed.  eg.  how the base govuk template allows the page title to be overridden through the `pageTitle` block. If there are only 2 variants, it's arguable that it's clearer to just have if statements for the 2 cases.

PR #580, opened 2020-06-08
[P4 1529] - Add cancellation reasons to allocation

barbaracassani:
## Proposed changes

This PR adds the reasons to the allocation cancel journey.

### What changed

The reasons fields were added as per prototype.

### Why did it change

We issued the cancel journey without reasons to reach the deadline for the MVP.

### Issue tracking

- [P4-1529]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd>Screenshot 2020-06-09 at 15 12 59</kbd> | <kbd>Screenshot 2020-06-09 at 15 21 41</kbd>| 
|  | <kbd>Screenshot 2020-06-09 at 15 22 42</kbd>| 
|<kbd>Screenshot 2020-06-09 at 15 13 05</kbd> | <kbd>Screenshot 2020-06-09 at 15 12 12</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/services/allocation.js
```diff
@@ -14,16 +14,17 @@ const allocationService = {
     'to_location',
   ],
 
-  cancel(allocationId) {
+  cancel(data) {
```
teneightfive, line 17: Could we use the same pattern here that is used to [cancel moves](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/9614729dec89d030652f2001eac27d386fbb11eb/common/services/move.js#L269-L283)  by where we keep the first argument as the ID and pass the data in as the secon?

app/allocation/controllers/cancel.test.js
```diff
@@ -41,9 +41,9 @@ describe('Cancel controller', function() {
       })
 
       it('should cancel the allocation', function() {
-        expect(allocationService.cancel).to.be.calledWithExactly(
-          mockAllocation.id
-        )
+        expect(allocationService.cancel).to.be.calledWithExactly({
+          id: mockAllocation.id,
```
teneightfive, line 45: Can we add a test to ensure this method is called with the correct data, and to ensure it removes the omitted values from the session model.

common/lib/api-client/index.js
```diff
@@ -17,6 +17,8 @@ module.exports = function() {
     logger: IS_DEV,
   })
 
+  instance.pluralize.addUncountableRule('cancel')
```
teneightfive, line 20: I feel this is rogue enough that it probably warrants some kind of comment as to why it's here.

And possibly even within the model itself as well to explain the direct relationship...?

test/e2e/allocation.cancel.test.js
```diff
@@ -16,11 +16,17 @@ test('Cancel allocation', async t => {
   await t
     .expect(allocationJourney.getCurrentUrl())
     .match(/\/allocation\/[\w]{8}(-[\w]{4}){3}-[\w]{12}\/cancel\/reason$/)
+  await allocationJourney.allocationCancelPage.selectReason('other', 'Flood')
```
teneightfive, line 19: Maybe use `faker` here as we have done elsewhere.

common/lib/api-client/index.js
```diff
@@ -24,6 +24,10 @@ module.exports = function() {
     logger: IS_DEV,
   })
 
+  // the endpoint for cancel is structured in a way that requires its own model. However,
+  // devour would pluralise it. This rule ensures that we get /cancel and not /cancels
```
teneightfive, line 28: I'm wondering whether actually the better solution is to just define the custom path within the model.

That would avoid having to have these comments in two places. I'm just conscious we may have to do this for other status change endpoints, like `/complete` and it might be better to keep those maintained within the models configuration.

So this:

```javascript
  cancel: {
    attributes: {
      timestamp: '',
      notes: '',
      cancellation_reason: '',
      cancellation_reason_comment: '',
    },
  },
```

would become this:

```javascript
  cancel: {
    attributes: {
      timestamp: '',
      notes: '',
      cancellation_reason: '',
      cancellation_reason_comment: '',
    },
    options: {
      collectionPath: 'cancel',
    },
  },
```

What do you think? I can just see us having 5/6/7 of these addUnacountableRules here and it being a bit harder to maintain...

solidgoldpig: Or...

common/lib/api-client/models.js
```js
module.exports = {
  ...
  cancel: {
    attributes: {
      ...
    },
    singular: true
  }
}
```

common/lib/api-client/index.js
```js
const singulars = Object.keys(models).filter(key => models[key].singular)
...
pluralize: (str) => singulars.includes(str) ? str : pluralize(str)
```

barbaracassani: @solidgoldpig I tried but the injected `pluralize` doesn't have a `.singular` method which is invoked further down the line by the api client. I am sure it can be made working with some hack but doesn't seem like a super convenient solution.

barbaracassani: @teneightfive seems an improvement in readability, so I have applied your suggestion.

PR #588, opened 2020-06-11
feat(allocations): Remove single move from allocation

barbaracassani:
## Proposed changes

This PR adds the possibility of removing a single personless move from an allocation. This can be done only by PMU and only if there are at least two slots -- e.g. you cannot create a 0 slots allocation with it.

### What changed

A small journey was added in allocations, that reused the move services.

### Why did it change

It's a way of editing the unfilled allocations

### Issue tracking

- [P4-1662]()

## Screenshots

The initial state of the allocation. Note the "Remove from allocation" links.
<kbd>Screenshot 2020-06-11 at 17 14 56</kbd>
After clicking on "Remove from allocation"
<kbd>Screenshot 2020-06-11 at 17 24 27</kbd>
The allocation has now two slots.
<kbd>Screenshot 2020-06-11 at 17 15 24</kbd>
After removing one more slot. Note that the "Remove from allocation" link is not present.
<kbd>Screenshot 2020-06-11 at 17 38 30</kbd>
Following @Andy-Millar feedback, I have added a commit that removes the link if the card doesn't include a person:
<kbd>Screenshot 2020-06-12 at 11 43 50</kbd>

</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/allocation/controllers/index.js
+++ b/app/allocation/controllers/index.js
@@ -2,6 +2,7 @@ const CancelController = require('./cancel')
 const AllocationCriteriaController = require('./create/allocation-criteria')
 const AllocationDetailsController = require('./create/allocation-details')
 const Save = require('./create/save')
+const RemoveMoveController = require('./remove-move')
 
 const createControllers = {
   AllocationCriteriaController,
@@ -12,7 +13,12 @@ const cancelControllers = {
   CancelController,
 }
 
+const removeMoveControllers = {
+  RemoveMoveController,
+}
+
 module.exports = {
   cancelControllers,
+  removeMoveControllers,
   createControllers,
 }
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -21,6 +21,18 @@ module.exports = function view(req, res) {
     )
   }
 
+  const removeMoveLink = move => {
+    if (
+      move.person ||
+      !permissions.check('allocation:cancel', userPermissions) ||
+      moves.length === 1
+    ) {
+      return
+    }
+
+    return `/allocation/${allocation.id}/${move.id}/remove`
+  }
+
   const locals = {
     // TODO: Find way to store the actual URL they came from: See similar solution within moves app
     dashboardUrl: '/allocations',
@@ -44,6 +56,7 @@ module.exports = function view(req, res) {
       .map(move => ({
         id: move.id,
         person: move.person,
+        removeMoveHref: removeMoveLink(move),
         card: moveToCardComponent(move),
       })),
   }
--- a/app/allocation/views/view.njk
+++ b/app/allocation/views/view.njk
@@ -123,6 +123,15 @@
           })}}
         </p>
       {% endif %}
+
+      {% if move.removeMoveHref %}
+        <div>
+          <a class="app-!-position-top-right govuk-!-margin-top-7 govuk-!-margin-right-3 app-font-weight-normal" href="{{move.removeMoveHref}}">
+            {{ t("actions::person_unassign_confirmation") }}
+          </a>
+        </div>
+      {% endif %}
+
     </div>
   {% else %}
     {{ appMessage({
--- a/common/presenters/move-to-card-component.js
+++ b/common/presenters/move-to-card-component.js
@@ -11,7 +11,7 @@ function moveToCardComponent({
   hrefSuffix = '',
 } = {}) {
   return function item({ id, reference, person, status }) {
-    const href = `/move/${id}${hrefSuffix}`
+    const href = person ? `/move/${id}${hrefSuffix}` : ''
     const excludedBadgeStatuses = ['cancelled']
 
     const showStatusBadge =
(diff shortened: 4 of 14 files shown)


common/presenters/move-to-card-component.js
```diff
@@ -10,7 +10,10 @@ function moveToCardComponent({
   showStatus = true,
   hrefSuffix = '',
 } = {}) {
-  return function item({ id, reference, person, status }) {
+  return function item(
+    { id, reference, person, status },
+    cancelAllocationHref
```
teneightfive, line 15: So this feels like a similar scenario Alex got into around "actions" within a card. I feel that the use of all the "allocation" specific terminology in here could be made a little more generic as the purpose of the components is to be more content agnostic.

It looks like what we're attempting to do is to say in this instance we want to provide actions, but it doesn't actually look like we're passing them down to the component so I'm not sure the presenter is the best place for this logic.

I've made a further comment about how you could potentially bypass all of these changes to the presenter, similar to how Alex implemented the remove button.

app/allocation/controllers/view.js
```diff
@@ -44,7 +56,12 @@ module.exports = function view(req, res) {
       .map(move => ({
         id: move.id,
         person: move.person,
-        card: moveToCardComponent(move),
+        card: moveToCardComponent(
+          {
+            ...move,
+          },
+          removeMoveLink(move)
```
teneightfive, line 63: This value could be passed directly into the map to avoid having to mess around with the presenters.

Having looked at the presenters, this value isn't send to the `card` component so there is no need to send this value into the presenter.

app/allocation/controllers/view.js
```diff
@@ -44,7 +56,12 @@ module.exports = function view(req, res) {
       .map(move => ({
         id: move.id,
         person: move.person,
-        card: moveToCardComponent(move),
+        card: moveToCardComponent(
```
teneightfive, line 59:
```suggestion
      .map(move => ({
        id: move.id,
        person: move.person,
        removeSlotHref: removeMoveLink(),
        card: moveToCardComponent(move),
```

teneightfive: Not sure this "suggestion" worked as such, but hopefully you see what I mean.

app/allocation/views/view.njk
```diff
@@ -123,6 +123,15 @@
           })}}
         </p>
       {% endif %}
+
+      {% if move.card.cancelAllocationHref %}
```
teneightfive, line 127: With the suggestion in the controller you can then call these values directly on `move.` and not have to worry about the `move.card`.

The output on `move.card` is only to be passed directly into the card component so that it can be rendered.

You can bypass needing to do all of that.

Andy-Millar:
Can I ask why 'Person to be added' is a link? Does it lead anywhere (it really shouldn't). My update would be remove the link.

app/allocation/views/view.njk
barbaracassani: Good point.

barbaracassani:
@Andy-Millar the link you mention is pre-existent to this PR. I will investigate why it is there.

Andy-Millar:
Thanks @barbaracassani !

app/allocation/controllers/cancel-move.js
```diff
@@ -0,0 +1,49 @@
...
+    super.middlewareLocals()
+    this.use(this.setAdditionalLocals)
+  }
+
+  setAdditionalLocals(req, res, next) {
+    const {
+      allocation: {
+        moves: { length: movesCount },
+      },
+    } = res.locals
```
solidgoldpig, line 16: ^^^ this is cool and all (and it is - it never crossed my mind one could access length in a destructuring), but as we're only grabbing one value, this seems a lot more confusing to read than:

`const movesCount = res.locals.allocation.moves.length`

solidgoldpig: (and it's 4 additional lines that codeclimate will nag us about ;)

app/allocation/controllers/index.js
```diff
@@ -1,4 +1,5 @@
 const CancelController = require('./cancel')
+const CancelMoveController = require('./cancel-move')
```
solidgoldpig, line 2: We seem to call it removing a move elsewhere. Is it worth calling this `RemoveMove` to differentiate it fully from `Cancel` (which is for the allocation itself, right?) (And yes, `RemoveMove` is kind of weird)

app/allocation/controllers/view.js
```diff
@@ -44,6 +56,7 @@ module.exports = function view(req, res) {
       .map(move => ({
         id: move.id,
         person: move.person,
+        cancelAllocationHref: removeMoveLink(move),
```
solidgoldpig, line 59: Oh boy - see my comment about differentiating the cancels and removes earlier. Is this for cancelling an allocation or removing a move?

barbaracassani: Ok -- I have renamed everything to use `remove move` 👍

PR #597, opened 2020-06-12
chore: Bump dependencies to latest versions

solidgoldpig:
Bumping prettier caused spaces to be inserted between function and ()

Updating `govuk-frontend` and `@ministryofjustice/frontend` has been left for separate PRs


package.json
```diff
@@ -77,19 +77,19 @@
     "cookie-parser": "1.4.5",
     "core-js": "3.6.5",
     "date-fns": "2.14.0",
-    "debug": "4.2.0",
```
teneightfive, line 80: This looks like it's going the other way? Guessing it might have been a rebase conflict or something?

solidgoldpig: Hmmm - look at the versions on https://www.npmjs.com/package/debug.

Anyhow, I've installed debug@4.2.0 rather than debug@latest and all is well...

solidgoldpig:
> Looks good. Just one comment about an upgrade that seemed to be a downgrade.

You ok with the spaces that have been introduced between `function` and `()`?

Issue #512, opened 2020-05-14
Add githook to prevent accidental commits or pushes to master


solidgoldpig:
Branch protection only prevent force pushes. It is (apparently) still all too easy to commit and push a change straight to master.

PR #598, opened 2020-06-12
chore: Prevent accidental commits to master #512

solidgoldpig:
### What changed

Adds an npm script that exits if it detects the current branch is master and runs that as the first step of the pre-commit hook.

### Why did it change

Accidental commits and pushes to master have been know to happen.

### Issue tracking

#512 

### Environment variables

### Other considerations

### DIFF
--- a/package.json
+++ b/package.json
@@ -189,6 +189,7 @@
     "coverage": "nyc --reporter=html --reporter=lcov npm test",
     "coverage:open": "npm run coverage && open-cli ./coverage/index.html",
     "develop": "run-p watch:js:*",
+    "git:checkbranch": "CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD); if [ \"$CURRENT_BRANCH\" = \"master\" ]; then echo 'Committing to the master branch is not allowed'; exit 1; fi",
     "lint": "eslint .",
     "lint-staged": "lint-staged",
     "postrelease": "npm run release:tags",
@@ -226,7 +227,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "lint-staged"
+      "pre-commit": "npm run git:checkbranch && lint-staged"
     }
   },
   "lint-staged": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -226,7 +227,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "lint-staged"
+      "pre-commit": "npm run git:checkbranch && lint-staged"
```
teneightfive, line 230: Soooo I this this is actually going to prevent us from running `npm run release` which does actually commit and push to master in one command.

Trying to think an alternative but nothing comes to mind at the moment, let's catch-up next week and have a think. But I don't think we can merge this if it prevents us running releases.

solidgoldpig: Aha. We might be able to pass the skip flag as an arg to the standard-version command.

solidgoldpig: https://github.com/conventional-changelog/standard-version/blob/4e4c0f37ef1e2254dfbb1ed70ce4a02459d7c087/lib/lifecycles/commit.js#L23

solidgoldpig: Alternatively

```js
    "git:checkbranch": "if [ \"$SKIP_MASTER_CHECK\" = \"true\" ]; then exit 0; fi; CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD); if [ \"$CURRENT_BRANCH\" = \"master\" ]; then echo 'Committing to the master branch is not allowed'; exit 1; fi",
...
    "release": "SKIP_MASTER_CHECK=true standard-version -a",
```

solidgoldpig: `standard-version --no-verify -a`

works just fine. Is there any reason we need to run the `husky` pre-commit check before the release's auto-commit?

teneightfive: >Is there any reason we need to run the husky pre-commit check before the release's auto-commit?

The only reason would be to ensure the staged files have been linted, but in theory they would have already been in previous commits. 

Does `--no-verify` stop anything else running? If not, go with that, otherwise the alternative also looks good

solidgoldpig: I commented out the actual commit line in `standard-release` and simply consoled out the command it would have run. All seemed to be ok

PR #610, opened 2020-06-22
[P4-1672] Rework create move flow to use profile

solidgoldpig:
### What changed

Save controller of create move now creates a profile and uses the resulting profile when creating the move.

### Why did it change

To support api-improvements

### Issue tracking

- [P4-1672 - Rework create move flow to use profile](https://dsdmoj.atlassian.net/browse/P4-1672)

## Screenshots

Functionally identical

#### Automated testing

#### Manual testing

- [x] Chrome

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/create/save.js
+++ b/app/move/controllers/create/save.js
@@ -3,7 +3,7 @@ const { get, omit, capitalize, flatten, values, some } = require('lodash')
 const analytics = require('../../../../common/lib/analytics')
 const courtHearingService = require('../../../../common/services/court-hearing')
 const moveService = require('../../../../common/services/move')
-const personService = require('../../../../common/services/person')
+const profileService = require('../../../../common/services/profile')
 
 const CreateBaseController = require('./base')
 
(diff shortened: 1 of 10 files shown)


app/move/controllers/create/save.js
```diff
@@ -23,14 +23,21 @@ class SaveController extends CreateBaseController {
         'errors',
         'errorValues',
       ])
-      const move = await moveService.create(data)
+
+      const profile = await profileService.create(data.person.id, {
+        assessment_answers: data.assessment,
+      })
+      delete data.person
+      delete data.assessment
```
barbaracassani, line 31: Isn't preferable to copy first and alter later?

teneightfive: Could also use `_.omit()` to remove the keys you don't want to send to `moveService.create()`?

solidgoldpig: Yes, good point. I was being lazy. I'll rejig it

PR #613, opened 2020-06-22
[P4-1389] Add support for regions within allocations

barbaracassani:
## Proposed changes

Currently the PMU user sees a list of "example locations" but in reality we want them to be able to browse by region. 

### What changed

We show the list of regions and we introduce the appropriate filter for allocations.

### Why did it change

PMU's work is region-based, not location-based.

### Issue tracking

- [P4-1389]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd>Screenshot 2020-06-23 at 10 44 00</kbd> | <kbd>Screenshot 2020-06-23 at 10 41 45</kbd> |
| <kbd>Screenshot 2020-06-23 at 10 44 11</kbd> | <kbd>Screenshot 2020-06-23 at 10 41 56</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -1,7 +1,15 @@
-const { sortBy } = require('lodash')
+const { get, sortBy } = require('lodash')
 
 function locations(req, res) {
-  const locations = sortBy(req.userLocations, 'title')
+  const userPermissions = get(req.session, 'user.permissions', [])
+
+  let locations
+
+  if (userPermissions.includes('allocation:create')) {
+    locations = req.session.regions
+  } else {
+    locations = sortBy(req.userLocations, 'title')
+  }
 
   res.render('locations/views/locations.njk', {
     locations,
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -11,7 +11,6 @@ async function locations(req, res, next) {
   if (userPermissions.includes('allocation:create')) {
     try {
       regions = await referenceDataService.getRegions()
-      req.session.regions = regions
     } catch (error) {
       next(new Error('Failed to retrieve the regions'))
       return
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -5,19 +5,16 @@ const referenceDataService = require('../../common/services/reference-data')
 async function locations(req, res, next) {
   const userPermissions = get(req.session, 'user.permissions', [])
 
-  let locations = []
   let regions = []
 
   if (userPermissions.includes('allocation:create')) {
     try {
       regions = await referenceDataService.getRegions()
     } catch (error) {
-      next(new Error('Failed to retrieve the regions'))
-      return
+      return next(error)
     }
-  } else {
-    locations = sortBy(req.userLocations, 'title')
   }
+  const locations = sortBy(req.userLocations, 'title')
 
   res.render('locations/views/locations.njk', {
     locations,
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -1,18 +1,28 @@
 const { get, sortBy } = require('lodash')
 
-function locations(req, res) {
+const referenceDataService = require('../../common/services/reference-data')
+
+async function locations(req, res, next) {
   const userPermissions = get(req.session, 'user.permissions', [])
 
-  let locations
+  let locations = []
+  let regions = []
 
   if (userPermissions.includes('allocation:create')) {
-    locations = req.session.regions
+    try {
+      regions = await referenceDataService.getRegions()
+      req.session.regions = regions
+    } catch (error) {
+      next(new Error('Failed to retrieve the regions'))
+      return
+    }
   } else {
     locations = sortBy(req.userLocations, 'title')
   }
 
   res.render('locations/views/locations.njk', {
     locations,
+    regions,
   })
 }
 
--- a/app/locations/index.js
+++ b/app/locations/index.js
@@ -20,7 +20,7 @@ router.use(setUserLocations)
 router.get('/', checkLocationsLength, setRegions, locations)
 router.get('/all', setAllLocations, redirect)
 router.get('/:locationId', setLocation, redirect)
-router.get('/region/prisons/:regionId', setRegion, redirect)
+router.get('/region/:regionId', setRegion, redirect)
 
 // Export
 module.exports = {
--- a/app/locations/index.js
+++ b/app/locations/index.js
@@ -9,15 +9,18 @@ const {
   setUserLocations,
   checkLocationsLength,
   setLocation,
+  setRegion,
   setAllLocations,
+  setRegions,
 } = require('./middleware')
 
 router.use(setUserLocations)
 
 // Define routes
-router.get('/', checkLocationsLength, locations)
+router.get('/', checkLocationsLength, setRegions, locations)
 router.get('/all', setAllLocations, redirect)
 router.get('/:locationId', setLocation, redirect)
+router.get('/region/prisons/:regionId', setRegion, redirect)
 
 // Export
 module.exports = {
--- a/app/locations/index.js
+++ b/app/locations/index.js
@@ -11,16 +11,15 @@ const {
   setLocation,
   setRegion,
   setAllLocations,
-  setRegions,
 } = require('./middleware')
 
 router.use(setUserLocations)
 
 // Define routes
-router.get('/', checkLocationsLength, setRegions, locations)
+router.get('/', checkLocationsLength, locations)
 router.get('/all', setAllLocations, redirect)
 router.get('/:locationId', setLocation, redirect)
-router.get('/region/:regionId', setRegion, redirect)
+router.get('/regions/:regionId', setRegion, redirect)
 
 // Export
 module.exports = {
--- a/app/locations/middleware.js
+++ b/app/locations/middleware.js
@@ -36,11 +36,11 @@ async function setRegion(req, res, next) {
   }
 
   try {
-    const regions = await referenceDataService.getRegions()
-    req.session.currentRegion = find(regions, { id: regionId }) ?? null
+    const region = await referenceDataService.getRegionById(regionId)
+    req.session.currentRegion = region ?? null
   } catch (error) {
     req.session.currentRegion = null
-    next(new Error('Failed to retrieve the regions'))
+    next(error)
     return
   }
 
--- a/app/locations/middleware.js
+++ b/app/locations/middleware.js
@@ -1,5 +1,7 @@
 const { find, get } = require('lodash')
 
+const referenceDataService = require('../../common/services/reference-data')
+
 function setUserLocations(req, res, next) {
   req.userLocations = get(req.session, 'user.locations', [])
   next()
@@ -15,9 +17,24 @@ function checkLocationsLength(req, res, next) {
 
 function setLocation(req, res, next) {
   const { locationId } = req.params
+
   const location = find(req.userLocations, { id: locationId })
 
   req.session.currentLocation = location
+  next()
+}
+
+function setRegion(req, res, next) {
+  const { regionId } = req.params
+
+  req.session.currentLocation = null
+
+  if (regionId) {
+    req.session.currentRegion = find(req.session.regions, { id: regionId })
+  } else {
+    req.session.currentRegion = undefined
+  }
+
   next()
 }
 
@@ -32,9 +49,20 @@ function setAllLocations(req, res, next) {
   next()
 }
 
+async function setRegions(req, res, next) {
+  try {
+    req.session.regions = await referenceDataService.getRegions()
+    next()
+  } catch (error) {
+    next(new Error('Failed to retrieve the regions'))
+  }
+}
+
 module.exports = {
   setUserLocations,
   checkLocationsLength,
   setLocation,
+  setRegion,
   setAllLocations,
+  setRegions,
 }
--- a/app/locations/middleware.js
+++ b/app/locations/middleware.js
@@ -1,5 +1,7 @@
 const { find, get } = require('lodash')
 
+const referenceDataService = require('../../common/services/reference-data')
+
 function setUserLocations(req, res, next) {
   req.userLocations = get(req.session, 'user.locations', [])
   next()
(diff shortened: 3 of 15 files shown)


common/services/reference-data.js
```diff
@@ -92,6 +92,10 @@ const referenceDataService = {
     })
   },
 
+  getRegions() {
+    return apiClient.findAll('region').then(response => response.data)
```
teneightfive, line 96: This might hit a snag if there are more than 100 regions as I believe that's the current API page size limit. We might need to do something more like the locations method to ensure we get all results.

app/locations/views/locations.njk
```diff
@@ -29,13 +29,15 @@
         <ul class="govuk-list govuk-!-font-size-24 govuk-!-font-weight-bold">
           {% for location in locations %}
             <li>
-              <a href="/locations/{{ location.id }}">
                 {% if location.type == "regions" %}
-                  {{ location.key }} {{ location.name }}
+                  <a href="/locations/{{ location.id }}?isRegion=true">
```
teneightfive, line 33: I wonder if it might be nicer to separate the two features. Does this approach mean that somebody who sees regions cannot access a single location anymore?

barbaracassani: That is something I did wonder about. Does PMU ever need to access a single location?

app/locations/middleware.js
```diff
@@ -17,20 +17,27 @@ function checkLocationsLength(req, res, next) {
...
   const location = find(req.userLocations, { id: locationId })
 
   req.session.currentLocation = location
   next()
 }
 
+function setRegion(req, res, next) {
+  const { regionId } = req.params
+
+  req.session.currentLocation = null
```
teneightfive, line 30: Is there any reason we set `currentLocation` to `null` but `currentRegion` to `undefined` when they're not set?

app/locations/index.js
```diff
@@ -9,17 +9,18 @@ const {
...
 } = require('./middleware')
 
 router.use(setUserLocations)
 
 // Define routes
-// todo apply only to PMU journeys
 router.get('/', checkLocationsLength, setRegions, locations)
 router.get('/all', setAllLocations, redirect)
 router.get('/:locationId', setLocation, redirect)
+router.get('/region/prisons/:regionId', setRegion, redirect)
```
teneightfive, line 23: Although we established the current regions from the API only return prisons I wonder if we do want to add that to the URL or not as the frontend should probably be more flexible and support other _regions_ in the future. So maybe this is just `/region/:regionId`?

app/locations/index.js
```diff
@@ -9,17 +9,18 @@ const {
...
   setLocation,
+  setRegion,
   setAllLocations,
   setRegions,
 } = require('./middleware')
 
 router.use(setUserLocations)
 
 // Define routes
-// todo apply only to PMU journeys
```
teneightfive: On a note for this, we may need to add another permissions to restrict the region routes and the display of them in the view behind that permission

app/locations/views/locations.njk
```diff
@@ -30,7 +30,7 @@
           {% for location in locations %}
             <li>
                 {% if location.type == "regions" %}
-                  <a href="/locations/{{ location.id }}?isRegion=true">
+                  <a href="/locations/region/prisons/{{ location.id }}">
                     {{ location.key }} {{ location.name }}
```
teneightfive, line 34: Would it be possible to move the key outside of the link a bit more like the prototype please:

image

I also wonder whether we can group the regions separately so that as we discussed, if a user has both permissions they still have the ability to navigate via a single location.

Maybe two separate blocks like this:

image

app/locations/middleware.test.js
```diff
@@ -266,26 +266,57 @@ describe('Locations middleware', function () {
...
     })
 
     context('when the API is available', function () {
-      it('should retrieve all regions', function () {
-        mockReferenceData.getRegions = sinon.stub().returns(Promise.resolve([]))
-        proxiedMiddleware.setRegions(req, {}, nextSpy)
-        expect(req.session).to.have.property('regions')
-        expect(req.session.regions).to.deep([])
+      it('should retrieve all regions', async function () {
+        mockReferenceData.getRegions = sinon.fake.returns(Promise.resolve([]))
```
solidgoldpig, line 273: `sinon.fake.resolves([])` ?

app/locations/middleware.js
danielepolencic: `find` returns `undefined` when the item isn't found in the list. I thought I'd keep the same return value. I can change it to `null`.

app/locations/views/locations.njk
danielepolencic: > Does this approach mean that somebody who sees regions cannot access a single location anymore?
This is addressed in a future ticket: P4-1607

app/locations/index.js
danielepolencic: I used to have `/region/:regionId` and changed it to `/region/prisons` because the endpoint deals only with prisons.

I think the route should describe what it does, but I'm happy to revert the change.

app/locations/index.js
danielepolencic: You mean the `/region/:regionId` route?

app/locations/views/locations.njk
danielepolencic: I didn't notice the difference, my bad. Moving the key outside.

As for the grouping, I asked and the answer is that it won't happen in this story. However, we have a future story designed to address the issue.

app/locations/middleware.test.js
danielepolencic: `getRegions` is an `async` function, I thought I could keep the same return type.
Happy to change it.

solidgoldpig: Fair enough.

app/locations/index.js
teneightfive: Yeah, I do agree. It's just now awkward because the API endpoints don't follow the same pattern.

The other thing that doesn't feel very nice is that we will have `404`s at both levels now and I tend to find that a bad design when building nested URLs within an application. Maybe it would be okay if we redirect those somewhere so they don't just have a dead-end. What do you think?

Basically my view is that you should normally be able to remove portions of URLs and still find meaningful things at those URLs. `/region` and `/region/prisons` now wouldn't contain anything.

app/locations/views/locations.njk
teneightfive: So the problem with currently only showing one or the other is that if a person how both a PMU role and any other kind of role, they can't use the rest of the service because the regions are only used for allocations. For example, you can't now view outgoing or incoming moves for a single location because it is only possible to select a region.

It also wouldn't be possible to create a new move because those journeys rely on a current location being set.

Would it be much work to change the nesting of the if on this template to just wrap regions and always show locations instead of it being an if/else?

app/locations/controllers.js
```diff
@@ -1,10 +1,28 @@
...
+async function locations(req, res, next) {
+  const userPermissions = get(req.session, 'user.permissions', [])
+
+  let locations = []
+  let regions = []
+
+  if (userPermissions.includes('allocation:create')) {
+    try {
+      regions = await referenceDataService.getRegions()
+      req.session.regions = regions
```
teneightfive, line 14: Do you need to store the regions on the session? You raised a good point about what needs to happen when serializing and deserializing the session so we can probably drop the need to set it here.

teneightfive:
I just tested the review app and the allocations seem to filter correctly. However, I don't think we're passing the regions to any endpoints to display moves (single requests, outgoing, incoming).

I'm not sure what the expected behaviour is, or if we can ignore this because these users might not have access to those areas, but I have a feeling PMU can view single requests so we might need to ensure that those are also correctly filtered. I'll add some screenshots to show what I mean.

The counts for outgoing and single requests remains the same but allocations count changes:

| South | London |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

app/locations/index.js
danielepolencic: I reverted to `/regions/:regionId`.

> Basically my view is that you should normally be able to remove portions of URLs and still find meaningful things at those URLs. `/region` and `/region/prisons` now wouldn't contain anything.

This is a bit at odds with the current practice of (ab)using middlewares. We have routes that exist only to set values (and redirect), but they have no actual meaning to users.

I checked and dropping the `prison` suffix solves the issue. When the `regionId` is not set, we still route the user to a valid page.

app/locations/views/locations.njk
danielepolencic: Unfortunately, I don't have enough context to make an informed decision (the proposed change seems small, though).
Also, the proposal changes the scope of the story and should be discussed with the rest of the team. I copy-pasted the comment into the Jira ticket so that we can keep the conversation going.

app/locations/controllers.js
danielepolencic: Refactored the `setRegion` middleware to _not_ use the values in the session.

app/locations/middleware.js
teneightfive: I think `undefined` makes sense. Maybe we set `req.session.currentLocation` to undefined too for consistency?

app/locations/index.js
teneightfive: >This is a bit at odds with the current practice of (ab)using middlewares. We have routes that exist only to set values (and redirect), but they have no actual meaning to users.

Yep, you're right in that there's nothing that actually lives _there_. What approach do you normally take for such actions? Would you just handle it within a middleware instead of a controller for that route? Ultimately the route needs some form of middleware/controller defined on it to handle that action.

>I checked and dropping the prison suffix solves the issue. When the regionId is not set, we still route the user to a valid page.

👍

app/locations/index.js
teneightfive: Yes, sorry, that's the route I meant. I suppose it depends how strict we want to be from an application perspective.

If the logic is handled in the view where the user would never see the link, how much do we feel we want to protect the route? In this instance, if a user goes to it manually the outcome doesn't feel as bad as if a user was trying to access a resource that they shouldn't have permission to do.

app/locations/views/locations.njk
teneightfive: Thanks. Have also followed up on the ticket to help with the conversation. I think my concern comes from knowing that there at, at the very least, test accounts (my own included) that contain both PMU and other roles. 

And this current implementation would mean that it would be harder for me to use the other parts of the service like creating a move if I sign in with that account. The change seems small to support both and I think that's a nicer approach moving forward. That permissions should add/remove things in most cases instead of completely changing feature behaviour.

app/locations/controllers.js
```diff
@@ -1,10 +1,27 @@
...
+
+  let locations = []
+  let regions = []
+
+  if (userPermissions.includes('allocation:create')) {
+    try {
+      regions = await referenceDataService.getRegions()
+    } catch (error) {
+      next(new Error('Failed to retrieve the regions'))
+      return
```
teneightfive, line 16: Could you just `return` next?

app/locations/controllers.js
```diff
@@ -1,10 +1,27 @@
...
+    }
+  } else {
+    locations = sortBy(req.userLocations, 'title')
```
teneightfive, line 19: Could you use this as the default value for `locations` instead of having to have an extra `else`?

app/locations/middleware.js
```diff
@@ -15,12 +17,36 @@ function checkLocationsLength(req, res, next) {
...
+    next()
+    return
+  }
+
+  try {
+    const regions = await referenceDataService.getRegions()
+    req.session.currentRegion = find(regions, { id: regionId }) ?? null
+  } catch (error) {
+    req.session.currentRegion = null
+    next(new Error('Failed to retrieve the regions'))
```
teneightfive, line 43: Could it be better to pass the error to next so that the stacktrace will contain a bit more information about the failure?

app/locations/views/locations.njk
```diff
@@ -34,6 +34,14 @@
               </a>
             </li>
           {% endfor %}
+          {% for region in regions %}
```
teneightfive, line 37: If we're combing both lists it might be better to list the regions first as there are likely to be less of them.

Some users who may have access to regions could have access to a lot of locations via NOMIS "caseloads" so the regions will appear a long way down the page.

I wonder if it might be nicer to use separate `ul`s too as the data within the list is technically different?

app/locations/controllers.js
teneightfive: Ah, I think this else is what is still causing a user with both roles to not be able to see locations.

I feel like we should always pass locations to the view and that regions should be behind a permission.

That way users still have the ability to see a particular location if they want but also the ability to see locations grouped by region.

app/locations/middleware.js
danielepolencic: I removed the `getRegions` and `find` and replaced it with a `getRegionById`.

app/locations/index.js
danielepolencic: I think `/region/:regionId` should display the following page:

screenshotprison

As it should `/locations/:locationId`.
If I understand the domain model correctly, moves belong to locations, so the current URL could change to:

```
PREVIOUS: http://localhost:3000/moves/day/2020-07-06/8fadb516-f10a-45b1-91b7-a256196829f9/outgoing
PROPOSAL http://localhost:3000/location/region/<regionId>/moves/?day=2020-07-06&type=outgoing
```

The day and the direction are not paths, but filters for the current view. They could be repurposed as query string arguments.

Notice how the proposed URL drills down into the resources:

```
locations -> regions -> moves
``` 

The structure simplifies retrieving (and link sharing) the current state of the UI. I can bookmark the URL and get back to it or I can share it with a coworker. I can also remove parts of the URL and navigate resources (i.e. `/locations/regions/:regionId` should display a list of all regions).
On the technical side, we can rebuild the entire state of your session from the URL (there's no magic middleware journey to follow).

I understand that things started with the best intention and grew organically to this. I'm not expecting to change _all_ routes now, neither in the near future.
However, I think it's important to discuss it so that we can mitigate some of the drawbacks such as misusing the user session.

app/locations/controllers.js
danielepolencic: I don't know how the backend works, but they will be able to see all moves for a region I believe.
Is the API checking the permissions for a user?

app/locations/index.js
danielepolencic: With the current code, you will be able to retrieve all moves for that region.
I'm not sure how the backend works — does it check the permissions for a user?
If it doesn't, anyone can use the full region dashboard as long as they know the ID of the region.

app/locations/views/locations.njk
danielepolencic: I implemented the list with your design and not how the ticket described.
I left a comment in the ticket.

app/locations/controllers.js
```diff
@@ -1,10 +1,25 @@
...
+  let regions = []
+
+  if (userPermissions.includes('allocation:create')) {
+    try {
+      regions = await referenceDataService.getRegions()
+    } catch (error) {
+      return next(error)
+    }
+  }
+  locations = sortBy(req.userLocations, 'title')
```
teneightfive, line 18: Any reason to not set this on line 9 (or even 7 next to the other `const` statement) and use `const`? Doesn't appear to be reset anyway so could combine that statement.

PR #617, opened 2020-06-24
[P4-1734] Update form wizard to support ingesting YAML framework

teneightfive:
### What changed

This addition means that the form wizard will always set the correct
`dependent` property on a field if it is added as a conditional
field within another field's options.

This PR also includes a small change to an existing helper and the addition of
page caption support within a form wizard step.

### Why did it change

Previously creating conditional fields was a two step process that
involved adding a field as a conditional option to a parent as well
as adding a dependent property to the child field to ensure validation
is only applied in the correct circumstances.

This change doesn't affect any previous implementations but makes this
process easier moving forward so that when dynamically creating fields,
like will be the case for the Person Escort Record framework, you
don't need to loop over the fields again to set the dependent values.
This is now taken care of within the form wizard itself.

### Issue tracking

- [P4-1734](https://dsdmoj.atlassian.net/browse/P4-1734)

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/controllers/form-wizard.js
+++ b/common/controllers/form-wizard.js
@@ -1,10 +1,57 @@
 const Sentry = require('@sentry/node')
 const { Controller } = require('hmpo-form-wizard')
-const { map, fromPairs } = require('lodash')
+const { flatten, map, fromPairs, forEach } = require('lodash')
 
 const fieldHelpers = require('../helpers/field')
 
 class FormController extends Controller {
+  middlewareSetup() {
+    super.middlewareSetup()
+    this.use(this.setupConditionalFields)
+  }
+
+  setupConditionalFields(req, res, next) {
+    const allFields = req.form.options.allFields
+    const stepFields = req.form.options.fields
+    const dependentFields = {}
+
+    forEach(stepFields, (field, fieldName) => {
+      if (!field.items) {
+        return
+      }
+
+      field.items.forEach(item => {
+        const conditionalFields = flatten([item.conditional || []])
+
+        conditionalFields.forEach(key => {
+          const conditionalField = allFields[key]
+
+          if (!conditionalField) {
+            return
+          }
+
+          dependentFields[key] = {
+            ...conditionalField,
+            // tell the form wizard to not output field initially as it will be nested
+            skip: true,
+            // set dependent object for validation
+            dependent: {
+              field: fieldName,
+              value: item.value,
+            },
+          }
+        })
+      })
+    })
+
+    req.form.options.fields = {
+      ...stepFields,
+      ...dependentFields,
+    }
+
+    next()
+  }
+
   getErrors(req, res) {
     const errors = super.getErrors(req, res)
     const errorList = map(errors, ({ key, type }) => {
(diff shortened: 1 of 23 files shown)


common/controllers/form-wizard.js
```diff
@@ -1,10 +1,51 @@
...
+    this.use(this.setupConditionalFields)
+  }
+
+  setupConditionalFields(req, res, next) {
+    const allFields = req.form.options.allFields
+    const stepFields = req.form.options.fields
+    const dependentFields = {}
+
+    forEach(stepFields, (field, key) => {
+      if (field.items) {
```
danielepolencic, line 19: I think you could have less nesting and simpler reasoning if you were to use a guard.

```js
if (!field.items) {
  return
}
// rest of the code
```

common/controllers/form-wizard.js
```diff
@@ -1,10 +1,51 @@
...
+        field.items.forEach(item => {
```
danielepolencic, line 20: The code boils down to:

```js
const dependentFields = {} // accumulator
conditionalFields.forEach(it => {
  /* code */
  /* add to accumulator /*
})
```

This is a hand-rolled `.reduce`. You could replace the `forEach` with `reduce` (or `fold`, not sure what's called in lodash).

common/controllers/form-wizard.js
teneightfive: I did consider that but the empty return and inconsistent behaviour sort of stopped me. 

I'll use the early return :)

common/controllers/form-wizard.js
teneightfive: The slight issue with that is that the `dependentFields` needs to be built outside of the parent loop of `conditionalFields`. I'll see if I can tweak the solution to use `reduce` though

teneightfive: @danielepolencic I wasn't really able to get it neatly working so have tweaked it to use early return but kept the `forEach` for now.

If you can post a code snippet I'd be happy to push another change to fix it after?

danielepolencic: No worries, I had a go at it and I understand your point better now. It's not straightforward.
The code has 3 nested layers of loops. It could be refactored and flattened like this:

```js
  _(stepFields)
    .chain()
    .filter(it => Array.isArray(it.items))
    .map((item, fieldName) => {
      return _(flatten([item.conditional || []]))
        .chain()
        .map(it => allFields[it])
        .filter(it => !!it)
        .map(it => {
          return {
            ...it,
            // tell the form wizard to not output field initially as it will be nested
            skip: true,
            // set dependent object for validation
            dependent: {
              field: fieldName,
              value: item.value,
            },
          }
        })
        .value()
    })
    .value()
```

I'm not terribly familiar with lodash, so the above might need tweaking (also, no idea on how to chain functions, I usually use the native `.map`, `.reduce`, etc.).

You don't have to implement the change and I don't see a problem if it's addressed later.

PR #622, opened 2020-06-25
[P4-1734] Add route handler to show framework steps

teneightfive:
### What changed

This creates a form wizard for each section of the person escort
framework passing in the transformed steps and questions.

The form wizard then handles next steps and form validation.

### Why did it change

This adds support to render the form steps generated from the [person escort record framework](https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks)

### Issue tracking

- [P4-1734](https://dsdmoj.atlassian.net/browse/P4-1734)

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/person-escort-record/index.js
+++ b/app/person-escort-record/index.js
@@ -1,0 +1,17 @@
+// NPM dependencies
+const router = require('express').Router()
+
+// Local dependencies
+const frameworksService = require('../../common/services/frameworks')
+
+const { defineFormWizards } = require('./router')
+
+// Define routes
+const framework = frameworksService.getPersonEscortFramework()
+defineFormWizards(framework, router)
+
+// Export
+module.exports = {
+  router,
+  mountpath: '/person-escort-record',
+}
--- a/app/person-escort-record/middleware.js
+++ b/app/person-escort-record/middleware.js
@@ -1,0 +1,10 @@
+function setFramework(framework) {
+  return (req, res, next) => {
+    req.framework = framework
+    next()
+  }
+}
+
+module.exports = {
+  setFramework,
+}
--- a/app/person-escort-record/middleware.test.js
+++ b/app/person-escort-record/middleware.test.js
@@ -1,0 +1,41 @@
+const middleware = require('./middleware')
+
+describe('Person escort record middleware', function () {
+  describe('#setFramework()', function () {
+    let mockRes, mockReq, nextSpy
+
+    beforeEach(function () {
+      nextSpy = sinon.spy()
+      mockRes = {}
+      mockReq = {
+        params: {},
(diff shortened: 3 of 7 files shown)


app/person-escort-record/controllers.js
```diff
@@ -0,0 +1,30 @@
...
+
+function sectionHandler(req, res, next) {
+  const { questions } = req.framework
+  const { key: sectionKey, steps: sectionSteps } = req.frameworkSection
+  const wizardConfig = {
+    buttonText: 'actions::save_and_continue',
+    controller: FormWizardController,
+    entryPoint: true,
+    journeyName: `person-escort-record-${sectionKey}`,
+    journeyPageTitle: 'Person escort record',
```
solidgoldpig, line 13: Get from `locales`? (Or am I misremembering what `journeyPageTitle` is used for?)

app/person-escort-record/controllers.js
```diff
@@ -0,0 +1,30 @@
+const wizard = require('hmpo-form-wizard')
+
+const FormWizardController = require('../../common/controllers/form-wizard')
+
+function sectionHandler(req, res, next) {
+  const { questions } = req.framework
+  const { key: sectionKey, steps: sectionSteps } = req.frameworkSection
```
solidgoldpig, line 7: Rather than passing the framework in via `req`, could you not just pass the frameworkSection wanted, and then require in the framework here?

I think it would also make for simpler mocking.

app/person-escort-record/controllers.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+  const steps = {
+    '/': {
+      reset: true,
+      resetJourney: true,
+    },
+    ...sectionSteps,
+  }
+
+  return wizard(steps, questions, wizardConfig)(req, res, next)
```
solidgoldpig, line 25: If you passed the framework section as a string (as above) rather than the whole object, we could cache and reuse the middleware returned by the wizard.

app/person-escort-record/controllers.js
teneightfive: Yeah, good point. Might need a person-escort-record.json file...

app/person-escort-record/controllers.js
teneightfive: So the purpose of that was for some changes coming in later PRs. Basically we will need to use it in other controllers and setting in a middleware felt like a cleaner way to re-use this object. 

It also means we only need to call the method to load the framework from files once which felt more efficient.

PR #623, opened 2020-06-25
[P4 1618] Add explicit include to nested personService methods

solidgoldpig:
### What changed
    
 - Add `location` to personService's `getActiveCourtCases` and `getTimetableByDate` methods.
   
### Why did it change

Default includes will be removed from all endpoint under v2.

See https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/633

`court_case` and 'timetable` had been missed and were still relying on the default rather than explicit behaviour.

### Issue tracking

- [P4-1618 - Remove default includes wherever they are defined
](https://dsdmoj.atlassian.net/browse/P4-1618)

## Screenshots

Functionaly identical

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/lib/api-client/middleware/request-include.js
+++ b/common/lib/api-client/middleware/request-include.js
@@ -19,7 +19,11 @@ module.exports = {
       include = include.sort().join(',')
     }
 
-    req.params.include = include
+    if (include) {
+      req.params.include = include
+    } else {
+      delete req.params.include
+    }
 
     return payload
   },
--- a/common/lib/api-client/middleware/request-include.test.js
+++ b/common/lib/api-client/middleware/request-include.test.js
@@ -44,6 +44,24 @@ describe('API Client', function () {
             })
           })
 
+          context('with include as an empty array', function () {
+            it('should not set the include', function () {
+              request.params.include = []
+              middleware.req({ req: request })
+
+              expect(request.params.include).to.not.exist
+            })
+          })
+
+          context('with include as an empty string', function () {
+            it('should not set the include', function () {
+              request.params.include = ''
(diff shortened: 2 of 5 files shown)


common/lib/api-client/middleware/request-include.js
```diff
@@ -19,7 +19,11 @@ module.exports = {
       include = include.sort().join(',')
     }
 
-    req.params.include = include
+    if (include) {
+      req.params.include = include
+    } else {
+      delete req.params.include
```
teneightfive, line 25: Would it not be enough to just wrap in the if block?

common/lib/api-client/middleware/request-include.js
```diff
@@ -19,7 +19,11 @@ module.exports = {
       include = include.sort().join(',')
     }
 
-    req.params.include = include
+    if (include) {
```
teneightfive, line 22: Also this could do with a unit test update?

common/lib/api-client/middleware/request-include.js
solidgoldpig: Nope. Because we need to remove the property if the value of include is falsy rather than just set it to undefined.

common/lib/api-client/middleware/request-include.js
solidgoldpig: Yep. That would have clarified what the purpose of the change was.

And if the tests continue passing after the change to make the code good, they can carry on passing if someone makes the change back the other way...

PR #625, opened 2020-06-26
[P4-1735] Support followup comments from framework

teneightfive:
### What changed

This adds support to create conditional comments fields from the manifest where an option has defined
followup comments. 

This change included updating the form wizard support to handle objects as part of a field options conditionals
and to ensure validation works correctly and is rendering correctly.

### Why did it change

For particular question types within the form wizard we want to support the ability to provide additional
comments. Rather than requiring a new question and therefore response for each item we wanted to support
these type of comments fields being dynamically created.

This required some of that heavy lifting to be done on the frontend within the form wizard.

**Note:** This PR doesn't address the error messages not currently bringing in the correct content.

### Issue tracking

- [P4-1735](https://dsdmoj.atlassian.net/browse/P4-1735)

## Screenshots

| Required | Optional |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/helpers/field/flatten-conditional-fields.js
+++ b/common/helpers/field/flatten-conditional-fields.js
@@ -1,0 +1,28 @@
+const { isObject, concat } = require('lodash')
+
+function flattenConditionalFields([key, field]) {
+  if (!field.items) {
+    return [key, field]
+  }
+
+  const items = field.items.map(item => {
+    if (!item.conditional) {
+      return item
+    }
+
+    const conditionals = concat([], item.conditional)
+    const conditionalKeys = conditionals.map(conditional => {
+      if (isObject(conditional)) {
+        return conditional.name
+      }
+
+      return conditional
+    })
+
+    return { ...item, conditional: conditionalKeys }
+  })
+
+  return [key, { ...field, items }]
+}
+
+module.exports = flattenConditionalFields
--- a/common/services/frameworks.js
+++ b/common/services/frameworks.js
@@ -1,7 +1,7 @@
 const fs = require('fs')
 const path = require('path')
 
-const { keyBy, set } = require('lodash')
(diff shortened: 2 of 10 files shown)


common/helpers/field/flatten-conditional-fields.js
```diff
@@ -0,0 +1,27 @@
...
+function flattenConditionalFields([key, field]) {
+  if (!field.items) {
+    return [key, field]
+  }
+
+  return [
+    key,
+    {
+      ...field,
+      items: field.items.map(item => {
```
solidgoldpig, line 12: My fave hobby horse -  I think doing the items assignment before the return statement would make it easier to understand both what's happening to the items and also what's being returned

common/services/frameworks.js
```diff
@@ -58,13 +58,28 @@ function transformQuestion(
...
   if (options) {
-    field.items = options.map(({ value, label, followup }) => {
-      return {
-        value,
-        text: label,
-        conditional: followup,
+    field.items = options.map(
+      ({ value, label, followup, followup_comment: followupComment }) => {
+        const commentField = {
+          rows: 4,
```
solidgoldpig, line 64: Will the `rows` and classes values always be the same?

teneightfive: Yes, for this "field" the intention is to always display them in the same manor. Basically the desire is to be able to capture "more details" about a particular yes/no question without having to create a new question within the framework and send it as a separate response.

In future some of these may be replaced with more specific fields to try and capture better data from users.

PR #627, opened 2020-06-26
[P4-1736] Support custom validation messages in form wizard

teneightfive:
### What changed

This adds support to define a custom validation message within the
fields configuration that can be used to render the validation
message.

### Why did it change

The form wizard currently uses the translation files to build form validation messages.

However, when importing the frameworks we want to keep the validation message tied to the field
configuration. The manifest defines that the message can be supplied as part of the validation object.

With this change we can pass this directly to the form wizard and then extract that message 
so that it can be used to render the message in the error summary and on the field.

### Issue tracking

- [P4-1736](https://dsdmoj.atlassian.net/browse/P4-1736)

## Screenshots

| Validation message output | Manifest file defintion |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/controllers/form-wizard.js
+++ b/common/controllers/form-wizard.js
@@ -54,10 +54,10 @@ class FormController extends Controller {
 
   getErrors(req, res) {
     const errors = super.getErrors(req, res)
-    const errorList = map(errors, ({ key, type }) => {
+    const errorList = map(errors, error => {
       return {
-        html: fieldHelpers.getFieldErrorMessage(key, type),
-        href: `#${key}`,
+        html: fieldHelpers.getFieldErrorMessage(error),
+        href: `#${error.key}`,
       }
     })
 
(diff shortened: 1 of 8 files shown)


solidgoldpig:
Shouldn't the summary error messages be in the same order as the fields?

teneightfive:
>Shouldn't the summary error messages be in the same order as the fields?
Ah, the sort of are, but also aren't...I think because we currently dynamically add the conditional fields to the end of fields object their validation is triggered after the top level fields.

Therefore that's what it looks like the ordering is off. I'll make a note and see if it's something that can be easily improved.

PR #633, opened 2020-07-01
[P4-1739] Create interruption card support for Person Escort Record framework

teneightfive:
### What changed

Adds support to use custom step types from the frameworks. This initial piece of work supports
using the [Interruption card](https://home-office-digital-patterns.herokuapp.com/patterns/interruption-card) pattern to direct users to things they must carry out offline.

To support this, it also involved adding markdown support to this PR.

### Why did it change

One of the journeys within the Person Escort Record requires the user to take action outside
of the digital service. The interruption card pattern allows us to describe those instructions in
a clear way for now. 

### Issue tracking

- [P4-1739](https://dsdmoj.atlassian.net/browse/P4-1739)

## Screenshots

<kbd></kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


nodemon.json
```diff
@@ -6,11 +6,10 @@
     "config/",
     "locales/",
     "server.js",
-    "start.js"
-  ],
-  "ignore": [
-    ".build/"
+    "start.js",
+    ".build/frameworks/**/*.json"
```
solidgoldpig, line 10: Is it worth adding `.env` too?

app/person-escort-record/views/form-step.njk
```diff
@@ -1,15 +1,19 @@
 {% extends "form-wizard.njk" %}
 
+{% set mainClasses = "interruption-card" if options.stepType == 'interruption-card' %}
```
solidgoldpig, line 3: Just a thought - is it worth having a `form-step` and `form-interruption-card` template?

nodemon.json
teneightfive: Not sure just restarting the server would be enough would it? Wouldn't you need to reload you shell session too?

app/person-escort-record/views/form-step.njk
teneightfive: Not a bad thought. Although would mean updating the render method to handle changing of the template and probably now having 3 templates (one base to inherit from, plus one for the two _types_).

Maybe one to consider if the type becomes more drastically different in what it renders but in this case it's only applying a class and hiding the save action.

What do you think?

PR #636, opened 2020-07-02
Migrate documents from move to profile

solidgoldpig:
### What changed

Store and fetch documents on/from profile rather than move

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/636/commits/fc45e930eddf510aee34c046d7c6121b550a3a08 can be reverted when we update to v2 of the API

### Why did it change

API improvements

### Issue tracking

n/a

## Screenshots

Functionally identical

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/update/document.js
+++ b/app/move/controllers/update/document.js
@@ -1,4 +1,4 @@
-const { get } = require('lodash')
+const { get, pick } = require('lodash')
 
 const profileService = require('../../../../common/services/profile')
 const DocumentUploadController = require('../create/document')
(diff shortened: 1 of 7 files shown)


app/move/controllers/update/document.js
```diff
@@ -50,8 +50,10 @@ class UpdateDocumentUploadController extends UpdateBase {
     }
 
     try {
-      await moveService.update({
-        id: req.getMoveId(),
+      const profile = req.getMove().profile
+
+      await profileService.update({
+        ...profile,
```
teneightfive, line 56: Do you need to send the whole profile, or can you just send the ID like before? Could prevent updating something you don't want to at this stage.

solidgoldpig: We also have to send the person id too. But dumping any props that aren't those now.

PR #643, opened 2020-07-03
P4 1854 move frontend to version 2

solidgoldpig:
### What changed

Adds version requested to Accept header

### Why did it change

Enabled migration to v2

### Issue tracking

- [P4-1854](https://dsdmoj.atlassian.net/browse/P4-1854)

## Screenshots

Functionally identical

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

- [x] Added to [continous integration](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-book-secure-move-frontend)
- [x] Added to staging environment (deployment repository)
- [x] Added to preproduction environment (deployment repository)
- [x] Added to production environment (deployment repository)

### Other considerations

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,8 @@ The TZ (timezone) environment variable is set to 'Europe/London' in `start.js`.
 | REDIS_HOST **(required)** | Redis hostname. Can be used instead of `REDIS_URL`. Will override `REDIS_URL` if set | |
 | REDIS_AUTH_TOKEN | Optional auth token for the Redis instance | |
 | API_BASE_URL **(required)** | Base URL for the backend API server for this service without any path | `http://localhost:3000` |
-| API_PATH **(required)** | Base base for the API | `/api/v1` |
+| API_PATH **(required)** | Base path for the API | `/api` |
+| API_VERSION **(required)** | API version to use | |
 | API_HEALTHCHECK_PATH **(required)** | Path to which healthcheck pings are sent | |
 | API_AUTH_PATH **(required)** | Path to which OAuth2 access token requests should be sent | |
 | API_CLIENT_ID **(required)** | Client ID used to authenticate with the backend API | |
--- a/common/lib/api-client/middleware/request.js
+++ b/common/lib/api-client/middleware/request.js
@@ -1,7 +1,7 @@
 const debug = require('debug')('app:api-client')
 const { get } = require('lodash')
-const uuid = require('uuid')
 
+const { API } = require('../../../../config')
 const redisStore = require('../../../../config/redis-store')
 const models = require('../models')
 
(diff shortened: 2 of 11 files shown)


config/index.js
```diff
@@ -1,6 +1,12 @@
 /* eslint no-process-env: "off" */
 require('dotenv').config()
 
+const API_VERSION = process.env.API_VERSION
+
+if (!API_VERSION) {
+  throw new Error('API_VERSION must be set')
```
teneightfive, line 7: Maybe add in that its the environment variable that needs setting so its clearer from the error message: `The `API_VERSION` environment variable must be set`?

README.md
```diff
@@ -214,7 +214,8 @@ The TZ (timezone) environment variable is set to 'Europe/London' in `start.js`.
 | REDIS_HOST **(required)** | Redis hostname. Can be used instead of `REDIS_URL`. Will override `REDIS_URL` if set | |
 | REDIS_AUTH_TOKEN | Optional auth token for the Redis instance | |
 | API_BASE_URL **(required)** | Base URL for the backend API server for this service without any path | `http://localhost:3000` |
-| API_PATH **(required)** | Base base for the API | `/api/v1` |
+| API_PATH **(required)** | Base path for the API | `/api` |
```
teneightfive, line 217: I realised looking at the code that we don't actually set this default. Should we remove it, or set the default in the config?

README.md
```diff
@@ -214,7 +214,8 @@ The TZ (timezone) environment variable is set to 'Europe/London' in `start.js`.
...
+| API_VERSION **(required)** | API version to use | |
```
teneightfive, line 218: I noticed that an error gets through if this isn't set but wondered if we should set a default? And then it actually got me to thinking whether we should support this being set by an environment variable at all...because if we set it to the wrong version, would that not cause the rest of the application to break because it is built against a specific version?

So should this actually just be more of a temporary thing, like when we've introduced feature flags, to help us migrate to the new version, and then once that is done, we specifically set the version in the config?

common/lib/api-client/middleware/request.js
```diff
@@ -24,18 +24,14 @@ function requestMiddleware({ cacheExpiry = 60, disableCache = false } = {}) {
       const pathname = new URL(req.url).pathname
 
       const searchString = new URLSearchParams(req.params).toString()
-      const key = `cache:${req.method}.${pathname}${
+      const key = `cache:v${API.VERSION}:${req.method}.${pathname}${
+        // const key = `cache:${req.method}.${pathname}${
```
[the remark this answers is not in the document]
solidgoldpig: yep

config/index.js
solidgoldpig: 👍

README.md
solidgoldpig: > So should this actually just be more of a temporary thing, like when we've introduced feature flags, to help us migrate to the new version, and then once that is done, we specifically set the version in the config?

Absolutely. This just gives us the flexibility to roll back instantly should the version change surface any unexpected issues.

PR #646, opened 2020-07-07
Upgrade dependencies - 2020-07-07

solidgoldpig:
### What changed

```
i18next                       19.5.3  →  19.5.4
conventional-github-releaser   3.1.4  →   3.1.5
eslint                         7.3.1  →   7.4.0
sass-loader                    9.0.0  →   9.0.2
testcafe                       1.8.7  →   1.8.8
webpack-merge                  4.2.2  →   5.0.8
```

### Environment variables

### Other considerations


common/parsers/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+const parsers = {
+  date(value) {
+    if (!value) {
+      return value
+    }
+
+    const enGBValue = value
+      .trim()
+      .replace(/^(\d{1,2})([/-])(\d{1,2})(([/-]\d{4}){0,1})$/, '$3$2$1$4')
```
teneightfive, line 11: Are these changes related to dep upgrades?

Also, my regex brain isn't great, what's the goal of this regex?

teneightfive: Thanks for the clarification in slack. As much as I hate them, I wonder if a small comment here might just clarify what we're doing here and why we've had to (because chrono no longer supports en_GB)?

solidgoldpig: Absolutely - just wanted to check this actually worked before testing, documenting and squashing up ;)

PR #648, opened 2020-07-07
[P4-1767] Update a response from the API

teneightfive:
### What changed

This change takes the values from the API and pre-fills the form step with those values.

### Why did it change

This allows a user to update an existing value for a question within the framework.

### Issue tracking

- [P4-1767](https://dsdmoj.atlassian.net/browse/P4-1767)

## Screenshots

<kbd></kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/person-escort-record/controllers/frameworks.js
+++ b/app/person-escort-record/controllers/frameworks.js
@@ -1,3 +1,5 @@
+const { kebabCase } = require('lodash')
+
 const FormWizardController = require('../../../common/controllers/form-wizard')
 const responseService = require('../../../common/services/framework-response')
 
@@ -17,6 +19,48 @@ class FrameworksController extends FormWizardController {
     req.form.options.buttonText = buttonText
 
     next()
+  }
+
+  reduceResponses(accumulator, { value, value_type: valueType, question }) {
+    const field = question.key
+
+    if (valueType === 'object') {
+      accumulator[field] = value.option
+      accumulator[`${field}--${kebabCase(value.option)}`] = value.details
+    }
+
+    if (valueType === 'collection') {
+      accumulator[field] = value.map(item => item.option)
+      value.forEach(item => {
+        accumulator[`${field}--${kebabCase(item.option)}`] = item.details
+      })
+    }
+
+    if (valueType === 'string' || valueType === 'array') {
+      accumulator[field] = value
+    }
+
+    return accumulator
+  }
+
+  getValues(req, res, callback) {
+    const fields = req.form.options.fields
+    const responses = req.personEscortRecord.responses
+    const savedValues = responses
+      .filter(response => fields[response.question.key])
+      .filter(response => response.value)
+      .reduce(this.reduceResponses, {})
+
+    super.getValues(req, res, (err, values) => {
+      if (err) {
+        return callback(err)
+      }
+
+      callback(null, {
+        ...savedValues,
+        ...values,
+      })
+    })
   }
 
   getResponses(formValues, allResponses) {
--- a/app/person-escort-record/controllers/frameworks.test.js
+++ b/app/person-escort-record/controllers/frameworks.test.js
@@ -75,6 +75,328 @@ describe('Person Escort Record controllers', function () {
       })
     })
 
+    describe('#reduceResponses', function () {
+      let response
+
+      context('with `object` value type', function () {
+        beforeEach(function () {
+          response = controller.reduceResponses(
+            {},
+            {
+              value: {
+                option: 'Yes, I do',
+                details: 'Lorem ipsum',
+              },
+              value_type: 'object',
+              question: {
+                key: 'object-response',
+              },
+            }
+          )
+        })
+
+        it('should set field for option', function () {
+          expect(response).to.have.property('object-response')
+          expect(response['object-response']).to.equal('Yes, I do')
+        })
+
+        it('should set field for details', function () {
+          expect(response).to.have.property('object-response--yes-i-do')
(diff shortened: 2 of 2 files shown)


app/person-escort-record/controllers/frameworks.js
```diff
@@ -19,6 +21,48 @@ class FrameworksController extends FormWizardController {
...
+        accumulator[`${field}--${kebabCase(item.option)}`] = item.details
+      })
+    }
+
+    if (valueType === 'string' || valueType === 'array') {
+      accumulator[field] = value
+    }
+
+    return accumulator
+  }
```
danielepolencic, line 44: _The following comment is based on personal opinions, so feel free to ignore it._

The intent of this code is to branch out to different values based on `valueType`.
A `switch` statement expresses the same intent but it's easier to scan with the eyes (you can't have expressions in `case`s, but you can in `if`s).

app/person-escort-record/controllers/frameworks.js
```diff
@@ -19,6 +21,48 @@ class FrameworksController extends FormWizardController {
...
+
+  getValues(req, res, callback) {
+    const fields = req.form.options.fields
+    const responses = req.personEscortRecord.responses
+    const savedValues = responses
+      .filter(response => fields[response.question.key])
+      .filter(response => response.value)
+      .reduce(this.reduceResponses, {})
```
danielepolencic, line 52: _This is also a very minor comment that can be easily ignored as the outcome doesn't change._

`reduceResponse` is just a pure function, there is no benefit on having it as a method on the class. You could have it as a single function and remove the need to `this.XXXX`. `this` in `reduce` statements is a bit tricky, and reading that lines sort of made me think... "does it uses class properties? why does it have to be a method on the class? is the scope relevant?"
A simple function would have not raised the same questions.

teneightfive: That's a valid point. So the reason for exposing it on the class was mainly to make it easier to test in isolation. That meant I could mock it when testing this function and test the functionality in that method in isolation.

If it's added as a private method I'd need to then test that functionality within `getValues` which I felt makes reading those tests harder and involves more boilerplate code in the tests.

Any thoughts on how I could test it in isolation? Maybe I could move it to `helpers`...?

app/person-escort-record/controllers/frameworks.js
teneightfive: Hmm, I tend to avoid `switch` statements as I find them harder to read. I think I went with this approach in a similar fashion of a return early pattern, but in this case there's no need to return early so maybe a switch might be easier to read...

```js
if (valueType === 'object') {
  accumulator[field] = value.option
  accumulator[`${field}--${kebabCase(value.option)}`] = value.details
}

if (valueType === 'collection') {
  accumulator[field] = value.map(item => item.option)
  value.forEach(item => {
    accumulator[`${field}--${kebabCase(item.option)}`] = item.details
  })
}

if (valueType === 'string' || valueType === 'array') {
  accumulator[field] = value
}
```

vs

```js
switch (valueType) {
  case 'object'
    accumulator[field] = value.option
    accumulator[`${field}--${kebabCase(value.option)}`] = value.details
    break
  case 'collection'
    accumulator[field] = value.map(item => item.option)
    value.forEach(item => {
      accumulator[`${field}--${kebabCase(item.option)}`] = item.details
    })
    break
  case 'string'
    accumulator[field] = value
    break
  case 'array'
    accumulator[field] = value
    break
}
```

The downside to the switch in this case is that we also need separate statements for `string` and `array` which set the same value on the accumulator so results in a bit of duplication...

@solidgoldpig @danielepolencic what are your thoughts after seeing them side-by-side?

solidgoldpig: Personally I prefer... object literals

```js
const accumulatorMethods = {
  object: (field, value) => {
    accumulator[field] = value.option
    accumulator[`${field}--${kebabCase(value.option)}`] = value.details
  },
  collection: (field, value) => {
    accumulator[field] = value.map(item => item.option)
    value.forEach(item => {
      accumulator[`${field}--${kebabCase(item.option)}`] = item.details
    })
  },
  string: (field, value) => {
    accumulator[field] = value
  },
  array: (field, value) => {
    accumulator[field] = value
  },
}

accumulatorMethods[valueType]
```

/ducks

Normally I would actually prefer something like that, but this seems overkill here. Unless we want to actually extract and test each of those accumulator bits ;)

Maybe it's just because I drank the Crockford kool-aid, but I don't like switch statements. I don't find the replacement easier to read - but maybe that's just because I'm not used to it.

OTOH, maybe we could do with adding some `elses` to the if statements ¯\_(ツ)_/¯

teneightfive: >Maybe it's just because I drank the Crockford kool-aid

😂 

I think I'll leave it as it is for now for speed and ease, unless there are any major hesitations?

app/person-escort-record/controllers/frameworks.js
teneightfive: I started to have a look at this but then it differs from what we've done elsewhere at the moment. I agree it would be nicer to not use a method on the class where we don't need access to the class properties but I think to avoid wasting too much time for now I'll leave it as it is.

Also, I'm aware I will need to extend this shortly once the endpoint is there to ensure values are processed correctly and sent in the same format so I will look to re-work it as part of that work.

PR #653, opened 2020-07-09
[P4-1910] Use profile for assessment answers

solidgoldpig:
### What changed

Amends create and update controllers to use profile rather than person for getting the previous assessment answers

### Why did it change

Required to enable transition from v1 to v2 of the API

### Issue tracking

- [P4-1910](https://dsdmoj.atlassian.net/browse/P4-1910)

## Screenshots

Functionally identical

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/create/assessment.js
+++ b/app/move/controllers/create/assessment.js
@@ -2,6 +2,7 @@ const { flatten, get, values } = require('lodash')
 
 const fieldHelpers = require('../../../../common/helpers/field')
 const presenters = require('../../../../common/presenters')
+const profileService = require('../../../../common/services/profile')
 const referenceDataService = require('../../../../common/services/reference-data')
 
 const CreateBaseController = require('./base')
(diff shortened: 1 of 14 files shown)


app/move/controllers/create/assessment.js
```diff
@@ -31,19 +32,26 @@ class AssessmentController extends CreateBaseController {
...
       customAssessmentGroupings = [],
       showPreviousAssessment,
     } = req.form.options
 
-    if (showPreviousAssessment) {
+    let profile = req.getProfile()
+
+    if (!profile.id) {
       const person = req.getPerson()
+      profile = await profileService.create(person.id, {})
```
teneightfive, line 46: I think we should create it at the same point at which we create a person. That way it's the safest point in time. As soon as we know we have a person, we create the profile for this move.

That would save you having to check for the id and prevent the issue I'm sure you saw where it would create a new one for each assessment step...

teneightfive: Creating it at the point of having a person also ensures that if we use anything else on a profile in the future it will exist rather than being tied to only these steps.

For example, there are also details on the move that only make sense when there is a _person_ so in future these could be moved to the profile to make it cleaner to remove a person and their specific information for that move from the move. If we create the profile sooner, it would ensure that wouldn't catch us out again in future.

common/services/profile/profile.unformat.js
```diff
@@ -0,0 +1,59 @@
+const getAnswer = (profile, field) => {
```
teneightfive, line 1: I think it would be nice to try and stick to either a folder or a file rather than maintaining both. I find having both creates extra confusion as there are now two places to look.

We did it for person with the vision to change it which I understand with other pressures we never got round to. So I'd rather not do the same if we can avoid it...

app/move/controllers/create/assessment.js
solidgoldpig: But, that's exactly why I put it here. We have journeys that create a person and journeys that don't. That's why I think this is the safest point to create the profile.

Would you prefer to create the profile?

a) on the move details page
b) i) when the person is selected from the results / ii) when the person is created on the person-details page

common/services/profile/profile.unformat.js
solidgoldpig: I was doing it like person precisely for that reason!

PR #665, opened 2020-07-15
Person Escort Record corrections

teneightfive:
### What changed

This PR contains a few tweaks that needed to be made once testing against the actual API endpoints.

In summary:

- update models
- improve framework response handling
- fix references to section progress

### Why did it change

The API endpoints are now more stable and some things in the frontend code required tweaking to
make the journeys work properly.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/framework-step-to-summary.js
+++ b/common/presenters/framework-step-to-summary.js
@@ -1,15 +1,26 @@
-const { filter, flatten } = require('lodash')
+const { filter, find, flatten, isArray } = require('lodash')
 
 const frameworksHelpers = require('../helpers/frameworks')
 
 const frameworkFieldToSummaryListRow = require('./framework-field-summary-list-row')
 
 function frameworkStepToSummary(allFields, responses, baseUrl) {
-  return ([key, step]) => {
+  return ([key, step], i, allSteps = []) => {
+    const allNextSteps = allSteps
+      .map(([key, step]) => step.next)
+      .filter(next => isArray(next))
+      .reduce((acc, it) => [...it, ...acc], [])
+    const unmetStepConditions = allNextSteps
+      .filter(condition => condition.next === step.slug)
+      .filter(item => {
+        const response = find(responses, ['question.key', item.field]) || {}
+        return item.value !== response.value
+      })
+
     const stepUrl = `${baseUrl}/${step.slug}`
     const stepFields = step.fields
 
-    if (!stepFields.length) {
+    if (stepFields.length === 0 || unmetStepConditions.length > 0) {
       return
     }
 
--- a/app/person-escort-record/controllers/framework-overview.js
+++ b/app/person-escort-record/controllers/framework-overview.js
@@ -1,11 +1,14 @@
 const presenters = require('../../../common/presenters')
 
 function frameworkOverview(req, res) {
-  const { originalUrl, framework, personEscortRecord } = req
-  const fullname = personEscortRecord?.profile?.person?.fullname
-  const taskList = presenters.frameworkToTaskListComponent(`${originalUrl}/`)(
-    framework
-  )
+  const { originalUrl, framework, personEscortRecord = {} } = req
+  const { meta, profile } = personEscortRecord
+  const fullname = profile?.person?.fullname
+  const taskList = presenters.frameworkToTaskListComponent({
+    baseUrl: `${originalUrl}/`,
+    frameworkSections: framework.sections,
+    sectionProgress: meta?.section_progress,
+  })
 
   res.render('person-escort-record/views/framework-overview', {
     taskList,
--- a/app/person-escort-record/controllers/framework-overview.test.js
+++ b/app/person-escort-record/controllers/framework-overview.test.js
@@ -12,13 +12,12 @@ const mockFramework = {
 
 describe('Person Escort Record controllers', function () {
   describe('#frameworkOverview()', function () {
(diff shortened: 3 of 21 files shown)


common/presenters/framework-step-to-summary.js
```diff
@@ -1,15 +1,30 @@
...
 
 const frameworksHelpers = require('../helpers/frameworks')
 
 const frameworkFieldToSummaryListRow = require('./framework-field-summary-list-row')
 
 function frameworkStepToSummary(allFields, responses, baseUrl) {
-  return ([key, step]) => {
+  return ([key, step], i, allSteps = []) => {
+    const unmetStepConditions = allSteps
+      .map(step => step[1].next)
```
danielepolencic, line 10: I'm not sure if this is still possible (or relevant to this PR), but having an object instead of a tuple is more readable.
It took me a while to realise that `[1]` is the actual object that belongs to the step.

I think step should be `{url: string, step: { next: ..., slug: string, fields, string[]}}` instead of `[string, object]`

common/presenters/framework-step-to-summary.js
```diff
@@ -1,15 +1,30 @@
...
+        const conditions = next
+          .filter(condition => isObject(condition))
+          .filter(condition => condition.next === step.slug)
+          .filter(item => {
+            const response = find(responses, ['question.key', item.field]) || {}
+            return item.value !== response.value
+          })
+
+        return conditions.length
+      })
```
danielepolencic, line 22: I think breaking this up could help a bit with understanding how it works.

1. `const allNext = allSteps.map(step => step[1].next).filter(next => isArray(next))`. It is the same for all steps.
1. `isArray(next) === isObect(condition)` since `isObject([]) === true`. This can be removed.
1. You're interested only in the set that is relevant to the current step. `const relevantNext = allNext.filter(/* */)`
1. Find the answers

You could have a function that retrieves all previousSteps

common/presenters/framework-step-to-summary.js
```diff
@@ -1,15 +1,30 @@
...
+
     const stepUrl = `${baseUrl}/${step.slug}`
     const stepFields = step.fields
 
-    if (!stepFields.length) {
+    if (!stepFields.length || unmetStepConditions.length) {
```
danielepolencic, line 27: `.length` is a number, but you're using it as a truthy/falsy value. I'd rather have `stepFields.length !== 0 || unmetStepConditions.length > 0`

common/presenters/framework-step-to-summary.js
teneightfive: I struggled a lot to get this neat too. Essentially what needs to happen is to remove any steps that are dependent on values within the responses. The only way to get it at this stage is to loop through `allSteps` which is all items in the map and search through the objects to find `next`. The tricky thing is that next can be a number of things too but for now we're only catering for basic conditional filtering.

I could convert it to an object but then you can't use native `map` and it would mean bringing in more lodash deps...I wonder if in that case this would make it more readable?

```javascript
.map(([key, step]) => step.next)
```

common/presenters/framework-step-to-summary.js
teneightfive: I'd love if you had time to post a snippet. I toyed with this a lot to get it to return what it actually needs to.

I think 2. isn't possible because they are checks on separate objects. When is check that `next` is in fact an array, because in most cases it's just a string. And then `isObject` is checking that the condition is an object and therefore is dependent on a responses value.

Here's an example of what nexts can look like from the test file:
```
[
  '/step-one',
  {
    next: [
      {
        field: 'conditionalField',
        value: 'Yes',
        next: 'step-two',
      },
      'step-three',
    ],
    slug: 'step-one',
    fields: ['fieldOne'],
  },
  ],
  [
    '/step-two',
    {
      next: 'step-three',
      slug: 'step-two',
      fields: ['fieldOne'],
    },
  ],
  [
    '/step-three',
    {
      slug: 'step-three',
      fields: ['fieldOne'],
    },
  ],
]
```

`next` supports the same logic from the [form wizard](https://github.com/UKHomeOffice/passports-form-wizard#next-steps). So the function of this is to remove steps from the display if the condition is not met.

common/presenters/framework-step-to-summary.js
teneightfive: Late night coding fail...

common/presenters/framework-step-to-summary.js
danielepolencic: > I could convert it to an object but then you can't use native map

You can always `Object.keys(obj).map(key => {})`?
But I agree, if the change is touching over components, it's probably wiser to defer it.

common/presenters/framework-step-to-summary.js
danielepolencic: This snippet passes the current tests:

```js
    const allNextSteps = allSteps
      .map(([key, step]) => step.next)
      .filter(next => isArray(next))
      .reduce((acc, it) => [...it, ...acc], [])

    const unmetStepConditions = allNextSteps
      .filter(condition => condition.next === step.slug)
      .filter(item => {
        const response = find(responses, ['question.key', item.field]) || {}
        return item.value !== response.value
      })
```

PR #670, opened 2020-07-16
[P4-1594] Person escort record start journey

teneightfive:
### What changed

This includes some updates to the existing draft journey built for create a new Person Escort Record. In summary:

- Removes before you start page from PER creation
- Brings section progress into move detail page
- Directs user to first step in a section from task list
- Sends user to overview at end of a section
- Add support for nested URL

URLs now supported for creating/viewing/editing a PER:

Nested move URL structure
```
/move/:moveId/person-escort-record/new
/move/:moveId/person-escort-record/
/move/:moveId/person-escort-record/:section
/move/:moveId/person-escort-record/:section/:step
```

Singular PER structure
```
/person-escort-record/:perId/
/person-escort-record/:perId/:section
/person-escort-record/:perId/:section/:step
```

**Note:** Creation is currently only supported from within a move as it requires a profile to already exist on a move

**[DEMO APP](https://booksecuremove-pr-670.herokuapp.com/)** (Sign in as Police User)

### Why did it change

This is to bring the draft more inline with the prototype but also trying to keep things a bit separated in terms of code structure and to keep inline with existing patterns already used within the service.

### Issue tracking

- [P4-1594](https://dsdmoj.atlassian.net/browse/P4-1594)
- [P4-1726](https://dsdmoj.atlassian.net/browse/P4-1726)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/person-escort-record/controllers/framework-overview.js
```diff
@@ -1,9 +1,10 @@
 const presenters = require('../../../common/presenters')
 
 function frameworkOverview(req, res) {
-  const { originalUrl, framework, personEscortRecord = {} } = req
-  const { meta, profile } = personEscortRecord
-  const fullname = profile?.person?.fullname
+  const { originalUrl, framework, personEscortRecord = {}, move = {} } = req
+  const { profile: moveProfile } = move
+  const { meta, profile: perProfile } = personEscortRecord
+  const fullname = perProfile?.person?.fullname || moveProfile?.person?.fullname
```
solidgoldpig, line 7: Why not
```js
const profile = perProfile || moveProfile
const fullname = profile.person.fullname
```
Why would there not a profile on the move?

app/person-escort-record/index.js
```diff
@@ -1,31 +1,31 @@
 // NPM dependencies
-const router = require('express').Router()
+const router = require('express').Router({ mergeParams: true })
```
solidgoldpig, line 2: Gah! I can't believe I didn't know this option existed. I think this means some rubbish code in the move updates can be removed

app/person-escort-record/views/framework-overview.njk
```diff
@@ -4,7 +4,9 @@
   <header class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <span class="govuk-caption-xl">
-        {{ t("person-escort-record::heading") }}
+        {{ t("person-escort-record::heading", {
+          context: "with_name"
+        }) }}
       </span>
       <h1 class="govuk-heading-xl">
```
solidgoldpig, line 11: Is this empty if there's no name?

(I see we still have the spans outside the h1 element...)

app/person-escort-record/middleware.js
```diff
@@ -14,12 +14,16 @@ function setFrameworkSection(frameworkSection) {
   }
 }
 
-async function setPersonEscortRecord(req, res, next, recordId) {
+async function setPersonEscortRecord(req, res, next) {
+  const recordId =
+    req.params?.personEscortRecordId || req?.personEscortRecord?.id
```
solidgoldpig, line 19: Not following this logic. The only place I can see where do we get `req.personEscortRecord` is later in this method

app/move/index.js
```diff
@@ -1,10 +1,14 @@
 // NPM dependencies
-const router = require('express').Router()
+const express = require('express')
+const router = express.Router()
+const moveRouter = express.Router()
```
solidgoldpig, line 4: Is it an idea to, er, move the moveRouter in to a separate module?

app/person-escort-record/controllers/framework-overview.js
teneightfive: Yes, that does read a little neater.

app/person-escort-record/index.js
teneightfive: Haha, yeah I remember realising it was a big game-changer when I discovered it ;)

app/person-escort-record/controllers/framework-overview.js
solidgoldpig: sorry, "Why would there not a profile on the move?" was a separate question rather than  a clarification of the code suggestion ;)

app/person-escort-record/views/framework-overview.njk
teneightfive: No, this is just setting the context for this use. Essentially it adds a `for` to the key in this use case.

The default without this context is `Person Escort Record` and with this context it uses `Person Escort Record for`

app/person-escort-record/middleware.js
teneightfive: This allows us to support the app both nested and on its own. So if the ID is set in the request params it will use that ID to get the record, otherwise it will look to see if an ID has been set within another middleware (currently we do this in the move app so that the router can be nested a whole).

At the moment it's not hugely efficient as we need to make another call to get the PER with all the responses. I didn't want to add the responses to the default `include` of `move` as it would've meant asking for them for each item in the GET all request we make as we haven't split our GET one vs GET all calls yet.

app/move/index.js
teneightfive: Yes...I almost went down that rabbit hole but tried to stop myself making this PR about too many different things.

I think it might be worth looking at the approach I've taken with nested apps within the `app/person-escort-record/app/` folder to see how we could make this area a bit better to maintain in the long run.

app/person-escort-record/controllers/framework-overview.js
teneightfive: It's less about the move and more about if its using `/person-escort-record/:recordId` instead of `/move/:moveId/person-escort-record`.

In the former there is no `req.move`, but instead `req.personEscortRecord`.

teneightfive: This essentially allows us to access the PER on its own or as part of a move.

PR #679, opened 2020-07-18
Use consistent error message pattern

teneightfive:
### What changed

- Update special vehicle error step to be more consistent.
- Update content to remove negative contractions.

### Why did it change

This brings this validation error message inline with the errors that are displayed
when a user attempts to create a move/allocation conflict so that we use a consistent pattern.

It also updates the language of the message to avoid using negative contractions inline
with the [GOV.UK Style guide advice](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#contractions).

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/components/message/_message.scss
+++ b/common/components/message/_message.scss
@@ -22,6 +22,10 @@ $_message-border-width: 5px;
     margin-top: 0;
   }
 
+  p {
+    @include app-reading-width;
+  }
+
   * + * {
     margin-top: govuk-spacing(1);
   }
@@ -87,6 +91,10 @@ $_message-border-width: 5px;
 
 .app-message--error {
   border-color: $govuk-error-colour;
+
+  p + p {
+    margin-top: govuk-spacing(2);
+  }
 }
 
 .app-message--muted {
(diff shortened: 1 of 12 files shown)


common/components/message/_message.scss
```diff
@@ -22,6 +22,10 @@ $_message-border-width: 5px;
     margin-top: 0;
   }
 
+  p {
```
solidgoldpig, line 25: Is there not a govuk mixin for this? Why 600px?

teneightfive: So that was to set the [optimal reading width](https://www.freshconsulting.com/uiux-principle-46-text-box-width-should-help-users-read/#:~:text=700%20pixels%20wide%20is%20a,(2.33%20inches)%20in%20width.) as the text shouldn't really stretch the whole width of the page.

I'll have a check see if there's something in GOV.UK Frontend but I couldn't see anything...

teneightfive: I'll extract it to something more useful in our app code :)

PR #684, opened 2020-07-21
Ensure dependent framework responses are filtered correctly

teneightfive:
### What changed

This change ensures that any dependent responses are not sent to be
saved if the dependent field doesn't match the condition the field
depends on.

This also includes a fix to remove any falsy values from Array like
objects which prevents the API from returning an error when an
array with one empty string is sent when no checkboxes are selected.

### Why did it change

Currently the API would return an error when the frontend would send values for responses that
aren't valid because the condition for a dependent question is not met.

This change ensures that we remove any dependent fields before sending to the API.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/helpers/field/is-allowed-dependent.js
+++ b/common/helpers/field/is-allowed-dependent.js
@@ -1,0 +1,40 @@
+const { intersection } = require('lodash')
+
+function isAllowedDependent(fields, key, values) {
+  const field = fields[key]
+
+  if (!field) {
+    return false
+  }
+
+  if (!field.dependent) {
+    return true
+  }
+
+  // set up dependent to an object of field and value
+  if (typeof field.dependent === 'string') {
+    field.dependent = {
+      field: field.dependent,
+      value: true,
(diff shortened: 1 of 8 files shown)


common/helpers/field/is-allowed-dependent.js
```diff
@@ -0,0 +1,40 @@
...
+
+  // set up dependent to an object of field and value
+  if (typeof field.dependent === 'string') {
+    field.dependent = {
+      field: field.dependent,
+      value: true,
+    }
+  }
+
+  if (!field.dependent) {
```
solidgoldpig, line 18: Couldn't this check be done before the string to object coercion

teneightfive: Yep, I'll move it and make sure the tests still pass.

PR #685, opened 2020-07-21
chore: Support v1 and v2 person differences

solidgoldpig:
Implemnents P4-1911

### What changed

- Flatten identifier properties in model
- Update personService transform, format and unformat methods to match
- Update personToSummaryList presenter to use flattened props

### Why did it change

Enables move from v1 -> v2 of API

### Issue tracking

- [P4-1911](https://dsdmoj.atlassian.net/browse/P4-1911)

## Screenshots

Functionally identical

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/person-to-summary-list-component.test.js
+++ b/common/presenters/person-to-summary-list-component.test.js
@@ -1,3 +1,4 @@
+const { API } = require('../../config')
 const i18n = require('../../config/i18n')
 const filters = require('../../config/nunjucks/filters')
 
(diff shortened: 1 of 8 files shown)


common/presenters/person-to-summary-list-component.test.js
```diff
@@ -1,3 +1,4 @@
+const { API } = require('../../config')
```
teneightfive, line 1: Should this be mocked so the two branches can be tested in isolation?

solidgoldpig: I don't think it's worth the effort. It should only run as one or the other and the forking stuff should be yanked out the moment we switch over and are happy everything's ok. Unless you have a way of running the e2es with the VERSION mocked too ;)

PR #686, opened 2020-07-22
[P4-1784] Render flags from the Person Escort Record instead of the move request

teneightfive:
### What changed

This set of changes amend the behaviour for display tags for a move. Tags will now only be shown when
all sections of the Person Escort Record are completed.

If they have not been completed or a PER has not yet been started it will display a message in place of the
tags showing what needs to be done.

### Why did it change

The Person Escort Record is a new feature that allows risk and health information to be collected in more
detail. With this new addition the previous behaviour for rendering flags no longer fits or is appropriate.

This new behaviour avoids the need for any complicated logic to combine the tags and instead will only
display if we know all the information we need.

### Issue tracking

- [P4-1784](https://dsdmoj.atlassian.net/browse/P4-1784)

### Dashboard

This shows the two variations for a complete PER and an incomplete PER

<kbd></kbd>

### Move detail (incomplete Person Escort Record)

This version display when no Person Escort Record has been started or any section is not complete.

<kbd></kbd>

### Move detail (complete Person Escort Record)

This version displays when all sections of the Person Escort Record have been completed.

<kbd></kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/framework-flags-to-tag-list.js
+++ b/common/presenters/framework-flags-to-tag-list.js
@@ -1,0 +1,20 @@
+const { sortBy, uniqBy, kebabCase } = require('lodash')
+
+const { FRAMEWORKS } = require('../../config')
+
+function frameworkFlagsToTagList(flags = [], hrefPrefix = '') {
+  const tags = uniqBy(flags, 'title').map(({ flag_type: category, title }) => {
+    const settings = FRAMEWORKS.FLAG_SETTINGS[category] || {}
+
+    return {
+      text: title,
+      href: `${hrefPrefix}#${kebabCase(title)}`,
+      classes: settings.tagClass || '',
+      sortOrder: settings.sortOrder || null,
+    }
+  })
+
+  return sortBy(tags, ['sortOrder', 'text'])
+}
+
+module.exports = frameworkFlagsToTagList
--- a/app/move/controllers/view.js
+++ b/app/move/controllers/view.js
@@ -1,4 +1,4 @@
-const { get, sortBy } = require('lodash')
+const { isEmpty, get, sortBy } = require('lodash')
 
 const presenters = require('../../../common/presenters')
 const frameworksService = require('../../../common/services/frameworks')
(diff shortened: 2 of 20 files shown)


common/presenters/framework-flags-to-tag-list.js
```diff
@@ -0,0 +1,20 @@
+const { sortBy, uniqBy, kebabCase } = require('lodash')
+
+const { FRAMEWORKS } = require('../../config')
+
+function frameworkFlagsToTagList(flags = [], hrefPrefix = '') {
+  const tags = uniqBy(flags, 'title').map(({ flag_type: category, title }) => {
+    const settings = FRAMEWORKS.FLAG_SETTINGS[category] || {}
```
danielepolencic, line 7: _DISCLAIMER: opinion-based comment_.

The function is not pure and it's fairly hard to test (as highlighted by the use of `proxyrequire`).

The function can be easily refactored to be pure by passing the settings as an extra argument or as a high order function.

Example:

```js
function createFrameworkFlagsToTagList(settings = {}) {
  return (flags = [], hrefPrefix = "") => {
    return {
      text: title,
      href: `${hrefPrefix}#${kebabCase(title)}`,
      classes: settings.tagClass || '',
      sortOrder: settings.sortOrder || null,
    }
  }
}
```

Testing the function it's easy (no need for `proxyrequire`) and the same input produces the same output.

The consequences of this change are more complex, though.
You will have to pass down the configuration instead of requiring it locally.

Both options require trade-offs, however, using pure functions means less magic and easier testing.

teneightfive: @danielepolencic Yep, I do think that is a much nicer approach. Thanks for just talking through with me too. However, not wanting to derail this change too much as I think the ideal approach would mean refactoring both the presenter and controllers I'm going to write a separate issue if you're happy with that?

teneightfive: Added an issue: #696

PR #690, opened 2020-07-22
build: Wait for master to be deployed before running e2e tests

solidgoldpig:
### What changed

On CircleCI E2E tests run against master branch poll ping endpoint to check same commit has been deployed before running the tests

Update ping endpoint to return build details to enable checking of commit.

### Why did it change

Tests against master on staging were kicked off the moment that the staging deployment task had finished.

This meant that:

a) tests early in the run may have been run against the old pods running the previous deployed commit of the app.
b) tests that ran against the wrong pods were in danger of having requests dropped if they had not been fulfilled by the time the pods were terminated which would cause the test to fail.

This should prevent tests being run against the wrong pods and having the pods switched out from under them.

### Issue tracking

n/a

## Screenshots

n/a

### Other considerations

### DIFF
--- a/app/healthcheck/controllers.js
+++ b/app/healthcheck/controllers.js
@@ -1,7 +1,23 @@
 const { some } = require('lodash')
 
-const { BUILD_DATE, BUILD_BRANCH, GIT_SHA } = require('../../config')
+const {
+  API,
+  APP_BUILD_DATE,
+  APP_BUILD_TAG,
+  APP_BUILD_BRANCH,
+  APP_GIT_SHA,
+} = require('../../config')
 const { version } = require('../../package.json')
+
+// https://github.com/ministryofjustice/ping.json
+const buildDetails = {
+  build_date: APP_BUILD_DATE,
+  build_tag: APP_BUILD_TAG,
+  commit_id: APP_GIT_SHA,
+  version_number: version,
+  branch: APP_BUILD_BRANCH,
+  api_version: API.VERSION,
+}
 
 module.exports = {
   get: (req, res) => {
@@ -12,15 +28,12 @@ module.exports = {
     res.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate')
     res.status(errors ? 503 : 200).json({
       status,
-      version,
       dependencies,
-      buildDate: BUILD_DATE,
-      buildTag: BUILD_BRANCH,
-      gitSha: GIT_SHA,
+      ...buildDetails,
     })
   },
   ping: (req, res) => {
     res.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate')
-    res.status(200).send('OK')
+    res.status(200).json(buildDetails)
   },
 }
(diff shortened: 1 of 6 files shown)


config/index.js
```diff
@@ -56,9 +56,10 @@ module.exports = {
   NO_CACHE: process.env.CACHE_ASSETS ? false : IS_DEV,
   FEEDBACK_URL: process.env.FEEDBACK_URL,
   SUPPORT_EMAIL: process.env.SUPPORT_EMAIL,
-  BUILD_DATE: process.env.APP_BUILD_DATE,
-  BUILD_BRANCH: process.env.APP_BUILD_TAG,
-  GIT_SHA: process.env.APP_GIT_COMMIT,
+  APP_BUILD_DATE: process.env.APP_BUILD_DATE,
+  APP_BUILD_BRANCH: process.env.APP_BUILD_BRANCH,
+  APP_BUILD_TAG: process.env.APP_BUILD_TAG,
+  APP_GIT_SHA: process.env.APP_GIT_COMMIT,
```
danielepolencic, line 62: I think this is bending the rules a bit.
According to the code above, I could change the GIT SHA, the build tag or the build date **at run time**.
I'm not sure if we practice immutable builds, but this going against it.
All of these properties are properties of the container and should be packaged (statically) with the app. You should not be able to change at runtime.
CC @mafonso

app/healthcheck/controllers.js
```diff
@@ -1,8 +1,24 @@
...
+const {
+  API,
+  APP_BUILD_DATE,
+  APP_BUILD_TAG,
+  APP_BUILD_BRANCH,
+  APP_GIT_SHA,
+} = require('../../config')
 const { version } = require('../../package.json')
 
+// https://github.com/ministryofjustice/ping.json
```
teneightfive, line 12: 👍

config/index.js
solidgoldpig: This wasn't introduced in this PR - can we raise an issue for it?

mafonso: Yeah, I guess we could render them into a file from the ARGs into a file, and then consume that file, as a matter of principle. 
But if we agree not to override them in runtime, it has the same effect in a simpler way. And doing it does not have a tremendous side effect. It might actually be useful in test environments.

test/e2e/parallel-runner.js
```diff
@@ -189,12 +196,45 @@ if (args.n) {
...
+          checkPing = false
+          process.stdout.write(`${E2E_BASE_URL} returned correct commit sha`)
+        }
+      } catch (error) {
+        process.stdout.write('.')
+        await sleep(2000)
+      }
+    }
+  }
+}
```
mafonso, line 225: I would probably put this in the circleCI workflow and outside the test.
After deployment wait until the new sha is confirmed to be deployed or fail the deploy.

Feels like the test could assert that the version is correct but should not have to wait until it is.

config/index.js
solidgoldpig: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/692

Issue #428, opened 2020-04-17
Use `time` element wherever possible


teneightfive:
## Problem

We are using dates and times in various places through the application. Currently in some instances we have tried to use the [`time`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) element but the use is inconsistent.

## Solution

Find places where we are displaying dates or times in the markup and ensure they are appropriately wrapped using the `time` element.

PR #691, opened 2020-07-23
[P4-1323] Displayed the current week in the dashboard

danielepolencic:
## Proposed changes

Amend the dashboard to include the current week

| Before | After |
| ------ | ----- |
| image | image |
| image | image |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### DIFF
--- a/app/home/controllers.js
+++ b/app/home/controllers.js
@@ -1,5 +1,6 @@
 const { get } = require('lodash')
 
+const dateHelpers = require('../../common/helpers/date')
 const permissions = require('../../common/middleware/permissions')
 
 function dashboard(req, res) {
@@ -11,32 +12,41 @@ function dashboard(req, res) {
     return res.redirect('/moves')
   }
 
+  const dateRange = dateHelpers.getCurrentWeekAsRange()
+  const today = new Date().toISOString()
+
   const sections = {
     outgoing: {
       permission: 'moves:view:outgoing',
       heading: 'dashboard::sections.outgoing.heading',
       filter: req.filterOutgoing,
+      period: 'day',
     },
     incoming: {
       permission: 'moves:view:incoming',
       heading: 'dashboard::sections.incoming.heading',
       filter: req.filterIncoming,
+      period: 'day',
     },
     singleRequests: {
       permission: 'moves:view:proposed',
       heading: 'dashboard::sections.single_requests.heading',
       filter: req.filterSingleRequests,
+      period: 'week',
     },
     allocations: {
       permission: 'allocations:view',
       heading: 'dashboard::sections.allocations.heading',
       filter: req.filterAllocations,
+      period: 'week',
     },
   }
 
   res.render('home/dashboard', {
     pageTitle: 'dashboard::page_title',
     sections,
+    dateRange,
+    today,
   })
 }
 
--- a/app/home/controllers.js
+++ b/app/home/controllers.js
@@ -12,7 +12,7 @@ function dashboard(req, res) {
     return res.redirect('/moves')
   }
 
-  const dateRange = dateHelpers.getCurrentWeekAsRange()
+  const currentWeek = dateHelpers.getCurrentWeekAsRange()
   const today = new Date().toISOString()
 
   const sections = {
@@ -45,7 +45,7 @@ function dashboard(req, res) {
   res.render('home/dashboard', {
     pageTitle: 'dashboard::page_title',
     sections,
-    dateRange,
+    currentWeek,
     today,
   })
 }
--- a/app/home/controllers.test.js
+++ b/app/home/controllers.test.js
@@ -43,30 +43,27 @@ describe('Home controllers', function () {
 
       it('should return locations sorted by title', function () {
         const params = res.render.args[0][1]
-        expect(params).to.deep.equal({
-          pageTitle: 'dashboard::page_title',
-          sections: {
-            outgoing: {
-              permission: 'moves:view:outgoing',
-              heading: 'dashboard::sections.outgoing.heading',
-              filter: req.filterOutgoing,
-            },
-            incoming: {
-              permission: 'moves:view:incoming',
-              heading: 'dashboard::sections.incoming.heading',
-              filter: req.filterIncoming,
-            },
-            singleRequests: {
-              permission: 'moves:view:proposed',
-              heading: 'dashboard::sections.single_requests.heading',
-              filter: req.filterSingleRequests,
-            },
-            allocations: {
-              permission: 'allocations:view',
-              heading: 'dashboard::sections.allocations.heading',
-              filter: req.filterAllocations,
-            },
-          },
+        const sections = [
+          'outgoing',
+          'incoming',
+          'singleRequests',
+          'allocations',
+        ]
+        expect(params).to.have.all.keys(
+          'pageTitle',
+          'sections',
+          'dateRange',
+          'today'
+        )
+        expect(params.pageTitle).to.equal('dashboard::page_title')
+        expect(params.sections).to.have.all.keys(sections)
+        sections.forEach(section => {
+          const fields = ['permission', 'heading', 'filter', 'period']
+          expect(params.sections[section]).to.have.all.keys(fields)
+          expect(params.sections[section].permission).to.be.a('string')
+          expect(params.sections[section].heading).to.be.a('string')
+          expect(params.sections[section].filter).to.be.an('array')
+          expect(params.sections[section].period).to.be.a('string')
         })
       })
     })
--- a/app/home/controllers.test.js
+++ b/app/home/controllers.test.js
@@ -52,7 +52,7 @@ describe('Home controllers', function () {
         expect(params).to.have.all.keys(
           'pageTitle',
           'sections',
-          'dateRange',
+          'currentWeek',
           'today'
         )
         expect(params.pageTitle).to.equal('dashboard::page_title')
--- a/app/home/dashboard.njk
+++ b/app/home/dashboard.njk
@@ -20,19 +20,9 @@
 
             <span class="govuk-!-display-block">
             {% if section.period === 'week' %}
-              {{ dateRange | formatDateRange }}
-              {{ appTime({
-                datetime: dateRange | formatISOWeek,
-                text: '(' + dateRange | formatDateRangeAsRelativeWeek | lower + ')',
-                classes: 'app-secondary-text-colour'
-                }) }}
+              {{ currentWeek | formatDateRangeWithRelativeWeek }}
             {% else %}
-              {{ today | formatDateWithDay }}
-              {{ appTime({
-                datetime: today,
-                text: '(' + today | formatDateAsRelativeDay | lower + ')',
-                classes: 'app-secondary-text-colour'
-                }) }}
+              {{ today | formatDateWithRelativeDay }}
             {% endif %}
             </span>
           </h2>
--- a/app/home/dashboard.njk
+++ b/app/home/dashboard.njk
@@ -15,8 +15,26 @@
     {% for key, section in sections %}
       {% if canAccess(section.permission) %}
         <section class="govuk-grid-column-one-half govuk-!-margin-bottom-6">
-          <h2 class="govuk-heading-m govuk-!-margin-bottom-3">
+          <h2 class="govuk-heading-m govuk-!-margin-bottom-1">
             {{ t(section.heading) }}
+
+            <span class="govuk-!-display-block">
+            {% if section.period === 'week' %}
+              {{ dateRange | formatDateRange }}
+              {{ appTime({
+                datetime: dateRange | formatISOWeek,
+                text: '(' + dateRange | formatDateRangeAsRelativeWeek | lower + ')',
+                classes: 'app-secondary-text-colour'
+                }) }}
+            {% else %}
+              {{ today | formatDateWithDay }}
+              {{ appTime({
+                datetime: today,
+                text: '(' + today | formatDateAsRelativeDay | lower + ')',
+                classes: 'app-secondary-text-colour'
+                }) }}
+            {% endif %}
+            </span>
           </h2>
 
           {{ appFilter({
--- a/common/templates/layouts/collection.njk
+++ b/common/templates/layouts/collection.njk
@@ -6,18 +6,18 @@
   }) }}'});
 {% endblock %}
 
+{% if period === 'week' %}
+  {% set displayDate = dateRange | formatDateRangeWithRelativeWeek %}
+{% else %}
+  {% set datetime = dateRange | first %}
+  {% set displayDate = datetime | formatDateWithRelativeDay %}
+{% endif %}
+
 {% block pageTitle %}
-  {% if period === 'week' %}
-    {{ t("collections::page_title_with_date", {
-      context: context,
-      date: dateRange | formatDateRangeAsRelativeWeek
-    }) }}
-  {% else %}
-    {{ t("collections::page_title_with_date", {
-      context: context,
-      date: datetime | formatDateAsRelativeDay
-    }) }}
-  {% endif %}
+  {{ t("collections::page_title_with_date", {
+    context: context,
+    date: displayDate
+  }) }}
 {% endblock %}
 
 {% block content %}
@@ -32,28 +32,7 @@
           }) }}
           </span>
 
-          <span class="govuk-!-display-block">
-          {% if period === 'week' %}
-            {{ dateRange | formatDateRange }}
-            {% if displayRelativeDate %}
-              {{ appTime({
-                datetime: dateRange | formatISOWeek,
-                text: '(' + dateRange | formatDateRangeAsRelativeWeek | lower + ')',
-                classes: 'app-secondary-text-colour'
-                }) }}
-            {% endif %}
-          {% else %}
-            {% set datetime = dateRange | first %}
-            {{ datetime | formatDateWithDay }}
-            {% if displayRelativeDate %}
-              {{ appTime({
-                datetime: datetime,
-                text: '(' + datetime | formatDateAsRelativeDay(' ') | lower + ')',
-                classes: 'app-secondary-text-colour'
-                }) }}
-            {% endif %}
-          {% endif %}
-          </span>
+          <span class="govuk-!-display-block">{{ displayDate }}</span>
         </h1>
 
         {% block dateNavigation %}
--- a/common/templates/layouts/collection.njk
+++ b/common/templates/layouts/collection.njk
@@ -6,19 +6,18 @@
   }) }}'});
 {% endblock %}
 
-{% if period === 'week' %}
-  {% set datetime = dateRange | formatISOWeek %}
-  {% set displayDate = dateRange | formatDateRangeAsRelativeWeek %}
-{% else %}
-  {% set datetime = dateRange | first %}
-  {% set displayDate = datetime | formatDateAsRelativeDay %}
-{% endif %}
-
 {% block pageTitle %}
-  {{ t("collections::page_title_with_date", {
-    context: context,
-    date: displayDate | lower
-  }) }}
+  {% if period === 'week' %}
+    {{ t("collections::page_title_with_date", {
+      context: context,
+      date: dateRange | formatDateRangeAsRelativeWeek
+    }) }}
+  {% else %}
+    {{ t("collections::page_title_with_date", {
+      context: context,
+      date: datetime | formatDateAsRelativeDay
+    }) }}
+  {% endif %}
 {% endblock %}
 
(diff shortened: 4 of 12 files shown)


app/home/controllers.js
```diff
@@ -34,8 +36,14 @@ function dashboard(req, res) {
     },
   }
 
+  const [fromDate, toDate] = dateHelpers.getCurrentWeekAsRange()
+
   res.render('home/dashboard', {
     pageTitle: 'dashboard::page_title',
+    week: `(${formatters.date(fromDate, 'd MMM')} - ${formatters.date(
```
teneightfive, line 43: It's best to use `to` instead of dashes in dates as per the [GOV.UK styleguide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates).

teneightfive: You can also just use [this nunjucks filters](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/ba14c1c25c69ba56750fd2b18ead7c3d4df82345/config/nunjucks/filters.js#L68) which already supports the use of `to` and will shorten the dates if they contain the same month or year.

teneightfive:
I also feel this is linked to #428. It would be nice to use the time element and maybe a `title` attribute on instances like `Today` or `This week` to reveal the actual date on hover.

An example from GitHub:

Screenshot 2020-07-23 at 10 00 28 (2)

Andy-Millar:
Should appear in listings pages for ingoing and outgoing.

For the overview page we should just add the dates after the words week / day (as we do in the listing pages) - so for the overview page will appear as shown below. It's repetitive, but removes the issue of ambiguity which we've observed in testing:

Screenshot 2020-07-23 at 10 05 53

Andy-Millar:
Just to note - I've just hacked the original screenshot

Andy-Millar:
No to date hover - it's a hidden interaction with no affordance. Don't have it at all if it's hidden is the general rule of thumb with things in this regards

teneightfive:
>No to date hover - it's a hidden interaction with no affordance. Don't have it at all if it's hidden is the general rule of thumb with things in this regards

@Andy-Millar sorry, I wasn't suggesting that as the only solution to solve what has been seen in testing. If users aren't comfortable just seeing today/tomorrow then we should be more explicit with the date. #428 is something that we should do from an accessibility and semantics point of view regardless. It's good practice to wrap any relative dates with a [`time`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#:~:text=The%20HTML%20element%20represents,on%20a%2024%2Dhour%20clock.) element to communicate what date we are actually referring to.

I also wonder if a more consistent solution in this case might be to always persist the date and move the relative date to the brackets - then it would also align with how we use relative dates in the move detail summary:

<kbd></kbd>

So something like this:

| | |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

With the date in brackets it could feel like the date is jumping around:

| | |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

Andy-Millar:
This is a nice idea - (having tomorrow / today / this week etc in brackets after the date) - let's do that then!
If we display the date AND have the hover that's all good with me - I'm just against hiding info without any clear affordance that I can interact with an element. Displaying the date persistently means we aren't hiding anything, so all good

danielepolencic:
Is this what we want:

image

Andy-Millar:
Nearly. Few things:

1. Style 1 - keep the style of the date (ie. 20 Jul 2020) the same style as the title (ie. **black**, **bold**, I think it's "**govuk-heading-m**"
2. Style 2 - for the brackets and the copy in brackets (ie. (today), (this week), etc) the same style as the title (ie. **bold**, I think it's "**govuk-heading-m**") but change the colour to **grey**. 
3. Copy 1: "**Outgoing for {Date} (today)**" - CORRECT
4. Copy 2: "**Incoming for {Date} (today)**" - CORRECT
5. Copy 3: "**Single requests sent {Date 1} to {Date 2} (this week)**"
6. Copy 4: "**Allocations for {Date 1} to {Date 2} (this week)**"

Andy-Millar:
LGTM

teneightfive:
Should we persist this change to the other "dashboards":

image

These views feel like the ones users would use more to manage moves or allocations and if "this week" or "today" isn't providing enough it feels like the change should also happen here.

@Andy-Millar which screen was causing more confusion when the actual date range wasn't shown? If it was just the overview dashboard then maybe we don't need to worry about the dashboards which show moves or allocations.

app/home/controllers.js
```diff
@@ -11,26 +13,37 @@ function dashboard(req, res) {
...
 
+  const week = dateHelpers.getCurrentWeekAsRange()
+  const [today] = dateHelpers.getCurrentDayAsRange()
+
   const sections = {
     outgoing: {
       permission: 'moves:view:outgoing',
       heading: 'dashboard::sections.outgoing.heading',
       filter: req.filterOutgoing,
+      time: formatDate(today),
```
teneightfive, line 24: There is a [`helper`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/27474d4c2c020708d358a52f6a9d1f7513f658d9/config/nunjucks/filters.js#L194) that will already combine the day and day if it’s either today or tomorrow.

app/home/controllers.js
```diff
@@ -11,26 +13,37 @@ function dashboard(req, res) {
     return res.redirect('/moves')
   }
 
+  const week = dateHelpers.getCurrentWeekAsRange()
+  const [today] = dateHelpers.getCurrentDayAsRange()
```
teneightfive, line 17: What about using `new Date()` instead of calling this helper?

app/home/controllers.test.js
```diff
@@ -43,30 +43,23 @@ describe('Home controllers', function () {
...
-          },
+        const sections = [
+          'outgoing',
+          'incoming',
+          'singleRequests',
+          'allocations',
+        ]
+        expect(params).to.have.all.keys('pageTitle', 'sections')
+        expect(params.pageTitle).to.equal('dashboard::page_title')
+        expect(params.sections).to.have.all.keys(sections)
```
teneightfive, line 54: This is a nicer approach to testing all the keys exist :+1

app/home/controllers.test.js
```diff
@@ -43,30 +43,23 @@ describe('Home controllers', function () {
...
+        sections.forEach(section => {
+          expect(params.sections[section]).to.have.all.keys([
```
teneightfive, line 56: I notice this refactor now removes that we were testing that the sections contained the correct values for things like permissions and the filter. 

Interested on the thoughts of this approach, do you think it’s safe to not test this kind of behaviour in unit tests? I’m often in two minds...testing to means updating tests if the permission changes. But if not, it makes it easier to break when just updating the code. Sometimes a test explains the reason behind a value and can alert you if you change something you shouldn’t have.

locales/en/dashboard.json
```diff
@@ -2,24 +2,28 @@
...
-      "heading": "$t(primary_navigation.outgoing) for today",
+      "heading": "$t(primary_navigation.outgoing) for",
       "summary": {
         "total": "outgoing move",
         "total_plural": "outgoing moves"
       }
     },
     "incoming": {
-      "heading": "$t(primary_navigation.incoming) for today",
+      "heading": "$t(primary_navigation.incoming) for",
```
teneightfive, line 12: I wonder if it would be better to pass the date into the heading. Otherwise a screen reader would just read out “heading one, incoming moves for”.

locales/en/dashboard.json
```diff
@@ -2,24 +2,28 @@
...
+      "heading": "$t(primary_navigation.single_requests) sent"
     },
     "allocations": {
-      "heading": "$t(primary_navigation.allocations) for this week"
+      "heading": "$t(primary_navigation.allocations) for"
     }
+  },
+  "time": {
+    "day": "today",
+    "week": "this week"
```
teneightfive, line 27: Might also be worth double checking the other translation files as I think this is already defined somewhere.

Andy-Millar:
@teneightfive @danielepolencic 

Ah this should be applied globally for all dashboard.

So in your screenshot @teneightfive, it should be:
{Date} (today)

(Both in the same style as current, no line break, date in black, today in grey)

app/home/controllers.test.js
danielepolencic: I don't think there's an easy answer here.
I usually strive to test the intent of the code. Testing that I'm using the same key twice (one in the code, one in the test) adds limited value.
As an example, if I want to refactor the code (perhaps renaming a key), I need to refactor the tests too because there is tight coupling with the rest of the code.
If the intent is to test the _shape_ of the objects (does the object look like a section?), I think there are better ways to do it with stronger guarantees (examples here are libraries such as [tcomb](https://github.com/gcanti/tcomb) or precompilers/linters such as Typescript).
Said that I have preferences, but I don't mind aligning with the rest of the codebase. If the tests above are too loose, I'm happy to revert to a stricter check of the keys/values.

app/home/controllers.js
danielepolencic: I have to wrap the `(today)` in a `<span>` to change the colour to grey, so I can't use the helper. I tried really hard to reuse what was already there but opted for the code above in the end.

teneightfive: Ah yes. Of course!

app/home/controllers.test.js
teneightfive: >As an example, if I want to refactor the code (perhaps renaming a key), I need to refactor the tests too because there is tight coupling with the rest of the code.

I think that's an interesting point. My view would be that if we didn't test the presence of particular keys to make refactoring easier it could open the door to regression. For example, if a view depends on an object having a particular key structure. If we don't test those keys (to make refactoring easier) that refactor could remove important keys and we wouldn't catch it.

I think my concern in this instance is how can we ensure that the filter value is correctly set to the value from the request object. This refactor removes the test that ensures that the filter key in this object is set from the request and if in a future change somebody no longer passes that value to this object, how would they know the expected behaviour has changed?

We could for example rely on the e2e tests picking that up but that also feels a bit slower and more fragile.

app/home/controllers.js
teneightfive: I wonder how @Andy-Millar feels about not having the relative value in grey...?

danielepolencic:
Amended.

| Before | After |
| ------ | ----- |
| image | image |

app/home/controllers.js
danielepolencic: I worked around the issue and found a way to reuse the existing helpers as well as the `<time>` tag.

locales/en/dashboard.json
danielepolencic: Refactored. I propagated the change to the rest of the dashboards since we had the same issue everywhere else.

Andy-Millar:
Can we retain the font styling for "Allocations for" as shown in the 'before' screenshot please.
Style "3 to 9 Aug (this week)" as shown in the 'after' screenshot please.
Thanks!

common/templates/layouts/collection.njk
```diff
@@ -6,36 +6,48 @@
...
+  {% endif %}
 {% endblock %}
 
 {% block content %}
 
   {% block pageHeader %}
     <header class="govuk-grid-row govuk-!-margin-bottom-7">
       <div class="govuk-grid-column-two-thirds">
-        <span class="govuk-caption-xl">
+        <h1 class="govuk-heading-xl govuk-!-margin-bottom-2">
```
teneightfive, line 28: So that's a good spot that it should be all one heading so it's ready better but I think it would be good to keep the same visual style as before where the `Allocations for` is a reduced size and muted colour as I think it looks a little large and jarring with this change. What do you think @Andy-Millar ?

danielepolencic: Fixed.

app/home/dashboard.njk
```diff
@@ -15,8 +15,27 @@
...
-          <h2 class="govuk-heading-m govuk-!-margin-bottom-3">
+          <h2 class="govuk-heading-m govuk-!-margin-bottom-1">
             {{ t(section.heading) }}
+
+            <span class="govuk-!-display-block">
+            {% if section.period === 'week' %}
+              {{ dateRange | formatDateRange }}
+              {{ appTime({
+                datetime: dateRange | formatISOWeek,
+                text: '(' + t('actions::current_week') | lower + ')',
```
teneightfive, line 26: I'd avoid lowercasing this I think. Then it's consistent with how it's being used in the side panel.

app/home/dashboard.njk
```diff
@@ -15,8 +15,26 @@
...
+            {% if section.period === 'week' %}
+              {{ dateRange | formatDateRange }}
+              {{ appTime({
+                datetime: dateRange | formatISOWeek,
+                text: '(' + dateRange | formatDateRangeAsRelativeWeek | lower + ')',
+                classes: 'app-secondary-text-colour'
+                }) }}
+            {% else %}
+              {{ today | formatDateWithDay }}
+              {{ appTime({
```
teneightfive, line 31: So it looks like we've lost the day from the original format and it now also displays the date in grey on every page:

this
image

should be this:
image

and this:
image

should be this:
image

common/templates/layouts/collection.njk
```diff
@@ -6,36 +6,50 @@
...
+          {% if period === 'week' %}
+            {{ dateRange | formatDateRange }}
+            {{ appTime({
+              datetime: dateRange | formatISOWeek,
+              text: '(' + dateRange | formatDateRangeAsRelativeWeek | lower + ')',
+              classes: 'app-secondary-text-colour'
+              }) }}
+          {% else %}
+            {% set datetime = dateRange | first %}
+            {{ datetime | formatDateWithDay }}
```
teneightfive, line 45: So it looks like we've lost the day from the original format and it now also displays the date in grey on every page:

this
image

should be this:
image

and this:
image

should be this:
image

common/templates/layouts/collection.njk
```diff
@@ -6,36 +6,54 @@
...
-  }) }}
+  {% if period === 'week' %}
+    {{ t("collections::page_title_with_date", {
+      context: context,
+      date: dateRange | formatDateRangeAsRelativeWeek
+    }) }}
+  {% else %}
+    {{ t("collections::page_title_with_date", {
+      context: context,
+      date: datetime | formatDateAsRelativeDay
```
teneightfive, line 18: This has caused the page title to lose the date when not a relative date. 

I do think it would have been nicer to keep some of the original logic and make use of the nunjucks filters to handle logic of how the date is output instead of doing it all in the controllers.

Doing this in controllers has caused duplication and seems to have added a bit more logic into these templates.

<kbd></kbd>

common/templates/layouts/collection.njk
```diff
@@ -6,36 +6,54 @@
...
+                classes: 'app-secondary-text-colour'
+                }) }}
+            {% endif %}
+          {% else %}
+            {% set datetime = dateRange | first %}
+            {{ datetime | formatDateWithDay }}
+            {% if displayRelativeDate %}
+              {{ appTime({
+                datetime: datetime,
+                text: '(' + datetime | formatDateAsRelativeDay(' ') | lower + ')',
```
teneightfive, line 51: What's the purpose of passing the empty space as the argument to `formatDateAsRelativeDay`? Is it to prevent it from displaying if not a relative date? That feels a little hard to read if so, wonder if there is a better way.

It would also be good to remove the lowercasing to keep consistency with other uses of the relative date:

<kbd></kbd>

common/templates/layouts/collection.njk
danielepolencic: I guess we could remove the date from the title? Is it strictly necessary?
Outgoing moves for -> Outgoing moves

> [...] make use of the nunjucks filters to handle logic of how the date is output instead of doing it all in the controllers.
> Doing this in controllers has caused duplication and seems to have added a bit more logic into these templates.

The template had logic even before this change as it was assigning variables and branching depending on the period.
And I agree that the logic should be encapsulated in the controller where is testable.

I can create another filter or modify the existing `appTime` macro.

common/templates/layouts/collection.njk
danielepolencic: Yes, the code breaks with an empty space and displays nothing.
That chunk of code requires some love as the empty space is not the only _ugly hack_ that comes to my mind.
Concatenating parenthesis, with a `formatDate` and `lower` suggest that there is quite a lot of logic being executed.

This goes back to the previous comment:

1. I can introduce a new filter that doesn't display the date if it is not {Today,Tomorrow,Yesterday} 
1. I can edit the macro and move the code there.

As for the upper case, in one of the last Government project, I've learnt that most words should not be capitalised for the sake of it as they are more difficult to read. I'm not sure if there's a GovUK toolkit equivalent for content design, but this includes some guidance on capitalisation https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style

However if the rest of the website uses the capital T, I will change the rest of the code to be aligned to that.

common/templates/layouts/collection.njk
teneightfive: >I guess we could remove the date from the title? Is it strictly necessary?

 I'd prefer to keep it to be honest. When using things like the back button in a browser or looking at the history it would always just say the same thing. I think it's also probably actually better that the page title always shows the full date rather than the relative given this.

>I can create another filter or modify the existing appTime macro.

I think I felt like a nunjucks filter would have been a better place to encapsulate and test this logic but given this PR has been open a while and the work has been done I'm happy to leave it in the controller.

common/templates/layouts/collection.njk
teneightfive: >I can introduce a new filter that doesn't display the date if it is not {Today,Tomorrow,Yesterday}

We already have a filter which could be re-used for this purpose: [`formatDateWithRelativeDay`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/73c90bdf70fc5c965d9c0e0d99612c7d8c7ad617/config/nunjucks/filters.js#L194). Granted this is only for a single date but this could be replicated to do the same for relative week display.

> As for the upper case, in one of the last Government project, I've learnt that most words should not be capitalised for the sake of it as they are more difficult to read. I'm not sure if there's a GovUK toolkit equivalent for content design, but this includes some guidance on capitalisation https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style
>
> However if the rest of the website uses the capital T, I will change the rest of the code to be aligned to that.

I had another look through here and found a [section on dates](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates) which uses lowercase as you suggested. It does however, use a relative date first, then the actual date in brackets. But let's stick with the way round we have it for now but can we change both instances to be consistent?

If you do re-use that filter then you could lowercase in in the filter once. I pushed [a branch](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/8008f87bf28ad6a3a50c8c8171617cfde952ad0f) which re-used the filter as an example. I think it would greatly reduce the logic around this change but I also understand this change has been pending a while now so I'll leave the decision up to you. 

I do think it would be good to remain consistent with date display and to avoid any knock-on regressions in page titles with this change.

common/templates/layouts/collection.njk
danielepolencic: Amended the title, added a new filter for the week, refactored the rest of the code.

common/templates/layouts/collection.njk
danielepolencic: Amended for consistent capitalisation.

jason-sims:
👍 LGTM - thanks mate

common/templates/layouts/collection.njk
```diff
@@ -7,11 +7,10 @@
...
-  {% set datetime = dateRange | formatISOWeek %}
-  {% set displayDate = dateRange | formatDateRangeAsRelativeWeek %}
+  {% set displayDate = dateRange | formatDateRangeWithRelativeWeek %}
 {% else %}
   {% set datetime = dateRange | first %}
-  {% set displayDate = datetime | formatDateAsRelativeDay %}
+  {% set displayDate = datetime | formatDateWithRelativeDay %}
 {% endif %}
 
 {% block pageTitle %}
```
solidgoldpig, line 16:
```
{{ t("collections::page_title_with_date", {
    context: context,
    date: displayDate | lower
  }) }}
```
Can we remove the `| lower` in the title? Now we're not saying "Outgoing moves for today", I don't think we need it. Now we get the somewhat weird "Outgoing moves for wednesday 19 aug 2020 (today)"

solidgoldpig (approved):

PR #697, opened 2020-07-28
[P4-1795] Refactor move detail to make room for Person Escort Record

teneightfive:
### What changed

This refactors the move detail template to make better use of includes so that the Person Escort Record changes
can be added soon after.

### Why did it change

The one move detail template was getting extremely long with lots of nested logic. It was becoming
harder to maintain so this extracts some of that into smaller files.

### Issue tracking

- [P4-1795](https://dsdmoj.atlassian.net/browse/P4-1795)

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/move/views/view.njk
```diff
@@ -219,45 +219,12 @@
 
       {% include "move/views/_includes/assessment.njk" %}
```
solidgoldpig, line 220: Might it be cleaner to have:
```njk
{% include "move/views/_includes/assessment.njk" %}
{% include "move/views/_includes/court-information.njk" %}
{% include "move/views/_includes/documents.njk" %}
```
and move the conditionals to the chunks themselves?

Alternatively, could the conditionals be set up in the controller?

```njk
{% if includeCourtInformation %}
  {% include "move/views/_includes/court-information.njk" %}
{% endif %}
```
That might make it easier to unit test that we're outputting the right bits at least

app/move/views/_includes/court-information.njk
```diff
@@ -0,0 +1,66 @@
+<section class="app-!-position-relative">
+  {% if move.from_location.location_type == 'prison' and not courtSummary.rows.length %}
```
solidgoldpig, line 2: Why has the conditional moved inside the `section` element?

app/person-escort-record/controllers/framework-overview.js
```diff
@@ -2,7 +2,7 @@ const presenters = require('../../../common/presenters')
 
 function frameworkOverview(req, res) {
   const { originalUrl, framework, personEscortRecord = {}, move } = req
-  const profile = personEscortRecord?.profile || move?.profile
+  const profile = move?.profile || personEscortRecord?.profile
```
solidgoldpig, line 5: Really struggling to understand how the move could lack a profile but a PER exists...

teneightfive: Sorry, got trigger happy on the merge! It was actually around when the person escort record app was included within the move app. 

So a PER will exist on the profile, but we haven't asked for the profile to be included as part of the request. We also aren't running the `transform` method on the profile when it comes back from the move so the `fullname` property wasn't available. Inverting this statement meant it would continue to work in both instances of the app being loaded.

Ideally, the best fix would be to utilise some sort of transformation level at devour level when each model is deserialised but unfortunately the library doesn't support that as easily as we would like.

app/move/views/_includes/court-information.njk
teneightfive: The original conditional has been moved to the `view.njk` template. However, there is still another condition that needs to remain to choose whether to display court hearings or court information as they aren't shown for all move types.

app/move/views/view.njk
teneightfive: Yeah, ultimately I've trying to get to the latter that you've mentioned. I think it would be good to test those conditions via unit tests in the controller.

It has unfortunately evolved quite quickly with minimal large refactor so I was trying to pick and choose my battles. My next piece of work will involve changing how these sections are displayed quite drastically so you may see this evolve a bit more over the next couple of days...

PR #699, opened 2020-07-29
FE changes required for GOLIVE

danielepolencic:
## Proposed changes

- filter moves by supplier id
- get supplier locations from new dedicated endpoint
- refresh supplier locations each time user sets/switches location
- get moves data as csv from API
- get move supplier info from the move itself rather than the location
- prevent fetching of moves when no current location
- redirect from dashboard to moves before other middleware runs

### What changed

- add property `supplierId` to user object (when the user is a supplier)
- pass `from_location_id` in conjunction with `supplier_id` only if current location is set
- update user session locations when supplier locations are refreshed
- get csv download from the backend rather than generating it in the FE
- add client to restfully access API
- refactor `api-client` to share headers needed for accessing API with restful client
- add `supplier` relationship to `moves` model

## Issue tracking

- [[P4-1882] Filter moves by supplier id](https://dsdmoj.atlassian.net/browse/P4-1940)
- [[P4-2044] Show a supplier's list of distinct locations from the moves they currently own](https://dsdmoj.atlassian.net/browse/P4-2044)
- [[P4-2063] API endpoint to return locations with supplier moves](https://dsdmoj.atlassian.net/browse/P4-2063)

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer


test/fixtures/moves.json
```diff
@@ -80,14 +74,8 @@
...
           },
-          {
-            "identifier_type": "prison_number",
-            "value": "796507"
-          },
-          {
-            "identifier_type": "athena_reference",
-            "value": "763307"
-          }
+          { "identifier_type": "prison_number", "value": "796507" },
```
solidgoldpig, line 77: Once again, and this is nothing to do with this PR, I feel that what's being tested is somewhat bogus.

If we're running against v2, the identiers shouldn't be present and these props are actually flat on the person obj.

Also, again hating the (to my mind) unnecessary document reflow caused by prettier line length  pfft pfft pfft...

app/moves/middleware/set-filter.moves.js
```diff
@@ -8,10 +8,16 @@ function setfilterMoves(items = [], bodyKey) {
   return async function buildFilter(req, res, next) {
     const promises = items.map(item =>
       moveService
-        .getActive({
-          ...get(req, `body.${bodyKey}`, {}),
-          isAggregation: true,
-        })
+        .getActive(
```
solidgoldpig, line 11: a) We're still passing the location ids through too - this means we'll still send both the location ids and the supplier id (not sure how that works at the backend) and also we'll still batch if the locations exceed the batch size
b) this [c|s]hould probably be better handled in `set-body.moves` (and similarly for the others).

common/services/move.js
```diff
@@ -114,17 +114,22 @@ const moveService = {
     fromLocationId,
     toLocationId,
     isAggregation = false,
+    supplierId = undefined,
```
solidgoldpig, line 117: `getActive` doesn't have the `supplierId` arg

common/services/user.test.js
```diff
@@ -93,6 +93,86 @@ describe('User service', function () {
...
+          supplierStub.key
+        )
+      })
+
+      it('returns an Array of location objects', function () {
+        expect(result).to.deep.equal(supplierStub.id)
+      })
+    })
+
+    context('when supplier does not exist', function () {
```
solidgoldpig, line 165: Added some extra tests to cover the other cases when there is either no supplier role or no auth groups are returned.

common/lib/api-client/models.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
       date: '',
       date_from: '',
       date_to: '',
+      supplier_id: '',
```
teneightfive, line 19: Can we delete this commit for cleanliness? It seems to be adding something that isn't available on the move?

app/home/controllers.js
```diff
@@ -3,15 +3,19 @@ const { get } = require('lodash')
 const dateHelpers = require('../../common/helpers/date')
 const permissions = require('../../common/middleware/permissions')
 
-function dashboard(req, res) {
+function movesRedirect(req, res, next) {
```
teneightfive, line 6: Should this live in `middleware` instead of `controllers`? Feels like it resembles what we've called middleware elsewhere in the app.

common/lib/api-client/rest-client.js
```diff
@@ -0,0 +1,26 @@
+const axios = require('axios')
+
+const {
+  API: { BASE_URL },
+} = require('../../../config')
+
+const auth = require('./auth')()
+const getRequestHeaders = require('./request-headers')
+
+const restClient = async (url, params, options = {}) => {
```
teneightfive, line 10: I wonder if there is a simpler way to achieve this for the CSV response. Could we not patch the [response middleware](https://github.com/twg/devour/blob/3ce573bb708d851069a3c6c9f6805fbad4adf3bb/src/middleware/json-api/res-deserialize.js#L12), in a similar way to how we currently [patch the post middleware](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/2721f4ecfb2590dd59f33a0a2299f2fc05039789/common/lib/api-client/middleware/post.js#L4) and just check the `Content-Type` header? If so, just return the data instead of deserializing.

common/services/reference-data.js
```diff
@@ -83,13 +84,18 @@ const referenceDataService = {
     })
   },
 
-  getLocationsBySupplierId(supplierId) {
-    return referenceDataService.getLocations({
-      filter: {
-        cache: false,
-        'filter[supplier_id]': supplierId,
-      },
+  async getLocationsBySupplierId(supplierId) {
```
teneightfive, line 87: Is there any reason other than the different URL as to why this needs to be handled by a different "REST" client?

After our conversation yesterday and us already having a model for this, it would seem more sensible for the API to only maintain one endpoint that does this action. This appears to have introduced a duplicate endpoint with similar functionality.

common/services/move.js
```diff
@@ -136,7 +141,7 @@ const moveService = {
           'filter[to_location_id]': toLocationId,
           'filter[supplier_id]': supplierId,
         },
-        isUndefined
+        isEmpty
```
teneightfive, line 146: I have a feeling this might have been using `isUndefined` for a specific reason...might have been something about retaining the ability to send an empty string, or `null` for particular relationships...

teneightfive: But if that were the case, I'd have expected there to be a test, so maybe it's fine.

app/move/views/confirmation.njk
```diff
@@ -35,7 +35,7 @@
           name: move.profile.person.fullname | upper,
           location: location,
           date: move.date | formatDateWithDay,
-          supplier: supplierNames | oxfordJoin
+          supplier: move.supplier.name if move.supplier.name else t("supplier_fallback")
```
teneightfive, line 38: I believe you can do `move.supplier.name or t("supplier_fallback")` in nunjucks. It checks for deep properties, and if they don't exist, use the fallback after the `or`.

common/lib/api-client/rest-client.js
solidgoldpig: I tried - I really did. I don't see how you can skip the deserializing.

But I also think that it's cleaner to have a restful client for thing that the API exposes in a restful way.

Maybe the csv or data  should be a model type and we could return that as a property of the returned object for a json-api call

common/services/move.js
solidgoldpig: plus the e2e tests still work fine...

I would suggest that the tests were simply missing... since it allowed an empty array to be passed which ended up as an empty string which ends up as a filter by value '' => no results

I think we need to move all the argument formation and sanitising to move#getAll

common/lib/api-client/models.js
solidgoldpig: Zapped. It's been there for the past 4 months!

common/services/reference-data.js
solidgoldpig: > Is there any reason other than the different URL as to why this needs to be handled by a different "REST" client?

No

> After our conversation yesterday and us already having a model for this, it would seem more sensible for the API to only maintain one endpoint that does this action. This appears to have introduced a duplicate endpoint with similar functionality
Yes.

app/home/controllers.js
solidgoldpig: Lazily just extracted the method to the same module... you're right

teneightfive (approved):

PR #727, opened 2020-08-05
fix(per): Group responses correctly under flag summary

teneightfive:
### What changed

This fix checks to see if the response already exists when it's a
collection and appends to that response instead of creating a duplicate question.

This mimics the behaviour used for the overview pages.

It also adds some more high level unit tests to the presenters that control
this behaviour.

### Why did it change

Previously if the response type was `collection` and multiple values
were selected which correlate to a "tag" needing to be shown the
question text was being duplicated.

This was because the response values were being filtered and then
appending separately.

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/framework-section-to-panel-list.js
+++ b/common/presenters/framework-section-to-panel-list.js
@@ -3,66 +3,89 @@ const { kebabCase, omit } = require('lodash')
 const frameworkResponseToMetaListComponent = require('./framework-responses-to-meta-list-component')
 
 function frameworkSectionToPanelList({
-  tagList,
-  framework,
+  tagList = [],
+  questions,
   personEscortRecord,
-  personEscortRecordUrl,
-}) {
+  personEscortRecordUrl = '',
+} = {}) {
   return section => {
     const flagsMap = personEscortRecord?.responses
+      // Remove responses with no flags
       .filter(response => response.flags.length > 0)
+      // Remove responses from other framework sections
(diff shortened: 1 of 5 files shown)


common/presenters/framework-section-to-panel-list.js
```diff
@@ -1,68 +1,91 @@
...
-            acc[flag.title] = []
-          }
+          const flagGroup = acc[flag.title] || []
+
+          // handle collections and arrays differently to other types
+          if (
+            response.value_type === 'collection' ||
+            response.value_type === 'array'
+          ) {
+            const existingIndex = findIndex(flagGroup, { id: response.id })
```
solidgoldpig, line 27: The `findIndex` saves us from doing...
```
const existingIndex = flagGroup.indexOf(flagGroup.find(f => f.id === response.id ))
```

Could we not just do?
```
const existingGroup = flagGroup.find(f => f.id === response.id )
...
if (existingGroup) {
    existingGroup.value.push(flagValue[0])
}
```

solidgoldpig: NB. this is totally nitpicking and you can safely ignore it ;)

PR #728, opened 2020-08-07
[P4-1997] SHC and STC transfer moves

danielepolencic:
Enable new moves YOI/SCH/STC to YOI/SCH/STC .

Caveats:
- The API doesn't return a type for a YOI. YOI locations are prisons.
- YOI to YOI transfers are prison transfers.
- When a move is created, it's only possible to update the destination in the same category (if you chose STC as the destination, you can only choose another STC).
- All YOI/SCH/STC moves have a mandatory "upload risk assessment" step.
- STC/SCH users should be granted permission to create prison moves.

## Screenshots

Screenshot 2020-08-20 at 16 36 52
Screenshot 2020-08-20 at 16 37 15
Screenshot 2020-08-20 at 16 37 38
Screenshot 2020-08-20 at 16 38 06
Screenshot 2020-08-20 at 16 38 38
Screenshot 2020-08-20 at 16 38 50
Screenshot 2020-08-20 at 16 39 03
Screenshot 2020-08-20 at 16 39 28

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### DIFF
--- a/app/move/steps/create.js
+++ b/app/move/steps/create.js
@@ -114,7 +114,7 @@ module.exports = {
   '/move-date-range': {
     pageTitle: 'moves::steps.move_date.heading',
     fields: ['date_from', 'has_date_to', 'date_to'],
-    next: 'prison-transfer-reason',
+    next: 'transfer-reason',
   },
   '/move-date': {
     editable: true,
@@ -144,7 +144,7 @@ module.exports = {
           {
             field: 'to_location_type',
             value: 'prison',
-            next: 'prison-transfer-reason',
+            next: 'transfer-reason',
           },
           'release-status',
         ],
@@ -154,7 +154,7 @@ module.exports = {
     controller: MoveDate,
     fields: ['date', 'date_type', 'date_custom'],
   },
-  '/prison-transfer-reason': {
+  '/transfer-reason': {
     controller: PrisonTransferReason,
     pageTitle: 'moves::steps.prison_transfer_reason.heading',
     fields: ['prison_transfer_type', 'prison_transfer_comments'],
--- a/app/move/steps/create.js
+++ b/app/move/steps/create.js
@@ -39,8 +39,33 @@ const riskStep = {
       value: 'prison',
       next: 'special-vehicle',
     },
+    ...WhenYouthTransferMove({
+      goTo: 'travel-special-vehicle',
+      orElse: 'health-information',
+    }),
     'health-information',
   ],
+}
+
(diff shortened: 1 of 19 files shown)


locales/en/fields.json
```diff
@@ -79,6 +79,12 @@
       "court_appearance": {
         "label": "Court"
       },
+      "secure_childrens_home": {
+        "label": "Secure Children's Home (SCH)"
```
teneightfive, line 83: Quite a minor thing but can we use an apostrophe (`’`) here instead of a prime. To be honest, I was never aware of the difference until somebody pointed it out to me but now it's something I can never un-see! Here's a [good article](https://carsonparkdesign.com/quotation-marks-apostrophes-versus-primes/#:~:text=Single%20primes%20are%20used%20to,and%20double%20primes%20for%20inches.&text=The%20apostrophe%2C%20or%20single%20closed,used%20for%20contractions%20or%20possessives.) explaining the difference.

We've also tried hard to use them correctly everywhere else in app so would be good to maintain that consistency.

locales/en/fields.json
```diff
@@ -93,6 +99,12 @@
   "to_location_prison_transfer": {
     "label": "Name of prison"
   },
+  "to_location_secure_childrens_home_transfer": {
+    "label": "Name of Secure Children's Home (SCH)"
```
teneightfive, line 103: Same here for apostrophe instead of a prime

app/move/steps/create.js
```diff
@@ -178,6 +178,30 @@ module.exports = {
...
+      {
+        field: 'from_location_type',
+        value: 'secure_childrens_home',
+        next: [
+          {
+            field: 'to_location_type',
+            value: 'secure_childrens_home',
+            next: 'move-date-range',
+          },
+          'move-date',
```
solidgoldpig, line 190: Are all these `move_date`s necessary? (I know that the original one here is for the prison type and that it's been cloned). Surely the final`move_date` value is sufficient?

solidgoldpig: > Are all these `move_date`s necessary? (I know that the original one here is for the prison type and that it's been cloned). Surely the final`move_date` value is sufficient?

Nope. I'm completely wrong - that is really unnecessarily verbose :(

solidgoldpig (approved):
LGTM - with the caveat that I don't fully understand the implications for the business rules and permissions ;)

PR #737, opened 2020-08-14
refactor: Set optional field label dynamically

teneightfive:
### What changed

This adds another field helper to determine whether a field is optional
or not and set the label suffix (`(optional)`) dynamically.

### Why did it change

The majority of our fields are required so the pattern is to mark any non-required fields as _optional_.

To accommodate the framework this has been added to the form wizard logic.

This change started in the framework but I moved it to the form wizard
to reduce the complexity of the importer and to share the logic across
all areas where we use the form wizard.

### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/helpers/field/set-optional-label.js
+++ b/common/helpers/field/set-optional-label.js
@@ -1,0 +1,31 @@
+const { cloneDeep, get, set, flatten } = require('lodash')
+
+const i18n = require('../../../config/i18n')
+
+function setOptionalLabel([key, field]) {
+  const translated = cloneDeep(field)
+  const validations = flatten([translated.validate])
+  const isOptional =
+    validations.filter(v => v?.type === 'required' || v === 'required').length <
+    1
+  const labelPaths = [
+    'label.text',
+    'label.html',
+    'fieldset.legend.text',
+    'fieldset.legend.html',
+  ]
+
+  if (isOptional) {
+    labelPaths.forEach(path => {
+      const key = get(translated, path)
+
+      if (key) {
+        set(translated, path, `${key} (${i18n.t('optional')})`)
+      }
+    })
+  }
+
+  return [key, translated]
+}
+
+module.exports = setOptionalLabel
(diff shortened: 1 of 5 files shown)


common/helpers/field/set-optional-label.js
```diff
@@ -0,0 +1,31 @@
...
+    'fieldset.legend.text',
+    'fieldset.legend.html',
+  ]
+
+  if (isOptional) {
+    labelPaths.forEach(path => {
+      const key = get(translated, path)
+
+      if (key) {
+        set(translated, path, `${key} (${i18n.t('optional')})`)
```
solidgoldpig, line 23: I think this is more than good enough for now. 

If we do need the ability to specify different strings or (even none at all), it would be easy enough to do something like:

```js
const defautOptional = i18n.t('optional')
const optional = i18n.t(`${path}_optional`, defaultOptional)
set(translated, path, `${key}${optional}`)
```

To that end, might be worth putting the space and parentheses in the optional text itself...

locales/en/default.json
```diff
@@ -1,6 +1,7 @@
 {
   "feedback_link": "Give us feedback",
   "age": "Age",
+  "optional": "optional",
```
solidgoldpig, line 4: as noted above:

```json
"optional": " (optional)",
```

teneightfive:
@solidgoldpig So after a bit more thought and to avoid extra code and complications for now I've added the `(optional)` string to the label for optional fields [in the framework](https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks/pull/19) instead (similar to how we do it currently). I think that gives us a bit more control and also makes it a bit easier to see what's going on.

So with that news...I'm going to close this PR.

PR #747, opened 2020-08-26
Add mock auth server

solidgoldpig:
### What changed

- Enable running app without redis
- Add mock auth server for CI
 
Amends E2E tests
- Relax path to match auth login screen
- Only test for presence of download rather than its contents
- Prevent e2e checkAssessment helper emitting warning

Removed `DISABLE_CACHE` env var

### Why did it change

- This enables the running of e2e tests against a local instance which removes the need to wait for the review app to spin up.
- The auth server has consistently been the main cause for e2e test failures

### Issue tracking

n/a

## Screenshots

n/a

#### Automated testing

#### Manual testing

n/a

### Environment variables

- [x] Added to [continous integration](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-book-secure-move-frontend)

### Other considerations


.circleci/config.yml
```diff
@@ -166,14 +166,25 @@ jobs:
...
+                  echo "export AUTH_PROVIDER_URL='http://localhost:3999'" >> $BASH_ENV
+                  echo "export NOMIS_ELITE2_API_URL='http://localhost:3999'" >> $BASH_ENV
+                  echo "export SERVER_HOST='localhost:3000'" >> $BASH_ENV
+                  echo "export E2E_BASE_URL_REVIEW_APP='http://localhost:3000'" >> $BASH_ENV
                   source $BASH_ENV
-                no_output_timeout: 600s
-      - run:
-          name: Run E2E tests
-          command: E2E_BASE_URL=$<<parameters.base_url_env>> npm run test-e2e:ci
+                  npm run build
```
teneightfive, line 177: Why not just run this as a separate `run` command? 

Might make it easier to read both here and in the CI steps. So something like:

```yaml
- run:
    name: Build app
    command: npm run build
- run:
    name: Start the app
    command: 
        node start.js 1>/dev/null &
        node mocks/auth-server.js &
        sleep 5
- run:
    name: Run E2E tests
    command: E2E_BASE_URL=$<<parameters.base_url_env>> npm run test-e2e:ci
```

.circleci/config.yml
```diff
@@ -166,14 +166,25 @@ jobs:
           condition: <<parameters.use_review_app>>
           steps:
             - run:
-                name: Set Heroku Review Apps URL for Pull Requests
+                name: Run E2E tests (same instance)
                 command: |
-                  echo "export E2E_BASE_URL_REVIEW_APP=$(sh .circleci/get-review-app-url.sh)" >> $BASH_ENV
+                  echo "export REDIS_URL=''" >> $BASH_ENV
```
teneightfive, line 171: I feel like it these are no longer dynamic then we can set them on the Job using [Circle's built-in support](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-job) for envs.

.circleci/config.yml
solidgoldpig: I tried that.

The processes get terminated. So they have to kick off in the same container process(?) as the actual e2e tests

.circleci/config.yml
solidgoldpig: Or they could remain dynamic… I was just getting it working. But happy to set them as you suggest

.circleci/config.yml
teneightfive: Ahhh gotcha! Makes sense

teneightfive: Could probably still split the build though?

.circleci/config.yml
teneightfive: Also true. They could come from CI still. Maybe that’s better too

teneightfive (approved):

PR #761, opened 2020-09-03
Display approved single moves by 'Date of travel'

danielepolencic:
### What changed

When I select the “approved” filter on the Single Requests table
“Date of travel” (move date) is shown & sorted on. 
This is different from the “Earliest date of travel” on the “pending review” & “rejected” filter

### Issue tracking
- P4-1955

## Screenshots

image

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### DIFF
--- a/app/moves/middleware/set-results.single-requests.js
+++ b/app/moves/middleware/set-results.single-requests.js
@@ -33,10 +33,7 @@ async function setResultsSingleRequests(req, res, next) {
 
     req.resultsAsTable = {
       active: presenters.singleRequestsToTableComponent(singleRequests),
-      cancelled:
-        cancelledMoves.length > 0
-          ? presenters.singleRequestsToTableComponent(cancelledMoves)
-          : [],
+      cancelled: presenters.singleRequestsToTableComponent(cancelledMoves),
     }
 
     next()
--- a/common/presenters/single-requests-to-table-component.js
+++ b/common/presenters/single-requests-to-table-component.js
@@ -1,3 +1,5 @@
+const { isNil } = require('lodash')
+
 const i18n = require('../../config/i18n')
 const filters = require('../../config/nunjucks/filters')
 const componentService = require('../services/component')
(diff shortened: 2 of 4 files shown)


common/presenters/single-approved-requests-to-table-component.js
```diff
@@ -0,0 +1,84 @@
...
+    {
+      head: {
+        text: i18n.t('collections::labels.to_location'),
+      },
+      row: {
+        text: 'to_location.title',
+      },
+    },
+    {
+      head: {
```
solidgoldpig, line 58: Do we really need this additional module?

Could we not pass an additional arg to singleRequestsToTable, eg. date_type

```
head: {
  text: i18n.t(`collections::labels.${date_type === 'actual' ? '' : 'earliest_' }move_date`),
  attributes: {
    width: '120',
  },
},
row: {
  text: data => filters.formatDate(data[date_type === 'actual' ? 'date' : 'date_from']),
},
```

solidgoldpig (approved):

common/presenters/single-approved-requests-to-table-component.js
teneightfive: I think I'm leaning towards what @solidgoldpig said in that this feels like a lot of extra code to change the presentation of one column.

In essence it's still about transforming a single request into a table format but it sounds like the desired business rules are that if an actual date has been defined, we want to display it instead of the proposed dates.

I'd probably consider updating the existing table with something like this:

```javascript
{
  head: {
    text: i18n.t(data.date ? 'collections::labels.move_date' : 'collections::labels.earliest_move_date'),
    attributes: {
      width: '120',
    },
  },
  row: {
    text: data => filters.formatDate(data.date || data.date_from),
  },
},
```

Deep down, the logic is probably better being determined by the presence of the `date` attribute rather than the status of the move. A move can't be "approved" without a date being sent as part of the approval.

solidgoldpig: yep, @teneightfive's suggestion to use the attribute itself to determine the column is a better way of handling it.

PR #764, opened 2020-09-04
[P4-2126] fix: Add validation to prison number field

merlinc:
## Proposed changes
Add form-wizard validation to the prison number field

### What changed
A validator was added to the prison number field

### Why did it change

There is an existing issue with "Prison number containing colon character - e.g. 08:00 returns nasty exception" [P4-2124](https://dsdmoj.atlassian.net/browse/P4-2124).

Recently a downstream service made some changes, and these errors were being surfaced in the `frontend` application as 5xx errors. These errors were also being thrown while running the E2E tests.

#763 works around the issue for the E2E tests, but the underlying bug is still present.

This fix applies validation to the field to ensure that only correctly formatted prison numbers are sent downstream.

### Issue tracking

- [P4-2126](https://dsdmoj.atlassian.net/browse/P4-2126)

## Screenshots

Screenshot from 2020-09-04 16-11-10
----
Screenshot from 2020-09-04 16-11-23
----

Screenshot from 2020-09-29 17-00-17

----

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/locales/en/validation.json
+++ b/locales/en/validation.json
@@ -22,6 +22,7 @@
   "default": "is required",
   "default_with_label": "{{label}} is required",
   "generic": "something went wrong",
+  "prisonNumber_with_label": "{{label}} must be of the format A1234AB",
   "move_conflict": {
     "heading": "This move cannot be created",
     "message": "A move <a href=\"{{href}}\">already exists</a> for <strong>{{name}}</strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong>.",
--- a/locales/en/validation.json
+++ b/locales/en/validation.json
@@ -22,7 +22,7 @@
   "default": "is required",
   "default_with_label": "{{label}} is required",
   "generic": "something went wrong",
-  "prisonNumber_with_label": "{{label}} must be of the format A1234AB",
+  "prisonNumber": "Enter a prison number in the correct format",
   "move_conflict": {
     "heading": "This move cannot be created",
     "message": "A move <a href=\"{{href}}\">already exists</a> for <strong>{{name}}</strong> to <strong>{{location}}</strong> on <strong>{{date}}</strong>.",
--- a/app/move/fields/filter.prison-number.js
+++ b/app/move/fields/filter.prison-number.js
@@ -12,6 +12,9 @@ const filterPrisonNumber = {
     html: 'fields::filter.prison_number.label',
     classes: 'govuk-label--s',
   },
+  hint: {
+    text: 'fields::filter.prison_number.hint',
+  },
   validate: ['required', validators.prisonNumber],
 }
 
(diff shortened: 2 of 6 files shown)


solidgoldpig (approved):

solidgoldpig:
All looks good - bar displaying a proper error message

merlinc:
@Andy-Millar Are you happy with the wording of this error message (as above)

Andy-Millar:
@merlinc Thanks for tagging me in this.

I'm just looking at the error message page in the design system - https://design-system.service.gov.uk/components/error-message/

I'm thinking it would be better to pre-empt the user getting the format wrong by adding hint text. So we update the form title to be:

[Title] **Prison number**
[Hint] For example, 'A1234AB'

Then can we have the error message display as:
[Error] Enter a prison number in the correct format

It would be good if we can follow this pattern for other journeys too (ie. Police) - is that also part of this work, or does it need a new ticket?

merlinc:
@Andy-Millar The other IDs we enter (e.g PNC), also should have validation - these would be a separate ticket as this one is only concerning the Prison Number field.

I'll look into getting the messages updated as per your content

Andy-Millar:
Excellent - thanks @merlinc

locales/en/validation.json
```diff
@@ -22,6 +22,7 @@
   "default": "is required",
   "default_with_label": "{{label}} is required",
   "generic": "something went wrong",
+  "prisonNumber_with_label": "{{label}} must be of the format A1234AB",
```
teneightfive, line 25: Just had a quick look at the Design System for [error messages](https://design-system.service.gov.uk/patterns/national-insurance-numbers/#error-messages) and I wonder if we should add a hint to the field and simplify the error message:

**Prison number**
The person's identifier within the prison system. For example, A1234AB.
_Enter a prison number in the correct format_
[.....................]

Andy-Millar: I don't think we need to add "The person's identifier within the prison system." It's working fine without this message, bit overkill otherwise. 

Content should be as previous:

> [Title] Prison number
> [Hint] For example, 'A1234AB'
> 
> Then can we have the error message display as:
> [Error] Enter a prison number in the correct format

teneightfive: Yep, think you're right. I wasn't sure on that either actually!

I also only just spotted you said the same thing a few days ago 🤦  Great minds 😉

merlinc:
@Andy-Millar 

Screenshot from 2020-10-06 15-29-57

Screenshot from 2020-10-06 15-30-08

Andy-Millar:
@teneightfive Haha I did wonder - I was scratching my head! We literally came up with the same content - so design system worked haha.

Andy-Millar:
Perfect - thanks @merlinc , great stuff!

PR #768, opened 2020-09-07
Prevent last property bag from being removed [P4-2071]

solidgoldpig:
### What changed

Adds minItems (defaults to 1) and maxItems to field schema for multiple items fields

- Show remove buttons when items exceeds minItems
- Show add another button unless maxItems exceeds items

### Why did it change

Prevent removal of last item when there should be at least one item

### Issue tracking

- [P4-2071](https://dsdmoj.atlassian.net/browse/P4-2071)

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/services/frameworks.js
+++ b/common/services/frameworks.js
@@ -100,8 +100,6 @@ const frameworksService = {
       questions,
       type,
       validations = [],
-      minItems,
-      maxItems,
     } = {}
   ) {
     if (!key) {
@@ -148,16 +146,7 @@ const frameworksService = {
       field.default = [{}]
 
       // default minimum items based on whether it's required
-      if (minItems === undefined) {
-        minItems = validations.includes('required') ? 1 : 0
-      }
-
-      field.minItems = minItems
-
-      // set maximum items if any
-      if (maxItems) {
-        field.maxItems = maxItems
-      }
+      field.minItems = validations.includes('required') ? 1 : 0
     }
 
     if (options) {
--- a/common/services/frameworks.js
+++ b/common/services/frameworks.js
@@ -100,7 +100,7 @@ const frameworksService = {
       questions,
       type,
       validations = [],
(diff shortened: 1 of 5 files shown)


common/services/frameworks.js
```diff
@@ -100,6 +100,8 @@ const frameworksService = {
       questions,
       type,
       validations = [],
+      minItems = 1,
```
teneightfive, line 103: We'll need to actually define these in the framework before we merge it. Slightly pedantic but the casing style from the framework has been snake case so we may need to update this once we get the definition into the framework.

teneightfive: @cwrw we should have a quick chat about how/if we actually want to define this in the framework at this stage.

teneightfive (commented):
The updates to the component look good but we'll need to co-ordinate the framework change with how we define it in the framework.

We could split this into two changes to get the component update in first, but up to you. Might be just as easy to wait.

common/services/frameworks.js
```diff
@@ -144,6 +146,13 @@ const frameworksService = {
       field['ignore-defaults'] = true
       // use `default` to start with one empty item
       field.default = [{}]
+      // default minimum items to 1
+      field.minItems = minItems
```
teneightfive, line 150: Had a quick chat with @cwrw and seeing as we don't really want to add it fully to the framework until we have actual use cases for needing custom minimum and maximum we thought let's just determine the min length based on the presence of required for now.

I think it's good to have the ability in the component but in this file we can do something simpler for now.

solidgoldpig: Updated to take the default from the validations.

teneightfive (approved):

PR #775, opened 2020-09-09
Show warning badges on PER print record

solidgoldpig:
### What changed

Adds warning badges/tags to PER print record

Font size is reduced to prevent the likelihood of wrapping the completed sections part on to the next page

### Issue tracking

- [P4-2003 - PER Print; show warning badges on the overview sheet](https://dsdmoj.atlassian.net/browse/P4-2003)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd>print-screen-old</kbd> | <kbd>print-screen-new</kbd> |
| <kbd>print-old</kbd> | <kbd>print-new</kbd> |

#### Automated testing

Tests will be added when addressing the lack of tests in next story addressing hard-coded values and lack of tests for this module

#### Manual testing

- [x] Chrome
- [x] Firefox
- [x] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/person-escort-record/views/print-record.njk
+++ b/app/person-escort-record/views/print-record.njk
@@ -123,6 +123,11 @@
                 reference: reference
               }) }}
             </span>
+            <div class="govuk-!-margin-top-2">
+              {% for tag in personEscortRecordTagList %}
+                {{ appTag(tag) }}
+              {% endfor %}
+            </div>
           </header>
 
           {# TODO: Make this view a bit more dynamic and built from the controller #}
--- a/app/person-escort-record/controllers/print-record.js
+++ b/app/person-escort-record/controllers/print-record.js
@@ -62,6 +62,9 @@ function printRecord(req, res) {
       }
     }
   )
+  const personEscortRecordTagList = presenters.frameworkFlagsToTagList(
+    personEscortRecord?.flags
+  )
   // TODO: Extract this logic to the framework
   // Checking for this value based on a key within the frontend is fragile
   // It could break if somebody simply renames a question key (filename) or option
@@ -120,6 +123,7 @@ function printRecord(req, res) {
     requiresMedicationDuringTransport,
     personalDetailsSummary,
     personEscortRecordSections,
+    personEscortRecordTagList,
   }
 
   res.render('person-escort-record/views/print-record', locals)
--- a/app/person-escort-record/controllers/print-record.test.js
+++ b/app/person-escort-record/controllers/print-record.test.js
@@ -22,7 +22,7 @@ describe('Person Escort Record controllers', function () {
     })
 
     it('should contain correct locals', function () {
-      expect(Object.keys(params)).to.have.length(16)
+      expect(Object.keys(params)).to.have.length(17)
       expect(Object.keys(params)).to.deep.equal([
         'moveId',
         'moveType',
@@ -40,6 +40,7 @@ describe('Person Escort Record controllers', function () {
         'requiresMedicationDuringTransport',
         'personalDetailsSummary',
         'personEscortRecordSections',
+        'personEscortRecordTagList',
       ])
     })
   })
--- a/common/components/message/_message.scss
+++ b/common/components/message/_message.scss
@@ -37,6 +37,10 @@ $_message-border-width: 5px;
   * + & {
     margin-top: govuk-spacing(4);
   }
+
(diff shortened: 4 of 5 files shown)


app/person-escort-record/views/print-record.njk
```diff
@@ -123,6 +123,11 @@
                 reference: reference
               }) }}
             </span>
+            <div class="govuk-!-margin-top-2" data-tag-list-source="move-request">
```
teneightfive, line 126: You can drop the `data` attribute. This was being used for an e2e check on the move detail page.

teneightfive (approved):

teneightfive:
I've just thought of a potential issue with this addition...this will now cause our "cover sheet" (first page) to break onto two pages if the person has the two warning banners as well as a number of flags:

| | |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

The banners and flags have also just come through in colour on mine. We should maybe look to override those on the print view too. (Something I did miss on the warning banners).

@Andy-Millar what's your view on this now causing the cover page to potentially break onto two pages?

I was actually having a look at these banners earlier and what the desired outcome is on the paper equivalent and I think we have some room to improve the display of them a little bit based on actual responses to the framework questions.

We could also make the design more compact to prevent the screenshots about too.

Andy-Millar:
Me and @solidgoldpig have been discussing this issue. It only occurs in the case that all the risks and the two banners are true - so should be extremely rare. I'd rather it break (as per your screenshots) so that it is always on the cover page.

I'd like to resolve it though - few things that should resolve the issue:
- Smaller sizing for the person's name
- smaller sizing for the banners at the top of the page
- smaller sizing for the 'completed sections' / 'attached docs' section

Alternatively, we could swap 'Move info' with the 'completed sections' / 'attached docs' as a quick fix, but that's still not an ideal solution

teneightfive:
> Me and @solidgoldpig have been discussing this issue. It only occurs in the case that all the risks and the two banners are true - so should be extremely rare. I'd rather it break (as per your screenshots) so that it is always on the cover page.
> 
> I'd like to resolve it though - few things that should resolve the issue:
> 
> * Smaller sizing for the person's name
> * smaller sizing for the banners at the top of the page
> * smaller sizing for the 'completed sections' / 'attached docs' section
> 
> Alternatively, we could swap 'Move info' with the 'completed sections' / 'attached docs' as a quick fix, but that's still not an ideal solution

@Andy-Millar Cool, if you're aware and happy that's perfect. As you say, might be more an edge case anyway.

We can look at some improvements after. I was discussing how we can reduce the fragility of it this morning with @cwrw and how we can move the content to the framework. We looked at the existing paper PER and it seems the areas on the box depend on responses to questions so we could look to dynamically show them based on responses in a more compact style.

I also noticed that it appears our escape risk is suggesting something a little different to the paper version.

From the paper version:

<kbd>image</kbd>

It looks like a small content tweak could solve ours for now - `Escape risk` -> `Escape list`. Is it also true that this will only be for prison use? If so, we can wrap it in condition that will only show it for prison journeys. This will also reduce the number of scenarios where all items will be displayed.

solidgoldpig:
> The banners and flags have also just come through in colour on mine. We should maybe look to override those on the print view too. (Something I did miss on the warning banners).

Chrome is insisting they're black (and no, I don't have print b+w turned on). But in Firefox I see the colours. And that's what the code says...

Should the banner borders also be black rather than coloured?

solidgoldpig:
> I've just thought of a potential issue with this addition...this will now cause our "cover sheet" (first page) to break onto two pages if the person has the two warning banners as well as a number of flags:

Firefox not only pushes the section on to the next page, but keeps the Completed sections/Attached docs intact!

Andy-Millar:
Good stuff.

Re. copy - we are using 'Escape risk' so that it also for police. We are simply flagging people who are an escape risk

Andy-Millar:
> > The banners and flags have also just come through in colour on mine. We should maybe look to override those on the print view too. (Something I did miss on the warning banners).
> 
> Chrome is insisting they're black (and no, I don't have print b+w turned on). But in Firefox I see the colours. And that's what the code says...
> 
> Should the banner borders also be black rather than coloured?

Ahh just spotted this - sorry @solidgoldpig 

Can we make it match what we do for the warning badges please

PR #776, opened 2020-09-10
[P4-2054] Ability to sort Single Requests table

danielepolencic:
### Why did it change

Ability to sort Single Requests table (follows the same convention as allocations):

- Name - Alphabetical (same as allocations)
- Sent on - Date (same as allocations)
- Move from - Alphabetical (same as allocations)
- Move to - Alphabetical (same as allocations)
- Move type - Alphabetical (same as allocations)

Note: 
- Earliest date of travel (or date of travel) - this is currently a composite of two columns, so is not being sorted as a part of this ticket

### Issue tracking
-  [P4-2054](https://dsdmoj.atlassian.net/browse/P4-2054)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### DIFF
--- a/common/presenters/single-requests-to-table-component.js
+++ b/common/presenters/single-requests-to-table-component.js
@@ -7,83 +7,95 @@ const componentService = require('../services/component')
 const moveToCardComponent = require('./move-to-card-component')
 const tablePresenters = require('./table')
 
-function singleRequestsToTable(moves) {
-  const tableConfig = [
-    {
-      head: {
-        text: i18n.t('name'),
-        attributes: {
-          width: '220',
+function singleRequestsToTable({ isSortable = true, query } = {}) {
+  return function (moves) {
+    const tableConfig = [
+      {
+        head: {
+          isSortable,
+          sortKey: 'name',
+          html: 'name',
+          attributes: {
+            width: '220',
+          },
+        },
+        row: {
+          html: data => {
+            const card = moveToCardComponent({
+              isCompact: true,
+              hrefSuffix: '/review',
+            })(data)
+            return componentService.getComponent('appCard', card)
+          },
+          attributes: {
+            scope: 'row',
+          },
         },
       },
-      row: {
-        html: data => {
-          const card = moveToCardComponent({
-            isCompact: true,
-            hrefSuffix: '/review',
-          })(data)
-          return componentService.getComponent('appCard', card)
+      {
+        head: {
+          isSortable,
+          sortKey: 'created_at',
+          html: 'collections::labels.created_at',
+          attributes: {
+            width: '120',
+          },
         },
-        attributes: {
-          scope: 'row',
-        },
-      },
-    },
-    {
-      head: {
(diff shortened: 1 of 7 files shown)


common/presenters/single-requests-to-table-component.test.js
```diff
@@ -179,7 +189,7 @@ describe('#singleRequestsToTableComponent()', function () {
 
   describe('its behaviour', function () {
     beforeEach(function () {
-      output = presenter(mockMoves)
+      output = presenter(mockMoves)(mockMoves)
```
merlinc, line 192: This looks weird, but I'm not sure what the alternative is.

The first invocation needs the data to determine the columns that are shown. This would create an empty table if used.
The second invocation uses that table, with the provided data to provide a normal table view.

teneightfive: Could you not call it with no args for the first call for the purpose of this test?

```javascript
output = presenter()(mockMoves)
```

common/presenters/single-requests-to-table-component.js
```diff
@@ -7,83 +7,95 @@ const componentService = require('../services/component')
...
 const tablePresenters = require('./table')
 
-function singleRequestsToTable(moves) {
-  const tableConfig = [
-    {
-      head: {
-        text: i18n.t('name'),
-        attributes: {
-          width: '220',
+function singleRequestsToTable({ isSortable = true, query } = {}) {
```
teneightfive, line 10: It might be good to add some tests around calling this method with the different properties.

It currently looks like the tests only test the default values for these options.

common/presenters/single-requests-to-table-component.test.js
merlinc: I'll add some tests for variants of columns - I'm not sure if I can meaningfully pass through empty column data, but I can at least test the dynamic columns feature.

merlinc: I've added tests for the various permutations of the columns - I still have the `(mockMoves)(mockMoves)` but I also have ones with `()(mockMoves)`

common/presenters/single-requests-to-table-component.js
merlinc: Tests have been added for the use of `isSortable` and `query`

teneightfive, line 10: I wonder if the default would be cleaner as `false`?

That way the call to these presenters in the middleware feels a little easier to read too. If the table is sortable, you need to pass in the query, if not, call with default options:

```javascript
req.resultsAsTable = {
  active: presenters.singleRequestsToTableComponent({
    isSortable: true,
    query,
  })(cancelledMoves),
  cancelled: presenters.singleRequestsToTableComponent()(cancelledMoves),
}
```

teneightfive (approved):

PR #777, opened 2020-09-10
[P4-2056] Add Cancelled by PMU

merlinc:
## Proposed changes
Add an additional cancellation reason of "cancelled by pmu", and a corresponding comment field

### Issue tracking

- [P4-2056](https://dsdmoj.atlassian.net/browse/P4-2056)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/home/dashboard.njk
```diff
@@ -35,4 +35,4 @@
       {% endif %}
     {% endfor %}
   </div>
-{% endblock %}
+{% endblock %}
```
teneightfive, line 38: This looks like an accidental change?

We should keep the newline at the end of the file in these instances too. Maybe look at installing an [editorconfig extension](output = presenter(mockMoves)(mockMoves)) and it should take care of this for you.

common/services/move.js
```diff
@@ -269,8 +269,7 @@ const moveService = {
       .update('move', {
         id,
         status: 'cancelled',
-        cancellation_reason: reason,
```
teneightfive, line 272: I think it would be better to keep the structure of this method the same so that the names of the API attributes are kept in one place and isolated.

By changing this we have made the controllers need to know about the API attribute names and if this ever changed we'd be having to update it in a number of different places.

The idea of these services was to separate that logic so that only the service needs to care about the API payload structures.

teneightfive: However, I know the _create_ does slightly contradict this concept...but I think it would be nice to not change this if possible.

app/move/controllers/cancel.js
```diff
@@ -39,12 +39,21 @@ class CancelController extends FormWizardController {
         Object.keys(req.form.options.allFields)
       )
 
-      data.cancellation_reason_comment = data.cancellation_reason_other_comment
+      if (
```
teneightfive, line 42: I find if/else if blocks a little harder to read and understand. 

I wonder if a switch statement might be easier to read in this instance:

```javascript
const reason = data.cancellation_reason
let comment

switch (reason) {
  case 'cancelled_by_pmu':
    comment = data.cancellation_reason_cancelled_by_pmu_comment
    break
  case 'other':
    comment = data.cancellation_reason_other_comment
    break
  default:
    comment = data.cancellation_reason_comment
    break
}

await moveService.cancel(moveId, {
  reason,
  comment,
})
```

app/home/dashboard.njk
merlinc: Good catch.

Editorconfig is installed and running - don't know why it didn't pick up on this first time around, when I'm editing it now, I can't make the newline dissapear. (Will remove this from the PR anyway)

app/move/controllers/cancel.js
merlinc: We only have a comment for 2 of the reasons, for the other 2 cases, we ignore that option.

The API has the field as optional, so we should only be sending it if we have it.

Would a switch statement without a default option be an acceptable substitute?

common/services/move.js
merlinc: So the two cancel methods (moves, allocations) were previously doing this differently. 

Allocation was passing the data all the way through, whereas move was "renaming" the properties.

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blame/5093d71a7a2a4e3d5de5ed90257979eabdd0835d/app/allocation/controllers/cancel.js

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blame/5093d71a7a2a4e3d5de5ed90257979eabdd0835d/app/move/controllers/cancel.js

This change was aiming to simplify this logic around conditional parameters, but it was made at the expense of bleeding through architectural tiers.

Based upon your comments above, how about I make a separate PR to change allocations to have a specific data structure - that should simplify the changes needed in this one.

teneightfive: >Based upon your comments above, how about I make a separate PR to change allocations to have a specific data structure - that should simplify the changes needed in this one.

This sounds like a good plan to me. I do understand in some areas we've bleed the architecture of the API into the form wizard and controllers but I think it would be nice to try to continue to avoid that where possible.

app/move/controllers/cancel.js
teneightfive: >We only have a comment for 2 of the reasons, for the other 2 cases, we ignore that option.

Ah, gotcha. Then in that case I think this is a good suggestion:

Would a switch statement without a default option be an acceptable substitute?

app/home/dashboard.njk
solidgoldpig: what on earth does `[editorconfig extension](output = presenter(mockMoves)(mockMoves))` mean?

merlinc: I assumed the presenter was an accidental `paste` from another comment

common/services/move.js
merlinc: The refactoring PR is https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/802

app/move/controllers/cancel.js
solidgoldpig: Personally I prefer object literals and hate the switches

```js
const commentType = {
  cancelled_by_pmu: 'cancellation_reason_cancelled_by_pmu_comment',
  other: 'cancellation_reason_other_comment',
  default: 'cancellation_reason_comment'
}

const reason = data.cancellation_reason
let comment = data[commentType[reason || 'default']]

await moveService.cancel(moveId, {
  reason,
  comment,
})
```

test/e2e/pages/cancel-move.js
```diff
@@ -39,7 +39,10 @@ class CancelMovePage extends Page {
           selector: this.fields.cancellationReasonOtherComment,
           value: comment,
         }
-      } else if (reason === 'Cancelled by Population Management Unit (PMU)') {
+      } else if (
+        reason === 'Cancelled by Population Management Unit (PMU)' &&
```
solidgoldpig: Where on earth does this reason come from?

solidgoldpig: aaaah. It's the e2e tests. Move along... nothing to see here

solidgoldpig (approved):

common/services/move.js
merlinc: Refactoring PR has been merged, so controllers now don't pass through properties into the service.

app/move/controllers/cancel.js
merlinc: @solidgoldpig In this scenario, we don't have a default option for the comment, so there'd be a few more undefined checks that would need to be done. Given that, I don't know if it would be cleaner than the switch statement I've implemented.

test/e2e/pages/cancel-move.js
merlinc: Yeah, they work off labels, so changing the `locale` files will break the E2E tests, which is nice.

app/move/controllers/cancel.js
solidgoldpig: I'm pretty sure that what I wrote is absolutely equivalent to the switch.

It's a bit of a running gag - any time someone uses a switch statement, I show that you could just as well do it with an object literal. The ghost of Douglas Crockford's "The Good Parts" still stalks this land...

teneightfive (approved):

app/move/controllers/cancel.js
merlinc: @solidgoldpig Ah, sorry - I was unclear. What I meant to say is that you've re-written @teneightfive 's code, which is great - but neither of them correctly solve the problem, as there is no default comment :)

solidgoldpig: sorry, here's the original rewritten then

```js
      const commentType = {
        cancelled_by_pmu: 'cancellation_reason_cancelled_by_pmu_comment',
        other: 'cancellation_reason_other_comment',
      }

      const reason = data.cancellation_reason
      const comment = data[commentType[data.cancellation_reason]]

      await moveService.cancel(moveId, {
        reason,
        ...(comment && {
          comment: comment,
        }),
      })
```

Even simpler!

solidgoldpig: or indeed

```js
      const reason = data.cancellation_reason
      const comment =
        data[`cancellation_reason_${data.cancellation_reason}_comment`]

      await moveService.cancel(moveId, {
        reason,
        ...(comment && {
          comment: comment,
        }),
      })
```

if we're sticking rigorously to that naming convention and don't need the map ;)

PR #781, opened 2020-09-14
Add user agent string to api client

solidgoldpig:
### What changed

Add user agent string to api client

### Why did it change

This enables us to distinguish our client requests from others using the same library

### Issue tracking

n/a

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/lib/api-client/request-headers.js
+++ b/common/lib/api-client/request-headers.js
@@ -2,10 +2,12 @@ const uuid = require('uuid')
 
 const {
   API: { VERSION },
+  APP_VERSION,
 } = require('../../../config')
 
 module.exports = (format = 'application/vnd.api+json') => {
   return {
+    'User-Agent': `hmpps-book-secure-move-frontend/${APP_VERSION}`,
     Accept: `${format}; version=${VERSION}`,
     'Accept-Encoding': 'gzip',
     'Idempotency-Key': uuid.v4(),
--- a/app/healthcheck/controllers.js
+++ b/app/healthcheck/controllers.js
@@ -6,15 +6,15 @@ const {
   APP_BUILD_TAG,
   APP_BUILD_BRANCH,
   APP_GIT_SHA,
+  APP_VERSION,
 } = require('../../config')
-const { version } = require('../../package.json')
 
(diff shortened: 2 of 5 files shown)


merlinc:
I think this is a good idea. 

My only question is how does the` api` layer want us to supply this sort of data? I did q quick search through the `api` codebase and didn't see anything relevant about `user-agent`.

merlinc (commented):
I think this is a good idea. 

My only question is how does the` api` layer want us to supply this sort of data? I did q quick search through the `api` codebase and didn't see anything relevant about `user-agent`.

solidgoldpig:
> I think this is a good idea.
> 
> My only question is how does the` api` layer want us to supply this sort of data? I did q quick search through the `api` codebase and didn't see anything relevant about `user-agent`.

It's a halfway-house quick fix. Rather than figuring out whether a particular `axios` request if from the frontend app or not, it will be self-evident. Obviously this does not go the whole hog and tie the api request back to a parent request by a user... but that's for laters.

common/lib/api-client/request-headers.js
```diff
@@ -2,10 +2,12 @@ const uuid = require('uuid')
 
 const {
   API: { VERSION },
+  APP_VERSION,
 } = require('../../../config')
 
 module.exports = (format = 'application/vnd.api+json') => {
   return {
+    'User-Agent': `hmpps-book-secure-move/${APP_VERSION}`,
```
teneightfive, line 10: Do we want to append `-fronted` to be more explicit: `hmpps-book-secure-move-frontend/${APP_VERSION}`?

teneightfive (approved):

common/lib/api-client/request-headers.js
solidgoldpig: Done

merlinc: @teneightfive Good idea, that makes it more obvious it is from our `frontend` client

PR #782, opened 2020-09-15
Update Person Escort Record framework version

teneightfive:
### What changed

This adds support for the latest PER framework version, [v1.2.0](https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks/releases/tag/v1.2.0).

This also includes a fix to support the use of emdashes in the framework which were previously being formatted out and then throwing a validation error as that value was no longer found in the list of available options.

### Why did it change

In order to support the use of the new version, we need to add it to the list of available packages.

The problem cropped up when one of the framework questions contained an emdash (`—`) in the value and the form wizard
[default formatters](https://github.com/UKHomeOffice/passports-form-wizard/blob/c8131187ddaafd220af89e75b8c2a400f4bdd78e/lib/controller/index.js#L31-L37) would [replace emdashes](https://github.com/UKHomeOffice/passports-form-wizard/blob/c8131187ddaafd220af89e75b8c2a400f4bdd78e/lib/formatting/formatters.js#L35-L37) with a hypen. This led to the formatted value not being seen as part of the allowed options and the field failing the [`equal`](https://github.com/UKHomeOffice/passports-form-wizard/blob/c8131187ddaafd220af89e75b8c2a400f4bdd78e/lib/validation/validators.js#L70-L78) validator which is applied by default to fields that contain options.

This also uncovered that there was no support for the `equal` validation code in the translation keys or that if a framework
field failed validation it would use `{key}.label` as the error message text. This meant extending the method to get the error
message to cater for fields that have come from the framework and that contain either `description` or `question`.

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/person-escort-record/router.js
+++ b/app/person-escort-record/router.js
@@ -30,6 +30,7 @@ function defineFormWizard(req, res, next) {
     name: `person-escort-record-${key}`,
     template: 'framework-step',
     templatePath: 'person-escort-record/views/',
+    defaultFormatters: ['trim', 'singlespaces', 'apostrophes', 'quotes'],
   }
 
   return wizard(wizardSteps, wizardFields, wizardConfig)(req, res, next)
--- a/common/helpers/field/get-field-error-message.js
+++ b/common/helpers/field/get-field-error-message.js
@@ -2,6 +2,8 @@ const i18n = require('../../../config/i18n')
 
 function getFieldErrorMessage({
   message,
+  description = '',
+  question = '',
   key: fieldKey,
   type: errorType,
 } = {}) {
@@ -13,12 +15,13 @@ function getFieldErrorMessage({
     return message
   }
 
-  const errorLabel = i18n.t(`fields::${fieldKey}.label`, {
+  const labelFallback = description || question
+  const errorLabel = i18n.t([`fields::${fieldKey}.label`, labelFallback], {
     context: 'with_error',
   })
   const fallback = i18n.t(`validation::${errorType}`, {
     context: 'with_label',
-    label: errorLabel.toLowerCase(),
+    label: errorLabel,
   })
   return i18n.t([`fields::${fieldKey}.error_message`, fallback], {
     context: errorType,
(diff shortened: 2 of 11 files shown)


merlinc (dismissed):

app/person-escort-record/router.js
```diff
@@ -30,6 +30,7 @@ function defineFormWizard(req, res, next) {
     name: `person-escort-record-${key}`,
     template: 'framework-step',
     templatePath: 'person-escort-record/views/',
+    // defaultFormatters: ['trim', 'singlespaces', 'apostrophes', 'quotes'],
```
solidgoldpig, line 33: ??

teneightfive: Yep. My bad 🤦

cwrw (approved):

common/helpers/field/get-field-error-message.js
```diff
@@ -13,12 +15,13 @@ function getFieldErrorMessage({
...
   }
 
-  const errorLabel = i18n.t(`fields::${fieldKey}.label`, {
+  const labelFallback = description || question
+  const errorLabel = i18n.t([`fields::${fieldKey}.label`, labelFallback], {
     context: 'with_error',
   })
   const fallback = i18n.t(`validation::${errorType}`, {
     context: 'with_label',
-    label: errorLabel.toLowerCase(),
```
merlinc, line 21: Are there any repercussions from removing the `toLowerCase` method ?

solidgoldpig (approved):

common/helpers/field/get-field-error-message.js
teneightfive: So I think the slight side-effect might now be `Enter Name of court` instead of `Enter name of court`, however, the lowercasing was having wider side-effects such as `Enter police national computer (pnc)` where we were lowercasing everything.

So to also not mess with the casing of labels coming from the framework I decided to remove it for now. We have the ability to tinker with error messages within the locales but I think that should be a separate task.

PR #800, opened 2020-09-21
Update dependencies

teneightfive:
```
@sentry/node                      5.22.3  →  5.24.2
govuk-frontend                     3.8.1  →   3.9.0
helmet                             4.1.0  →   4.1.1
hmpo-form-wizard                  11.7.1  →  11.8.0
markdown-it                       11.0.0  →  11.0.1
query-string                      6.13.1  →  6.13.2
yargs                             15.4.1  →  16.0.3
@babel/core                       7.11.5  →  7.11.6
@commitlint/cli                    9.1.2  →  11.0.0
@commitlint/config-conventional   10.0.0  →  11.0.0
copy-webpack-plugin                6.1.0  →   6.1.1
eslint                             7.8.0  →   7.9.0
husky                              4.2.5  →   4.3.0
lint-staged                      10.2.13  →  10.4.0
mini-css-extract-plugin           0.11.0  →  0.11.2
npm-check-updates                  8.0.3  →   9.0.1
prettier                           2.1.1  →   2.1.2
sass-loader                       10.0.1  →  10.0.2
terser-webpack-plugin              4.1.0  →   4.2.2
testcafe                           1.9.1  →   1.9.3
webpack                           4.44.1  →  4.44.2
```


package.json
```diff
@@ -97,18 +97,18 @@
     "express-session": "1.17.1",
     "filesize": "6.1.0",
     "form-data": "3.0.0",
-    "govuk-frontend": "3.8.1",
+    "govuk-frontend": "3.9.0",
```
solidgoldpig, line 100: I would do this as a separate update so that it can be marked as a `feat` change

package.json
```diff
@@ -75,7 +75,7 @@
     "@hmpps-book-secure-move-frameworks/1.1.0": "git+https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks.git#v1.1.0",
```
solidgoldpig, line 75: `ncu` throws an error for me when it's `git+https://` rather than `https://` but it obviously works ok for you

teneightfive: Ahhh, I did fix this, then apparently running the command overwrote that. 

I think the reason it no longer complains though is because I added it to the `reject` list. I needed to do that regardless because otherwise it will try and update the previous versions to all point to the latest, which would break our backwards compatibility.

package.json
teneightfive: Any particular reason? If we _always_ want to do that I can also add this package to the list of ignore updates for now?

solidgoldpig: I'll restate what I said ;)

I think upgrading `govuk-frontend` is worth a line in the release notes (just as `moj-frontend` would be). The other dependencies are not of interest to the business, but as `govuk-frontend` can change the look and feel and/or behaviour...

ie:  `feat(app): Upgrade govuk-frontend to 3.9.0`

solidgoldpig: Agreed to disagree offline - proceeding as is

solidgoldpig (approved):

PR #801, opened 2020-09-22
Reorganise api client middleware

solidgoldpig:
### What changed

- checks for cached result first - if cached result found skips following middleware
- uses `lru-cache` rather than simple closure
- suppresses devour logging in dev in favour of outputting on request via debug

### Why did it change

- simplifies request middleware
- prepares ground to add instrumentation

### Issue tracking

n/a

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/lib/api-client/middleware/request-timeout.js
```diff
@@ -2,6 +2,10 @@ module.exports = function requestTimeout(timeout) {
   return {
     name: 'request-timeout',
     req: payload => {
+      if (payload.res) {
+        return payload
+      }
+
       if (payload.req) {
```
solidgoldpig, line 9: Is there a reason why we check for the presence of the `req` object? Surely the `payload` will always have it if passed by `devour`?

common/lib/api-client/auth.js
```diff
@@ -49,9 +50,14 @@ Auth.prototype = {
       },
     }
 
+    debug('AUTH REQUEST', this.config.authUrl)
     return axios
       .post(this.config.authUrl, data, config)
       .then(response => response.data)
+      .catch(error => {
+        debug('AUTH ERROR', error)
```
teneightfive, line 58: Is there much benefit in us adding this extra debugging? In the case of an error, the express error handler would show that anyway, no?

common/lib/api-client/cache.js
```diff
@@ -0,0 +1,39 @@
...
+
+  if (typeof data === 'string') {
+    data = JSON.parse(data)
+  }
+
+  return data
+}
+
+async function set(key, response = {}, expiry, useRedisCache) {
+  const { data } = response
```
teneightfive, line 24: Any reason to pass in the whole response object?

This look like the `set` will return this format:

```json
{
  "data": {
    "foo": "bar"
  }
}
```

but `get` will return:

```json
{
  "foo": "bar"
}
```

Might be more consistent to pass the data expected to be set directly into `set` and then that structure will be exactly what's return from `get`.

common/lib/api-client/index.js
```diff
@@ -22,7 +24,7 @@ module.exports = function () {
 
   instance = new JsonApi({
     apiUrl: API.BASE_URL,
-    logger: IS_DEV,
+    logger: false,
```
teneightfive, line 27: Any reason for turning this off now, even for development?

common/lib/api-client/middleware/cache-key.js
```diff
@@ -0,0 +1,30 @@
+const { get } = require('lodash')
+
+const models = require('../models')
+
+function cacheKeyMiddleware({ apiVersion = 2 } = {}) {
```
teneightfive, line 5: Not sure it's a good idea to set the default API version here as well as in the configuration.

I think it would be better to not include a default here, and if we want one, it should be defined in the configuration file.

common/lib/api-client/middleware/cache-key.js
```diff
@@ -0,0 +1,30 @@
...
+
+      if (tryCache) {
+        const pathname = new URL(req.url).pathname
+
+        const searchString = new URLSearchParams(req.params).toString()
+        const key = `cache:v${apiVersion}:${req.method}.${pathname}${
+          searchString ? `?${searchString}` : ''
+        }`
+
+        payload.cacheKey = key
```
teneightfive, line 22: Would this be better placed on the request object seeing as it relates to the request - `payload.req.cacheKey`?

common/lib/api-client/middleware/get-cache.js
```diff
@@ -0,0 +1,30 @@
+const debug = require('debug')('app:api-client:get-cache')
+
+const cache = require('../cache')
+
+function getCacheMiddleware({ useRedisCache = false } = {}) {
```
teneightfive, line 5: Same with this default. I think if we want a _default_ that value should live in the configuration and be passed into this function.

common/lib/api-client/middleware/cache-key.js
solidgoldpig: I don't see that it makes much odds but I think it's better on the payload

common/lib/api-client/middleware/request-headers.js
```diff
@@ -1,8 +1,12 @@
 const getRequestHeaders = require('../request-headers')
 
 module.exports = {
-  name: 'req-headers',
+  name: 'request-headers',
   req: function req(payload = {}) {
+    if (payload.res) {
```
teneightfive, line 6: This might be nicer to combine with the following lines:

```javascript
const { req, res } = payload

if (!req || res) {
  return payload
}

...
```

common/lib/api-client/cache.js
solidgoldpig: 👍

common/lib/api-client/middleware/request-include.js
```diff
@@ -1,8 +1,12 @@
 const { get } = require('lodash')
 
 module.exports = {
-  name: 'req-include',
+  name: 'request-include',
   req: function req(payload = {}) {
+    if (payload.res) {
```
teneightfive, line 6: Same suggestion here about combining the conditions and early returns.

common/lib/api-client/auth.js
solidgoldpig: Not in the case of anything that we don't rethrow to the express error handler - eg. the 404s and 422s for the person images.

I think this is a massive improvement since by default we do not drown in devour messages - only if we explicitly ask for them. (Whereas before the choice was to pretend it wasn't DEV or hack devour to exclude certain cases). I have already found this much more useful and agreeable

(NB. the catch block will also be used for the instrumentation to record the failure)

common/lib/api-client/index.js
solidgoldpig: See https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/801#discussion_r492631417

common/lib/api-client/middleware/cache-key.js
solidgoldpig: 👍

solidgoldpig: Should it throw if not passed a version?

common/lib/api-client/middleware/cache-key.js
solidgoldpig: As per https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/801#discussion_r492632004

Should it throw if not passed a defined value?

common/lib/api-client/middleware/get-cache.js
solidgoldpig: As per https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/801#discussion_r492632004

Should it throw if not passed a defined value?

common/lib/api-client/middleware/request-headers.js
solidgoldpig: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/801#discussion_r492511072

I think it would be better to remove the unnecessary `req` check ;)

common/lib/api-client/middleware/request-include.js
solidgoldpig: and same answer

common/lib/api-client/index.js
merlinc: Offtopic, so don't want to derail this PR - but `devour` using its own internal logger will get in the way of any json logging approach we'll want to do in the future.

common/lib/api-client/middleware/cache-key.js
merlinc: Would that end up being an error on startup? That would be more clear than a silent failure and continuing.

common/lib/api-client/middleware/request-timeout.js
teneightfive: Not off the top of my head, but I'd be inclined to say there was a reason.

merlinc: @solidgoldpig Figuring this out might be the solution to the req check vs return payload question! :)

solidgoldpig: `Devour` always has a 'req` object on the `payload`. The only reason it wouldn't be there is if we explicitly deleted it.

I think it's overeager guard. Please, someone show me otherwise

NB. the example at https://github.com/twg/devour#your-first-middleware does not bother with the check

common/lib/api-client/index.js
solidgoldpig: exactly

common/lib/api-client/middleware/cache-key.js
solidgoldpig: Good point

teneightfive: >Should it throw if not passed a version?

I doubt this method should care. It's just to build a key, right? If not version, it just doesn't include that within the key.

And as @merlinc suggests, if our app **needs** a version, we should throw an error on startup, or set a default in the config.

common/lib/api-client/middleware/cache-key.js
teneightfive: It just feels more closely aligned to the request than the payload in this instance. Like `req.url`, `req.params`, `req.model`....the cache key is more tied to the request in this instance than the payload.

common/lib/api-client/middleware/get-cache.js
teneightfive: Same response here I'd say. I don't think this function needs to be that aggressive. If we want a default it should be set in the config and then there is no case of it not having a value.

The reason for passing that into this function here is to make it a little more pure so that it doesn't need to reach into the config, so I think it's okay to not throw in this instance.

common/lib/api-client/middleware/request-headers.js
teneightfive: It feels cleaner and easier to understand to check for a request if we're setting things on the request, no? 

Rather than assuming that the inclusion of a  response means it is a request.

solidgoldpig: > Rather than assuming that the inclusion of a response means it is a request.
I don't understand this comment

The req object is there is pass the request args to jsonApi.axios. In the case of the cached value being returned, we don't want to invoke  jsonApi.axios.

The response handler looks at payload.res. Therefore the cache middleware return the payload with the res property attached. The other middleware know not to bother doing anything and just return the payload as is if that property is present. The payload.res checks are valid. The payload.req ones are pointless.

common/lib/api-client/middleware/get-cache.js
solidgoldpig: in which case there's little difference between `{ useRedisCache = false }` and `{ useRedisCache }` other than necessitating an extra test to check that undefined behaves the same as false

common/lib/api-client/middleware/cache-key.js
solidgoldpig: `req` is what get passed to `axios`. The `cacheKey` has nothing to do with it. We are proper 🚲  🏠 ing here

teneightfive (approved):

PR #802, opened 2020-09-22
refactor: Refactor moves to use api cancel method

merlinc:
Cancelling moves and cancelling allocations were developed at different times by different people. This PR brings them both up to date with the current approach for making requests.

## Proposed changes

- Refactor services.move to use api cancel rather than api update
- Explicitly pass reason and comment parameters to service

### Why did it change

Code that performs the same task, but is approached differently, makes further work more complicated than it needs to be.

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/services/allocation.js
+++ b/common/services/allocation.js
@@ -52,11 +52,15 @@ const allocationService = {
 
     const timestamp = dateFunctions.formatISO(new Date())
 
-    return apiClient.one('allocation', id).all('cancel').post({
-      timestamp,
-      cancellation_reason: reason,
-      cancellation_reason_comment: comment,
-    })
+    return apiClient
+      .one('allocation', id)
+      .all('cancel')
+      .post({
+        timestamp,
+        cancellation_reason: reason,
+        cancellation_reason_comment: comment,
+      })
+      .then(response => response.data)
   },
 
   format(data) {
--- a/common/services/allocation.js
+++ b/common/services/allocation.js
@@ -45,21 +45,20 @@ function getAll({
 }
 
 const allocationService = {
-  cancel(id, data) {
+  cancel(id, { reason, comment } = {}) {
     if (!id) {
       return Promise.reject(new Error('No allocation id supplied'))
     }
 
     const timestamp = dateFunctions.formatISO(new Date())
 
-    return apiClient
-      .one('allocation', id)
-      .all('cancel')
-      .post({
-        timestamp,
-        ...data,
-      })
+    return apiClient.one('allocation', id).all('cancel').post({
+      timestamp,
+      cancellation_reason: reason,
+      cancellation_reason_comment: comment,
+    })
   },
+
   format(data) {
     const booleansAndNulls = ['complete_in_full', 'sentence_length']
     const relationships = ['to_location', 'from_location']
--- a/common/services/allocation.test.js
+++ b/common/services/allocation.test.js
@@ -3,6 +3,7 @@ const proxyquire = require('proxyquire')
 const apiClient = require('../lib/api-client')()
 const moveService = require('../services/move')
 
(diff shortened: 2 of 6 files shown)


common/services/allocation.js
```diff
@@ -45,21 +45,20 @@ function getAll({
...
     const timestamp = dateFunctions.formatISO(new Date())
 
-    return apiClient
-      .one('allocation', id)
-      .all('cancel')
-      .post({
-        timestamp,
-        ...data,
-      })
+    return apiClient.one('allocation', id).all('cancel').post({
```
merlinc, line 55: Oh `prettier`, you so funny sometimes :)

common/services/allocation.test.js
```diff
@@ -78,6 +82,26 @@ const mockAllocations = [
 
 describe('Allocation service', function () {
   describe('cancel', function () {
+    context.skip('without data args', function () {
```
merlinc, line 85: The two `without data args` tests now fail in a strange way, so have been skipped.

Are these providing value, and if so, what should the response be for a invocation missing parameters?

```
 1) Allocation service
       cancel
         without data args
           "before each" hook for "should call update method with data":
     Error: Not Found
      at error (common/lib/api-client/middleware/errors.js:13:19)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)

  2) Move Service
       #cancel()
         with move ID
           without data args
             should call update method with data:
     AssertionError: expected post to have been called exactly once with exact arguments {
  cancellation_reason: undefined,
  cancellation_reason_comment: undefined,
  id: "b695d0f0-af8e-4b97-891e-92020d6820b9",
  timestamp: "#timestamp"
}
{
  cancellation_reason: undefined,
  cancellation_reason_comment: undefined,
  id: "b695d0f0-af8e-4b97-891e-92020d6820b9",
  timestamp: "#timestamp"
}
      at Context.<anonymous> (common/services/move.test.js:1056:40)
```

common/services/move.test.js
```diff
@@ -1039,53 +1039,53 @@ describe('Move Service', function () {
...
+        it('should call the service to post a cancel', function () {
+          expect(apiClient.post).to.be.calledOnceWithExactly({
             cancellation_reason: 'other',
             cancellation_reason_comment: 'Reason for cancelling',
+            timestamp: sinon.match.string,
           })
         })
 
         it('should return move', function () {
-          expect(move).to.deep.equal(mockResponse.data)
```
merlinc, line 1088: This should be restored to `mockResponse.data` as all the other services unpack the response with an equivalent of `      .then(response => response.data)`

common/services/allocation.test.js
teneightfive: Maybe they can be removed?

common/services/move.test.js
```diff
@@ -1039,24 +1039,23 @@ describe('Move Service', function () {
...
         data: {
           ...mockMove,
-          status: 'cancelled',
         },
       }
       let move
 
       beforeEach(async function () {
-        sinon.stub(apiClient, 'update').resolves(mockResponse)
+        sinon.stub(apiClient, 'post').resolves(mockResponse)
```
teneightfive, line 1047: I'm not sure this will work because the `post` method isn't being called directly. You may have to stub `one` and `all` like the allocations test file.

teneightfive (approved):

common/services/move.test.js
merlinc: It was being set in a different `forEach` a few lines below, but that was obviously unclear, so I've moved those lines next to this one

common/services/allocation.test.js
merlinc: Done! :)

PR #812, opened 2020-09-23
[P4-1888] Display NOMIS information alongside framework questions

teneightfive:
### What changed

This changes the way the questions are rendered for the framework to include any NOMIS information if the response returned from the API has NOMIS mappings.

### Why did it change

A large number of users within the prison service use NOMIS information to complete the Person Escort Record. By gaining access the NOMIS via their API and importing information related to that person users will not have to manually enter that information into our system.

### Issue tracking

- [P4-1888](https://dsdmoj.atlassian.net/browse/P4-1888)

## Screenshots

<kbd></kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/person-escort-record/app/new/controllers.js
```diff
@@ -22,7 +22,7 @@ class NewPersonEscortRecordController extends FormWizardController {
 
   async saveValues(req, res, next) {
     try {
-      req.record = await personEscortRecordService.create(req.move.profile.id)
+      req.record = await personEscortRecordService.create(req.move.id)
```
merlinc, line 25: This is the only current usage of `move.profile.id`. There's already a lot of usage of `move.id` 

I'm wondering, there's other usage of `move.profile` throughout the system - do any of these need to be updated to use data off the `move` as opposed to the `profile` ?

merlinc (commented):
I'm not seeing anything that jumps out at me. 

And obviously it's only a draft, so the missing tests will appear as it evolves.

app/person-escort-record/app/new/controllers.js
teneightfive: I'm pretty sure any existing implementations using move ID and profile ID should be okay left as is.

This change is quite specifically related to the way Person Escort Records now need to be created via the API.

Previously, the API expected the profile ID, but, with the new requirement to import information from NOMIS the API needs to know the origin type (if a Prison, import from NOMIS). However, we have a current business rule that will allow profiles to be used across multiple moves, so the API would not be able to determine that if the client application only send the profile ID.

This change essentially unlocks the API to be able to determine its logical correctly. This will be the final piece that will allow the API to start pulling information from NOMIS and I think I'm going to extract this change so that the other functionality can be merged separately.

merlinc: > extract this change so that the other functionality can be merged separately.

Good call - I was thinking of suggesting the same - that way if there are any side effects, it'll be easier to see what caused it.

common/helpers/frameworks/render-nomis-mappings-to-field.test.js
```diff
@@ -0,0 +1,125 @@
...
+        output = helper([mockResponse])(mockField)
+      })
+
+      it('should return original field', function () {
+        expect(output).to.deep.equal(mockField)
+      })
+    })
+
+    context('with NOMIS mappings', function () {
+      const mockResponseWithMappings = {
```
merlinc, line 50: I've had weird runtime issues in the past with setting variables inside a `describe/context` block but outside of a `before/beforeEach/it` block, so now I declare a variable outside, and use a `before/beforeEach` to set it - even if they're not going to change.

teneightfive: Hmm, that's interesting. I've not experienced that before with defining strings/objects but have had that when defining stubs that might need resetting. 

I think we also use this approach in quite a few places within our test suites. Have you had any of those issues running the tests in this project yet? If it's cropped up it might be good to look into to see why it's happening.

I'm happy to change this up, but also wary that we do use this approach in quite a few other places so if it's something we should be doing, maybe we should do it as a global change and then we can also explain the reasons behind it alongside the change?

merlinc: No worries about doing it this change, I guess its now more of an fyi.

What happens is that mocha runs all the code outside of the before blocks while starting up the tests, which can cause other unrelated tests to fail. It's an edge case, but Ive seen it in several projects - haven't seen it causing issues here yet.

Happy to ignore it for now.

merlinc (approved):

PR #823, opened 2020-09-25
Progressively enhance add another

solidgoldpig:
### What changed

Adding and removing items not happens on the client-side (if supported) using the ’Add Another’ pattern from the MoJ Design System/Frontend package

### Why did it change

It... "can be more jarring for a user if they are adding many items as they need to wait for the page to reload after each addition."

### Issue tracking

- [P4 2076 - Enhance "Add another" with JavaScript component from MOJ Design System](https://dsdmoj.atlassian.net/browse/P4-2076)

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/person-escort-record/controllers/framework-step.js
+++ b/app/person-escort-record/controllers/framework-step.js
@@ -93,10 +93,6 @@ class FrameworkStepController extends FormWizardController {
       // call parent saveValues to handle storing new values in the session
       super.saveValues(req, res, next)
     } catch (error) {
-      if (req.body['js-items-length']) {
-        return res.redirect(req.originalUrl)
-      }
-
       next(error)
     }
   }
--- a/app/person-escort-record/controllers/framework-step.js
+++ b/app/person-escort-record/controllers/framework-step.js
@@ -93,6 +93,10 @@ class FrameworkStepController extends FormWizardController {
       // call parent saveValues to handle storing new values in the session
       super.saveValues(req, res, next)
     } catch (error) {
+      if (req.body['js-items-length']) {
+        return res.redirect(req.originalUrl)
+      }
+
       next(error)
     }
   }
--- a/common/assets/javascripts/application.js
+++ b/common/assets/javascripts/application.js
@@ -21,6 +21,7 @@ const accessibleAutocomplete = require('accessible-autocomplete')
 const { initAll } = require('govuk-frontend')
 const StickySidebar = require('sticky-sidebar/dist/sticky-sidebar')
 
+const AddAnother = require('../../components/add-another/add-another')
 const Header = require('../../components/internal-header/internal-header')
 const Message = require('../../components/message/message')
 const MultiFileUpload = require('../../components/multi-file-upload/multi-file-upload')
@@ -29,6 +30,9 @@ const Analytics = require('./analytics')
 const { nodeListForEach } = require('./utils')
 
 initAll()
+
+// eslint-disable-next-line no-new
+new AddAnother('.moj-add-another')
 
 new Analytics().init()
 
--- a/common/components/add-another/add-another.js
+++ b/common/components/add-another/add-another.js
@@ -1,31 +1,11 @@
 const MOJFrontend = require('@ministryofjustice/frontend/moj/all')
 const { initAll } = require('govuk-frontend')
 
-MOJFrontend.AddAnother.prototype.setItemsLength = function () {
-  let itemsLength = this.container.find('[name="js-items-length"')
-
-  if (!itemsLength.length) {
-    this.container.append('<input type="hidden" name="js-items-length">')
-    itemsLength = this.container.find('[name="js-items-length"')
-  }
-
-  itemsLength.val(this.getItems().length)
-}
-
 const _onAddButtonClick = MOJFrontend.AddAnother.prototype.onAddButtonClick
 
 MOJFrontend.AddAnother.prototype.onAddButtonClick = function (e) {
   _onAddButtonClick.apply(this, [e])
-  this.setItemsLength()
   initAll()
-}
-
-const _onRemoveButtonClick =
-  MOJFrontend.AddAnother.prototype.onRemoveButtonClick
-
-MOJFrontend.AddAnother.prototype.onRemoveButtonClick = function (e) {
-  _onRemoveButtonClick.apply(this, [e])
-  this.setItemsLength()
 }
 
 MOJFrontend.AddAnother.prototype.updateAttributes = function (index, item) {
--- a/common/components/add-another/add-another.js
+++ b/common/components/add-another/add-another.js
@@ -14,7 +14,7 @@ MOJFrontend.AddAnother.prototype.updateAttributes = function (index, item) {
     el.name = $el.attr('data-name').replace(/%index%/, index)
     el.id = $el.attr('data-id').replace(/%index%/, index)
 
-    // BASM changes to this method from here
+    // Book a secyre move changes to this method from here
     var label = $el.siblings('label')[0] || $el.parents('label')[0]
 
     if (label) {
--- a/common/components/add-another/add-another.js
+++ b/common/components/add-another/add-another.js
@@ -2,10 +2,17 @@ const MOJFrontend = require('@ministryofjustice/frontend/moj/all')
 const { initAll } = require('govuk-frontend')
 
 const _onAddButtonClick = MOJFrontend.AddAnother.prototype.onAddButtonClick
+const _onRemoveButtonClick =
+  MOJFrontend.AddAnother.prototype.onRemoveButtonClick
 
 MOJFrontend.AddAnother.prototype.onAddButtonClick = function (e) {
   _onAddButtonClick.apply(this, [e])
   initAll()
+}
+
+MOJFrontend.AddAnother.prototype.onRemoveButtonClick = function (e) {
+  $('.govuk-error-summary').remove()
+  _onRemoveButtonClick.apply(this, [e])
 }
 
 MOJFrontend.AddAnother.prototype.updateAttributes = function (index, item) {
@@ -14,7 +21,7 @@ MOJFrontend.AddAnother.prototype.updateAttributes = function (index, item) {
     el.name = $el.attr('data-name').replace(/%index%/, index)
     el.id = $el.attr('data-id').replace(/%index%/, index)
 
-    // Book a secyre move changes to this method from here
+    // Book a secure move changes to this method from here
     var label = $el.siblings('label')[0] || $el.parents('label')[0]
 
     if (label) {
--- a/common/components/add-another/add-another.js
+++ b/common/components/add-another/add-another.js
@@ -1,0 +1,66 @@
+const MOJFrontend = require('@ministryofjustice/frontend/moj/all')
+const { initAll } = require('govuk-frontend')
+
+MOJFrontend.AddAnother.prototype.setItemsLength = function () {
+  let itemsLength = this.container.find('[name="js-items-length"')
+
+  if (!itemsLength.length) {
+    this.container.append('<input type="hidden" name="js-items-length">')
+    itemsLength = this.container.find('[name="js-items-length"')
+  }
+
+  itemsLength.val(this.getItems().length)
+}
+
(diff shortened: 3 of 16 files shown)


common/assets/javascripts/application.js
```diff
@@ -17,19 +17,150 @@ import '../images/tick-icon@2x.png'
...
 
+const AddAnother = require('../../components/add-another/add-another')
 const Header = require('../../components/internal-header/internal-header')
 const Message = require('../../components/message/message')
 const MultiFileUpload = require('../../components/multi-file-upload/multi-file-upload')
 
 const Analytics = require('./analytics')
 const { nodeListForEach } = require('./utils')
 
+// // MOJFrontend.AddAnother = function (container) {
```
solidgoldpig, line 36: GAH!

app/person-escort-record/controllers/framework-step.js
```diff
@@ -93,6 +93,10 @@ class FrameworkStepController extends FormWizardController {
       // call parent saveValues to handle storing new values in the session
       super.saveValues(req, res, next)
     } catch (error) {
+      if (req.body['js-items-length']) {
```
teneightfive, line 96: What is the error instance when this occurs? It feels a little odd to use a condition that checks for something in the request body to determine whether to ignore the error or not.

Would it not be better to maybe at least extend the condition for the error instance in which it occurs?

teneightfive: Also, redirecting straight away will skip saving the values to the session, is that the desired effect?

common/controllers/form-wizard.js
```diff
@@ -17,14 +18,37 @@ class FormController extends Controller {
     next()
   }
 
+  setupArbitraryMultipleFields(req, res, next) {
```
teneightfive, line 21: I wonder if we can find a better name for this method...

> **arbitrary**
> based on random choice or personal whim, rather than any reason or system.

☝️ that use of this doesn't really fill me with confidence in this method...

common/controllers/form-wizard.js
```diff
@@ -17,14 +18,37 @@ class FormController extends Controller {
...
+    if (req.method === 'POST') {
```
teneightfive, line 22: If this is the case, is this not something that should happen in the [`post`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/7bfa5edb1af9f2b02cbe3ffcdc4b6e91ec9d9027/common/controllers/form-wizard.js#L55) method. This seems exactly it's purpose.

common/controllers/form-wizard.js
```diff
@@ -17,14 +18,37 @@ class FormController extends Controller {
...
+
+        // add items
+        while (values[jsItemsProperty].length < itemsLength) {
+          values[jsItemsProperty].push({})
+        }
+
+        // remove items
+        values[jsItemsProperty].length = itemsLength
+
+        req.sessionModel.set(values)
```
teneightfive, line 38: This does also feel like it's doing similar things to what the non-js version does. Is there a way we could find a way to neaten it and share the logic rather than requiring extra JS-only values to be sent in the form?

common/templates/layouts/base.njk
```diff
@@ -145,5 +145,6 @@
 {% endblock %}
 
 {% block bodyEnd %}
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
```
teneightfive, line 148: 😬 How much does this bump up our page size?

server.js
```diff
@@ -170,6 +170,7 @@ app.use(
         scriptSrc: [
           "'self'",
           "'unsafe-inline'",
+          'cdnjs.cloudflare.com',
```
teneightfive, line 173: Wonder if it would be better to include this manually using npm and webpack?

teneightfive:
Also found a scenario which isn't fully working as I'd expect:

- Add another empty bag
- Submit form without any fields filled in

<kbd>image</kbd>

It seems to not register the fields for the second bag and looks a lot like some issues I had with the no-JS version.

It's really got me thinking there must be a way to re-use some of the way things work for the non-JS version. Is there a way that we can ensure the number of items is set correctly when it's a JS submission?

I feel we're doing a lot of heavy lifting for what should essentially still be treated as a form submission. And maybe the only thing that needs to be supported is that multiple items are submitted at once rather than the session needing to add empty items when it's a non-JS submission.

teneightfive:
The JS version also seems to remove the validation from a field if you interact with the "Add" and "Remove" actions too. However, the error summary still remains. 

<kbd>validation</kbd>

<kbd>validation-removal</kbd>

app/person-escort-record/controllers/framework-step.js
solidgoldpig: If a user ends up with fewer fields than expected or more fields than expected but all the info filled in with no validation errors, everything just works.

This covers the case where a user has added more fields than expected and then does not correctly fill out those fields (or invalidates an existing field). The PER record is rejected with a 422 unprocessable error, but if the user is simply redirected back to the original page as a get, the PER has now been updated to have the correct number of fields... and so can display the expected error messages (which it cannot if the user is simply presented with a 422 error) - the values do get saved to the session and do persist

- Yes, it should indeed be added in to not do the same for unexpected errors (when the js adds are in effect, that is)
- Yes, it would be nice/better to fix the PER BE to actually handle this, but...
- Yes, it's odd.

I did also try to actually add and remove the fields as ajax-style calls when the user clicks the buttons but that gets way more complicated

1. CSRF tokens
2. Identifying all the correct form fields and their values that need to be sent
3. Handling the cases where the items fail to get added or removed by the backend
4. Handling the cases where we've failed to keep the client and server version in sync

But I may have missed a trick there

common/controllers/form-wizard.js
solidgoldpig: It's simply too late to do this in the `post` middleware.

It 
a) needs to happen before form wizard proper gets its mitts on the fields
b) even more crucially needs to happen before setting up the multiple and conditional fields

common/controllers/form-wizard.js
solidgoldpig: Suggestions on a postcard, please

common/templates/layouts/base.njk
solidgoldpig: 70k I would guess.

"Happy" to take another day rewriting the MoJ Design System to not rely on jquery. Unfortunately using a lighter-weight api-compliant alternative is not an option because of the heavy usage of `$.proxy

server.js
solidgoldpig: Personally I don't think so. But, that is just an opinion and happy for whatever consensus is arrived at.

solidgoldpig:
> It seems to not register the fields for the second bag and looks a lot like some issues I had with the no-JS version.

Well, exactly. There are issues with the no-JS version. I'm not quite sure what magic the frontend is supposed to do to make those go away.

> It's really got me thinking there must be a way to re-use some of the way things work for the non-JS version. Is there a way that we can ensure the number of items is set correctly when it's a JS submission?

I looked into trying to update the PER response itself... but flailed

> I feel we're doing a lot of heavy lifting for what should essentially still be treated as a form submission. And maybe the only thing that needs to be supported is that multiple items are submitted at once rather than the session needing to add empty items when it's a non-JS submission.

yes, it would be nice if you could just say "Hey PER, here's a bunch of answers" ;)

common/controllers/form-wizard.js
solidgoldpig: Arbitrary in that it depends on the personal whim of the user.

`setupExplicitNumberOfMultipleFields`?

solidgoldpig:
> Also found a scenario which isn't fully working as I'd expect:

Yay! I was trying to find things that didn't work when I was showing it to Jade (but couldn't)

app/person-escort-record/controllers/framework-step.js
solidgoldpig: This must actually have been some side-effect of bumpiness in the backend. I don't see the 422s any more and there's no need to do anything special

common/controllers/form-wizard.js
solidgoldpig: as it no longer applies just to `POST`s and simply sets the values... it's now `setupAddMultipleFieldsValues`

common/controllers/form-wizard.js
solidgoldpig: it's positioning is still the same but it no longer applies just to `POST`s

common/controllers/form-wizard.js
solidgoldpig: Considerably rewritten  - I think it's simpler and it certainly no longer requires the js-only values

solidgoldpig:
@teneightfive Can you try and break it again please?

common/components/add-another/template.njk
```diff
@@ -9,40 +9,44 @@
 {% set itemName = params.itemName or t("item") %}
 {% set items = params.items %}
 
+<div class="moj-add-another">
```
teneightfive, line 12: I might just be GH's diffs but I can't see a closing tag for this?

common/controllers/form-wizard.js
```diff
@@ -17,14 +18,32 @@ class FormController extends Controller {
     next()
   }
 
+  setupAddMultipleFieldsValues(req, res, next) {
+    const values = req.sessionModel.toJSON()
+
+    const addAnotherProps = Object.keys(req.form.options.fields).filter(
```
teneightfive, line 24: You could destructure/assign `req.form.options.fields` as you use it a few times in this method?

common/controllers/form-wizard.js
```diff
@@ -17,14 +18,32 @@ class FormController extends Controller {
...
+  }
+
   setupAddMultipleFields(req, res, next) {
     const allFields = req.form.options.allFields
     const values = req.sessionModel.toJSON()
     const itemFields = Object.entries(req.form.options.fields).reduce(
       fieldHelpers.reduceAddAnotherFields(allFields, values),
       {}
     )
-
```
teneightfive, line 27: Any reason for this? Might be nice to keep the line break between blocks for readability.

common/helpers/field/reduce-add-another-fields.js
```diff
@@ -18,13 +18,35 @@ function reduceAddAnotherFields(allFields = {}, values = {}) {
...
+        const dataName = `${field.name}[%index%][${itemField}]`
+        const dataId = dataName.replace(/\[|\]/gi, '-')
         const opts = {
           // tell form wizard to not render field at top level
           skip: true,
           idPrefix: id,
           prefix,
           name,
           id,
+          ...(fieldTemplate.component !== 'govukCheckboxes' && {
```
teneightfive, line 30: Is the checkboxes component the only one we want to exclude? Not sure I fully understand the purpose of this condition?

common/components/add-another/add-another.js
```diff
@@ -0,0 +1,46 @@
...
+  initAll()
+}
+
+MOJFrontend.AddAnother.prototype.updateAttributes = function (index, item) {
+  item.find('[data-name]').each(function (i, el) {
+    var $el = $(el)
+    el.name = $el.attr('data-name').replace(/%index%/, index)
+    el.id = $el.attr('data-id').replace(/%index%/, index)
+
+    // BASM changes to this method from here
```
teneightfive, line 17: 🚨 acronym 🚨  😭

teneightfive (approved):
Nice work on taking another look at this. Have tried the previous scenarios that failed and seems to be working much more as expected now.

Have left a few comments, let me know your thoughts.

common/components/add-another/template.njk
solidgoldpig: Nope - I missed it in the various rebase/conflict dances

common/controllers/form-wizard.js
solidgoldpig: No more or less than the other methods in the `form-wizard` module. ;)

common/controllers/form-wizard.js
solidgoldpig: Reinstated

Issue #712, opened 2020-07-31
Upgrade dependencies with breaking changes


solidgoldpig:
The following outdated dependencies are stuck at their current version

```
 css-loader     3.6.0  →  4.x
 webpack-merge  4.2.2  →  5.x
```

- `css-loader`
  build fails (see error below)
- `webpack-merge`
  `smart` method has been deprecated

```sh
ERROR in ./common/assets/scss/application-ie8.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/css-loader/dist/cjs.js):
Error: Can't resolve '/assets/images/govuk-crest.png' in '<PATH>/hmpps-book-secure-move-frontend/common/assets/scss'
```

PR #825, opened 2020-09-29
Create dependabot.yml

merlinc:
## Proposed changes

Add dependabot support to manage our dependency updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    ignore:
+      - dependency-name: '@hmpps-book-secure-move-frameworks*'
+      - dependency-name: 'govuk-frontend'
+      - dependency-name: 'moj-frontend'
+      - dependency-name: 'css-loader'
+        versions: ['>=4']
+      - dependency-name: 'webpack-merge'
+        versions: ['>=5']
+    commit-message:
+      prefix: 'build'
+      include: 'scope'
+  - package-ecosystem: 'docker'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+      day: 'sunday'


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
merlinc, line 11: This will run daily, at 5am. Once there are 5 update PRs in play, no more will be created until one is merged or closed.

.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+    ignore:
+      - dependency-name: "govuk-frontend"
+    commit-message:
+      prefix: "fix"
+      prefix_development: "chore"
+      include: "scope"
+- package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
merlinc, line 21: How often do we want to update docker?

.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "govuk-frontend"
+    commit-message:
+      prefix: "fix"
+      prefix_development: "chore"
+      include: "scope"
```
merlinc, line 17: This will give `devDependencies` a commit with `chore(list of deps)` and production `dependencies` a commit with `fix(list of deps)`

Is this what we would like?

merlinc:
@merlinc The commit message needs to be fixed to match our standards

.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "govuk-frontend"
```
teneightfive, line 13: I think @solidgoldpig wanted us to do the same for `moj-frontend`.

And there are [two packages](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/7bfa5edb1af9f2b02cbe3ffcdc4b6e91ec9d9027/.ncurc.json#L2) that cause failures that we might also want to ignore for now:
- `css-loader`
- `webpack-merge`

.github/dependabot.yml
teneightfive: It feels like they are both a `chore`, unless of course increasing the version _is_ fixing a bug?

solidgoldpig: Are they not.... `build`? 

I don't like adding the dependencies to the change log. Except for the govuk and moj frontend packages (and someone disagreed with me on that).

Oh, we're ignoring them. Ignore this comment then. Except for the build bit ;)

.github/dependabot.yml
solidgoldpig: We'll have to ignore them for now if we don't want the app to be broken.

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/712

.github/dependabot.yml
solidgoldpig: 5 update PRs created by dependabot or in total?

Is it possible to have all the updates as one PR?

merlinc: 5 update PRs created by dependabot.

Looking at their docs, it doesn't seem like they can batch the updates - which is a shame, as I know that other similar things, like Renovate, have that ability.

.github/dependabot.yml
teneightfive: Yeah, `build` also makes sense.

.github/dependabot.yml
teneightfive: Yep, that was my suggestion for now. We add them to the ignore list as well.

.github/dependabot.yml
merlinc: This is now configured to be in the format of `build(express)`

.github/dependabot.yml
merlinc: I've set the `css-loader` and `webpack-merge` to be below the next version that's a breaking change. This should mean that any patch updates will come through.

`moj-frontend` and `govuk-frontend` are both set to completely ignore.

solidgoldpig (approved):

.github/dependabot.yml
merlinc: Also added `@hmpps-book-secure-move-frameworks*` to the list, as that is also an intentional upgrade we'd need to perform

.github/dependabot.yml
merlinc: Let's see what weekly looks like for now

PR #847, opened 2020-10-07
fix: Allow prison number to be case insensitive

teneightfive:
### What changed

This relaxes the validation to allow lowercase letters to be used.

It also updates the hint text to use quotes instead of primes:
https://carsonparkdesign.com/quotation-marks-apostrophes-versus-primes/

### Why did it change

The API supports case insensitive search on prison number so we
shouldn't force uppercase validation on the frontend.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/locales/en/fields.json
+++ b/locales/en/fields.json
@@ -9,7 +9,7 @@
     },
     "prison_number": {
       "label": "Prison number",
-      "hint": "For example, 'A1234AB'"
+      "hint": "For example, ‘A1234AB’"
     }
   },
   "people": {
--- a/app/move/validators.js
+++ b/app/move/validators.js
@@ -30,7 +30,7 @@ module.exports = {
   prisonNumber(value) {
     return (
       value === '' ||
-      Controller.validators.regex(value, /^[A-Z][0-9]{4}[A-Z]{2}$/)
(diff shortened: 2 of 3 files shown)


merlinc (approved):

locales/en/fields.json
```diff
@@ -9,7 +9,7 @@
     },
     "prison_number": {
       "label": "Prison number",
-      "hint": "For example, 'A1234AB'"
+      "hint": "For example, ‘A1234AB’"
```
merlinc, line 12: Which sort of quotes should be used inside a string?

merlinc: *reads the description*

Ah! Do we do this everywhere else?

teneightfive: Yes, it's something I've always tried to do. Bit of a personal quest after a content designer I worked with pointed out the difference to me. But I don't think we have any form of linting to enforce it so it's usually just something I look out for in PR reviews. 

GOV.UK always tries to do it well and you'll see examples in some of [their guides](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#single-quotes) too. A common one I always try and make sure is correct is to use `don’t` (an apostrophe) instead of `don't` (a prime).

PR #870, opened 2020-10-13
docs: Add docs on module usage

merlinc:
## Proposed changes

Add some documentation on how the modules are configured and used.

#### Automated testing

- ~[ ] Added unit tests to cover changes~
- ~[ ] Added end-to-end tests to cover any journey changes~

#### Manual testing

~Has been tested in the following browsers:~

~- [ ] Chrome~
~- [ ] Firefox~
~- [ ] Edge~
~- [ ] Internet Explorer~

### Environment variables

### Other considerations

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -1,6 +1,6 @@
 # App modules
 
-App modules are subsections of the main site, dealing with a specific area , e.g locations or moves. They are constructed using Express routers.
+App modules are subsections of the main site, dealing with a specific area , e.g locations or moves. They are constructed using [Express](https://expressjs.com) [routers](https://expressjs.com/en/api.html#router), [param middleware](https://expressjs.com/en/api.html#router.param) and [shared middleware](https://expressjs.com/en/api.html#router.use).
 
 They can have a variety of different parts, in either a root file, or a folder with multiple files, and exposed via an `index.js` in the folder.
 
--- a/app/README.md
+++ b/app/README.md
@@ -1,0 +1,105 @@
+# App modules
+
+App modules are subsections of the main site, dealing with a specific area , e.g locations or moves. They are constructed using Express routers.
+
+They can have a variety of different parts, in either a root file, or a folder with multiple files, and exposed via an `index.js` in the folder.
+
+- controllers
+  - Express
+- constants
+  - Predefined default settings for a module, including routing paths, actions, and other defaults
+- fields
+  - Form-wizard field definitions
+- middleware
+  - Express middleware for the module, rendering against the views
+- steps
+  - Form-wizard step definitions
+- validators
+  - Form-wizard field validator functions
+- views
+  - Nunjucks templates
+- index (routing)
+  - Main entrypoint for the module, tying together all the parts of the module in the form of an Express router
+
+## Index / Routing Setup
+
+### params middleware
+
+- set req vars
+  - **setDateRange**
+    - req.period
+    - req.dateRange
+- redirect
+
+### shared middleware
+
+- **saveUrl**
+
+### routing middleware
+
+#### types
+
+- redirects
+- simple routes
+- routes with form-wizard
+- sub-routers
+- arrays of grouped middleware
+  - collections middleware (`setActions`,`setContext`,`setPagination`)
+  - data setup, retrieval and transformation middleware (`setBody` `setFilter` `setResults`)
+
+#### router.`HTTP_METHOD`
+
+##### Generic middleware
+
+- **path/with/:tokens**
+- **protectRoute**
+- redirects
+
+##### Collections setup
+
+- setActions
+- setContext
+  - permission
+  - text
+  - href
+- setPagination
+
+##### Data setup, retrieval and transformation
+
+- setBody*Something*
+  - sets `req.body.something`
+    - structure includes `status`, `sortBy`, `sortDirection`, `[locationType]`
+- set*Something*Filters
+  - calls services
+  - uses body prop
+  - sets `req.filter`
+- set*Something*Results
+  - calls services
+  - uses body prop
+  - sets `req.resultsSomething` or `req.resultsSomethingAsTable`
+
+##### Render
+- `render` method from controller, e.g. `home/controllers.js:dashboard`
+  - `res.render` with template e.g. `home/dashboard`
+    - actions
+    - context
+    - dateRange
+    - filter
+    - pagination
+    - period
+    - resultsVars
+
+# Questions
+
+- what is the difference between setBody, setFilter and setResults
+- what is context? It's used by Express and by Nunjucks, are they exactly the same thing?
+- ````javascript
+  content: {
+          html: t("collections::no_results", {
+            context: context
+          })
+        }```
+          - ```javascript
+      {{ t("errors::" + errorKey + ".content", { context: journeyName }) | safe }}
+      ```
+  ````
--- a/app/README.md
+++ b/app/README.md
@@ -4,22 +4,14 @@ App modules are subsections of the main site, dealing with a specific area , e.g
 
 They can have a variety of different parts, in either a root file, or a folder with multiple files, and exposed via an `index.js` in the folder.
 
-- controllers
-  - Express
-- constants
-  - Predefined default settings for a module, including routing paths, actions, and other defaults
-- fields
-  - Form-wizard field definitions
-- middleware
-  - Express middleware for the module, rendering against the views
-- steps
-  - Form-wizard step definitions
-- validators
-  - Form-wizard field validator functions
-- views
-  - Nunjucks templates
-- index (routing)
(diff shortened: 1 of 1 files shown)


app/README.md
```diff
@@ -0,0 +1,105 @@
...
+- setContext
+  - permission
+  - text
+  - href
+- setPagination
+
+##### Data setup, retrieval and transformation
+
+- setBody*Something*
+  - sets `req.body.something`
```
solidgoldpig, line 70: sets `req.body.requested`...

Dependent on the *Something* though, a different serivce method gets invoked with the params set up thusly

app/README.md
```diff
@@ -0,0 +1,105 @@
...
+    - structure includes `status`, `sortBy`, `sortDirection`, `[locationType]`
+- set*Something*Filters
+  - calls services
+  - uses body prop
+  - sets `req.filter`
+- set*Something*Results
+  - calls services
+  - uses body prop
+  - sets `req.resultsSomething` or `req.resultsSomethingAsTable`
```
solidgoldpig, line 79: `req.result*Something*` or `req.results<Something>` to make it more obvious?

app/README.md
```diff
@@ -0,0 +1,105 @@
...
+    - context
+    - dateRange
+    - filter
+    - pagination
+    - period
+    - resultsVars
+
+# Questions
+
+- what is the difference between setBody, setFilter and setResults
```
solidgoldpig, line 94: Did I manage to answer this ok on Slack the other day?

app/README.md
```diff
@@ -0,0 +1,105 @@
+# App modules
+
+App modules are subsections of the main site, dealing with a specific area , e.g locations or moves. They are constructed using Express routers.
+
+They can have a variety of different parts, in either a root file, or a folder with multiple files, and exposed via an `index.js` in the folder.
+
+- controllers
+  - Express
```
teneightfive, line 8: Could these maybe use an emdash after the _thing_ instead of a nested list? 

Might be easier to scan the _things_ as a list of bullets with a description after. Semantically they also feel like the same item rather than a child item.

We did something similar in the [framework docs](https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks/tree/master/frameworks/person-escort-record#step-options)

app/README.md
```diff
@@ -0,0 +1,105 @@
...
+- validators
+  - Form-wizard field validator functions
+- views
+  - Nunjucks templates
+- index (routing)
+  - Main entrypoint for the module, tying together all the parts of the module in the form of an Express router
+
+## Index / Routing Setup
+
+### params middleware
```
teneightfive, line 26: We could link to the express docs for [param middleware](https://expressjs.com/en/api.html#router.param) and [shared middleware](https://expressjs.com/en/api.html#router.use) (below) to explain their usage in more depth.

app/README.md
```diff
@@ -0,0 +1,105 @@
...
+  - sets `req.filter`
+- set*Something*Results
+  - calls services
+  - uses body prop
+  - sets `req.resultsSomething` or `req.resultsSomethingAsTable`
+
+##### Render
+- `render` method from controller, e.g. `home/controllers.js:dashboard`
+  - `res.render` with template e.g. `home/dashboard`
+    - actions
```
teneightfive, line 84: Might be worth being more generic in this case and referring to the [express render docs](https://expressjs.com/en/api.html#res.render) and just document that this is called with any locals required for that view.

app/README.md
teneightfive: This is the current structure:

`setBody` — used to build the shared body sent to the service calls to generate the filter and the results. Contains a shared query for date filters, location filters, etc. Attached result to `req.body[type]`

`setFilter` — builds the filter component for that collection. Uses `req.body[type]` as a base and overrides the filter based on the desired counts to retrieve, usually status. Attached the result as a filter component to `req.filter`.

`setResults` — makes a call to the relevant service method to return the results that should be displayed for the current view. Uses `req.body[type]` which is based on the current params and filters in the URL. Also responsible for calling the relevant presenter for this resource type and sets the results as `req.resultsAsTable` or `req.resultsAsCards` to be used by the controller/template.

app/README.md
```diff
@@ -0,0 +1,105 @@
...
+    - dateRange
+    - filter
+    - pagination
+    - period
+    - resultsVars
+
+# Questions
+
+- what is the difference between setBody, setFilter and setResults
+- what is context? It's used by Express and by Nunjucks, are they exactly the same thing?
```
teneightfive, line 95: [Context](https://www.i18next.com/translation-function/context) is a concept used by the i18n library [i18n](https://www.i18next.com/).

We use it in certain circumstances to change the wording we need to use based on particular contexts within the application. The collection is one example, by passing the type as the context we can change the text to be more specific to something like a "single request" or "allocation" but use a single key within the template.

app/README.md
```diff
@@ -0,0 +1,57 @@
...
+### Common Collections Middleware
+
+- `redirectDefaultQuery` - redirect with a default query if provided
+- `redirectView` - redirect to a path of the format `timePeriod/date/location/view?query`
+- `setActions` - sets potential actions during the routing setup phase
+- `setContext` - sets the `context` used by [i18n]() to specific content within a generalised template
+- `setDateRange` - sets the `dateRange` used by the weekly views
+- `setPagination` - sets the pagination object used for day/week pagination. Note: This is not for numeric pagination of table data.
+
+## Orchestration
```
merlinc, line 44: I'm not really happy with this name, but I couldn't think of another grouping for these three kinds of middleware

app/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+There are also three specific kinds of orchestration middleware used by this application to more clearly orchestrate fetching and arranging of data. When multiple instances of this middleware are used, they share data through the `req.body[moduleName]` property.
+
+- `setBody` — used to build the shared body data that is used to call service(s) which then supplies data for the filter and the results. Contains a shared query for date filters, location filters, etc. Attaches data to `req.body[moduleName]`.
```
merlinc, line 48: @solidgoldpig FYI These are mostly @teneightfive 's words, slightly reworded. If there's anything that you think needs clearing up, let me know.

app/README.md
```diff
@@ -0,0 +1,57 @@
...
+- `setFilter` — builds the filter component for collection modules. Uses `req.body[moduleName]` as a base and overrides the filter based on the desired counts to retrieve - usually status. Attaches data as a filter component to `req.filter`.
+- `setResults` — makes a call to the relevant service method to return the results that should be displayed for the current view. Can use `req.body[moduleName]` which is based on the current params and filters in the URL. If required for the view it will also call the relevant [presenter]('../common/presenters') for this resource type and sets the results as `req.resultsAsTable` or `req.resultsAsCards` to be used by the controller/template.
+
+## General notes
```
merlinc, line 52: Are there any other gotchas we can think of?

app/README.md
merlinc: Reworked this section based upon slack / other PR comments

app/README.md
merlinc: Reworked this section based upon slack / other PR comments

app/README.md
merlinc: Reworked this section based upon slack / other PR comments

app/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+They can have a variety of different parts, in either a root file, or a folder with multiple files, and exposed via an `index.js` in the folder.
+
+- `controllers` - Express controllers
+- `constants` - Predefined default settings for a module, including routing paths, actions, and other defaults
+- `fields` - Form-wizard field definitions
+- `middleware` - Express middleware for the module, rendering against the views
+- `steps` - Form-wizard step definitions
+- `validators` - Form-wizard field validator functions
+- `views` - Nunjucks templates`
```
teneightfive, line 13: Rogue backtick?

app/README.md
```diff
@@ -0,0 +1,57 @@
...
+- `setLocations` - sets locations on the current request
+- `setPrimaryNaviation` - sets the primaryNavigation based upon the current users permissions
+- `setUser` - sets the user on the current request from the session
+
+### Common Collections Middleware
+
+- `redirectDefaultQuery` - redirect with a default query if provided
+- `redirectView` - redirect to a path of the format `timePeriod/date/location/view?query`
+- `setActions` - sets potential actions during the routing setup phase
+- `setContext` - sets the `context` used by [i18n]() to specific content within a generalised template
```
teneightfive, line 40:
```suggestion
- `setContext` - sets the `context` used by [i18next](https://www.i18next.com/translation-function/context) to specific content within a generalised template
```

app/README.md
teneightfive:
```suggestion
- `views` - Nunjucks templates
```

teneightfive (approved):

PR #888, opened 2020-10-16
[P4-2337] Create view for timeline component

solidgoldpig:
### What changed

- Adds move events timeline view
- Adds tabs to move view to access it

### Why did it change

Enables the review of events linked to a move

### Issue tracking

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd>Move_details_for_MELLIE__ROGER_–_Book_a_secure_move</kbd> |
| <kbd>Move_details_for_MELLIE__ROGER_–_Book_a_secure_move</kbd> | <kbd>Move_details_for_MELLIE__ROGER_–_Book_a_secure_move</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


config/nunjucks/filters.js
```diff
@@ -114,6 +114,7 @@ function formatDateRange(value, delimiter = 'to') {
  * @example {{ "2019-02-21" | formatDateRange("DD/MM/YY") }}
  */
 function formatDateRangeAsRelativeWeek(value) {
+  const i18n = require('../i18n')
```
solidgoldpig, line 117: This is obviously awful - just a stopgap to prompt discussion of how to prevent the circular dependency between config/118n and nunkucks/filters

(I did point out before that I didn't like the i18n.t lurking in here)

As no lang information is actually passed into the function, the i18n value returned would always be the default language. Is there any reason not to just grab the value for the string from the locales file directly? Or perhaps better, a specific config/locales file for the filters>

Or should we pass the i18n method in to either initialise the filters or return an instance?

...

Also should we move the contents of `nunjucks/filters` to `common/lib/filters` and then load that from the nunucks and the i18n modules

app/move/views/view.njk
```diff
@@ -279,7 +297,7 @@
 
       {% if personEscortRecordIsEnabled and personEscortRecord %}
         {% include "move/views/_includes/person-escort-record-summary.njk" %}
-      {% else %}
+      {% else %}COCK
```
teneightfive, line 300: 🚨

solidgoldpig: AYE

config/nunjucks/filters.js
solidgoldpig, line 117: This is obviously awful - just a stopgap to prompt discussion of how to prevent the circular dependency between config/118n and nunkucks/filters

(I did point out before that I didn't like the i18n.t lurking in here)

As no lang information is actually passed into the function, the i18n value returned would always be the default language. Is there any reason not to just grab the value for the string from the locales file directly? Or perhaps better, a specific config/locales file for the filters>

Or should we pass the i18n method in to either initialise the filters or return an instance?

...

Also should we move the contents of nunjucks/filters to common/lib/filters and then load that from the nunucks and the i18n modules

app/move/controllers/view.js
```diff
@@ -5,6 +5,7 @@ const updateSteps = require('../steps/update')
 
 const getUpdateLinks = require('./view/view.update.links')
 const getUpdateUrls = require('./view/view.update.urls')
+const setYouthMove = require('./view/view.youth-move')
```
solidgoldpig, line 9: I originally extracted this because I thought the timeline would need to share the code - it doesn't (for now anyway).

common/lib/api-client/middleware/process-response.js
```diff
@@ -0,0 +1,27 @@
+const findUnpopulatedResources = require('../../find-unpopulated-resources')
+
+const processResponseMiddleware = {
+  name: 'process-response',
+  req: function req(response) {
+    const { req = {} } = response
+
+    // TODO: populateResources probably better as preserveResourceReferences
+    if (req.populateResources) {
```
solidgoldpig, line 9: If devour didn't ruthlessly remove unincluded relationships... If...

common/lib/api-client/index.js
```diff
@@ -49,6 +50,10 @@ module.exports = function () {
     })
   )
 
+  if (API.USE_MOCKS) {
+    insertRequestMiddleware(requestMocks)
```
solidgoldpig, line 54: Is there a better way? Surely…

app/move/controllers/timeline.js
```diff
@@ -0,0 +1,187 @@
...
+
+const { event: eventModel } = require('../../../common/lib/api-client/models')
+const populateResources = require('../../../common/lib/populate-resources')
+const presenters = require('../../../common/presenters')
+const moveService = require('../../../common/services/move')
+const i18n = require('../../../config/i18n')
+
+const getTabsUrls = require('./view/view.tabs.urls')
+
+const addMockValuesToEvents = (req, moveEvents) => {
```
solidgoldpig, line 11: Horrible. Should probably be removed before merging. Or enabling only if `API_USE_MOCKS` is on

app/move/controllers/view.js
solidgoldpig, line 9: I originally extracted this because I thought the timeline would need to share the code - it doesn't (for now anyway).

solidgoldpig: (Not sure why this is showing up on the tabs commit too :( )

app/move/controllers/view/view.tabs.urls.js
```diff
@@ -0,0 +1,9 @@
+const getTabsUrls = move => {
+  const tabsUrls = {
+    view: `/move/${move.id}`,
```
teneightfive, line 3: Could you use relative URLs within the view rather than needing these?

app/move/controllers/timeline.js
```diff
@@ -0,0 +1,33 @@
...
+  const { move } = req
+  const { profile } = move
+
+  const urls = {
+    tabs: getTabsUrls(move),
+  }
+
+  const { assessment_answers: assessmentAnswers = [] } = profile || {}
+
+  const { timeline_events: moveEvents } = await moveService.getById(move.id, {
```
teneightfive, line 17: This will introduce the frontend calling the API for the move twice with this change.

Can we instead, make a change to [`setMove`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/903978823a955b1af54c08d3b7e6c07ca4b24820/app/move/middleware.js#L5-L16) to dynamically ask for the relevant includes.

I'd suggest making that middleware do this:

```javascript
  setMove: async (req, res, next, moveId) => {
    if (!moveId) {
      return next()
    }

    try {
      req.move = await moveService.getById(moveId, {
        include: req.moveInclude,
      })
      next()
    } catch (error) {
      next(error)
    }
  },
```

Then a middleware called only for this route which sets that extra include:

```javascript
function setTimelineInclude (req, res, next) {
  req. moveInclude = 'timeline_events'
  next()
}

moveRouter.get('/timeline', protectRoute('move:view'), setTimelineInclude, timeline)
```

I'd also suggest keeping the `populateResources: true` within the service as the controller/middleware shouldn't have to care about that. We know this is just a patch for **only** events until the API supports the correct includes so I'd rather see that isolated to the service.

app/move/views/view.njk
```diff
@@ -264,7 +264,25 @@
   {% endif %}
 {% endmacro %}
 
+{% macro tabs(urls, active) %}
+<div class="govuk-tabs">
+  <ul class="govuk-tabs__list">
+    <li class="govuk-tabs__list-item {{- ' govuk-tabs__list-item--selected' if active == 'view'}}">
```
teneightfive, line 270: Could a loop be used here to avoid the markup duplication?

app/move/controllers/view/view.tabs.urls.js
solidgoldpig: So...

```
  const tabsUrls = {
    view: `.../${move.id}`,
    timeline: `timeline`,
  }
```
??

with the actual move link being wrong when it's the current page?

solidgoldpig: Can we have a centralised url thingy already?

app/move/controllers/timeline.js
solidgoldpig: If it's moved in to the service then we have to check every move for unpopulated resources which is totally unnecessary.

And as an isolated pair of commits, it can be reverted in one go.

As for getting round the setMove conundrum, I think this is much cleaner
```js
router.get(
  `/:moveId(${uuidRegex})/timeline`,
  protectRoute('move:view'),
  timeline
)
```

and then in timeline
```js
  const move = await moveService.getById(req.params.moveId, {
    include: 'timeline_events',
    populateResources: true,
  })
```

solidgoldpig: (I already pushed a commit doing so)

teneightfive (approved):

PR #896, opened 2020-10-19
feat: Add support for in-session feature flags

merlinc:
## Proposed changes

Currently feature flags are set one time as the server starts up. This means that in order to change feature flags, the server needs to be restarted. This means it is not possible to have a browser test that uses feature flags. Middleware can be tested on its own, but combinations of middleware, such as when used with an Express router, cannot be tested. 

This PR uses a [form-wizard feature](https://github.com/HMPO/hmpo-form-wizard#session-injection) called session injection. This allows a variety of request variables to be changed on a per request basis, storing the result in the session object.

This should hopefully also allow work behind feature flags to be previewed on heroku.

Note: Unit tests should continue to test individual usage of feature flags, this new feature is only to enable testing them at a browser level.

### What changed

There is currently only one feature flag in current use - `PERSON_ESCORT_RECORD` - these updates are the changes to the scaffolding required to support this feature, plus the changes required to support the `PERSON_ESCORT_RECORD` feature.

- Load feature flags from env vars into req.session.featureFlags
- Change uses of config.featureFlags to use req.session.featureFlags
- Update tests to use req.session.featureFlags

### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/allocation/index.js
+++ b/app/allocation/index.js
@@ -3,7 +3,6 @@ const wizard = require('hmpo-form-wizard')
 
 const FormWizardController = require('../../common/controllers/form-wizard')
 const { protectRoute } = require('../../common/middleware/permissions')
-const { FEATURE_FLAGS } = require('../../config')
 const { setMove } = require('../move/middleware')
 
 const { createControllers, viewAllocation } = require('./controllers')
@@ -70,11 +69,7 @@ router.use(
   wizard(cancelSteps, cancelFields, cancelConfig)
 )
 
-router.get(
-  '/:allocationId',
-  protectRoute('allocations:view'),
-  viewAllocation(FEATURE_FLAGS.PERSON_ESCORT_RECORD)
-)
+router.get('/:allocationId', protectRoute('allocations:view'), viewAllocation())
 
 // Export
 module.exports = {
--- a/app/moves/middleware/set-results.moves.test.js
+++ b/app/moves/middleware/set-results.moves.test.js
@@ -41,6 +41,9 @@ describe('Moves middleware', function () {
         },
         session: {
           currentLocation: '#location',
+          featureFlags: {
+            PERSON_ESCORT_RECORD: true,
+          },
         },
       }
     })
@@ -134,7 +137,7 @@ describe('Moves middleware', function () {
         })
       })
 
-      context('with `personEscortRecordFeature`', function () {
+      context('with `personEscortRecordFeature` enabled', function () {
         context('with correct user permission', function () {
           beforeEach(async function () {
             req.canAccess.withArgs('person_escort_record:view').returns(true)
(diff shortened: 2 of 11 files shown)


app/allocation/index.js
```diff
@@ -70,11 +69,7 @@ router.use(
   wizard(cancelSteps, cancelFields, cancelConfig)
 )
 
-router.get(
-  '/:allocationId',
-  protectRoute('allocations:view'),
-  viewAllocation(FEATURE_FLAGS.PERSON_ESCORT_RECORD)
-)
+router.get('/:allocationId', protectRoute('allocations:view'), viewAllocation())
```
merlinc, line 72: This feature flag no longer needs to be applied on the router level,

server.js
```diff
@@ -151,6 +151,17 @@ if (config.IS_DEV) {
   app.use(development.setUserLocations(config.USER_LOCATIONS))
 }
 
+// Session injection middleware
+app.all('*', function (req, res, next) {
```
merlinc, line 155: This should obviously be pulled out as a separate middleware

server.js
```diff
@@ -151,6 +151,17 @@ if (config.IS_DEV) {
...
+  req.session.featureFlags = config.FEATURE_FLAGS
+  next()
+})
+
+if (config.SESSION_INJECTION_ENABLED) {
```
merlinc, line 160: Only if the session injection feature is enabled, do we want to use these routes.

One thing to think about for the future, is do we want to have this functionality on production in order to test in production without requiring a restart for environment variables. However, without a separate single source of truth for this configuration, to enable the feature in production for everyone will require changing environment variables and restarting.

app/moves/middleware/set-results.moves.test.js
```diff
@@ -174,6 +173,49 @@ describe('Moves middleware', function () {
       })
     })
 
+    context('with `personEscortRecordFeature` disabled', function () {
```
merlinc, line 176: I didn't see a test for this feature being disabled, so it seemed like a good idea to add one now.

merlinc:
This new functionality also needs to be documented in the [feature flags documentation](./docs/feature-flags)

merlinc:
Due to issues with partially clearing a users session, together with the amount of data that is normally stored in it, this approach is likely to introduce bugs in the system. Also there is no mechanism to control feature flags without restarting the service, which also reduces some of the usefulness of quickly enabling/disabling a feature in an environment.

Given these issues, other workarounds have been used for existing feature development, and so this PR should be closed.

PR #902, opened 2020-10-20
[P4-2317] fix(allocation): Stop person removal from an active or completed move

teneightfive:
### What changed

This change adds a check to the form wizard that handles the removal
of a person and displays a message to the user if this action cannot be
done.

This change still persists the button on the allocation page to remove a person instead of removing it
as simply removing might confused the user as to the circumstances in which it is not available.

Another option would be to disable the button and add a message as to why this action isn't possible
but this part of the interface already felt quite complicated and introducing this message might add extra
things to digest on one screen.

This is also based on a per move basis so can't be done as a single action on the allocation. Therefore, the
message has been moved to the step when a user attempts to complete the action. This was also much more
straightforward to implement a fix at this stage to prevent the error continually appearing in Sentry.

### Why did it change

Currently the API does not allow a person (or profile) to be removed
from a move that has been started (`in_transit`) or that has taken
place(`completed`).

### Issue tracking

- [P4-2317](https://dsdmoj.atlassian.net/browse/P4-2317)

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/unassign.js
+++ b/app/move/controllers/unassign.js
@@ -4,6 +4,7 @@ const moveService = require('../../../common/services/move')
 class UnassignController extends FormWizardController {
   middlewareChecks() {
     this.use(this.checkAllocation)
+    this.use(this.checkEligibility)
     super.middlewareChecks()
   }
 
@@ -19,6 +20,20 @@ class UnassignController extends FormWizardController {
     }
 
     next()
+  }
+
+  checkEligibility(req, res, next) {
+    const { allocation, id, profile, status } = req.move
+
+    if (['requested', 'booked'].includes(status)) {
+      return next()
+    }
+
+    res.render('move/views/unassign-ineligible', {
+      allocation,
+      moveId: id,
+      fullname: profile?.person?.fullname,
+    })
   }
 
   middlewareLocals() {
--- a/app/move/controllers/unassign.test.js
+++ b/app/move/controllers/unassign.test.js
@@ -10,6 +10,8 @@ describe('Move controllers', function () {
     describe('#middlewareChecks()', function () {
       beforeEach(function () {
         sinon.stub(FormWizardController.prototype, 'middlewareChecks')
+        sinon.stub(UnassignController.prototype, 'checkAllocation')
+        sinon.stub(UnassignController.prototype, 'checkEligibility')
         sinon.stub(controller, 'use')
 
         controller.middlewareChecks()
@@ -21,8 +23,14 @@ describe('Move controllers', function () {
       })
 
       it('should call checkAllocation middleware', function () {
-        expect(controller.use).to.have.been.calledOnceWithExactly(
+        expect(controller.use).to.have.been.calledWithExactly(
           controller.checkAllocation
+        )
+      })
+
+      it('should call checkEligibility middleware', function () {
+        expect(controller.use).to.have.been.calledWithExactly(
+          controller.checkEligibility
         )
       })
     })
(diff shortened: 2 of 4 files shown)


teneightfive:
@Andy-Millar @JadeNConroy I have made this change to prevent this scenario throwing a Sentry error each time it occurs. I had to different a little bit from the ticket because the allocation doesn't have a state of being "in transit" or "completed", that is determined by each move individually. 

With that in mind I've had to implement this fix at the level when a user attempts to remove a person from that particular move. This will also prevent anyone from being able to remove someone from a cancelled or proposed move now too.

I re-used the pattern we have used elsewhere for conflicts but maybe the level "error" could be reduced in this instance. Here's a couple of examples using different levels of notifications:

| Warning | Info |
| ------ | ----- |
| <kbd><kbd>image</kbd></kbd> | <kbd><kbd>image</kbd></kbd> |

Let me know your thoughts.

app/move/controllers/unassign.js
```diff
@@ -21,6 +22,20 @@ class UnassignController extends FormWizardController {
...
+    const { allocation, id, profile = {}, status } = req.move
+
+    if (['requested', 'booked'].includes(status)) {
+      return next()
+    }
+
+    res.render('move/views/unassign-ineligible', {
+      allocation,
+      moveId: id,
+      fullname: profile.person?.fullname,
```
solidgoldpig, line 35: We have person directly on the move. Might it be better to get it from there?

solidgoldpig (approved):

app/move/controllers/unassign.js
teneightfive: ...erm, should that actually be the case? I thought the whole point of moving to a "profile" was to remove the association of "person" from a move?!

teneightfive: The API docs also suggest that's not the case:

<kbd>image</kbd>

Or are you saying that's something that we are doing on the frontend in the service transformer/middleware?

teneightfive: It [looks like](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/18dc3637ae5a9ba6cd72874975b76ce2505c3559/common/services/move.js#L62) we're assuming the API returns this relationship in the service, but [the model suggests](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/18dc3637ae5a9ba6cd72874975b76ce2505c3559/common/lib/api-client/models.js#L3-L55) this relationship also doesn't exist.

I wonder if this is related to us transitioning from `V1` to `V2` of the API and whether we should remove this?

solidgoldpig: I can't remember for sure whether we added the person to the move in the transform for backwards compatibility or as I seem to remember as a convenience... We can remove. But I'd like that to wait until after we've done all these overhauls we have in motion…

Andy-Millar:
@teneightfive Nice idea - like it!

Andy-Millar (approved):

app/move/controllers/unassign.js
teneightfive: >We can remove. But I'd like that to wait until after we've done all these overhauls we have in motion…

Yep, totally agree this is a future fix and not to be bundled as part of this. Have created [an issue](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/903) to keep track of it for now...

JadeNConroy (approved):

PR #906, opened 2020-10-21
Add req context to services and api client

solidgoldpig:
This is where I got to when I add req to every single service call…

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/compare/api-client/request-id

but the amount of awful calls with undefined or empty objects in them led me to despair. Yes, `req` could be moved to be first argument to reduce the what position should it be in discombobulation and the need for unnecessary args, but it just felt as if it was totally orthogonal to (indeed, irrelevant to) the actual service methods

So I went back to the drawing board and came up with the changes in this PR

tldr;
```js
const moveService = require('./common/services/move')
```
still works but you can now also do
```js
const MoveService = require('./common/services/move')
const moveService = MoveService.addRequestContext(req)
```
This allows for the individual usages of the service calls to be updated individually rather than having to do it in one massive big bang. I’ve update the move cancel controller and move middleware, but nowt else. (Potentially, after all the service uses had been updated, the services could be refactored to be classes? Or do it now and bit the bullet?)

There is now `req.services` that exposes a method for each of the services registered in `common/services/index` (Was that even being used before?) - those methods add the req context automatically and also reuse any cached instance.
Beyond that, the last 2 commits show possible automated/unified handling of logging+metrics for the services.

Anyhow, comments or screams and curses welcome

### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/lib/api-client/index.js
+++ b/common/lib/api-client/index.js
@@ -15,17 +15,13 @@ const {
 } = require('./middleware')
 const models = require('./models')
 
-let instance
-
-module.exports = function () {
-  if (instance) {
-    return instance
-  }
-
-  instance = new JsonApi({
+module.exports = function (req) {
+  const instance = new JsonApi({
     apiUrl: API.BASE_URL,
     logger: false,
   })
+
+  instance._originalReq = req
 
   instance.replaceMiddleware('errors', errors)
   instance.replaceMiddleware('POST', post(FILE_UPLOADS.MAX_FILE_SIZE))
--- a/common/lib/api-client/request-headers.test.js
+++ b/common/lib/api-client/request-headers.test.js
@@ -43,6 +43,10 @@ describe('API Client', function () {
         it('should return the `Idempotency-Key` header', function () {
           expect(headers['Idempotency-Key']).to.equal('#uuid')
         })
+
+        it('should return the `X-Transaction-Id` header', function () {
+          expect(headers['X-Transaction-Id']).to.equal('auto-#uuid')
+        })
       })
     })
 
@@ -50,7 +54,7 @@ describe('API Client', function () {
       let headers
 
       beforeEach(function () {
-        headers = getRequestHeaders('format/foo')
+        headers = getRequestHeaders({ format: 'format/foo' })
       })
 
       it('should return `Accept` header with specified format and the api version', function () {
@@ -58,19 +62,23 @@ describe('API Client', function () {
           `format/foo; version=${config.API.VERSION}`
         )
       })
+    })
 
-      it('should return the `User-Agent` header', function () {
-        expect(headers['User-Agent']).to.equal(
-          'hmpps-book-secure-move-frontend/chuck-d'
-        )
-      })
+    context('when called with a request', function () {
+      let headers
 
-      it('should return the `Accept-Encoding` header', function () {
-        expect(headers['Accept-Encoding']).to.equal('gzip')
+      beforeEach(function () {
+        headers = getRequestHeaders({
+          req: { transactionId: '#transactionId' },
+        })
       })
 
       it('should return the `Idempotency-Key` header', function () {
         expect(headers['Idempotency-Key']).to.equal('#uuid')
+      })
+
+      it('should return req’s transactionId as the `X-Transaction-Id` header', function () {
+        expect(headers['X-Transaction-Id']).to.equal('#transactionId')
       })
     })
   })
--- a/common/lib/monitoring.js
+++ b/common/lib/monitoring.js
@@ -1,0 +1,77 @@
+/* eslint-disable no-console */
+const timer = require('./timer')
+
+const wrapMethodAsync = (methodName, method, methodOptions, context) => {
+  return async (...args) => {
+    const methodTimer = timer()
+
+    try {
+      const result = await method.apply(null, args)
+      const duration = methodTimer()
+      console.log(
+        `${methodName}, success, took ${duration}, tId:${context?.transactionId}`
+      )
+      // TODO: metrics call
+      // TODO: log success
+      return result
+    } catch (error) {
+      const duration = methodTimer()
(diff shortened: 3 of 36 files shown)


common/middleware/set-services.js
```diff
@@ -0,0 +1,25 @@
+const services = require('../services')
+
+const setServices = (req, res, next) => {
+  const servicesCache = {}
```
solidgoldpig, line 4: Might be worth renaming `servicesCache` to `servicesRequestCache` or  something that makes it more explicit that this is a short-term, per-request rather than long-lived cache.

common/lib/api-client/index.js
```diff
@@ -55,6 +61,18 @@ module.exports = function () {
...
+    req: payload => {
+      const { jsonApi } = payload
+
+      if (jsonApi._originalReq) {
+        delete jsonApi._originalReq
+      }
+
+      return payload
+    },
+  })
```
Nimphal, line 75: I am definitely missing something here, but what is this achieving?

common/middleware/set-services.js
Nimphal: Maybe something like servicesRequestContext or something?

common/lib/monitoring.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable no-console */
```
merlinc, line 1: Is this monitoring part of the PR, or is it a holdover from your previous promtheus PR?

server.js
```diff
@@ -52,6 +54,12 @@ if (config.IS_PRODUCTION) {
   app.enable('trust proxy')
 }
 
+// Ensure we have a useful transaction id
```
merlinc, line 57: This would be cleaner and more obvious if it was a transactionID middleware, e.g.
```
app.use(setTransactionId`

File: common/lib/api-client/request-headers.test.js [hunk 5]
```diff
@@ -43,34 +43,38 @@ describe('API Client', function () {
...
-        expect(headers['Accept-Encoding']).to.equal('gzip')
+      beforeEach(function () {
+        headers = getRequestHeaders({
+          req: { transactionId: '#transactionId' },
+        })
       })
 
-      it('should return the `Idempotency-Key` header', function () {
-        expect(headers['Idempotency-Key']).to.equal('#uuid')
+      it('should return req’s transactionId as the `X-Transaction-Id` header', function () {
```
merlinc, line 76: Are we ok with removing the `Idempotency-Key` test here?

merlinc (commented):
You were probably waiting for this - but I'd like some documentation on how all this works :grinning: 

Definitely with how things work with the services running off the `req` object, but you might find that while writing that up, you've accidentally documented the current system too. That's also good :)

It'd also be good to get a bit more background on what the monitoring is doing in this PR, is it a reworking of the existing code, or is a new addition for this new approach?

common/lib/api-client/index.js
solidgoldpig: If it's the block of code I just removed, the answer is nothing :). (I based this off of an initial investigation by Dom and had forgotten to remove it)

common/lib/monitoring.js
solidgoldpig: > Beyond that, the last 2 commits show possible automated/unified handling of logging+metrics for the services

Sorry, the monitoring part is just a demonstration of how we could add it in for all methods we wanted as essentially config. (What with the numerous rather gnarly rebases I've had to do, I didn't fancy spinning off another branch on top of this one)

I would actually remove those commits from this PR and put them in a new one.

server.js
solidgoldpig: Yes it totally would.

solidgoldpig: and stuff like this be moved out too ;)

```
if (config.SENTRY.DSN) {
  Sentry.init({
    dsn: config.SENTRY.DSN,
    environment: config.SENTRY.ENVIRONMENT,
    release: config.SENTRY.RELEASE,
  })

  app.use(
    Sentry.Handlers.requestHandler({
      // Ensure we don't include `data` to avoid sending any PPI
      request: ['cookies', 'headers', 'method', 'query_string', 'url'],
      user: ['id', 'username', 'permissions'],
    })
  )
  app.use(sentryRequestId)
}
```

solidgoldpig: and done

common/middleware/set-services.js
solidgoldpig: Unless anyone has other suggestions, `servicesRequestContext` it shall be

common/lib/api-client/request-headers.test.js
solidgoldpig: That looks like a straight up mistake to me. Good catch

solidgoldpig: Aah. It's not really removed - it's been moved to line 43 in the previous context block. I'll add in the dupe test though.

common/services/person.js
```diff
@@ -1,129 +1,137 @@
...
-    }
-  },
-
-  format(data) {
-    const formatted = mapValues(data, (value, key) => {
-      if (typeof value === 'string') {
-        if (relationshipKeys.includes(key)) {
-          return { id: value }
-        }
+const addRequestContext = req => {
```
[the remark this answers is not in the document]
solidgoldpig: Ouch

merlinc:
Building on this PR, if I wanted to add a custom header to all downstream requests, how would I do that?

i.e. I make a request to `http://localhost:3000` with a `X-PECS-SCENARIO=pmu-dashboard` header, and I want that header passed down to the api layer.

PR #908, opened 2020-10-21
chore: Add ability to specify different env vars for each PR

solidgoldpig:
### What changed

Allow environment variables set at the project level tp be overridden for current PR

eg. Given a Github PR number of 123

if set, the env var PR123_E2E_VIDEO will be used in place of E2E_VIDEO

Also fixes a bug where env vars set as `false` ended up being true - process.env[foo] values are always stringified (which makes sense)

### Why did it change

Enable specific PRs to change the env vars they are using to run the e2e tests without having to change the values for all runs

### Issue tracking

n/a

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/test/e2e/parallel-runner.js
+++ b/test/e2e/parallel-runner.js
@@ -4,13 +4,58 @@ const concurrently = require('concurrently')
 const glob = require('glob')
 const yargs = require('yargs')
 
-const {
-  E2E_MAX_PROCESSES,
-  E2E_SKIP,
-  E2E_VIDEO,
-  E2E_FAIL_FAST,
-  E2E_BASE_URL,
-} = process.env
+/**
+ * Allow environment variables set at the project level tp be overridden for current PR
+ * eg. Given a CIRCLE_PR_NUMBER of 123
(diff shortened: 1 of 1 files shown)


test/e2e/parallel-runner.js
```diff
@@ -4,13 +4,58 @@ const concurrently = require('concurrently')
...
+const getEnvVar = key => {
+  let value = process.env[key]
+
+  if (value === undefined) {
+    return
+  }
+
+  value = value.trim()
+
+  if (value === 'false') {
```
teneightfive, line 45: Maybe a `switch` statement might be easier to read in this instance?

```javascript
switch (value.trim()) {
  case value === 'false':
    value = false
    break
    
  case value === 'true':
    value = true
    break

  default:
    value = value.trim()
    break
}
```

teneightfive (approved):

PR #923, opened 2020-10-26
feat(allocation): Improve allocations listing performance

solidgoldpig:
This PR depends on this BE PR
https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1070 (which has now been merged)

### What changed

Switch to using allocation's `metadata` to determine slots rather than having to ask for `moves` and `moves.profile`.

Additonally displays 'CANCELLED' rather than '0' for a cancelled allocation's progress when listing it.

Also displays the number of moves on the individual allocation view rather than '0'.

### Why did it change

Reduce the number of includes required to render the list of allocations

### Issue tracking

- [P4-2341 - FE - Support Metadata in FE client for Allocation Dashboard](https://dsdmoj.atlassian.net/browse/P4-2341)

## Screenshots

| Before | After |
| ------ | ----- |
|<kbd>Allocations_for_26_Oct_to_1_Nov_2020__This_week_</kbd>|<kbd>Allocations_for_26_Oct_to_1_Nov_2020__This_week_</kbd>|
|<kbd>Allocation_for_0_people_from_WETHERBY__HMPYOI__to_DARTMOOR__HMP__on_28_Oct_2020</kbd>|<kbd>Allocation_for_0_people_from_WETHERBY__HMPYOI__to_DARTMOOR__HMP__on_28_Oct_2020</kbd>|

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -5,7 +5,7 @@ const presenters = require('../../../common/presenters')
 function viewAllocation(personEscortRecordFeature = false) {
   return (req, res) => {
     const { allocation, canAccess } = req
-    const { moves, status } = allocation
+    const { moves, status, totalSlots } = allocation
     const bannerStatuses = ['cancelled']
     const movesWithoutProfile = moves.filter(move => !move.profile)
     const movesWithProfile = moves.filter(move => move.profile)
@@ -48,7 +48,7 @@ function viewAllocation(personEscortRecordFeature = false) {
       unassignedMoveId: movesWithoutProfile.length
         ? movesWithoutProfile[0].id
         : undefined,
-      totalCount: moves.length,
+      totalCount: totalSlots,
       remainingCount: movesWithoutProfile.length,
       addedCount: movesWithProfile.length,
       moves: sortBy(
--- a/app/allocation/views/view.njk
+++ b/app/allocation/views/view.njk
@@ -77,79 +77,81 @@
     {% endif %}
   </section>
 
-  <section class="govuk-!-margin-top-8">
-    <h2 class="govuk-heading-m govuk-!-margin-bottom-1">
-      {{ t("allocation::view.people.heading") }}
-    </h2>
+  {% if allocation.status != "cancelled" %}
+    <section class="govuk-!-margin-top-8">
+      <h2 class="govuk-heading-m govuk-!-margin-bottom-1">
+        {{ t("allocation::view.people.heading") }}
+      </h2>
 
-    <h3 class="govuk-heading-s">
-      {% if remainingCount === 0 %}
-        <span class="app-icon app-icon--tick app-green">{{ t("allocation::view.people.complete") }}</span>
-      {% elif canAccess('allocation:person:assign') %}
-        {{ t("allocation::view.people.progress", {
-          count: remainingCount,
-          context: "remaining"
-        }) }}
-      {% else %}
-        {{ t("allocation::view.people.progress", {
-          count: totalCount,
-          addedCount: addedCount
-        }) }}
-      {% endif %}
-    </h3>
+      <h3 class="govuk-heading-s">
+        {% if remainingCount === 0 %}
+          <span class="app-icon app-icon--tick app-green">{{ t("allocation::view.people.complete") }}</span>
+        {% elif canAccess('allocation:person:assign') %}
+          {{ t("allocation::view.people.progress", {
+            count: remainingCount,
+            context: "remaining"
+          }) }}
+        {% else %}
+          {{ t("allocation::view.people.progress", {
+            count: totalCount,
+            addedCount: addedCount
+          }) }}
+        {% endif %}
+      </h3>
 
-    {% if canAccess('allocation:person:assign') and remainingCount %}
-      <p>
-        {{ t("allocation::view.people.special_vehicle_instruction", {
-          href:"/move/new"
-        }) | safe }}
-      </p>
-
-      {{ govukButton({
-        href: "/move/" + unassignedMoveId + "/assign",
-        html: t("actions::add_person", {
-          context: "another" if addedCount !== 0
-        })
-      }) }}
-    {% endif %}
-  </section>
-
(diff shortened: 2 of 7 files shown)


common/services/allocation.js
```diff
@@ -83,8 +83,17 @@ const allocationService = {
   },
   transform({ includeCancelled = false } = {}) {
```
teneightfive, line 84: So, I think with this change you should be able to strip out all the work this option is doing. It was only really there to try and come up with an accurate number of moves on the dashboard. But if the API is now doing this correctly we can simplify a lot of this logic.

common/services/allocation.js
```diff
@@ -83,8 +83,17 @@ const allocationService = {
...
+      delete result.meta
+
       return {
         ...result,
+        totalSlots,
+        filledSlots,
+        unfilledSlots: totalSlots ? totalSlots - filledSlots : undefined,
         moves: result.moves
           .filter(
             move => includeCancelled || !['cancelled'].includes(move.status)
```
teneightfive, line 99: We may also be able to remove this as I don't think we need to do any logic with `allocations.moves`.

teneightfive (commented):
So with this change we should look at what we currently do with `allocation.moves` in terms of getting progress and replace with the new properties.

common/services/allocation.js
smoothcontract: Mentioned previously to @solidgoldpig - happy to include an `unfilled` count in the response meta if it means simplifying this code; it's trivial to add in the backend now and won't impact performance. Let me know if you'd like that 😄

teneightfive: @smoothcontract That might actually be nice just to avoid us having to calculate in certain situations.

The main purpose of this code though was to try and provide an accurate number for "cancelled" allocations. As with cancelled allocations we needed to not filter out cancelled moves to try and get an accurate original count. Otherwise they all just showed as `0 people`. But I'm guessing that even if the allocation is cancelled, the new meta data will show the number of requested moves before the allocation was cancelled?

solidgoldpig:
Gone back to the drawing board as the BE gets slightly overhauled

common/services/allocation.js
solidgoldpig: Will attempt to remove this when the single allocation is updated to return just the uncancelled moves

app/allocation/controllers/view.js
```diff
@@ -5,7 +5,7 @@ const presenters = require('../../../common/presenters')
 function viewAllocation(personEscortRecordFeature = false) {
   return (req, res) => {
     const { allocation, canAccess } = req
-    const { moves, status } = allocation
+    const { moves, status, totalSlots } = allocation
     const bannerStatuses = ['cancelled']
     const movesWithoutProfile = moves.filter(move => !move.profile)
```
teneightfive, line 10: Look at what these two variables (`movesWithoutProfile` and `movesWithProfile`) are trying to establish, it looks remarkably like `filledSlots` and `unfilledSlots`...could we replace them with the counts now coming from the API so the frontend does less of that heavy lifting?

teneightfive (approved):

app/allocation/controllers/view.js
solidgoldpig: Not yet. Because we don't have those values when we ask for a single allocation.... yet.

merlinc: @solidgoldpig If it's a separate improvement, maybe we should capture it on the backlog so we can keep track of it?

app/allocation/controllers/view.js
```diff
@@ -48,7 +48,7 @@ function viewAllocation(personEscortRecordFeature = false) {
       unassignedMoveId: movesWithoutProfile.length
         ? movesWithoutProfile[0].id
         : undefined,
-      totalCount: moves.length,
+      totalCount: totalSlots,
```
merlinc, line 51: Did this change not require any updates to the tests?

common/presenters/allocations-to-table-component.js
```diff
@@ -4,6 +4,13 @@ const componentService = require('../services/component')
 
 const tablePresenters = require('./table')
 
+function showProgress(data, showRemaining) {
```
merlinc, line 7: This function could do with some tests around it.

Also, this function and the ones below it look like they're pure functions, so should be easy to extract out and test individually.

common/services/allocation.js
merlinc: Again, I'd say if it's a separate improvement, maybe we should capture it on the backlog so we can keep track of it?

app/allocation/views/view.njk
```diff
@@ -77,79 +77,81 @@
...
-
-      {{ govukButton({
-        href: "/move/" + unassignedMoveId + "/assign",
-        html: t("actions::add_person", {
-          context: "another" if addedCount !== 0
-        })
-      }) }}
-    {% endif %}
-  </section>
+  {% if allocation.status != "cancelled" %}
```
merlinc, line 80: I don't expect anything in this PR, but it makes me sad that the current state of the browser tests being E2E tests mean they're too slow to verify these screen behave as we'd expect, and that there aren't any component tests we could leverage for similar testability.

merlinc (approved):
LGTM apart from some minor testing improvements - and it looks like both Code Climate and I are in agreement with wanting more tests, so I'm happy to approve, given you'll have to add them to make the robot happy :)

common/presenters/allocations-to-table-component.js
solidgoldpig: Covered the scenario. I would usually agree with you about how it should be done, but presenters take input and spit back output and I'm following the way its been done.

app/allocation/views/view.njk
solidgoldpig: It would also be nice if we had unit tests for the views and the conditionals were handled in the view/controller code ;)

PR #916, opened 2020-10-23
[P4-2294][P4-2066] Show previous answer for a Person Escort Record question

teneightfive:
### What changed

This adds the ability to show a previous answer to a question on the Person Escort Record journey.

This is done as part of the rendering step within the framework step controller.

### Why did it change

In some cases users are moving the same person multiple times within a short period, for example a move to court each day of the week. In many cases the answers to the assessment remain the same. This feature allows users to see the previous answer to a question and review it rather than having to fill the information in again.

### Issue tracking

- [P4-2294](https://dsdmoj.atlassian.net/browse/P4-2294)
- [P4-2066](https://dsdmoj.atlassian.net/browse/P4-2066)

## Screenshots

<kbd>booksecuremove-pr-916 herokuapp com_move_347296fb-ef85-4cf7-a5e5-d19b642ae2ad_person-escort-record_risk-information_security</kbd>

#### Automated testing

- [x] Added unit tests to cover changes
- [ ] Added end-to-end tests to cover any journey changes

#### Manual testing

Has been tested in the following browsers:

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

- [x] No environment variables were added or changed

### Other considerations

Commit messages with a `fix` or `feat` type are automatically used to generate the [changelog](ministryofjustice/hmpps-book-secure-move-frontend/blob/master/CHANGELOG.md) and
[GitHub release notes](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/releases) during the release task. Please make sure they will read well on their own in a
summary of changes and that the commit body gives a more detailed description of those changes if necessary.

- [x] No new [Personally Identifiable Information (PII)](https://support.google.com/analytics/answer/6366371?hl=en) is being sent via analytics
- [x] Update [README](ministryofjustice/hmpps-book-secure-move-frontend/blob/master/README.md) with any new instructions or tasks

PR #931, opened 2020-10-27
Support new frameworks version

teneightfive:
### What changed

This supports version 1.5.0 of the Person Escort Record framework,
which enables the ability to prefill from a previous record.

### Why did it change

To support the prefilling of the Person Escort Record we need to increase the latest version of the framework in the frontend.


jason-sims (approved):

merlinc (approved):

teneightfive:
This has been included into #916 so is no longer needed

PR #936, opened 2020-10-28
fix(per): Correct styling used to indicate when a question is prefilled

teneightfive:
### What changed

This change updates the styling to ensure the correct hierarchy and
that the correct classnames are used.

### Why did it change

The classnames were refactored but this meant some of the desired
styling was lost.

## Screenshots

| Before | After |
| ------ | ----- |


Andy-Millar (approved):
Thanks for super fast turnaround!

merlinc (approved):

solidgoldpig (approved):

PR #948, opened 2020-10-30
[P4-1420] feat: Add location-free-spaces service

merlinc:
## Proposed changes

Add a new service to connect to the `locations_free_spaces` endpoint. This endpoint provides availability data as meta properties based on the existing location reference data.

- `getLocationsFreeSpaces` - general method for all free spaces
- `getPrisonFreeSpaces` - convenience method for the main use case of prison spaces

Also contains model code and tests

### Issue tracking

- [P4-1420](https://dsdmoj.atlassian.net/browse/P4-1420)

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/lib/api-client/models.js
+++ b/common/lib/api-client/models.js
@@ -237,6 +237,11 @@ module.exports = {
       defaultInclude: ['suppliers'],
     },
   },
+  locations_free_spaces: {
+    options: {
+      cache: false,
+    },
+  },
   region: {
     fields: {
       key: '',
--- a/common/services/locations-free-spaces.js
+++ b/common/services/locations-free-spaces.js
@@ -1,10 +1,6 @@
-// https://hmpps-book-secure-move-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk/api/v1/allocations?filter[status]=unfilled&filter[locations]=a4db0d71-04d8-4313-b781-abc1530c2874%2C0c037872-48ad-42fa-84b2-b6f624480af2%2C7e8cc42f-0720-4123-8080-35d41798338c%2C5686a356-993b-47f0-ab1b-168e150410dc%2C186a2439-ee9e-4a34-8ddf-63ba52ba3ed1%2C056eed18-772b-45b3-b64a-4d7d563641f7%2Ce0bd4d4f-72f4-45e5-ba59-01d30dd4cdef%2C26bd697d-c90f-4d67-92d1-261df02f7f7a%2C16a549f8-8cd5-47e8-8272-2cfd5e459344%2Ce935e36b-6981-4213-ab34-bbfe36e6d566%2C689c75b2-23aa-4a0c-a331-33c43583a89a%2C28c06bf3-1791-4ed1-b58f-9bcdc656e16f%2C7366191b-fe72-4c01-a47f-f1a968bad97e%2Ccf12873e-545d-471a-b72f-cae6cec3a44f&filter[date_from]=2020-10-05&filter[date_to]=2020-10-11&page=1&per_page=1
-
 const { flattenDeep, sortBy } = require('lodash')
 
 const apiClient = require('../lib/api-client')()
-
-// getLocations({ filter, combinedData, page = 1, include } = {}) {
 
 const locationsFreeSpacesService = {
   getLocationsFreeSpaces({
--- a/common/services/locations-free-spaces.js
+++ b/common/services/locations-free-spaces.js
@@ -1,4 +1,4 @@
-const { flattenDeep, sortBy } = require('lodash')
+const { flattenDeep } = require('lodash')
 
 const apiClient = require('../lib/api-client')()
 
@@ -29,7 +29,7 @@ const locationsFreeSpacesService = {
         const hasNext = links.next && data.length !== 0
 
         if (!hasNext) {
-          return sortBy(locations, 'title')
+          return locations
         }
 
         return locationsFreeSpacesService.getLocationsFreeSpaces({
--- a/common/services/locations-free-spaces.js
+++ b/common/services/locations-free-spaces.js
@@ -42,6 +42,15 @@ const locationsFreeSpacesService = {
         })
       })
   },
+  getPrisonFreeSpaces({ dateFrom, dateTo } = {}) {
+    return locationsFreeSpacesService.getLocationsFreeSpaces({
+      dateFrom,
+      dateTo,
+      filter: {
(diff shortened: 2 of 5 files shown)


common/services/locations-free-spaces.js
```diff
@@ -0,0 +1,51 @@
...
+      .then(response => {
+        const { data, links } = response
+        const locations = combinedData
+          ? flattenDeep([combinedData, ...response.data])
+          : data
+
+        const hasNext = links.next && data.length !== 0
+
+        if (!hasNext) {
+          return sortBy(locations, 'title')
```
teneightfive, line 36: I wonder if it's better to do the sorting on the API side? But maybe it's not supported yet...?

common/services/locations-free-spaces.js
```diff
@@ -0,0 +1,51 @@
+// https://hmpps-book-secure-move-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk/api/v1/allocations?filter[status]=unfilled&filter[locations]=a4db0d71-04d8-4313-b781-abc1530c2874%2C0c037872-48ad-42fa-84b2-b6f624480af2%2C7e8cc42f-0720-4123-8080-35d41798338c%2C5686a356-993b-47f0-ab1b-168e150410dc%2C186a2439-ee9e-4a34-8ddf-63ba52ba3ed1%2C056eed18-772b-45b3-b64a-4d7d563641f7%2Ce0bd4d4f-72f4-45e5-ba59-01d30dd4cdef%2C26bd697d-c90f-4d67-92d1-261df02f7f7a%2C16a549f8-8cd5-47e8-8272-2cfd5e459344%2Ce935e36b-6981-4213-ab34-bbfe36e6d566%2C689c75b2-23aa-4a0c-a331-33c43583a89a%2C28c06bf3-1791-4ed1-b58f-9bcdc656e16f%2C7366191b-fe72-4c01-a47f-f1a968bad97e%2Ccf12873e-545d-471a-b72f-cae6cec3a44f&filter[date_from]=2020-10-05&filter[date_to]=2020-10-11&page=1&per_page=1
+
+const { flattenDeep, sortBy } = require('lodash')
+
+const apiClient = require('../lib/api-client')()
+
+// getLocations({ filter, combinedData, page = 1, include } = {}) {
```
teneightfive, line 7: Rogue comment?

common/services/locations-free-spaces.js
```diff
@@ -0,0 +1,51 @@
+// https://hmpps-book-secure-move-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk/api/v1/allocations?filter[status]=unfilled&filter[locations]=a4db0d71-04d8-4313-b781-abc1530c2874%2C0c037872-48ad-42fa-84b2-b6f624480af2%2C7e8cc42f-0720-4123-8080-35d41798338c%2C5686a356-993b-47f0-ab1b-168e150410dc%2C186a2439-ee9e-4a34-8ddf-63ba52ba3ed1%2C056eed18-772b-45b3-b64a-4d7d563641f7%2Ce0bd4d4f-72f4-45e5-ba59-01d30dd4cdef%2C26bd697d-c90f-4d67-92d1-261df02f7f7a%2C16a549f8-8cd5-47e8-8272-2cfd5e459344%2Ce935e36b-6981-4213-ab34-bbfe36e6d566%2C689c75b2-23aa-4a0c-a331-33c43583a89a%2C28c06bf3-1791-4ed1-b58f-9bcdc656e16f%2C7366191b-fe72-4c01-a47f-f1a968bad97e%2Ccf12873e-545d-471a-b72f-cae6cec3a44f&filter[date_from]=2020-10-05&filter[date_to]=2020-10-11&page=1&per_page=1
```
teneightfive, line 1: Rogue comment?

merlinc: Oops!

Deleted :)

common/services/locations-free-spaces.js
merlinc: Oops!

Deleted :)

common/services/locations-free-spaces.js
merlinc: Sorting of the location / establishment is part of the design, is supported on the backend , but it's being skipped for the moment while the rest of the functionality is built up.

I guess I could pass them through as default `sort[by]` and `sort[direction]` parameters, and then I don't need this particular sortBy line

merlinc: Done :)

teneightfive (approved):

common/services/locations-free-spaces.js
merlinc: To be clear, this has been removed from this PR and will be restored in a future one.

common/lib/api-client/models.js
```diff
@@ -237,6 +237,11 @@ module.exports = {
       defaultInclude: ['suppliers'],
     },
   },
+  locations_free_spaces: {
```
merlinc, line 240: As far as I can tell, adding this model made no difference - however it does seem safer to tie down our api-client to using known endpoints

common/lib/api-client/models.js
```diff
@@ -237,6 +237,11 @@ module.exports = {
...
+    options: {
```
merlinc, line 241: Are there any other model configuration options I should be including here? Default property names, types? It's a shallow data object as we only really care about the location name and the associated meta object with availability.

teneightfive: `defaultInclude` is the main one that you might want to define. This is passed to the api client as the list of resources to include when making a request.

I'm guessing there might be some relationships with your call so you might want to define that. Alternatively, this can be defined in the services and then you don't need to worry about.

This work was done to allow the transition from v1 to v2 of the API where v1 always included everything by default and v2 included nothing. So this addition was to allow us to switch with less refactoring elsewhere.

teneightfive (approved):

common/lib/api-client/models.js
merlinc: `locations_free_spaces.getLocationFreeSpaces` is very shallow as it's a jumping point onto other parts of the system:

_sample request url:_
`${host}/api/v1/locations_free_spaces?filter[location_type]=prison&filter[location_id]=${locationIds}&page=1&per_page=100&date_from=2020-11-02&date_to=2020-11-08`

so I don't think there's any relationships to be defined, at least at the moment.

PR #949, opened 2020-10-30
fix(per): Ensure add another responses are rendered correctly

teneightfive:
### What changed

This passes a responded property to the framework response component
for add another fields.

### Why did it change

The framework response component was updated to know about both
responded and prefilled and now requires that the value be passed
as true in order to render the value.

### Issue tracking

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/framework-field-summary-list-row.test.js
+++ b/common/presenters/framework-field-summary-list-row.test.js
@@ -611,7 +611,7 @@ describe('Presenters', function () {
               {
                 value: test.value,
                 valueType: test.valueType,
-                responded: false,
+                responded: true,
                 prefilled: false,
                 questionUrl: `${mockStepUrl}#${test.id}`,
               }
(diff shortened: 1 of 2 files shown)


common/presenters/framework-field-summary-list-row.test.js
```diff
@@ -611,7 +611,7 @@ describe('Presenters', function () {
               {
                 value: test.value,
                 valueType: test.valueType,
-                responded: false,
+                responded: true,
```
merlinc, line 614: Do you need a test for the `false` value too?

teneightfive: In this instance it should never actually be `false`. That was the issue. When this presenter is called, for the add another fields it should always set responded to `true` so that the value gets displayed.

cwrw (approved):
tested on heroku app and looks good, thanks for the quick fix

PR #787, opened 2020-09-16
chore: Ignore Jetbrains IDE settings

merlinc:
### Proposed changes

Ignore Jetbrains IDE settings (IntelliJ & Webstorm)

### Issue tracking

- [P4-XXXX]()

#### Automated testing

- [ ] Added unit tests to cover changes
- [ ] Added end-to-end tests to cover any journey changes

#### Manual testing

Has been tested in the following browsers:

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

- [x] No environment variables were added or changed

### Other considerations

Commit messages with a `fix` or `feat` type are automatically used to generate the [changelog](ministryofjustice/hmpps-book-secure-move-frontend/blob/master/CHANGELOG.md) and
[GitHub release notes](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/releases) during the release task. Please make sure they will read well on their own in a
summary of changes and that the commit body gives a more detailed description of those changes if necessary.

- [x] No new [Personally Identifiable Information (PII)](https://support.google.com/analytics/answer/6366371?hl=en) is being sent via analytics

PR #963, opened 2020-11-05
[P4-2400] Add clear-cache script

Thource:
### What changed

Added clear-cache-reference-data.js, a script that can be used to clear reference data out of the Redis cache.

### Why did it change

This has been added because currently, all reference data is being cached for 1 week. While this is fine for data that won't change often (gender, ethnicity, etc), it's not great for data that can change more often (locations, etc).

This script can be used manually, or scheduled as a task in CircleCI and can be ran with the following:

`node tasks/clear-cache-reference-data.js DATA_NAME`

OR

`npm run clear-cache-reference-data -- DATA_NAME`

`DATA_NAME` can be replaced with any reference data name, full list:
```
genders
ethnicities
assessment_questions
locations
regions
suppliers
prison_transfer_reasons
allocation_complex_cases
\*
```

### Issue tracking
- [P4-2400](https://dsdmoj.atlassian.net/browse/P4-2400)

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/lib/clear-cache-reference-data.js
+++ b/common/lib/clear-cache-reference-data.js
@@ -1,0 +1,26 @@
+const CONFIG = require('../../config')
+const redisStore = require('../../config/redis-store.js')
+
+const scanAndDelete = (cursor, pattern) => {
+  const client = redisStore().client
+  return client
+    .scanAsync(cursor, 'MATCH', pattern, 'COUNT', '100')
+    .then(async ([nextCursor, keys]) => {
+      await Promise.all(keys.map(key => client.del(key)))
+
+      if (nextCursor !== '0') {
+        return keys.length + (await scanAndDelete(nextCursor, pattern))
+      }
+
+      return keys.length
+    })
+}
+
+const clearCacheReferenceData = (referenceName = '') => {
+  return scanAndDelete(
+    '0',
+    `cache:v${CONFIG.API.VERSION}:GET./api/reference/${referenceName}*`
+  )
+}
+
+module.exports = clearCacheReferenceData
--- a/common/lib/clear-cache-reference-data.test.js
+++ b/common/lib/clear-cache-reference-data.test.js
@@ -1,0 +1,133 @@
+const proxyquire = require('proxyquire').noPreserveCache()
+
+const API_VERSION = 99
+
+const redisClient = {}
+
+const clearCacheReferenceData = proxyquire('./clear-cache-reference-data', {
+  '../../config/redis-store.js': () => {
+    return {
+      client: redisClient,
+    }
+  },
+  '../../config': {
+    API: {
+      VERSION: API_VERSION,
+    },
+  },
+})
+
+describe('Clear cache reference data', function () {
+  beforeEach(function () {
+    redisClient.scanAsyncData = []
+    redisClient.scanAsync = sinon.stub().returns({
+      then: f => f(redisClient.scanAsyncData.shift() || ['0', []]),
+    })
+    redisClient.del = sinon.stub().resolves()
+  })
+
+  context('when called without an arg', function () {
+    beforeEach(function () {
+      redisClient.scanAsyncData = [
+        ['1', [`cache:v${API_VERSION}:GET./api/reference/locations/2`]],
+        [
+          '0',
+          [
+            `cache:v${API_VERSION}:GET./api/reference/locations/1`,
+            `cache:v${API_VERSION}:GET./api/reference/genders/1`,
+          ],
+        ],
+      ]
+    })
+
+    it('returns the deleted count and calls the redis functions with the correct args', async function () {
+      expect(await clearCacheReferenceData()).to.eq(3)
+      expect(redisClient.scanAsync).to.have.been.be.calledWithExactly(
+        '0',
+        'MATCH',
+        `cache:v${API_VERSION}:GET./api/reference/*`,
+        'COUNT',
+        '100'
+      )
+      expect(redisClient.scanAsync).to.have.been.be.calledWithExactly(
+        '1',
+        'MATCH',
+        `cache:v${API_VERSION}:GET./api/reference/*`,
+        'COUNT',
+        '100'
+      )
+      expect(redisClient.scanAsync.callCount).to.eq(2)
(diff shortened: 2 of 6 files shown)


solidgoldpig (approved):

.gitignore
```diff
@@ -72,3 +72,6 @@ typings/
 
 # Test cafe reports
 report*.json
+
+# JetBrains IDE config
+.idea/
```
teneightfive, line 77: I think this convo came up before actually but it might be better to keep IDE specific configs in your local `.gitignore` file. It feels more the responsibility of the individual developer than the project, as we can't know what every developer's preferred IDE choice is.

solidgoldpig: See

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/437#discussion_r412227978

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/787

But it's up to Adam and Nev now, innit?

Thource: I agree with this, I'll change it

tasks/clear-cache-reference-data.test.js
```diff
@@ -0,0 +1,118 @@
...
+      it('logs some useful information and exits with code 0', async function () {
+        await main()
+
+        expect(process.stdout.write).to.have.been.called
+        expect(process.exit).to.have.been.calledOnceWithExactly(0)
+      })
+    })
+  })
+
+  context('when called without an arg', function () {
```
Nimphal, line 76: Is this different from the context above `when no arg is passed`?

Thource: The contexts above are in `when ran via the command line` and call `main` as to emulate the script being ran from the command line, instead of directly calling `clearCacheReferenceData`

tasks/clear-cache-reference-data.js
```diff
@@ -0,0 +1,57 @@
+const redisStore = require('../config/redis-store.js')
+
+const scanAndDelete = async (cursor, pattern) => {
+  const client = await redisStore().client
+  return await client
```
teneightfive, line 5: I don't think you need the `await` if you are returning a promise in this instance.

tasks/clear-cache-reference-data.js
```diff
@@ -0,0 +1,57 @@
+const redisStore = require('../config/redis-store.js')
+
+const scanAndDelete = async (cursor, pattern) => {
+  const client = await redisStore().client
```
teneightfive, line 4: `client` doesn't return a promise, so I don't think you need the `await` when defining this variable.

tasks/clear-cache-reference-data.js
```diff
@@ -0,0 +1,57 @@
...
+  return await client
+    .scanAsync(cursor, 'MATCH', pattern, 'COUNT', '100')
+    .then(async ([nextCursor, keys]) => {
+      keys.forEach(key => client.del(key))
```
teneightfive, line 8: Does `client.del` return a promise?

tasks/clear-cache-reference-data.js
```diff
@@ -0,0 +1,57 @@
...
+      return keys.length
+    })
+}
+
+const clearCacheReferenceData = async referenceName => {
+  if (!referenceName) {
+    return -1
+  }
+
+  return await scanAndDelete(
```
teneightfive, line 23: Same here. If you're returning a promise you don't need to use `async/await`.

tasks/clear-cache-reference-data.js
```diff
@@ -0,0 +1,57 @@
...
+    '0',
+    `cache:v2:GET./api/reference/${referenceName}*`
```
teneightfive, line 25: You might want to incorporate the configuration variables that are used [when building this key](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/c1c8c3ae9967b4dd9c17bc94aa8b311bfbb70ba9/common/lib/api-client/middleware/cache-key.js#L18).

If we change the version on the API or the API path I don't think this will work any longer.

teneightfive (changes requested):

README.md
```diff
@@ -76,6 +76,29 @@ npm run develop
 
 The app will be running at [http://localhost:3000](http://localhost:3000) or with browsersync at [http://localhost:3001](http://localhost:3001).
 
+### Clearing cached reference data
```
teneightfive, line 79: Just thinking about the placement of this. @merlinc did a bit of work to split out our documentation and try and clean up the base readme. I wonder if this should live somewhere in the [docs folder](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/master/docs), maybe `development.md` or `releasing.md`?

Thource: Moved this to releasing.md

tasks/clear-cache-reference-data.js
Thource: I've added the version number from the config, but I couldn't find `api/reference` defined anywhere

tasks/clear-cache-reference-data.js
Thource: It does, resolved with Promise.all

common/lib/clear-cache-reference-data.js
```diff
@@ -0,0 +1,26 @@
+const CONFIG = require('../../config')
+const redisStore = require('../../config/redis-store.js')
+
+const scanAndDelete = async (cursor, pattern) => {
```
teneightfive, line 4: The `async` can be removed not that there is no `await` within this function.

common/lib/clear-cache-reference-data.test.js
```diff
@@ -0,0 +1,128 @@
+const proxyquire = require('proxyquire').noPreserveCache()
+
+const API_VERSION = require('../../config').API.VERSION
```
teneightfive, line 3: It might be better to mock this within the `proxyquire` call too.

Thource: I did consider it, done

teneightfive (approved):

PR #969, opened 2020-11-09
[P4-1437] Add daily free spaces page

merlinc:
## Proposed changes

This adds the daily free spaces page to the population part of the application. For a given day this will display all free space data and a count of transfers in and out.f

A subsequent PR will deal with editing this data.

### Issue tracking

- [P4-1437](https://dsdmoj.atlassian.net/browse/P4-1437)

#### With free space data
Screenshot_2020-11-19 Population for Monday 9 Nov 2020

#### Without free space data
Screenshot_2020-11-19 Population for Tuesday 10 Nov 2020

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


config/index.js
```diff
@@ -98,6 +98,7 @@ module.exports = {
     SHORT: 'd M yyyy',
     LONG: 'd MMM yyyy',
     WITH_DAY: 'EEEE d MMM yyyy',
+    WITH_DAY_AND_TIME: 'EEEE d MMM yyyy, h:mm:ss aa',
```
teneightfive, line 101: I wonder if the seconds are ever really needed in this case? 

The format we've used elsewhere within the service when there is a time and date has been `(at) 3:30pm on Wed 10 Oct 2020`. This is a little more aligned with the [GOV.UK content guidelines](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times) too.

But the addition here is a good one, and maybe a format from config should be used to ensure that consistency.

locales/en/population.json
```diff
@@ -1,4 +1,25 @@
 {
+  "daily_view": {
+    "missing_value": "-",
```
teneightfive, line 3: Having a look at the screenshots I'm wondering if placing a `-` is the best way to communicate that we don't have any values for this item yet.

It might be better to replace the data layouts with a single message saying that there are no values for that location and to include a call to action (I'm assuming to the edit step?) to encourage that data to be provided?

app/population/middleware/set-body.free-spaces.js
```diff
@@ -0,0 +1,48 @@
...
+    const { populationId } = req.body.population
+
+    if (!populationId) {
+      req.body.population.freeSpaces = {
+        date: req.params.date,
+      }
+      return next()
+    }
+
+    const {
```
teneightfive, line 14: Is there a need to destructure all of these variables and pass them in again?

Is `req.body.population.freeSpaces` used anywhere else other than the view? 

It might be good to jump on a call and walk through some of this as I think it strays a little from how things have been set within the middleware/controller/template layers.

app/population/view/daily.njk
```diff
@@ -0,0 +1,65 @@
...
+      <h1 class="govuk-heading-m">{{ t("population::daily_view.labels.free_spaces") }}</h1>
+      <h2 class="govuk-heading-xl">{{ t("population::spaces_with_count", {count:resultsAsDailySummary.freeSpaces}) if resultsAsDailySummary.freeSpaces else t("population::daily_view.not_calculated") }}</h2>
+      <p class="govuk-body">{{ t("population::daily_view.free_space_description") }}</p>
+      <p class="govuk-body"><b>{{  t("population::daily_view.labels.last_updated") }} {{  resultsAsDailySummary.updatedAt | formatDateWithDayAndTime if resultsAsDailySummary.updatedAt else t("population::daily_view.not_updated")}}</b></p>
+    </div>
+  </div>
+  {%  endcall %}
+
+  <div class="govuk-grid-row govuk-!-padding-top-4">
+    <div class="govuk-grid-column-one-half">
```
teneightfive, line 27: This block looks like it contains quite a bit of duplication, especially around markup.

I wonder if we can extract some of this into the controller/a presenter so that view gets a list of section/data groups that it is then responsible for displaying.

From looking at the screenshot I also wonder if we could re-use the [data component](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/master/common/components/data) that we already have to display this information. 

(I'm aware the lack of ability to view all components probably made it harder to realise this existed though)

But it might be worth a chat with @Andy-Millar about re-using that to display this information.

It's actually currently used to render the location here, but can easily be re-used in this instance:

<kbd>image</kbd>

Another thing to consider would be flipping the number/label order so that the number appears first. This is something that benefits how things read (for example, with a screenreader when there is no visual style). Ie, `5 overnights in` reads more naturally than `overnights in 5`. 

Here's some examples of other parts of govt that use this pattern for data/statistics:

http://alphagov.github.io/digitalmarketplace-frontend-toolkit/statistics.html
https://govuk-elements.herokuapp.com/data/#data-visualisation

And here's another example of GOV.UK Notify summarise counts:

<kbd>dashboard-summary</kbd>

config/index.js
Andy-Millar: Agree - match the format we have globally

locales/en/population.json
Andy-Millar: Agree - dealing with this with content is better. 
We say "No moves" on the dashboard when there are no moves booked for that day, and "No Health information warnings" when no info has been added to the PER - I think matching this would be best. "No numbers entered"? 

Alternatively we could use it to promote adding the number - ie. we say "Warnings will display once a Person Escort Record has been completed" on the record page... so could be "To be completed"?

Yeap, there's an edit state to be added.

app/population/view/daily.njk
Andy-Millar: Not sure I agree with reusing (top screenshot). If we can have the font size-m (not small, as in the card screenshot). If that's not an issue though, then all good to reuse.

Don't flip the label - since we are having the edit state trigger everything into fields but in single column. This page isn't really to be a dashboard - it's just showing ins and outs in a concise way, allowing users to enter / update numbers as required.

locales/en/population.json
teneightfive: >We say "No moves" on the dashboard when there are no moves booked for that day, and "No Health information warnings" when no info has been added to the PER - I think matching this would be best. "No numbers entered"?

Yep, I think this would be a nice pattern to re-use. And within that banner style we can add a call to action to "add numbers"?

Andy-Millar: Yeah, that's exactly what it is with the edit pattern - green button

app/population/view/daily.njk
teneightfive: >Not sure I agree with reusing (top screenshot). If we can have the font size-m (not small, as in the card screenshot). If that's not an issue though, then all good to reuse.

Yes, sorry I meant more just re-use the component in terms of how we write the HTML in the markup. We have a few "modifiers" that we can attach to that component to tweak is visually, for things like font size or font weight so I think we can try and match it visually. But it would be good if we can re-use it from a code perspective.

>Don't flip the label - since we are having the edit state trigger everything into fields but in single column. This page isn't really to be a dashboard - it's just showing ins and outs in a concise way, allowing users to enter / update numbers as required.

Sure. But would it not be okay to still provide a different view when editing? A lot of our other screens don't match exactly to how the data is captured.

app/population/middleware/set-body.free-spaces.js
solidgoldpig: what about something like?

```js
const { camelCase } = require('lodash')
...
const population = await populationService.getById(populationId)
const freeSpaces = Object.keys(population).reduce((acc, key) => {
  acc[camelCase(key)] = pop[key]
  return acc
}, {})
```

app/population/view/daily.njk
Andy-Millar: Reusing code - gotcha all good.

Flipping - no, the user anchors to the data title then the number. Leave as spec-ed please

locales/en/population.json
Andy-Millar: To confirm: "No numbers entered"

app/population/middleware/set-body.free-spaces.js
teneightfive: I think my question is more around whether we actually _need_ to do this?

app/population/view/daily.njk
teneightfive: > Flipping - no, the user anchors to the data title then the number. Leave as spec-ed please

@Andy-Millar Ok re unavailable space counts. But...could we maybe flip it for the summary of the total spaces though? Doesn't look like the user enters this figure so doesn't need the form field "anchor" and it feels like using the same pattern used across government would be good in this instance?

So 
```
70
free spaces
```
instead of 
```
Free spaces
70 spaces
```

Andy-Millar: I'd dispute that it is used _across_ government - both patterns are used widely depending on the context.

But...

70
free spaces

...is fine though. 

Would perhaps be better to do away with both options actually, and just go for:

70 free spaces

...all on one line?

Andy-Millar: ie.

Screenshot 2020-11-12 at 16 11 59

teneightfive: Yep, also an option actually. I just find that it reads better (if you take away the styling) by having the number first instead of second.

merlinc: @teneightfive The `data` component displays its content with different sized text - I think the example you're using is the `card`, which is using some custom styles on `dt/dd` tags - https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/components/card/template.njk#L39-L54

As an example: 

Screenshot from 2020-11-16 16-12-27

From left to right we have:

- Data component
- Card component

and then

- Previous implementation

merlinc: @Andy-Millar I've updated the "free spaces panel" to look more like this now:

Screenshot from 2020-11-16 16-12-48

config/index.js
merlinc: I've changed this to match the "time at date" format, as seen in the screenshot below

locales/en/population.json
merlinc: The dashboard with no moves is showing "0 outgoing moves". 

Screenshot from 2020-11-16 16-22-06

Also, I'm not sure what this would look like in practice as there are multiple rows of data - so I've currently hidden the sections with no numbers. It looks like this:

Screenshot from 2020-11-16 16-19-14

@Andy-Millar Can you make a quick mockup of what you'd like to have displayed, and attach it to the Jira story - then I'll make it match up accordingly.

Andy-Millar: > The dashboard with no moves is showing "0 outgoing moves".
> 
> Screenshot from 2020-11-16 16-22-06

- I thought we agreed to just go straight to the overview without the blue blocks?

IPTC-MVP-00-01 Copy

Andy-Millar: When there are no numbers I think we should do this:

Screenshot 2020-11-17 at 10 13 55

Alternatively could do this, but I prefer the above option:

Screenshot 2020-11-17 at 10 14 14

app/population/view/daily.njk
Andy-Millar: Looks good to me

locales/en/population.json
Andy-Millar: Could also argue that once the numbers are added, the 'Update numbers' button becomes a link - 'Change numbers' - more in line with what we do elsewhere

Andy-Millar: ie:

Screenshot 2020-11-17 at 10 19 27

app/population/view/daily.njk
```diff
@@ -0,0 +1,140 @@
...
+{% endblock %}
+
+{% block content %}
+  <header>
+    <h1 class="govuk-heading-xl">
+      {{ resultsAsDailySummary.date | formatDateWithDay }}
+    </h1>
+  </header>
+
+  {%  call appPanel({classes: 'govuk-grid-row govuk-!-margin-bottom-4 govuk-!-padding-bottom-0'}) %}
```
teneightfive, line 16: You can drop the `govuk-grid-row` class as this is causing a double negative margin and the panel to sit outside of the content area.

app/population/index.js
```diff
@@ -6,14 +6,29 @@ const {
...
+  setResultsAsDailySummary,
+  setBodyPopulationId,
+  setBodyFreeSpaces,
+} = require('./middleware')
 
 router.param('date', setDateRange)
 
 router.get('/', redirectBaseUrl)
 
+router.use(
```
teneightfive, line 23: Should this be `router.get`?

app/population/index.js
```diff
@@ -6,14 +6,29 @@ const {
...
+  DAILY_PATH,
+  setContext('population'),
+  setBodyPopulationId,
+  setBodyFreeSpaces,
+  setResultsAsDailySummary,
+  daily
+)
+
 router.use(
```
teneightfive, line 32: This should probably also be `router.get`

app/population/constants.js
```diff
@@ -1,10 +1,13 @@
-const { dateRegex } = require('../../common/helpers/url')
+const { dateRegex, uuidRegex } = require('../../common/helpers/url')
 
 const MOUNTPATH = '/population'
 
 const BASE_PATH = `/:period(week|day)/:date(${dateRegex})`
 
+const DAILY_PATH = `/day/:date(${dateRegex})/:locationId(${uuidRegex})`
```
teneightfive, line 7: Does this need to be a constant?

common/lib/api-client/models/population.js
```diff
@@ -0,0 +1,20 @@
...
+    overnights_out: '',
+    out_of_area_courts: '',
+    discharges: '',
+    free_spaces: '',
+    updated_by: '',
+    created_at: '',
+    updated_at: '',
+  },
+  options: {
+    cache: false,
```
teneightfive, line 18: The default is false so you can drop the whole options object for this model.

app/population/middleware/set-population-id.js
```diff
@@ -0,0 +1,24 @@
+const locationsFreeSpacesService = require('../../../common/services/locations-free-spaces')
+
+async function setResultsPopulationTable(req, res, next) {
```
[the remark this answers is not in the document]
teneightfive: With [router.param middleware](https://expressjs.com/en/api.html#router.param) you can also use the 4th argument as the value. So this could become:

```javascript
async function setPopulationId(req, res, next, locationId) {
  ...
}
```

app/population/middleware/set-population-id.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    const { locationId, date } = req.params
+
+    const freeSpaces = await locationsFreeSpacesService.getPrisonFreeSpaces({
+      dateFrom: date,
+      dateTo: date,
+      locationIds: locationId,
+    })
+
+    const population = freeSpaces?.[0]?.meta?.populations?.[0] || {}
```
teneightfive, line 13: Looking at this, can a location have multiple populations? Are we sure the first item will always be the one we want?

app/population/middleware/set-population-id.js
```diff
@@ -0,0 +1,24 @@
...
+    const { id: populationId } = population
```
teneightfive, line 14: Could this be wrapped in a check so that `req.populationId` isn't set if there is no poplulation?

teneightfive: Although, looking at the test it will still be set as undefined I guess so maybe it doesn't need it.

app/population/middleware/set-results.daily-population.js
```diff
@@ -0,0 +1,82 @@
+const populationService = require('../../../common/services/population')
+
+async function setResultsDailyPopulation(req, res, next) {
+  try {
+    req.transfersIn = [
```
teneightfive, line 5: This looks like dummy data? I wonder if this should/could be added separately/later on when this exists?

teneightfive: This will avoid it accidentally living for ever in a live environment...

app/population/middleware/set-results.daily-population.js
```diff
@@ -0,0 +1,82 @@
...
+      {
+        establishment: 'Adipiscing Elit',
+        count: 1,
+      },
+    ]
+
+    const { populationId } = req
+
+    if (!populationId) {
+      req.details = {
```
teneightfive, line 34: Is there any need to assign this to a new object? Can't you just access `req.params` in the controller and pass the values to the view that way?

app/population/middleware/set-results.daily-population.js
```diff
@@ -0,0 +1,82 @@
+const populationService = require('../../../common/services/population')
+
+async function setResultsDailyPopulation(req, res, next) {
```
teneightfive, line 3: Similar to some other comments, I think we shouldn't confused the collection concept with this part of the app so we can do away with `set-results.{thing}.js`.

I wonder if we even need to do this here too and whether we can assign the population in `set-population-id.js`. If could simplify a lot of this and we can handle some of the data structure directly in the controller.

Let's pair after standup and I can explain more what I mean.

locales/en/population.json
merlinc: Resolving this thread based upon separate conversations with @teneightfive  and @Andy-Millar

app/population/middleware/set-population-id.js
merlinc: This request only retrieves one day worth of data, so the first item is definitely the one we want.

app/population/middleware/set-results.daily-population.js
merlinc: This is sample data to show what the empty screen looks like, it will be removed before merging the PR in.

app/population/middleware/set-results.daily-population.js
merlinc: Resolving as this has been discussed during pairing

app/population/middleware/set-results.daily-population.js
merlinc: Resolving as this has been discussed during pairing

app/population/middleware/set-population-id.js
merlinc: Resolving as this has been discussed during pairing and been changed into a "setPopulation" param middleware

app/population/middleware/set-population-id.js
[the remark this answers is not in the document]
merlinc: This middleware needs both the passed in param and the date param which is not passed in - so there's not a lot of point using one as an argument, and then retrieving the other direct from req.params

app/population/middleware/set-results.daily-summary.js
```diff
@@ -0,0 +1,7 @@
+async function setResultsDailySummmary(req, res, next) {
+  req.resultsAsDailySummary = req.body.population?.freeSpaces
```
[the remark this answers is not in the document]
merlinc: Middleware has been removed.

app/population/view/daily.njk
```diff
@@ -0,0 +1,119 @@
...
+  {%  for transfer in transfers %}
+    {{ transfer.count }} from {{  transfer.establishment }}<br>
+  {% endfor %}
+  </p>
+{% endmacro %}
+
+{% set displayDate = details.date | formatDateWithRelativeDay %}
+
+{% block pageTitle %}
+  {{ t("collections::page_title_with_date", {
```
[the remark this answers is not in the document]
merlinc: Moved these usages into the population locale file.

locales/en/collections.json
```diff
@@ -8,12 +8,13 @@
   "page_title_with_date_allocations": "Allocations for {{date}}",
   "page_title_with_date_outgoing_moves": "Outgoing moves for {{date}}",
   "page_title_with_date_single_requests": "Single requests sent {{date}}",
+  "page_title_with_date_population": "Population for {{date}}",
```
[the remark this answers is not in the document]
merlinc: Moved these into the population locales file

common/lib/api-client/models/population.js
merlinc: Removed property from model

Andy-Millar:
Small one for me on the "Change numbers" link - can we make a size bigger (so it's the same size as "Projected... etc" and regular weight not bold please @merlinc 

I've noticed we use a mix of weights for change links - regular on the record and bold on the PER 'overview' pages - Can't see why we would use diff weights - I'm of the view we should be using regular weight globally (cc @teneightfive)

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
...
+  <h3 class="govuk-heading-m">{{ label }}</h3>
+
+  <p class="govuk-body">
+  {%  for transfer in transfers %}
+    {{ transfer.count }} from {{  transfer.establishment }}<br>
+  {% endfor %}
+  </p>
+{% endmacro %}
+
+{% set displayDate = details.date if details.date else date | formatDateWithRelativeDay %}
```
teneightfive, line 35: This ternary looks like a hangover from the previous implementation. `details` is gone so this could become:

```nunjucks
{% set displayDate = date | formatDateWithRelativeDay %}
```

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
...
+  }) }}
+{% endblock %}
+
+
+{% block content %}
+  <header class="govuk-grid-row govuk-!-margin-bottom-7">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl govuk-!-margin-bottom-2">
+        <span class="govuk-caption-xl">
+            {{ t("population::daily_view.subheading") }}
```
teneightfive, line 49: Some misaligned indentation here that it would be nice to sort out.

app/population/index.js
```diff
@@ -1,20 +1,30 @@
...
+  setResultsAsPopulationTable,
+} = require('./middleware')
 
 router.param('date', setDateRange)
+router.param('locationId', setPopulationId)
 
 router.get('/', redirectBaseUrl)
 
-router.use(
+dailyRouter.get('/', setContext('population'), daily)
```
teneightfive, line 23: I think you can remove `setContext` now as it's no longer being used in this view.

app/population/index.js
```diff
@@ -1,20 +1,30 @@
...
+
+router.use(DAILY_PATH, dailyRouter)
+
+router.get(
   BASE_PATH,
   setContext('population'),
```
teneightfive, line 29: Obviously not necessarily related to this specific change but I wonder if `setContext` is serving any purpose on this route either?

app/population/middleware/index.js
```diff
@@ -1,7 +1,9 @@
 const redirectBaseUrl = require('./redirect-base-url')
+const setPopulationId = require('./set-population')
```
teneightfive, line 2: Should this be `setPopulation` instead of `setPopulationId`?

app/population/controllers/daily.js
```diff
@@ -0,0 +1,17 @@
+const populationToGrid = require('../../../common/presenters/population-to-grid')
+const transfersToGrid = require('../../../common/presenters/transfers-to-grid')
+
+function daily(req, res) {
+  const { population } = req
+  const transfers = transfersToGrid()
+
```
teneightfive, line 5: Any reason for this linebreak? I thought eslint should catch things like this...

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
+{% extends "layouts/base.njk" %}
+{% from "panel/macro.njk" import appPanel %}
```
teneightfive, line 2: These are already defined in [`base.njk`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/098a1522281d1d26953114c2d534c5f80d1bb09e/common/templates/layouts/base.njk#L37) so you shouldn't need to declare them here too.

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
...
+{% from "data/macro.njk" import appData %}
```
teneightfive, line 3: It doesn't look like you're using this anymore but it's also already defined globally so you shouldn't need to define it again.

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
...
+
+{% macro transfersGroup(headingProperty, transfers) %}
+  {%  set key = "population::daily_view.labels." + headingProperty %}
+  {%  set label = t(key) %}
+
+  <h3 class="govuk-heading-m">{{ label }}</h3>
+
+  <p class="govuk-body">
+  {%  for transfer in transfers %}
+    {{ transfer.count }} from {{  transfer.establishment }}<br>
```
teneightfive: After our discussion around counts I wonder if this macro can be removed in favour of `statsGroup` as it looks very similar apart from this line.

And if transfers now also display a number they will be the same. And a url could be pass in as a 3rd argument.

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
...
+          <div>
+            <a href="" class="govuk-link">
+              <strong>{{ t("population::daily_view.actions.change_numbers", {
+                context: "with_visually_hidden_text"
+              }) | safe }}</strong>
+            </a>
+          </div>
+        {% endif %}
+
+        {%  if not spaces.details.updated_at %}
```
teneightfive, line 73: Any reason this can't be an `else` instead of 2 separate `if` blocks?

app/population/view/daily.njk
```diff
@@ -0,0 +1,116 @@
...
+          </div>
+
+          <p class="govuk-body">{{ t("population::daily_view.free_space_description") }}</p>
+        {%  endif %}
+      </div>
+    </div>
+  {%  endcall %}
+
+  {% if spaces.details.updated_at %}
+  <div class="govuk-grid-row">
```
teneightfive, line 89: Could do with being knocked 2 spaces :)

app/population/view/daily.njk
```diff
@@ -0,0 +1,114 @@
...
+{% endmacro %}
+
+{% macro transfersGroup(headingProperty, transfers) %}
+  {%  set key = "population::daily_view.labels." + headingProperty %}
+  {%  set label = t(key) %}
+
+  <h3 class="govuk-heading-m">{{ label }}</h3>
+
+  <p class="govuk-body">
+    {{  transfers }}
```
teneightfive, line 31: This feels a little similar to the macro above (`statsGroup`). Is there a way these could actually be shared or made to be similar?

common/presenters/population-to-grid.js
```diff
@@ -0,0 +1,34 @@
+function populationToGrid({ population } = {}) {
+  if (population === undefined) {
+    return {}
+  }
+
+  return {
+    details: {
+      date: population.date,
```
teneightfive, line 8: Does date actually need to come from here? I'm guessing (in theory) this should match the date from the `req.params` because it was returned as part of the search on that date. 

So maybe for consistency and simplicity the view just uses the one variable?

app/population/view/daily.njk
```diff
@@ -0,0 +1,114 @@
+{% extends "layouts/base.njk" %}
+{% from "panel/macro.njk" import appPanel %}
+{% from "data/macro.njk" import appData %}
+
+{%  macro statsEntry(property, value) %}
+  {%  set key = "population::daily_view.labels." + property %}
+  {%  set label = t(key) %}
+
+  <h4 class="govuk-heading-s">{{ label }}</h4>
```
teneightfive, line 9: From the screenshots it looks like the numbers are a little disconnected from this heading visually. It might be nice to see if you can use some [spacing helper classes](https://design-system.service.gov.uk/styles/spacing/) to create a tighter visual link.

app/population/middleware/set-population.js
```diff
@@ -0,0 +1,58 @@
...
+    if (populationId) {
+      req.population = await populationService.getById(populationId, {
+        include: ['moves_from', 'moves_to', 'location'],
+      })
+
+      transfersIn = req.population.moves_to.length
+      transfersOut = req.population.moves_from.length
+    }
+
+    if (!req.population) {
```
teneightfive, line 32: This could be an `if/else`?

app/population/middleware/set-population.js
```diff
@@ -0,0 +1,58 @@
...
+      }
+    )
+
+    const freeSpacePopulation =
+      dailyFreeSpace?.[0]?.meta?.populations?.[0] || {}
+    const { id: populationId } = freeSpacePopulation
+
+    if (populationId) {
+      req.population = await populationService.getById(populationId, {
+        include: ['moves_from', 'moves_to', 'location'],
```
teneightfive, line 25: So, previously we've tried to avoid defining includes within middleware/controllers and to keep these hidden within the service layer (which is responsible for interacting with the API).

I think it would be safer, and cleaner to set that include within `populationService.getById` rather than in this middleware.

app/population/middleware/set-population.js
```diff
@@ -0,0 +1,58 @@
...
+      })
+
+      transfersIn = req.population.moves_to.length
+      transfersOut = req.population.moves_from.length
+    }
+
+    if (!req.population) {
+      ;[transfersIn, transfersOut] = await Promise.all([
+        moveService.getActive({
```
teneightfive, line 34: You could extract the shared options here to reduce duplication: 

```javascript
const options = { dateRange: [date, date], isAggregation: true }
const [transfersIn, transfersOut] = await Promise.all([
  moveService.getActive({ ...options, toLocationId: locationId }),
  moveService.getActive({ ...options, fromLocationId: locationId }),
])
```

app/population/view/daily.njk
```diff
@@ -0,0 +1,114 @@
...
+            {{ t("population::daily_view.subheading") }}
+            </span>
+        <span class="govuk-!-display-block">{{ displayDate }}</span>
+      </h1>
+    </div>
+  </header>
+
+  {%  call appPanel({classes: 'govuk-!-margin-bottom-4'}) %}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-half">
```
teneightfive, line 58: It feels like it might be more appropriate for this to be `two-thirds` like above. Looking at the screenshots it feels like it wraps too soon in some cases when this is only `one-half`.

app/population/view/daily.njk
```diff
@@ -0,0 +1,114 @@
...
+
+        {% if spaces.details.updated_at %}
+          <h2 class="govuk-heading-xl">{{ t("population::daily_view.free_spaces_with_count", {count:spaces.details.free_spaces}) }}</h2>
+          <p class="govuk-body">{{ t("population::daily_view.free_space_description") }}</p>
+          <p class="govuk-body"><b>{{  t("population::daily_view.labels.last_updated") }} {{  spaces.details.updated_at | formatDateWithTimeAndDay }}</b></p>
+          <p class="govuk-body-m">
```
teneightfive, line 64: I don't think there is a `govuk-body-m` class actually. I think it's the same as [`govuk-body`](https://design-system.service.gov.uk/styles/typography/).

I also had a quick look and the same style is applied to the `p` element so you can actually short all occurrences of `<p class="govuk-body">` or `<p class="govuk-body-m">` to `<p>` in this template.

teneightfive: This particular element also has a set style of a bottom margin applied, which is leading to the spacing on the inside of the box to be uneven:

<kbd>image</kbd>

It can be resolved by adding `govuk-!-margin-bottom-0` to this element.

merlinc: Removed usage of `govuk-body-m` - I spotted `govuk-body-s` and `govuk-body-l` and figured there was a `govuk-body-m` - I was wrong :)
Removed usaged of govuk-body on `<p>` elements.
Spacing issues resolved on both screens with and without free space data.

app/population/view/daily.njk
merlinc: Replaced heading grid size with `two-thirds`

app/population/view/daily.njk
merlinc: Replaced with an `else` block

app/population/view/daily.njk
merlinc: Fixed

app/population/view/daily.njk
merlinc: Macro has been removed and replaced with usage of `statsEntry`

app/population/view/daily.njk
merlinc: Macro has been removed and replaced with usage of `statsEntry`

app/population/view/daily.njk
merlinc: Removed unneeded global imports

app/population/view/daily.njk
merlinc: Removed unneded global imports

app/population/controllers/daily.js
merlinc: I think this line is eviewing old code, but I've removed a linebreak anyway.

app/population/middleware/index.js
merlinc: Renamed

app/population/index.js
merlinc: Removed.

app/population/view/daily.njk
merlinc: Aligned more correctly.

common/presenters/population-to-grid.js
merlinc: We only have the date returned from the populations endpoint, if there was population data. Otherwise we have to use the date as provided as the `req.params`.

I shouldn't make any different, but I was coding defensively and using the known data from the backend in the situation where it was available.

@teneightfive  If you're prefer to only use `req.params.date` I can change it to use that everywhere we're displaying a date.

app/population/middleware/set-population.js
merlinc: Replaced with an `if/else`

common/presenters/population-to-grid.js
merlinc: Changed this to only use the `req.params.date`

app/population/view/daily.njk
merlinc: Removed the usage of `spaces.details.date` and simplified this line as suggested.

app/population/index.js
merlinc: All of this `population` section can be revisited and refactored once the development patterns have become a little clearer.

The main population grid is definitely not part of this PR

app/population/middleware/set-population.js
merlinc: Shared options extracted.

app/population/controllers/daily.test.js
```diff
@@ -0,0 +1,94 @@
+const proxyquire = require('proxyquire')
+
+const populationToGridStub = sinon.stub()
+const transfersToGridStub = sinon.stub()
+
+const controller = proxyquire('./daily', {
```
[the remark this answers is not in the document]
merlinc: Refactored to no longer use `proxyquire`

app/population/view/daily.njk
merlinc: Adjusted the spacing of the statsEntry heading

app/population/middleware/set-population.js
merlinc: Moved this into a `getByIdWithMoves` method.

common/presenters/population-to-grid.js
teneightfive: Yeah I think it feels cleaner to use the date that the user is in theory viewing (which is set in the URL) for both instances.

I guess the worry is that the API returns a different date as part of the population, but even doing it this way, we wouldn't really _know_.

teneightfive (approved):
This looks great! Thanks for working through the changes and apologies that it took a little while to get through.

solidgoldpig (approved):

PR #982, opened 2020-11-13
Add event model and update move model [P4-2337]

solidgoldpig:
### What changed

- Add `event` model
- Add `timeline_events` to `move` model and update move tests
- Remove previous (and now incorrect) `events` relationship from `allocation` model

### Why did it change

Brings models back in line with BE and prepares way for move event timeline 

### Issue tracking

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/lib/api-client/models.js
```diff
@@ -349,9 +349,50 @@ module.exports = {
...
   event: {
     fields: {
-      event_name: '',
-      timestamp: '',
+      event_type: '',
       notes: '',
+      occurred_at: '',
+      recorded_at: '',
+      details: '',
+      eventable: {
```
merlinc, line 357: Event Table? Event Able? No, wait - it's eventable (verb) :joy:

test/fixtures/api-client/move.find.json
```diff
@@ -59,6 +59,9 @@
       },
       "original_move": {
         "data": null
+      },
+      "timeline_events": {
+        "data": []
```
merlinc, line 64: Would it be useful to have a fixture with a fleshed out `timeline_events` structure?

common/lib/api-client/models.js
```diff
@@ -319,10 +323,6 @@ module.exports = {
         jsonApi: 'hasMany',
         type: 'moves',
       },
-      events: {
```
merlinc, line 322: Are there likely to be any effects from removing this property from the allocations model?

merlinc (commented):
It seems ok to me, but I'm not convinced I completely grasp how all the models work together, so I'm not going to approve it myself.

common/lib/api-client/models.js
solidgoldpig: Nope

common/lib/api-client/models.js
solidgoldpig: ¯\_(ツ)_/¯

test/fixtures/api-client/move.find.json
solidgoldpig: I am following the pattern as established in `documents` and `court_hearings`. And analogous to the hasOne relationships such as `supplier` using null.

Given that the tests (currently) only check for the presence of the property, I don't believe it is useful

common/lib/api-client/models.js
solidgoldpig: To clarify, the allocations BE does not return events for an allocation since it does not have any. 

This is an artifact of when `events` in a previous iteration was used as a catch-all generic way to update resource. Cancelling an allocation used to be done by posting a cancel event to the `/allocations/:id/event` endpoint. We now post to `/allocations/:id/cancel`.

test/fixtures/api-client/move.find.json
merlinc: @solidgoldpig explained offline in a bit more detail what these `api-client` tests cover, so I'm happy with these fixtures as they are.

teneightfive (approved):

merlinc (approved):

PR #983, opened 2020-11-13
Transform move events into data consumable by timeline component

solidgoldpig:
Adds presenter for data transformation

Adds events locales to provide content

### What changed

Adds presenter for transformation of events to timeline format

Takes details from each event and mixes in
- move and person
- who performed the event (agency)
- context if any
  - value determined by a `contextKey` for the corresponding event type as defined in `events.json`
- event's relationships
- event's `eventable` relationship as a singularised property
  - eg. an eventable with type `journeys' is mixed in as `journey`
- common event properties
  - eg. `notes` and `occurred_at` timestamp

Also adds status change pseudo-events which are displayed as a GDS badge rather than as the standard heading

<kbd>Move_details_for_MELLIE__ROGER_–_Book_a_secure_move</kbd>

### Why did it change

### Issue tracking

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/events-to-timeline-component.js
+++ b/common/presenters/events-to-timeline-component.js
@@ -1,0 +1,63 @@
+const i18n = require('../../config/i18n')
+const addTriggeredEvents = require('../helpers/events/add-triggered-events')
+const setEventDetails = require('../helpers/events/set-event-details')
+
+const getItem = ({
+  component,
+  labelClasses,
+  heading,
+  description,
+  timestamp,
+  byline,
+}) => {
+  return {
+    label: {
+      classes: labelClasses,
+      html: heading,
+    },
+    classes: 'app-timeline__item',
+    html: description,
+    datetime: {
+      timestamp,
+      type: 'datetime',
+    },
+    byline: {
+      html: byline,
+    },
+  }
+}
+
+const eventsToTimelineComponent = (moveEvents, move) => {
+  const items = addTriggeredEvents(moveEvents).map(moveEvent => {
+    const event = setEventDetails(moveEvent, move)
+
+    const { details, event_type: eventType } = event
+    const statusChange = i18n.exists(`events::${eventType}.statusChange`)
+    const labelClasses = statusChange ? 'moj-badge' : undefined
+
+    const heading = i18n.t(`events::${eventType}.heading`, details)
+    const description = i18n
+      .t(`events::${eventType}.description`, details)
+      .replace(/^<br>/, '')
+    // Some strings that get added together are prefixed with a <br>.
+    // This enables them to be used as the first item or where a previous string doesn't get output
+    // and not insert additional space without having to make the <br> conditional
+
+    // TODO: when we have user info for event, update the following with something like
+    // const byline = i18n.t('events::byline', details)
+    const byline = ''
+
+    const timestamp = event.occurred_at
+
+    return getItem({
+      labelClasses,
+      heading,
+      description,
+      timestamp,
+      byline,
+    })
+  })
+  return { items }
+}
+
+module.exports = eventsToTimelineComponent
--- a/locales/en/events.json
+++ b/locales/en/events.json
@@ -1,0 +1,412 @@
+{
+  "additional": "{{user.name}}",
+  "select_agency": "$t(events::agency, {\"context\": \"{{agency}}\"})",
+  "agency": "",
+  "agency_user": "user",
+  "agency_pmu": "PMU",
+  "agency_supplier": "{{move.supplier.name}}",
+  "default_agency": "pmu",
+  "select_authorised_by": "$t(events::authorised_by, {\"context\": \"{{authorised_by, exists}}\"})",
+  "authorised_by": "",
+  "authorised_by_true": "<br>Authorised by {{authorised_by}}",
+  "select_fault_classification": "Incident has been classified as: $t(events::fault_classification, {\"context\": \"{{fault_classification}}\"})",
+  "fault_classification": "{{fault_classification}}",
+  "fault_classification_supplier": "Supplier’s fault",
+  "fault_classification_was_not_supplier": "Not supplier’s fault",
+  "fault_classification_investigation": "Under investigation",
+  "select_incident_vehicle": "$t(events::incident_vehicle, {\"context\": \"{{vehicle_reg, exists}}\"})",
+  "incident_vehicle": "",
+  "incident_vehicle_true": "The event occurred on vehicle {{vehicle_reg}}",
+  "incident_reported": "The incident was reported by {{supplier_personnel_numbers, oxfordJoin}} at {{reported_at, formatTime}}",
+  "incident_reported_classified": "<br>$t(events::incident_reported)<br>$t(events::select_fault_classification)",
+  "incident_level_details": "{{incident_level}} incident involving {{person.fullname}} occurred $t(events::select_incident_location_preposition) $t(events::select_incident_location_vehicle) $t(events::incident_reported_classified)",
+  "select_incident_location_preposition": "$t(events::incident_location_preposition, {\"context\": \"{{vehicle_reg, exists}}\"})",
+  "incident_location_preposition_true": "on",
+  "incident_location_preposition_false": "at",
+  "select_incident_location_vehicle": "$t(events::incident_location_vehicle, {\"context\": \"{{vehicle_reg, exists}}\"})",
+  "incident_location_vehicle_true": "vehicle {{vehicle_reg}}",
+  "incident_location_vehicle_false": "{{location.title}}",
+  "incident_escaped": "{{person.fullname}} escaped from $t(events::select_incident_location_vehicle) $t(events::incident_reported_classified)",
+  "no_events": "No events",
+  "JourneyAdmitThroughOuterGate": {
+    "heading": "Admitted through outer gate",
+    "description": "Vehicle {{journey.vehicle.registration}} admitted through outer gate of {{journey.to_location.title}}"
+  },
+  "JourneyArriveAtOuterGate": {
+    "heading": "Arrived at outer gate",
+    "description": "Vehicle {{journey.vehicle.registration}} arrived at outer gate of {{journey.to_location.title}}"
+  },
+  "JourneyCancel": {
+    "heading": "Journey cancelled",
+    "description": "Journey from {{journey.from_location.title}} to {{journey.to_location.title}} cancelled"
+  },
+  "JourneyChangeVehicle": {
+    "heading": "Vehicle changed",
+    "description": "{{person.fullname}} moved from {{previous_vehicle_reg}} to {{journey.vehicle.registration}}"
+  },
+  "JourneyComplete": {
+    "heading": "Journey completed",
+    "description": "Journey from {{journey.from_location.title}} to {{journey.to_location.title}} completed"
+  },
+  "JourneyExitThroughOuterGate": {
+    "heading": "Exited from outer gate",
+    "description": "Vehicle {{journey.vehicle.registration}} exited through outer gate of {{journey.from_location.title}}"
+  },
+  "JourneyHandoverToDestination": {
+    "heading": "Handover to establishment",
+    "description": "{{person.fullname}} handed over by {{move.supplier.name}} (Personnel number {{supplier_personnel_number}}) at {{journey.to_location.title}}"
+  },
+  "JourneyPersonBoardVehicle": {
+    "heading": "Boarded escort vehicle",
+    "description": "{{person.fullname}} boarded escort vehicle {{journey.vehicle.registration}}"
+  },
+  "JourneyPersonLeaveVehicle": {
+    "heading": "Left escort vehicle",
+    "description": "{{person.fullname}} left escort vehicle {{journey.vehicle.registration}}"
+  },
+  "JourneyReadyToExit": {
+    "heading": "Ready to exit",
+    "description": "Escort vehicle {{journey.vehicle.registration}} boarding complete, ready to exit from {{journey.from_location.title}}"
+  },
+  "JourneyReject": {
+    "heading": "Journey rejected",
+    "description": "Journey from {{journey.from_location.title}} to {{journey.to_location.title}} rejected by {{move.supplier.name}}"
+  },
+  "JourneyStart": {
+    "heading": "Journey started",
+    "description": "Journey from {{journey.from_location.title}} to {{journey.to_location.title}} started by {{move.supplier.name}}"
+  },
+  "JourneyUncancel": {
+    "heading": "Journey uncancelled",
+    "description": "Cancelled Journey from {{journey.from_location.title}} to {{journey.to_location.title}} revoked by {{move.supplier.name}}"
+  },
+  "JourneyUncomplete": {
+    "heading": "Journey uncompleted",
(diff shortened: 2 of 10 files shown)


common/presenters/events-to-timeline/add-triggered-events.test.js
```diff
@@ -0,0 +1,43 @@
...
+    context('When there are no triggered events', function () {
+      beforeEach(function () {
+        events = addTriggeredEvents(mockEvents)
+      })
+      it('should leave the events untouched', function () {
+        expect(events).to.deep.equal(mockEvents)
+      })
+    })
+
+    context('When there are triggered events', function () {
```
merlinc, line 28: Could we get a test for a mix of triggered and untriggered events?  That way we can be sure that the filter code is doing the correct job, and not just returning if it finds an untriggered event.

common/presenters/events-to-timeline/event-details-helpers.js
```diff
@@ -0,0 +1,84 @@
...
+
+// if an event has a supplier, the supplier was the agent
+// if not, check the locales for a default agent
+// finally, use PMU as default agent
+const getEventAgency = event => {
+  let agency = event.supplier ? 'supplier' : undefined
+
+  if (!agency) {
+    const defaultAgencyKey = `events::${event.event_type}.defaultAgency`
+    agency = i18n.exists(defaultAgencyKey) ? i18n.t(defaultAgencyKey) : 'pmu'
```
merlinc, line 25: Could we have the `pmu` string defined in config somewhere? Magic strings in codebases have a tendency to go wild

common/presenters/events-to-timeline/event-details-helpers.test.js
```diff
@@ -0,0 +1,251 @@
...
+            )
+          })
+
+          it('should get default agency', function () {
+            expect(i18n.t).to.be.calledOnceWithExactly(
+              'events::foo.defaultAgency'
+            )
+          })
+
+          it('should return pmu as the agency', function () {
```
merlinc, line 97: I think this is a typo and should probably be `should return the default agency`

common/presenters/events-to-timeline/event-details-helpers.js
```diff
@@ -0,0 +1,84 @@
...
+    details[pluralize.singular(event.eventable.type)] = event.eventable
+  }
+
+  return details
+}
+
+// add standard common properties of the event itself
+const getEventProperties = event => {
+  // eslint-disable-next-line camelcase
+  const { notes, occurred_at } = event
```
merlinc, line 70: Does this `occurred_at` property get passed into the timeline component?

common/presenters/events-to-timeline/event-details-helpers.test.js
```diff
@@ -0,0 +1,251 @@
...
+          person: mockMove.profile.person,
+        })
+      })
+
+      it('should return just the move if it has no profile', function () {
+        const mockMove = { id: 'move' }
+        const details = getMoveDetails(mockMove)
+        expect(details).to.deep.equal({
+          move: mockMove,
+          person: undefined,
```
merlinc, line 58: Any thoughts on having a `person: undefined` vs the property not existing at all?

common/presenters/events-to-timeline/event-details-helpers.test.js
```diff
@@ -0,0 +1,251 @@
...
+})
+
+describe('Timeline events', function () {
+  describe('Event details helpers', function () {
+    beforeEach(function () {
+      sinon.stub(i18n, 't').returnsArg(0)
+      sinon.stub(i18n, 'exists').returns(false)
+    })
+
+    context('#getMoveDetails', function () {
```
merlinc, line 43: I know its syntactic sugar, but should these be `describes` instead of `context`? I think that's how it is in the rest of the codebase?

merlinc (changes requested):

common/presenters/events-to-timeline/add-triggered-events.js
```diff
@@ -0,0 +1,18 @@
+const i18n = require('../../../config/i18n')
+
+// Add any faux events triggered by other events
```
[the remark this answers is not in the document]
solidgoldpig: It's an event that's displayed in the timeline but which was not actually returned from the BE. They are added when events of the corresponding event type are encountered eg. `MoveCancel` triggers a `StatusMoveCancelled` event as in the image in the initial PR description

common/presenters/events-to-timeline/event-details-helpers.js
solidgoldpig: Not as is, no.

It gets used

a) to set the timestamp on the event item passed in to the timeline component

`const timestamp = event.occurred_at`
https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/d159d58e4d9537ff31c5246524d795dbaf89dd1a/common/presenters/events-to-timeline-component.js#L50

b) part of the vars passed into strings for interpolation by i18next

eg. 
`"description": "{{person.fullname}} was offered $t(events::PerPrisonerWelfare.select_subtype) at {{occurred_at, formatTime}}...`
https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/d159d58e4d9537ff31c5246524d795dbaf89dd1a/locales/en/events.json#L336

common/presenters/events-to-timeline/event-details-helpers.test.js
solidgoldpig: Erm, it's actually a bit all over the place. If only we had a linter that enforced a convention...   

(Will change these instances if you confirm ;) )

common/presenters/events-to-timeline/add-mock-values-to-events.js
```diff
@@ -0,0 +1,32 @@
+const addMockValuesToEvents = (req, moveEvents) => {
```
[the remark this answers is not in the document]
solidgoldpig: Good spot.

Somehow crept back in when I was shuffling things around. (It's for the demo/mocks post the final real PR)

common/presenters/events-to-timeline/add-triggered-events.test.js
solidgoldpig: Done

common/presenters/events-to-timeline/event-details-helpers.test.js
solidgoldpig: i18n which is doing to use it doesn't care

common/presenters/events-to-timeline/event-details-helpers.test.js
solidgoldpig: Didn't bother to wait. Done

common/presenters/events-to-timeline/event-details-helpers.js
```diff
@@ -0,0 +1,86 @@
...
+  return {
+    notes,
+    occurred_at,
+  }
+}
+
+module.exports = {
```
teneightfive, line 79: Should these potentially be in `./common/helpers`? The name sort of suggests it should be...

common/presenters/events-to-timeline/event-details-helpers.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// add eventable resources to details object
+// eg event.eventable.type === 'journeys'
+// => event.details.journey
+const getEventableDetails = event => {
+  const details = {}
+
+  if (event.eventable && event.eventable.type) {
+    details[pluralize.singular(event.eventable.type)] = event.eventable
```
teneightfive, line 63: Could the pluralization be handled using [i18n plural helpers](https://www.i18next.com/translation-function/plurals)?

solidgoldpig: Why would we do that rather than use the same library as `devour`. This is to construct another call with the api-client, not translate a string.

common/presenters/events-to-timeline/event-details-helpers.test.js
solidgoldpig: However, just in case something like `person.fullname` ends up somewhere that does care, I'm returning an empty object instead now

common/presenters/events-to-timeline/event-details-helpers.js
solidgoldpig: Moved.

common/presenters/events-to-timeline-component.js
```diff
@@ -0,0 +1,66 @@
...
+      .t(`events::${eventType}.description`, details)
+      .replace(/^<br>/, '')
+    // TODO: when we have user info for event, update the following with something like
+    // const byline = i18n.t('events::byline', details)
+    const byline = ''
+
+    const timestamp = event.occurred_at
+
+    // constant for now, but designs allow for variations
+    const component = 'appTimelineItemAlt'
```
teneightfive, line 52: Can we remove this now we've discussed removing this concept from the component?

common/presenters/events-to-timeline-component.js
```diff
@@ -0,0 +1,66 @@
...
+
+  const items = moveEvents.map(event => {
+    const { details, event_type: eventType } = event
+    const statusChange = i18n.exists(`events::${eventType}.statusChange`)
+    const labelClasses = statusChange ? 'moj-badge' : undefined
+
+    const heading = i18n.t(`events::${eventType}.heading`, details)
+    const description = i18n
+      .t(`events::${eventType}.description`, details)
+      .replace(/^<br>/, '')
```
teneightfive, line 44: Whats the reason behind stripping `br`s here? Are they not something we are manually adding in the translation content anyway?

locales/en/events.json
```diff
@@ -0,0 +1,412 @@
...
+    "heading": "Supplier vehicle systems failed",
+    "description": "Vehicle {{vehicle_reg}} transporting {{person.fullname}} has had a system failure $t(events::incident_reported_classified)"
+  },
+  "StatusMoveBooked": {
+    "statusChange": true,
+    "heading": "Booked",
+    "description": "Move has been booked with {{move.supplier.name}}"
+  },
+  "StatusMoveCancelled": {
+    "statusChange": true,
```
teneightfive, line 388: I think this is another thing that stands out that suggests we're using translations for more than they should be. 

I understand you are under time pressures to deliver this work but it feels like a cleaner approach would be to define the event "types" somewhere else (maybe in the presenter itself) that are classed as "status changes".

teneightfive: The more I do think about these "status" events, I think it would make more sense to use the translations we've already defined within [`statuses`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/locales/en/statuses.json). That also ensures we keep consistent language.

I'll post a suggestion in the presenter.

teneightfive: On second thoughts, this might be a larger change. 

Does the API by any chance show the previous/current status as a separate value when these event types are returned?

Could we not just look at the "type" for something like a status change that returns the current value? That way the presenter could make sure of the existing keys.

Failing that, could we re-use the values here so that we are consistent across the system:

```json
  "StatusMoveCompleted": {
    "statusChange": true,
    "heading": "$(statuses::completed)",
    "description": "Move has been completed by {{move.supplier.name}}"
  },
```

common/presenters/events-to-timeline-component.js
```diff
@@ -0,0 +1,66 @@
...
+      type: 'datetime',
+    },
+    byline: {
+      html: byline,
+    },
+  }
+}
+
+const eventsToTimelineComponent = (moveEvents, move) => {
+  moveEvents = addTriggeredEvents(moveEvents)
```
teneightfive, line 32: Rather than passing the whole thing in, could these not use `.map()`?

It would make testing the helpers simpler too because they just have simple input/output too.

common/presenters/events-to-timeline-component.js
solidgoldpig: Some strings that get added together are prefixed with a `<br>`. This enables them to be used as the first item or where a previous string doesn't get output and not insert additional space.

common/presenters/events-to-timeline-component.js
solidgoldpig: No, we have to pass the whole thing into `addTriggeredEvents` because it inserts the additional faux triggered events.

locales/en/events.json
solidgoldpig: Parsing all that and taking away the point that we should be doing

`"heading": "$t(statuses::completed)"`

I agree in principle. Just getting it checked off

merlinc (approved):
LGTM

merlinc (approved):
LGTM

common/presenters/events-to-timeline-component.js
solidgoldpig: Added a comment to explain the intent

locales/en/events.json
solidgoldpig: 👍   Event statuses now aliased to the corresponding status strings

common/presenters/events-to-timeline-component.js
```diff
@@ -0,0 +1,65 @@
...
+
+  moveEvents = moveEvents.map(event => setEventDetails(event, move))
```
teneightfive, line 33: Rather than re-assign this can you move this map to line 35 with the other map.

common/presenters/events-to-timeline-component.js
teneightfive: Okay. Can we assign this to a new variable so that we don't risk mutating the argument.

common/presenters/events-to-timeline-component.js
teneightfive: That doesn't feel like a great approach but if it's working with the sets of data we can leave it for now. I might have been better to wrap them in a containing element and use CSS to control spacing.

locales/en/events.json
```diff
@@ -0,0 +1,412 @@
...
+  },
+  "StatusMoveCompleted": {
+    "statusChange": true,
+    "heading": "$t(statuses::completed)",
+    "description": "Move has been completed by {{move.supplier.name}}"
+  },
+  "StatusMoveInTransit": {
+    "statusChange": true,
+    "heading": "$t(statuses::in_transit)",
+    "description": "Move is in transit with {{move.supplier.name}}"
```
teneightfive, line 400: Seems a mix of tenses with the status description. This uses present tense whereas the rest seem to use past (has). I wonder if they should all use past, because in theory they have all occurred. Maybe `Move was started with {{supplier}}`?

teneightfive (approved):

locales/en/events.json
solidgoldpig: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/998

common/presenters/events-to-timeline-component.js
solidgoldpig: I actually think that make it less clear, but done

common/presenters/events-to-timeline-component.js
solidgoldpig: I don't see how this risks mutating the argument. Anyhow, I've passed the addTriggeredEvents directly to the map, removing the need for the reassignment.

PR #984, opened 2020-11-13
Make app filters available to as i18n formatters [P4-2337]

solidgoldpig:
### What changed

Enables i18next interpolation to use same filters as nunjucks

Also adds a formatter `exists` which returns a boolean if the value passed is undefined or an empty string

### Why did it change

Filters sharing because there's no need to reinvent the wheel

`exists` formatter allows for construction of dynamic if/else style contexts when calling nested strings

### Issue tracking

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/config/nunjucks/filters.js
+++ b/config/nunjucks/filters.js
@@ -17,6 +17,7 @@ const { filter, kebabCase, startCase } = require('lodash')
 const pluralize = require('pluralize')
 
 const parse = require('../../common/parsers')
+const { current_week: currentWeek } = require('../../locales/en/actions.json')
 const { DATE_FORMATS } = require('../index')
 
 function _isRelativeDate(date) {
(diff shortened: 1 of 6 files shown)


config/nunjucks/filters.js
```diff
@@ -114,6 +114,7 @@ function formatDateRange(value, delimiter = 'to') {
  * @example {{ "2019-02-21" | formatDateRange("DD/MM/YY") }}
  */
 function formatDateRangeAsRelativeWeek(value) {
+  const i18n = require('../i18n')
```
solidgoldpig, line 117: This is obviously awful - just a stopgap to prompt discussion of how to prevent the circular dependency between config/118n and nunkucks/filters

(I did point out before that I didn't like the i18n.t lurking in here)

As no lang information is actually passed into the function, the i18n value returned would always be the default language. Is there any reason not to just grab the value for the string from the locales file directly? Or perhaps better, a specific config/locales file for the filters>

Or should we pass the i18n method in to either initialise the filters or return an instance?

...

Also should we move the contents of nunjucks/filters to common/lib/filters and then load that from the nunucks and the i18n modules

solidgoldpig: Next commit implements the first suggested approach - grabbing the value for the string from the locales file directly

teneightfive (approved):

PR #985, opened 2020-11-13
Extract setting youth move to own module [P4-2337]

solidgoldpig:
### What changed

Extracts the setting of youth moves out of view controller into separate module

### Why did it change

Allows the code to be reused by other controllers

### Issue tracking

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/view.js
+++ b/app/move/controllers/view.js
@@ -5,6 +5,7 @@ const updateSteps = require('../steps/update')
 
 const getUpdateLinks = require('./view/view.update.links')
 const getUpdateUrls = require('./view/view.update.urls')
+const setYouthMove = require('./view/view.youth-move')
 
 module.exports = function view(req, res) {
   const { move, originalUrl, framework = {} } = req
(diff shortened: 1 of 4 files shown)


app/move/controllers/view.js
```diff
@@ -5,6 +5,7 @@ const updateSteps = require('../steps/update')
 
 const getUpdateLinks = require('./view/view.update.links')
 const getUpdateUrls = require('./view/view.update.urls')
+const setYouthMove = require('./view/view.youth-move')
```
merlinc, line 8: This stuck out as being different - so I'm wondering if there is a way to phrase this that fits the `get_Something_` pattern we're already using for these view controllers

merlinc (commented):
Apart from my one minor question, the rest of this is a simple lift-and-shift :100:

app/move/controllers/view.js
solidgoldpig: 1. It is literally updating the move type
2. It should stick out - it's technical debt that we even have to do this

So, disagreeing ;-p ...

teneightfive: I'm inclined to agree with @merlinc about this kind of sticking out a bit. And I'm not sure I agree with just because it's solving some technical debt from the API that it justifies us maybe putting it here.

If the purpose is to determine the move type correctly I wonder if we can extend the [transform method](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/146a7fbb3e24184ef0dff150f5cb97e89cbf290d/common/services/move.js#L50) in the move service to take care of this logic instead of extracting it to here?

**Edit**: After looking up the move service it actually looks like the transform method is [already doing something similar](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/146a7fbb3e24184ef0dff150f5cb97e89cbf290d/common/services/move.js#L52-L57)? So I'm wondering what extra purpose this serves?

solidgoldpig: https://talent-intl.slack.com/archives/C010Z1KP555/p1605521639006900

So leaving this open but ignoring for now until we can resolve this properly. It's not really relevant to the events timeline PR so can safely be ignored for now

PR #986, opened 2020-11-13
Add timeline component [P4-2337]

solidgoldpig:
*Depends on https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/984*

### What changed

Introduces new component `appTimeline`.

This is derived from the `mojTimeline` component (https://moj-design-system.herokuapp.com/components/timeline)

<kbd>Move_details_for_MELLIE__ROGER_–_Book_a_secure_move</kbd>

<kbd>Move_details_for_MELLIE__ROGER_–_Book_a_secure_move</kbd>

### Why did it change

### Issue tracking

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/assets/scss/application.scss
+++ b/common/assets/scss/application.scss
@@ -40,6 +40,7 @@ $moj-assets-path: "@ministryofjustice/frontend/moj/assets/";
 @import "@ministryofjustice/frontend/moj/components/primary-navigation/primary-navigation";
 @import "@ministryofjustice/frontend/moj/components/badge/badge";
 @import "@ministryofjustice/frontend/moj/components/filter/filter";
+@import "@ministryofjustice/frontend/moj/components/timeline/timeline";
 
 // Import app settings/tools/helpers
 @import "settings/all";
--- a/common/assets/scss/application.scss
+++ b/common/assets/scss/application.scss
@@ -40,7 +40,6 @@ $moj-assets-path: "@ministryofjustice/frontend/moj/assets/";
 @import "@ministryofjustice/frontend/moj/components/primary-navigation/primary-navigation";
 @import "@ministryofjustice/frontend/moj/components/badge/badge";
 @import "@ministryofjustice/frontend/moj/components/filter/filter";
-@import "@ministryofjustice/frontend/moj/components/timeline/timeline";
 
 // Import app settings/tools/helpers
 @import "settings/all";
--- a/common/components/timeline/_timeline.scss
+++ b/common/components/timeline/_timeline.scss
@@ -1,40 +1,108 @@
 /* ==========================================================================
+   #TIMELINE
+   ========================================================================== */
+
+   .app-timeline {
+    margin-bottom: govuk-spacing(4);
+    overflow: hidden;
+    position: relative;
+
+    &:before {
+      background-color: $govuk-brand-colour;
+      content: "";
+      height: 100%;
+      left: 0;
+      position: absolute;
+      top: govuk-spacing(2);
+      width: 5px;
+    }
+
+  }
+
+  .app-timeline--full {
+    margin-bottom: 0;
+    &:before {
+      height: calc(100% - 75px);
+    }
+  }
+
+  .app-timeline__item {
+    padding-bottom: govuk-spacing(6);
+    padding-left: govuk-spacing(4);
+    position: relative;
+
+    &:before {
+      background-color: $govuk-brand-colour;
+      content: "";
+      height: 5px;
+      left: 0;
+      position: absolute;
+      top: govuk-spacing(2);
+      width: 15px;
+    }
+
+  }
+
+  .app-timeline__title {
+    @include govuk-font($size: 19, $weight: bold);
+    display: inline;
+  }
+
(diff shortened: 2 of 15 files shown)


merlinc (commented):
Looking good and simple so far - I could do with getting my component explorer finished so we could also see these components in isolation in all their permutations

common/assets/scss/application.scss
```diff
@@ -26,6 +26,7 @@ $moj-assets-path: "@ministryofjustice/frontend/moj/assets/";
 @import "govuk-frontend/govuk/components/skip-link/index";
 @import "govuk-frontend/govuk/components/summary-list/index";
 @import "govuk-frontend/govuk/components/table/index";
+@import "govuk-frontend/govuk/components/tabs/index";
```
solidgoldpig, line 29: Not sure this is needed now. Will double-check

solidgoldpig: It is needed obviously. Just not by this.

solidgoldpig: and zapped

common/components/timeline/macro.njk
```diff
@@ -0,0 +1,27 @@
+{% macro appTimeline(params) %}
+  {%- include "./template.njk" -%}
+{% endmacro %}
+
+{% macro appTimelineItemHeading(params, item) %}
```
teneightfive, line 5: Is there a reason for needing to create all these separate macros, and to then need to import them all globally in `base.njk`?

As far as I can tell they're only going to be used within `template.njk` in this component?

common/components/timeline/timeline--item-alt.njk
```diff
@@ -0,0 +1,7 @@
+{{ appTimelineItemHeading(params, item) }}
+
+<div class="moj-timeline__content">
```
teneightfive, line 3: What's the need for the `-alt` version of this? I wonder if it would not be a bit simpler to use the same style already used elsewhere in the MOJ Design system?

Is there a case where we would actually need to use both of these variations together? It would be good to understand the thinking behind moving of the date and whether we actually need to support both of these variations or we just consistently use one approach to position of dates.

common/components/timeline/_timeline.scss
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  & .moj-timeline__description {
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(0);
+
+    & br {
+      content: '';
+      display: block;
+      padding-top: 10px;
```
teneightfive, line 37: Should we keep consistency here and use `govuk-spacing`?

common/components/timeline/macro.njk
```diff
@@ -0,0 +1,27 @@
...
+
+{% macro appTimelineItemAdditional(params, item) %}
+  {%- include "./timeline--additional.njk" -%}
+{% endmacro %}
+
+{% macro appTimelineItem(params, item) %}
+{%- include "./timeline--item.njk" -%}
+{% endmacro %}
+
+{% macro appTimelineItemAlt(params, item) %}
```
teneightfive, line 25: @solidgoldpig I had a quick conversation with @Andy-Millar to help me better understand these changes.

It sounds like he feels we will **not** use the proposed format from the MOJ Design system. So in that vein I would suggest we don't need to extract all of this logic in order to support both and just build this component in the format that we have decided.

That would simplify what it is we need to build at this stage.

common/components/timeline/macro.njk
solidgoldpig: Macros that are not meant to be available globally have been zapped from `base.njk`

teneightfive (approved):

common/components/timeline/template.njk
```diff
@@ -0,0 +1,37 @@
...
+  {%- for item in params.items %}
+  <div class="app-timeline__item {{- ' ' + item.classes if item.classes}}" {%- for attribute, value in item.attributes -%} {{ attribute }}="{{ value }}"{% endfor %}>
+
+    <div class="app-timeline__header">
+      <h{{ params.headingLevel | default(2) }} class="app-timeline__title {{- ' ' + item.label.classes if item.label.classes}}">
+        {{- item.label.html | safe if item.label.html else item.label.text -}}
+      </h{{ params.headingLevel | default(2) }}>
+    </div>
+
+    <div class="app-timeline__content">
```
teneightfive, line 12: This can be removed as it's not being reference within the css anymore.

common/components/timeline/timeline--item-alt.njk
solidgoldpig: Zapped as part of simplification and cutting ties with original component

common/components/timeline/macro.njk
solidgoldpig: As previous comment

common/components/timeline/template.njk
solidgoldpig: Zapped

PR #757, opened 2020-08-31
[P4-1997 bis] SHC and STC transfer moves

danielepolencic:
This ticket includes the changes in #728 and adds more:

- STC/SCH users should have a dashboard with incoming, outgoing and single requests
- STC/SCH should be able to cancel proposed moves
- STC/SCH should not be able to edit youth transfer moves 

### Issue tracking

- P4-1997
- P4-2107
- P4-2106
- P4-1932

## Screenshots

Screenshot 2020-08-31 at 17 21 54

#### Automated testing

- [x] Added unit tests to cover changes
- [ ] Added end-to-end tests to cover any journey changes

#### Manual testing

Has been tested in the following browsers:

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

PR #1001, opened 2020-11-17
Transform models individually

teneightfive:
### What changed

This looks to address occurrences where we want to do something with the deserialized response data from the API.

### Why did it change

In some instances we've wanted to attached extra attributes that are only used on the frontend, for example full name.

This has meant doing so within various services once the data is returned but this can lead to unforeseen circumstances when it is left off of new methods.

This change moves that logic to the model itself. So anytime a particular resource is deserialized in the API client, this transformer is run on the data.

Closes #985


solidgoldpig (approved):

app/move/controllers/view/view.locals.js
```diff
@@ -16,16 +16,6 @@ function getViewLocals(req) {
     cancellation_reason_comment: cancellationComments,
     rejection_reason: rejectionReason,
   } = move
-
-  // We have to pretend that 'secure_childrens_home', 'secure_training_centre' are valid `move_type`s
```
merlinc, line 20: Do we no longer need to pretend that these are valid move types?

common/lib/api-client/models/person.js
```diff
@@ -1,3 +1,5 @@
+const { _transform, personTransformer } = require('../transformers')
```
merlinc, line 1: If we're using `_transform` externally, then it looks weird when its using the "private variable" convention for naming.

common/lib/api-client/transformers/_transform.test.js
```diff
@@ -0,0 +1,93 @@
...
+        })
+
+        it('should deserialize model', function () {
+          expect(_this.deserialize.resource).to.have.been.calledWithExactly(
+            item,
+            included
+          )
+        })
+
+        it.skip('should delete model deserializer', function () {
```
merlinc, line 84: Skipped test?

locales/en/messages.json
```diff
@@ -3,6 +3,7 @@
     "success": {
       "heading": "Move requested",
       "heading_proposed": "Move sent for review",
+      "heading_youth_transfer": "Move sent for review",
```
merlinc, line 6: Does this addition to the locales need to be part of this PR?

app/move/views/confirmation.njk
```diff
@@ -30,8 +30,8 @@
         {% set context = "assign" %}
       {% else %}
 
-        {% if move.status === "proposed" and (move.move_type === "secure_training_centre" or move.move_type === "secure_childrens_home") %}
-          {% set context = move.move_type %}
+        {% if move.status === "proposed" and (move.from_location.location_type === "secure_training_centre" or move.from_location.location_type === "secure_childrens_home") %}
```
merlinc, line 33: There's a lot of these `move.move_type` to `move.from_location.location_type` changes - what the difference between this and the `move?.to_location?.location_type` on the SCH removal above?

common/lib/api-client/transformers/index.js
```diff
@@ -0,0 +1,9 @@
+const _transform = require('./_transform')
+const personEscortRecordTransformer = require('./person-escort-record.transformer')
+const personTransformer = require('./person.transformer')
+
+module.exports = {
+  _transform,
```
merlinc, line 6: As mentioned above, this naming is styled like a private variable, but it is being exported. I wonder if there's a better name for it instead.

common/lib/api-client/transformers/person.transformer.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function personTransformer(data) {
+  return {
+    ...data,
+    image_url: `/person/${data.id}/image`,
+    fullname: [data.last_name, data.first_names].filter(Boolean).join(', '),
```
merlinc, line 5: This is the code from `common/services/person` extracted out into a separate transformer

:+1:

app/move/controllers/view/view.locals.js
solidgoldpig: It's in the Nunjucks template now

common/lib/api-client/models/person.js
solidgoldpig: 👍

app/move/controllers/view/view.locals.js
teneightfive: Yeah, what @solidgoldpig said. 

It was introduced in #757 and looked like it was to cater for variations to the [pending review content](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/a7efddb5e67a08870ebbcfd7fc258d4a9200f366/locales/en/messages.json#L17-L18).

After digging into it, there were no other variations based on `move_type`, so instead, it now passes the `from_location.location_type` in as context which allows the same overrides to be set.

It essentially felt more dangerous to be overriding the value from the API to update content, instead it feels clearer to pass in the context we want to control content based on.

common/lib/api-client/models/person.js
teneightfive: Yeah, I wasn't convinced either and went round a few options for the name...I can't help keep thinking of Optimus Prime each time I was writing them too...

Open to naming suggestions, or maybe just removing the underscore prefix might be the better option for now?

common/lib/api-client/transformers/_transform.test.js
teneightfive: Yes, good catch! So, I was struggling to get this test to pass. 

Basically I'm trying to ensure that the deserializer is removed from the context before the devour deserializer is called again, otherwise it will result in an infinite loop.

Not sure if either of you can see any reason why the comparison doesn't work?

common/lib/api-client/transformers/person.transformer.js
teneightfive: Yep. The benefit here is that it will be called **every** time devour deserializers it. Previously if something like a person resource was nested in a relationship (`move` -> `profile` -> `person`), we would have to do that manually for the move and person even if neither of these required any kind of transformation themselves.

So this is more consistent and hopefully harder to miss.

locales/en/messages.json
teneightfive: That's true. I toyed with splitting this into separate PRs but making the second change means no need to add the move transformer in the end so grouped them together.

I'm happy to split them up if it's easier but the second will depend on the first?

app/move/views/confirmation.njk
teneightfive: Yes, I've tried to explain this within the [commit message](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1001/commits/a7efddb5e67a08870ebbcfd7fc258d4a9200f366) as to why these changes are happening.

Essentially, I've removed the value from the API being updated in two places and used a different piece of date (`from_location.location_type`) to provide the same override setup to the translation library.

app/move/controllers/view/view.locals.js
merlinc: Ok, sounds good.

app/move/views/confirmation.njk
merlinc: Commit message links don't survive rebasing off master :wink: 

Anyway, that makes sense (as does the commit messages)

merlinc (approved):
LGTM

PR #1016, opened 2020-11-19
[P4-2444] Setup support for youth risk assessment

teneightfive:
### What changed

This is mainly a refactor to extract the logic that will be used to support both the Person Escort Record and Youth risk assessments individually.

### Why did it change

The initial work was around supporting the Person Escort Record. This now needs to be extending to support multiple assessment simultaneously and concurrently so this work extracts the logic that will be used by both.

### Issue tracking

- [P4-2444](https://dsdmoj.atlassian.net/browse/P4-2444)


solidgoldpig (approved):
lgtm

PR #1019, opened 2020-11-20
Check prisoner category before creating a move

solidgoldpig:
### What changed

After a user selects a prisoner to be moved, checks whether they should be moved or not.

If not, displays a page explaining that the person cannot be moved and why not.

### Why did it change

Prevents prisoners being inadvertently moved. 

### Issue tracking

- [P4-2310 - Check category at "Who is being moved step"](https://dsdmoj.atlassian.net/browse/P4-2310)

## Screenshots

<kbd>Error__ARCAMONE__AMERY_cannot_be_moved_using_this_service</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/create/base.js
+++ b/app/move/controllers/create/base.js
@@ -1,5 +1,6 @@
 const FormWizardController = require('../../../../common/controllers/form-wizard')
 const presenters = require('../../../../common/presenters')
+const personService = require('../../../../common/services/person')
 
 class CreateBaseController extends FormWizardController {
   middlewareSetup() {
@@ -10,6 +11,7 @@ class CreateBaseController extends FormWizardController {
   middlewareChecks() {
     super.middlewareChecks()
     this.use(this.checkCurrentLocation)
+    this.use(this.checkMoveSupported)
   }
 
   middlewareLocals() {
@@ -40,6 +42,25 @@ class CreateBaseController extends FormWizardController {
     this._setModels(req)
     this._addModelMethods(req)
     next()
+  }
+
+  async checkMoveSupported(req, res, next) {
+    const person = req.getPerson()
+
+    if (!person?.prison_number) {
+      return next()
+    }
+
+    const category = await personService.getCategory(person.id)
+
+    if (!category || category.move_supported) {
+      return next()
+    }
+
+    return res.render('move/views/create/move-not-supported', {
+      person,
+      category,
+    })
   }
 
   setButtonText(req, res, next) {
--- a/app/move/views/create/move-not-supported.njk
+++ b/app/move/views/create/move-not-supported.njk
@@ -1,0 +1,10 @@
+{% extends "layouts/action-prevented.njk" %}
+
+{% set actionPreventedTitle = t("validation::move_not_supported.heading", {
+  name: person.fullname
+}) %}
+
+{% set actionPreventedMessage = t("validation::move_not_supported.message", {
+  category: category.key
+}) %}
+
--- a/common/lib/api-client/models/index.js
+++ b/common/lib/api-client/models/index.js
@@ -2,6 +2,7 @@ const allocation = require('./allocation')
 const allocationComplexCase = require('./allocation-complex-case')
 const assessmentQuestion = require('./assessment-question')
 const cancel = require('./cancel')
+const category = require('./category')
 const courtCase = require('./court-case')
 const courtHearing = require('./court-hearing')
 const document = require('./document')
@@ -34,6 +35,7 @@ module.exports = {
   allocation_complex_case: allocationComplexCase,
   assessment_question: assessmentQuestion,
   cancel,
+  category,
   court_case: courtCase,
   court_hearing: courtHearing,
   document,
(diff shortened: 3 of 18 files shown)


app/move/views/create/move-not-supported.njk
```diff
@@ -0,0 +1,28 @@
...
+
+{% block actionPreventedMessage %}
+  {{ t("validation::move_not_supported.message", {
+    category: category.key
+  }) | safe }}
+{% endblock %}
+
+{% block actionPreventedInstructions %}{% endblock %}
+
+{# {% block contentHeader %}
```
solidgoldpig, line 25: Remove the example bits Alex!!!

common/services/person.js
```diff
@@ -56,6 +58,25 @@ const personService = {
...
+
+  getById(id) {
+    const include = personService.defaultInclude
+    return personService._getById(id, { include })
+  },
+
+  async getCategory(id) {
+    const include = ['category']
+    const person = await personService._getById(id, { include })
+    return person.category
```
teneightfive, line 77: This could also return a promise inline with the other service methods:

```javascript
getCategory(id) {
  return personService
    ._getById(id, { include: ['category'] })
    .then(person => person.category)
}
```

app/move/controllers/create/base.js
```diff
@@ -42,6 +44,23 @@ class CreateBaseController extends FormWizardController {
     next()
   }
 
+  async checkMoveSupported(req, res, next) {
```
teneightfive, line 47: I've found these kind of methods easier to read when written using an early return pattern but that's purely personal preference:

```javascript
async checkMoveSupported(req, res, next) {
  const person = req.getPerson()

  if (!person || !person.prison_number) {
    return next()
  }

  const category = await personService.getCategory(person.id)
  if (!category || category.move_supported) {
    return next()
  }

  return res.render('move/views/create/move-not-supported', {
    person,
    category,
  })
}
```

The person check could even be simplified to (ie, if the person doesn't exist, by default they can't have a prison number?):

```javascript
  if (!person?.prison_number) {
    return next()
  }
```

teneightfive (approved):
Few minor style comments but looks good.

common/templates/layouts/action-prevented.njk
```diff
@@ -0,0 +1,42 @@
...
+{% endblock %}
+
+{% block beforeContent %}
+  {% block actionPreventedBackLink %}{% endblock %}
+  {{ super() }}
+{% endblock %}
+
+{% block contentHeader %}
+  {% set title %}
+    {% block actionPreventedTitle %}{% endblock %}
```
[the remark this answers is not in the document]
solidgoldpig: Good call. I'd misremembered that you couldn't set stuff outside of the blocks (except at load time). That's much, much cleaner...

app/move/views/create/move-not-supported.njk
```diff
@@ -0,0 +1,14 @@
+{% extends "layouts/action-prevented.njk" %}
+
+{% set actionPreventedTitle %}
```
teneightfive, line 3: A small note, but these can also be defined without the need for a block:

```nunjucks
{% set actionPreventedTitle = t("validation::move_not_supported.heading", {
  name: person.fullname
}) %}
```

The benefit of this is that you won't be injecting unnecessary whitespace into the final markup.

common/lib/api-client/models/index.js
```diff
@@ -34,6 +35,7 @@ module.exports = {
   allocation_complex_case: allocationComplexCase,
   assessment_question: assessmentQuestion,
   cancel,
+  category,
```
teneightfive, line 38: This might be a little too late to determine if the API side has already been built but `category` feels a little ambiguous. I wonder if it would be better named as `prison_category` or `prisoner_category` (I can't remember which the Prison API uses..)?

common/lib/api-client/models/person.js
```diff
@@ -23,6 +23,10 @@ module.exports = {
       jsonApi: 'hasMany',
       type: 'profiles',
     },
+    category: {
```
teneightfive, line 26: I'd also say, even if we don't want to rename the model. JSON:API allows the ability to define relationships with different names. So maybe at a minimum this relationship should be `prison_category` (or `prisoner_category` is that's the correct term).

common/lib/api-client/models/index.js
solidgoldpig: Will see what can be done

common/lib/api-client/models/person.js
solidgoldpig: > Will see what can be done

jason-sims:
LGTM

common/lib/api-client/models/index.js
solidgoldpig: Apparently the use of `category` is intentional as locations have also been rejigged to have them.

PR #1021, opened 2020-11-20
Show both NOMIS "High Security Hospitals" & "Hospital" location types for type ahead on Hospital moves [P4-2408]

solidgoldpig:
### What changed

`setLocationItems` can now be passed an array of location types which will be combined and sorted alphabetically.

Hospital location items are now made up of locations matching both `high_security_hospital` and `hospital`

### Why did it change

Users could not select a non high security hospital

### Issue tracking

- [P4-2408 - Show both NOMIS "High Security Hospitals" & "Hospital" location types for type ahead on Hospital moves](https://dsdmoj.atlassian.net/browse/P4-2408)

## Screenshots

As before. Just adds extra items to hospital dropdown

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/middleware/set-location-items.js
+++ b/common/middleware/set-location-items.js
@@ -13,14 +13,35 @@ function setLocationItems(locationType, fieldName) {
     }
 
     try {
-      const locations = await referenceDataService.getLocationsByType(
-        locationType
+      if (!Array.isArray(locationType)) {
+        locationType = [locationType]
+      }
+
+      const locationString = locationType[0]
+
+      const locations = (
+        await Promise.all(
+          locationType.map(x => referenceDataService.getLocationsByType(x))
+        )
       )
+        .reduce((acc, val) => acc.concat(val))
+        .sort((a, b) => {
+          if (a.title < b.title) {
+            return -1
+          }
+
+          if (a.title > b.title) {
+            return 1
+          }
+
+          return 0
+        })
+
       const items = fieldHelpers.insertInitialOption(
         locations
           .filter(referenceDataHelpers.filterDisabled())
           .map(fieldHelpers.mapReferenceDataToOption),
-        locationType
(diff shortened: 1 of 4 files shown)


common/middleware/set-location-items.js
```diff
@@ -13,14 +13,35 @@ function setLocationItems(locationType, fieldName) {
...
+
+      const locationString = locationType[0]
+
+      const locations = (
+        await Promise.all(
+          locationType.map(x => referenceDataService.getLocationsByType(x))
+        )
       )
+        .reduce((acc, val) => acc.concat(val))
+        .sort((a, b) => {
```
teneightfive, line 28: This may have already been thought of and decided against, but, could we have the `location_type` filter in the API support multiple values? Similar to what we have done with other filters like location IDs...?

Then we wouldn't need to really do a lot of this multiple call + sort heavy lifting?

solidgoldpig: Very good point. Will raise on Monday

merlinc (commented):
This all seems ok, but @teneightfive 's comment is valid, so I won't give it a :+1: until we've got an answer on if the BE team is going to simplify things for us

common/middleware/set-location-items.js
jason-sims: FE LGTM. The location type filter is not currently in scope. I will put this as an improvement for a future sprint.

teneightfive (approved):
Have created a [separate issue](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/1028) to improve the way the filtering is done once API support is added.

In the meantime, this looks good to merge.

PR #1030, opened 2020-11-23
Add rejected info to rejected single request banners [P4-2414]

solidgoldpig:
### What changed

Adds rebook and additional info (if any) to rejected single request’s banner.

### Why did it change

> PMU are providing this information but users cannot see it causing issues between prisons and PMU.

### Issue tracking

- [P4-2414 - Add Rebook & Additional Information to banner for rejected Single Requests](https://dsdmoj.atlassian.net/browse/P4-2414)
- [P4-2527 - Single request rejection additional info is currently not getting saved](https://dsdmoj.atlassian.net/browse/P4-2527)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd>Move_details_for_O__39_CONNELL__HAL_–_Book_a_secure_move</kbd> | <kbd>Move_details_for_O__39_CONNELL__HAL_–_Book_a_secure_move</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


locales/en/statuses.json
```diff
@@ -16,12 +16,13 @@
   "cancelled_rejected": "Single request rejected",
   "description": "",
   "description_made_in_error": "Reason — $t(fields::cancellation_reason.items.made_in_error.label)",
-  "description_supplier_declined_to_move": "Reason — $t(fields::cancellation_reason.items.supplier_declined_to_move.label)",
-  "description_cancelled_by_pmu": "Reason — $t(fields::cancellation_reason.items.cancelled_by_pmu.label) - {{comment}}",
+  "description_supplier_declined_to_move": "Reason — $t(fields::cancellation_reason.items.supplier_declined_to_move.label) JJJ?",
```
merlinc, line 19: Are these two lines typos?

merlinc (changes requested):

locales/en/statuses.json
solidgoldpig: Yep

merlinc (approved):
LGTM

app/move/controllers/view.js
```diff
@@ -1,6 +1,19 @@
+const moveService = require('../../../common/services/move')
+
 const getViewLocals = require('./view/view.locals')
 
-module.exports = function view(req, res) {
+module.exports = async function view(req, res) {
+  const { id, status, cancellation_reason: cancellationReason } = req.move
+
+  // TODO: remove this second call when backend returns rebook info as part of move
+  if (status === 'cancelled' && cancellationReason === 'rejected') {
```
teneightfive, line 9: Could we do this in the new "transformer" for now instead? If it's essentially something that we're going to expect on the move that feels like a nice place to do this patch for now.

app/move/controllers/view/view.locals.js
```diff
@@ -129,6 +129,7 @@ function getViewLocals(req) {
       ? req.t('statuses::' + status, { context: cancellationReason })
       : undefined,
     messageContent: req.t('statuses::description', {
+      ...move,
```
teneightfive, line 132: This seems quite overkill. Can we be more specific with what properties the translation string needs rather than passing the whole thing through please?

teneightfive (commented):
Just a bit of feedback on commits, I know we've talked about small atomic commits to tell a story but I think this PR could have benefit from a few less commits. The last 3 feel like they're tied together so I feel it would have been easier to review those as one.

For example, changing the banner to use `html` instead of `text` was related to the other two commits and would tell a better long term history if those were grouped.

app/move/controllers/view.js
solidgoldpig: I disagree. This is a kludge until the backend does the right thing. Moreover, moving it the transformer would mean that we would (unnecessarily) perform this transformation for every rejected single request when listing them.

PR #1032, opened 2020-11-23
[P4-2444] Set assessment banners

teneightfive:
### Issue tracking

- [P4-2444](https://dsdmoj.atlassian.net/browse/P4-2444)

## Screenshots

| Person Escort Record | Youth Risk Assessment |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/presenters/move-to-message-banner-component.js
```diff
@@ -0,0 +1,164 @@
...
+          href: `${baseUrl}/new`,
+          text: i18n.t('actions::start_assessment', {
+            context,
+            type,
+          }),
+        })}
+      `
+    }
+
+    return _returnComponent(title, content)
```
Nimphal, line 62: Perhaps `_returnComponent` could return undefined if there's no content passed - it would remove the need for multiple returns here and result in just one exit point for the larger function whose responsibility then becomes to just figure out what title and content are rather than also whether the component exists or not.

teneightfive: So, I originally had something similar where these if blocks weren't returning anything and there was a single call at the end.

However, that meant in some circumstances two blocks were being entered before the return.

I was trying to take more of a [return early pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8) instead of requiring the use of `else` or more statements within each `if` block.

common/presenters/move-to-message-banner-component.js
```diff
@@ -0,0 +1,147 @@
...
+    title: {
+      text: title,
+    },
+    content: {
+      html: content,
+    },
+  }
+}
+
+function moveToMessageBannerComponent({
```
teneightfive, line 22: I think this is the main point of discussion. I've tried to extract the logic that was being done in the template to a presenter so that it simply passes the relevant component structure to the view. This means it can also be tested faster using unit tests.

I've tried to use a return early pattern but it does still feel quite complex to cater for the different scenarios.

One thing I was considering was extracting the contents of each if block out to separate methods/presenters but I wasn't sure if that would make it harder to follow.

Any thoughts around how to structure this method would be very welcome :)

common/presenters/move-to-message-banner-component.js
```diff
@@ -0,0 +1,147 @@
...
+  const context = assessmentType
+
+  if (['proposed'].includes(status) || !profileId) {
+    return undefined
+  }
+
+  let title
+  let content
+
+  // No PER exists
```
merlinc, line 40: I wonder if this could be broken up into smaller presenters that are then assembled in this parent one. There's already a fairly clear separation which you've had to describe with comments, this could be even clearer if it was code.

e.g. 
```
if (perStatusUtils.noPer(status, assessment)) {
  return renderNoPer(any, params, that, are, needed
}

if ((perStatusUtils.perConfirmed(status, assessment)) {
  return renderConfirmedPer(other, params, that, are, needed)
}
```

This might also deal with the CodeClimate complaints about complexity and return statements.

common/presenters/move-to-message-banner-component.js
```diff
@@ -0,0 +1,147 @@
...
+    content = `
+      <p>
+        ${i18n.t(`messages::assessment.${perStatus}.content`, {
+          context,
+          date: filters.formatDateWithTimeAndDay(confirmedAt),
+        })}
+      </p>
+
+      <p>
+        <a href="${baseUrl}/print" class="app-icon app-icon--print">
```
merlinc, line 80: There already seems to be 3 different implementations of "print" (allocations, move, collections) - perhaps this now should be made consistently using a component.

locales/en/actions.json
```diff
@@ -41,12 +41,10 @@
   "save_and_return_to_overview": "Save and return to overview",
   "view_more_information": "View more information",
   "review": "Review",
-  "start_person_escort_record": "Start Person Escort Record",
+  "start_assessment": "Start $t({{context}})",
+  "print_assessment": "Print $t({{context}})",
   "provide_confirmation": "Provide confirmation",
   "confirm_and_complete_record": "Confirm and complete record",
-  "continue_person_escort_record": "Continue record",
-  "view_person_escort_record": "View Person Escort Record",
```
merlinc, line 48: Did this get replaced with a more generic context style message anywhere?

common/controllers/framework/framework-step.js
```diff
@@ -13,14 +12,9 @@ class FrameworkStepController extends FormWizardController {
...
-    const isEditable = req.assessment?.isEditable
-    const userPermissions = req.user?.permissions
-    const canEdit = permissionsControllers.check(
-      'person_escort_record:update',
-      userPermissions
-    )
+    const { isEditable, framework } = req.assessment
 
-    if (!isEditable || !canEdit) {
+    if (!isEditable || !req.canAccess(`${snakeCase(framework.name)}:update`)) {
```
merlinc, line 17: I can see the need for this here, but I'm not thrilled about the swapping between snake_cake and camelCase throughout the system. However, that's a much bigger issue than this PR, so I guess ... :+1: ?

app/move/controllers/view/view.locals.js
```diff
@@ -20,39 +20,32 @@ function getViewLocals(req) {
...
+  })
   const personEscortRecordIsEnabled = req.canAccess('person_escort_record:view')
   const personEscortRecordIsCompleted =
     !isEmpty(personEscortRecord) &&
     !['not_started', 'in_progress'].includes(personEscortRecord?.status)
-  const personEscortRecordIsConfirmed = ['confirmed'].includes(
-    personEscortRecord?.status
-  )
   const personEscortRecordUrl = `${moveUrl}/person-escort-record`
-  const showPersonEscortRecordBanner =
```
merlinc, line 38: I'm liking all this removal of code :100:

app/move/views/view.njk
```diff
@@ -93,97 +93,8 @@
     }) }}
   {% endif %}
 
-  {% if canStartPersonEscortRecord %}
```
merlinc, line 96: ... as well as all of this removal of code too! :100:

common/presenters/move-to-message-banner-component.js
```diff
@@ -0,0 +1,147 @@
...
+
+  const {
+    _framework = {},
+    meta = {},
+    status: perStatus,
+    confirmed_at: confirmedAt,
+  } = assessment || {}
+
+  // PER has been confirmed
+  if (assessment && perStatus === 'confirmed') {
```
merlinc, line 69: This block and the one below both check for `assessment`, but we've already partially checked to see if this is empty beforehand.

Perhaps this existence check could be done first, and then an appropriate action taken instead? This might make the logic clearer as well.

package.json
```diff
@@ -84,6 +84,7 @@
     "@hmpps-book-secure-move-frameworks/1.5.1": "git+https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks.git#v1.5.1",
     "@hmpps-book-secure-move-frameworks/1.5.2": "git+https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks.git#v1.5.2",
     "@hmpps-book-secure-move-frameworks/1.5.3": "git+https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks.git#v1.5.3",
+    "@hmpps-book-secure-move-frameworks/2.2.0": "git+https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks.git#youth-risk-assessment-updated",
```
merlinc, line 87: Obviously a line to replace with a proper version, once the next version of the framework is available. Noting it down here, so you can have the pleasure of dismissing this comment once that's done :)

merlinc (changes requested):

locales/en/actions.json
teneightfive: These aren't actually used anywhere. They snuck into an old commit which ended up refactoring out the use of these keys.

So this is more of a tidy up.

common/presenters/message-banner/move-to-confirmed-assessment-banner.js
```diff
@@ -0,0 +1,35 @@
...
+    <p>
+      <a href="${baseUrl}/print" class="app-icon app-icon--print">
+        ${i18n.t('actions::print_assessment', {
+          context,
+        })}
+      </a>
+    </p>
+  `
+
+  return {
```
teneightfive, line 23: These blocks do feel slightly duplicated. I was toying with the idea of move the consisted properties back out to the parent method and maybe each of these just returns title/content?

But then I don't know in what sort of format it should return.

a flatter structure: 
```javascript
return {
  title: 'title',
  content: '<p>content</p>'
}

const presenter = bannerPresenter({ ... })
return {
  allowDismiss: false,
  classes: 'app-message--instruction govuk-!-padding-right-0',
  title: {
    text: presenter.title,
  },
  content: {
    html: presenter.content,
  },
}
```

or the same structure as the component: 
```javascript
return {
  title: {
    text: 'title',
  },
  content: {
    html: '<p>content</p>',
  }
}

return {
  allowDismiss: false,
  classes: 'app-message--instruction govuk-!-padding-right-0',
  ...presenter,
}
```

common/presenters/move-to-message-banner-component.js
solidgoldpig: What Robin said.

Also, I'm a bit concerned about all the generation of html within the presenter. Personally I would prefer to have that lifting done by an actual template. (and not have logic in our templates ;) )

teneightfive: I've made an attempt at this approach with the latest commit. Let me know your thoughts.

common/presenters/move-to-message-banner-component.js
teneightfive: Hmm. My opinion would be that a component might be a bit overkill for this.

The implementations all require different text and the only thing that we would be moving into a component would be the anchor element and the classnames. The current instances use a different href value too so I don't see a component offering much more benefit/customisation that using anchor tags with the classname applied at this stage.

solidgoldpig (approved):

merlinc (approved):

Issue #807, opened 2020-09-22
Allow permissions/locations to be set during development


teneightfive:
## Problem

During local development it is often required to use different permissions or a mixture of different permissions to perform and test particular actions within the application.

Currently the way to do this is to set the permissions required as an environment variable. But this then makes it hard to see what permissions are currently set or to switch them around.

## Possible solution

Support some form of session injection to be able to update the current set of permissions via the interface during local development.

Issue #804, opened 2020-09-22
Document permissions and roles


teneightfive:
## Problem

Permissions are currently maintained within [a module](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/8b12fe7276ffb42b9ca94d9fa31707da7be4a9f1/common/lib/permissions.js#L1) in the application. This current method makes it hard to anyone to see what permissions exist within the app at any given time and what those permissions allow users to do.

## Possible solution

- Document all current permissions used
- Document what permissions current Auth and NOMIS roles receive
- Add a prompt to the [Pull Request template](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/.github/pull_request_template.md) to ensure future permissions are documented

PR #1039, opened 2020-11-24
chore: Don't set development permissions if they already exist

teneightfive:
### What changed

Only set development permissions if they don't exist.

### Why did it change

This prevents the permissions being overridden on every request.

It also means that they can be updated via another method like session
injection and won't be overridden.

### DIFF
--- a/common/middleware/development.js
+++ b/common/middleware/development.js
@@ -15,8 +15,9 @@ function bypassAuth(bypass) {
 
 function setUserPermissions(permissions) {
   return (req, res, next) => {
-    if (permissions) {
-      req.session.user = req.session.user || {}
+    req.session.user = req.session.user || {}
+
+    if (permissions && !req.session.user.permissions) {
       req.session.user.permissions = permissions.split(',')
     }
 
--- a/common/middleware/development.test.js
+++ b/common/middleware/development.test.js
@@ -59,11 +59,7 @@ describe('Development specific middleware', function () {
 
     beforeEach(function () {
       req = {
-        session: {
-          user: {
-            permissions: [],
-          },
-        },
+        session: {},
       }
     })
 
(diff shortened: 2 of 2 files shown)


merlinc:
> It also means that they can be updated via another method like session
injection and won't be overridden.

I'm a big fan of this concept!

So, to be clear in my mind, I've got a couple of questions that I've answered - can you confirm that I've got this all correct. 

- How does changing of user permissions happen at the moment? 
  - Users have roles returned as part of the auth process, these are then matched up with hard coded lists of permissions. We do not have "per user" changes to these permissions.
- Does this introduce any changes to the current process
  - No, because to currently change user permissions, we could need to change code and redeploy anyway

teneightfive:
> * How does changing of user permissions happen at the moment?
>   
>   * Users have roles returned as part of the auth process, these are then matched up with hard coded lists of permissions. We do not have "per user" changes to these permissions.

Currently, the only way permissions are updated for a user is by them signing out (therefore killing their session) and singing back in again. 

On sign in, we get their role from the auth service and we match that to our [permissions definitions](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/lib/permissions.js). This is essentially the map between what particular roles can do within our application.

Application permissions are much more fine grained and it allows us to chop and change those between roles as we need to.

This change will hopefully also open up some progress on #804 and #807.

> * Does this introduce any changes to the current process
>   
>   * No, because to currently change user permissions, we could need to change code and redeploy anyway

Nope, this module [isn't actually loaded](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/6386a9695ff9f5ce8d7b65aa51935570f73ef7ef/server.js#L145-L152) in production and is only used for local development.

But this current structure prevents any form of updating of these permissions dynamically (during development or tests) so this fix opens that up too.

Here's a WIP branch I have to show a way we can do this during development: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/compare/development/update-permissions

merlinc (approved):

PR #1040, opened 2020-11-24
feat: Add population create/edit wizard

merlinc:
## Proposed changes

The second major part of the population dashboard is the ability to add or edit population details for a particular day. This PR adds the UI to enable this, plus the relevant service methods for create and update.

### What changed

- Add edit route
- Add wizard
- Add create and update to population service

### Issue tracking

- [P4-1431](https://dsdmoj.atlassian.net/browse/P4-1431)

### Editing
Screenshot_2020-11-24 Population Details - create_population

### With Error messages
Screenshot_2020-11-24 Error Population Details - create_population

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/population/middleware/set-population.js
```diff
@@ -44,6 +44,9 @@ async function setPopulation(req, res, next) {
       transfersOut,
     }
 
+    req.locationId = locationId
```
merlinc, line 47: When attaching the wizard to the router, all the params are lost, and the only one accessible is the `/edit` param.

These changes allow these variables to be used later on in the workflow.

app/population/controllers/edit/details.js
```diff
@@ -0,0 +1,4 @@
+const EditPopulationBaseController = require('./base')
+
+class DetailsController extends EditPopulationBaseController {}
```
teneightfive, line 3: I left a comment on the steps object around this, but if you move the initial values method and the save method to this controller you can reduce the amount of inheritance and files needed.

Am I right in assuming this is only ever going to be a one-step form?

app/population/fields/bedwatch.js
```diff
@@ -0,0 +1,16 @@
...
+  validate: ['required', 'numeric'],
+  component: 'govukInput',
+  label: {
+    text: 'fields::bedwatch.label',
+    classes: 'govuk-label--s',
+  },
+  id: 'bedwatch',
+  name: 'bedwatch',
+  classes: 'govuk-input--width-5',
+  autocomplete: 'off',
```
teneightfive, line 11: A lot of these fields seem to share some similar config. You might be able to reduce the duplication by creating a [template](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/app/move/fields/common.date.js) and [extending it](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/07b8b4dc197b790a985f63944b08de81eb5b1087/app/move/fields/date-to.js#L6) for each field.

common/services/population.js
```diff
@@ -18,6 +20,34 @@ const populationService = {
       include: ['moves_from', 'moves_to', 'location'],
     })
   },
+
+  create(locationId, date, data) {
```
teneightfive, line 24: You might be able to simplify these methods by just passing data in similar to how the [moves service works](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/07b8b4dc197b790a985f63944b08de81eb5b1087/common/services/move.js#L239-L253).

You can still do some checks to ensure an ID is present but maybe just a property within the `data` argument.

app/population/index.js
```diff
@@ -1,6 +1,8 @@
 const router = require('express').Router()
 const dailyRouter = require('express').Router({ mergeParams: true })
+const wizard = require('hmpo-form-wizard')
```
teneightfive, line 3: I did a [recent fix](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1022) which replaced calls to the form wizard to allow us to support unique wizards per _thing_ in different tabs.

It might be better to use the [internal wizard call](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/07b8b4dc197b790a985f63944b08de81eb5b1087/app/allocation/index.js#L4) similar to [other app loaders](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/07b8b4dc197b790a985f63944b08de81eb5b1087/app/allocation/index.js#L31). It allows you to specify a request param to create a unique wizard. I think this is something that would be useful in this instance too.

app/population/middleware/set-population.js
teneightfive: Have you tried using the [`mergeParams`](https://expressjs.com/en/api.html#express.router) option on the router? It might persist these through.

We've [used it elsewhere](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/3d5a283e30f8f22705ab76d23a32da8bace7519b/app/move/index.js#L4) to achieve something similar.

app/population/fields/unlock.js
```diff
@@ -0,0 +1,14 @@
+const unlock = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/overnights-out.js
```diff
@@ -0,0 +1,14 @@
+const overnightsOut = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/discharges.js
```diff
@@ -0,0 +1,14 @@
+const discharges = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/overnights-in.js
```diff
@@ -0,0 +1,14 @@
+const overnightsIn = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/usable-capacity.js
```diff
@@ -0,0 +1,14 @@
+const usableCapacity = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/bedwatch.js
```diff
@@ -0,0 +1,16 @@
+const bedwatch = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/out-of-area-courts.js
```diff
@@ -0,0 +1,16 @@
+const outOfAreaCourts = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/fields/operational-capacity.js
```diff
@@ -0,0 +1,16 @@
+const operationalCapacity = {
```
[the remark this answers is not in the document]
merlinc: Refactored

app/population/steps/edit.js
```diff
@@ -0,0 +1,32 @@
...
+      'unlock',
+      'bedwatch',
+      'overnights_in',
+      'overnights_out',
+      'out_of_area_courts',
+      'discharges',
+    ],
+    next: 'save',
+  },
+  '/save': {
```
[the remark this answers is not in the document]
merlinc: Refactored to only use one controller.

app/population/controllers/edit/details.js
merlinc: Refactored to only use one controller.

app/population/fields/bedwatch.js
merlinc: Refactored to use a construction function as that created less duplication than extending it for each field

app/population/index.js
merlinc: Update this to use the `unique-form-wizard`.

Unfortunately there isn't a unique single parameter to use - the location and date are params, and are used when editing to load a population by `id`. Would this work with being a `population.id`? Also, on create, there is no population to load, so there's no request param that can be used - how would this work in that case?

app/population/middleware/set-population.js
merlinc: Yeah, `mergeParams` is already being used on the subrouter, and it should work as its working on other parts of this route.

I've tried refactoring this a few times, changing the order of adding and applying and its broken every time :broken_heart:  

So given that it currently works, I don't really want to change this until I have E2E tests in place to protect me.

app/population/controllers/edit/base.js
```diff
@@ -0,0 +1,16 @@
+const { omit } = require('lodash')
+
+const FormWizardController = require('../../../../common/controllers/form-wizard')
+
+class EditPopulationBaseController extends FormWizardController {
+  setInitialValues(req, res, next) {
+    if (req.journeyModel?.get('history').length === 1 && req.population) {
```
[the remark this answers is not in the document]
merlinc: So in the end, I've used the same line that you've used.

As previously discussed, I'm not thrilled about this - it definitely feels like a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction). And one thing I've noticed is that if the session expires, then bugs in this function will prevent the normal flow from occuring (e.g. in my code, if the history doesn't exist an error appears, not the session expired page).

This is a bad smell to me, so I'm going to keep thinking about it - but icky working code is better then perfect imaginary code :smile:

app/population/controllers/edit/base.js
```diff
@@ -0,0 +1,16 @@
...
+      const values = omit(req.population, ['moves_from', 'moves_to'])
```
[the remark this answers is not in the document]
merlinc: No need anymore, due to refactoring

app/population/views/edit.njk
```diff
@@ -0,0 +1,5 @@
+{% extends "form-wizard.njk" %}
+
+{% block fields %}
+  {{ super() }}
```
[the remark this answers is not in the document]
merlinc: Removed the template, I'll add it back again when its needed in the next PR

common/services/population.js
merlinc: Refactored to use the move service approach

app/population/index.js
```diff
@@ -22,6 +26,17 @@ router.get('/', redirectBaseUrl)
...
+const editConfig = {
+  controller: FormWizardController,
+  name: 'edit-population',
+  templatePath: 'population/views/edit/',
+  template: '../../../form-wizard',
+  journeyName: 'edit-population',
+  journeyPageTitle: 'actions::create_population',
+}
+
+dailyRouter.use('/edit', wizard(editSteps, editFields, editConfig))
```
teneightfive, line 38: I think it might be useful to use the 4th argument of this method to specify a unique value for the wizard. 

I think with the default setup it will suffer from what we experienced with the other instances. Where if you were to open multiple edit pages in different tabs they will be sharing the same session and journey models. This will lead to errors/values being shared across tabs incorrectly.

app/population/index.js
teneightfive: Ah, just seen this comment after leaving one on this file again.

Yes, I see the issue with your route, you don't have a single property as you're using a combination of date and location ID. You could create a property in the middleware that sets those that you can then use in this instance. It could even just combine the two:

```javascript
# middleware
req.wizardKey = `${req.locationId}-${req.date}`

# index loader
dailyRouter.use('/edit', wizard(editSteps, editFields, editConfig, 'wizardKey'))
```

app/population/middleware/set-population.js
teneightfive: Yeah, so I think it might actually need to be the [router created within the form wizard](https://github.com/HMPO/hmpo-form-wizard/blob/c8131187ddaafd220af89e75b8c2a400f4bdd78e/lib/wizard.js#L11) that needs that option, which might be why you can't access it.

Think this is good for now 👍

app/population/steps/edit.js
```diff
@@ -0,0 +1,26 @@
...
+    fields: [
+      'operational_capacity',
+      'usable_capacity',
+      'unlock',
+      'bedwatch',
+      'overnights_in',
+      'overnights_out',
+      'out_of_area_courts',
+      'discharges',
+    ],
```
teneightfive, line 24: You might want to update the button text for this step. I noticed it currently uses the default of "Continue" which suggests there's another step.

We currently have a `confirm_and_save` action but might want to check with @Andy-Millar on what's the best copy here. Maybe just "Save".

But once you have the action key you can add this line:

```
'/details': {
  ...
  buttonText: 'actions::confirm_and_save',
  ...
}
```

teneightfive (approved):

app/population/steps/edit.js
merlinc: Good idea - set it to use "Confirm and Save" for now, if @Andy-Millar wants it changed it's easy enough to tweak later :)

app/population/index.js
merlinc: *nods* I'd thought about using a composite key, hadn't seen any other usage of it in the system though - I've gone with your suggestion :)

Andy-Millar:
@merlinc Just picking this up now - some feedback on this from me:

- H1 should be 'Add numbers' (if come from 'Add numbers' link) or 'Change numbers' (if come from 'Change numbers' link)
- Button should be 'Save and calculate' 
- 'Cancel' link next to the button would be nice to have

PR #1048, opened 2020-11-25
chore: Add groundwork for req context

Thource:
[P4-1776] Add groundwork for the req context, to allow passing data from req to headers

## Proposed changes
All of this work was done by Alex, and has been picked from his PR https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/906

The intention of this PR is that it will act as a base PR, and I will make a new PR for each service, to add the request context to them individually. The reasoning behind this is that it should make it easier to understand what's going on in each PR.

### What changed
Added `ApiClient#_originalReq`, which is the current request
Added set-services.js, which sets `request.services`, this contains the functions needed to get service objects with request context via `req.services.move()` for example
Added set-transaction-id.js which sets `request.transactionId` to `req.get('x-request-id') || uuid.v4()`
Changed request-headers.js to take the request and include X-Transaction-Id in the headers

### Why did it change
This has changed so that we can get information about the request (user, headers, etc) more easily, especially in request-headers.js.

### Issue tracking

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/lib/api-client/index.js
+++ b/common/lib/api-client/index.js
@@ -16,17 +16,13 @@ const {
 } = require('./middleware')
 const models = require('./models')
 
-let instance
-
-module.exports = function () {
-  if (instance) {
-    return instance
-  }
-
-  instance = new JsonApi({
+module.exports = function (req) {
+  const instance = new JsonApi({
     apiUrl: API.BASE_URL,
     logger: false,
   })
+
+  instance._originalReq = req
 
   instance.replaceMiddleware('errors', errors)
   instance.replaceMiddleware('POST', post(FILE_UPLOADS.MAX_FILE_SIZE))
--- a/common/middleware/set-services.js
+++ b/common/middleware/set-services.js
@@ -1,0 +1,26 @@
+const services = require('../services')
+
+const setServices = (req, res, next) => {
+  const servicesRequestContext = {}
+
+  req.services = Object.keys(services).reduce((acc, serviceKey) => {
+    if (!servicesRequestContext[serviceKey]) {
+      const originalService = services[serviceKey]
+
+      if (!originalService.addRequestContext) {
+        servicesRequestContext[serviceKey] = originalService
+      } else {
+        servicesRequestContext[serviceKey] = originalService.addRequestContext(
+          req
+        )
+      }
+    }
+
+    acc[serviceKey] = () => servicesRequestContext[serviceKey]
+
+    return acc
+  }, {})
+  next()
+}
+
+module.exports = setServices
--- a/common/middleware/set-services.test.js
+++ b/common/middleware/set-services.test.js
@@ -1,0 +1,57 @@
+const proxyquire = require('proxyquire').noCallThru()
+
+const services = {
+  foo: {
+    addRequestContext: sinon.stub().returns({ method: 'foo - with context' }),
+  },
+  bar: {
+    method: 'bar - without context',
+  },
+}
+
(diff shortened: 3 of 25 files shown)


common/middleware/set-services.js
```diff
@@ -0,0 +1,26 @@
+const services = require('../services')
+
+const setServices = (req, res, next) => {
+  const servicesRequestContext = {}
```
teneightfive, line 4: Is the intention of this to cache the services? I think this placement will result it being cleared on each request.

common/lib/api-client/index.js
```diff
@@ -16,18 +16,14 @@ const {
...
-let instance
-
-module.exports = function () {
-  if (instance) {
-    return instance
-  }
-
-  instance = new JsonApi({
+module.exports = function (req) {
+  const instance = new JsonApi({
```
teneightfive, line 20: What's the reason for removing the singleton pattern here?

By removing this is means we'll be creating a new instance for every call. I'm trying to rack my brain but we had some import issues with creating a new instance each time.

I think it would be neater if we can retain the API client as a singleton but find a way to define the req with a particular call.

common/middleware/set-services.test.js
```diff
@@ -0,0 +1,57 @@
+const proxyquire = require('proxyquire').noCallThru()
+
+const services = {
+  foo: {
```
merlinc, line 4: I find that in test code using metasyntactic variables like foo,bar,baz to be less readable than variables that describe what they mean or do.

For example here, you have a `serviceWithContext` and `serviceWithoutContext`

common/services/move.test.js
```diff
@@ -1,10 +1,18 @@
 const proxyquire = require('proxyquire')
 
-const apiClient = require('../lib/api-client')()
+const PersonService = {}
```
merlinc, line 3: Perhaps @teneightfive or @solidgoldpig could correct me, but I haven't seen stub test code structured like this, in this particular repository.

I wonder if there's a different way that more closely matches the test structure that is already present. (Or if there is a compelling reason for us to refactor other tests)

common/services/reference-data.test.js
```diff
@@ -375,7 +373,7 @@ describe('Reference Data Service', function () {
           apiClient.findAll.resolves(mockResponse)
           await referenceDataService.getLocations({ include: ['foo', 'bar'] })
         })
-        it('should pass include paramter to api client', function () {
+        it('should pass include parameter to api client', function () {
```
merlinc, line 376: yptos! :grinning:

server.js
```diff
@@ -32,6 +32,8 @@ const sentryEnrichScope = require('./common/middleware/sentry-enrich-scope')
 const sentryRequestId = require('./common/middleware/sentry-request-id')
```
merlinc, line 32: Completely not the scope of this PR, but with 205 lines of code in `server.js` I wonder if we should start writing some tests for it - and it is separated out from the running of the server, which makes this a lot easier to do as well

common/middleware/set-services.js
```diff
@@ -0,0 +1,26 @@
...
+
+  req.services = Object.keys(services).reduce((acc, serviceKey) => {
```
merlinc, line 6: Now we're dynamically adding services, I wonder if adding a small debug logger line to indicate what's been added, might be useful to understand what's happening?

PR #1051, opened 2020-11-26
API client fixes

teneightfive:
### What changed

Includes two fixes the API client:

- don't store `undefined` in cache keys
- include querystring in URL during debug logging

### Why did it change

The cache keys should only return params that are sent as part of the URL.

Exposing the full URL allows us to see the differences in calls to the API. In some cases it differs on querystring filters like `page` so its useful to be able to distinguish those requests.

#### Automated testing

- [ ] ~Added end-to-end tests to cover any journey changes~

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/lib/api-client/middleware/request.js
```diff
@@ -13,8 +13,10 @@ function requestMiddleware({ cacheExpiry = 60, useRedisCache = false } = {}) {
       }
 
       const { req, jsonApi, cacheKey } = payload
+      const searchString = new URLSearchParams(req.params).toString()
```
Thource, line 16: Does this want to use pickBy too?

teneightfive: Interesting point actually. The request middleware is actually after the [`requestInclude`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/2181f093f537546a9b04ce4a8dc1a05cb00a9aa1/common/lib/api-client/middleware/request-include.js#L22-L26) middleware that cleans up that key.

I wonder if we actually want to have something at the start of the middleware that cleans the params of any undefined values... @solidgoldpig what are your thoughts as I think you added the request include middleware?

teneightfive:
@martyn-w this should solve the `include=undefined` you spotted in the redis cache.

common/lib/api-client/middleware/request.js
solidgoldpig: axios strips them out. Do it again here if you want ;)

solidgoldpig (approved):

PR #1052, opened 2020-11-26
Add ability to navigate to specific date in dashboard [P4-1824]

solidgoldpig:
### What changed

Adds ability to ‘Jump to date’

- Adds route to enable 'Jump to date' 
  - takes input and creates new date string in format `yyyy-MM-DD`
  - redirects to referring url, replacing original date with new date
- Adds 'Jump to date' link to date pagination
  - links to date-select route with original url appended as the referrer

### Why did it change

Feature requested by user and validated by design/research

### Issue tracking

- [P4-1824 - Ability to navigate to a specific date in the dashboard](https://dsdmoj.atlassian.net/browse/P4-1824)

## Screenshots

| After | Before |
| ------ | ----- |
| <kbd>Outgoing_moves_for_Thursday_26_Nov_2020__Today_</kbd> | <kbd>Outgoing_moves_for_Thursday_26_Nov_2020__Today_</kbd> |
| <kbd>Jump_to_date</kbd> |  |
| <kbd>Jump_to_date</kbd> |  |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


Andy-Millar:
Looking at this - this is an instance where entering "next friday" would be really useful for users I think...

I wonder if we expand our hint text in two ways:
1. to be more context (ie. 28 October **2020** - rather than 1990, which is more relevant for DOBs) - can we have this update to display a day +/- 7 days away from the present date? I understand this might be a bit out there... 
2. surface the "next Friday" command

ie.
For example 20/11/2020, 20 November 2020 or next Friday ["next Friday"]
?

test/e2e/date-select.test.js
```diff
@@ -0,0 +1,49 @@
...
+    {
+      selector: page.nodes.dateSelectInput,
+      value,
+    },
+  ])
+  await page.submitForm()
+  await t.expect(page.getCurrentUrl()).contains(expectedDate)
+}
+
+fixture('Jump to date').beforeEach(async t => {
```
merlinc, line 31: I've manually tested this work with the new population route, and it does. I'm still working on the population E2E tests, so I'll add a test in here once that's ready.

app/date-select/controller.js
```diff
@@ -0,0 +1,34 @@
+const FormWizardController = require('../../common/controllers/form-wizard')
+const { dateRegex } = require('../../common/helpers/url')
+
+const DateRegExp = new RegExp('/' + dateRegex)
+
+class DateSelectController extends FormWizardController {
```
merlinc, line 6: I thought we weren't going to use the FormWizard for this control, as it's a single UI element, not a wizard collecting data?

app/date-select/controller.js
```diff
@@ -0,0 +1,34 @@
...
+  get(req, res, next) {
+    if (req.query.referrer) {
+      req.sessionModel.reset()
```
merlinc, line 9: I think this might cause issues with users with multiple tabs.

Scenario: 
I start entering a move, allocation or escort record
I open a new tab to view a previous move I want to copy data from
I use the jump to date feature to get to the move I want
I go back to my previous form, and it has now been reset.

app/date-select/controller.js
merlinc: Ah! This is a separate form for entering the date that is being jumped to!

app/date-select/controller.js
merlinc: Maybe we could set this up as a different journey instead?

app/date-select/controller.js
```diff
@@ -0,0 +1,34 @@
...
+      return res.redirect(req.form.options.fullPath)
+    }
+
+    const referrer = req.sessionModel.get('referrer')
+
+    if (!referrer) {
+      return res.redirect('/')
+    }
+
+    res.locals.cancelUrl = referrer
```
teneightfive, line 20: Maybe these should be set in a [`middlewareLocals`](https://github.com/UKHomeOffice/passports-form-wizard/wiki/HMPO%20Forms%20Flow.pdf) middleware?

app/date-select/controller.js
teneightfive, line 9: I think this should be handled using a redirect first step like mentioned on the config file. Then you can use a `middlewareSetup` middleware to handle setting the query value to the session.

app/date-select/controller.js
```diff
@@ -0,0 +1,34 @@
...
+  get(req, res, next) {
+    if (req.query.referrer) {
+      req.sessionModel.reset()
+      req.sessionModel.set('referrer', req.query.referrer)
+      return res.redirect(req.form.options.fullPath)
+    }
+
+    const referrer = req.sessionModel.get('referrer')
+
+    if (!referrer) {
```
teneightfive, line 16: This feels like a check so maybe should be in a `middlewareChecks` middleware?

app/date-select/controller.js
```diff
@@ -0,0 +1,34 @@
...
+
+    res.locals.cancelUrl = referrer
+    res.locals.backLink = referrer
+
+    super.get(req, res, next)
+  }
+
+  successHandler(req, res, next) {
+    const { date_select: dateSelect, referrer } = req.sessionModel.toJSON()
+    req.sessionModel.reset()
```
teneightfive, line 28: Should we also kill the `journeyModel`?

app/date-select/controller.js
```diff
@@ -0,0 +1,34 @@
+const FormWizardController = require('../../common/controllers/form-wizard')
+const { dateRegex } = require('../../common/helpers/url')
+
+const DateRegExp = new RegExp('/' + dateRegex)
+
+class DateSelectController extends FormWizardController {
+  get(req, res, next) {
+    if (req.query.referrer) {
+      req.sessionModel.reset()
+      req.sessionModel.set('referrer', req.query.referrer)
```
teneightfive, line 10: You could maybe even use `forwardQuery` and then just always pull it from the querystring rather than needing to pull it from the session.

common/controllers/form-wizard/index.js
```diff
@@ -135,20 +135,7 @@ class BaseController extends Controller {
...
-        }),
-        href: `#${fields[error.key].id}`,
-      }
-    })
-
-    return {
-      ...errors,
-      errorList,
-    }
+    return fieldHelpers.addErrorListToErrors(errors, fields)
```
teneightfive, line 138: Is this change being used anywhere still? It looks like it was put in place for a different approach that you then changed.

Maybe we can even drop this commit/change until we have a need for it?

app/date-select/controllers.test.js
```diff
@@ -0,0 +1,144 @@
+// const filters = require('../../common/controllers/filters')
```
teneightfive: It would be good if we can tidy up this commit, maybe using an interactive rebase so that we don't need to worry about committing this approach to the history.

Andy-Millar:
> Looking at this - this is an instance where entering "next friday" would be really useful for users I think...
> 
> I wonder if we expand our hint text in two ways:
> 
> 1. to be more context (ie. 28 October **2020** - rather than 1990, which is more relevant for DOBs) - can we have this update to display a day +/- 7 days away from the present date? I understand this might be a bit out there... Also good if we did the same for booking move dates...?
> 2. surface the "next Friday" command
> 
> ie.
> For example 20/11/2020, 20 November 2020 or next Friday ["next Friday"]
> ?

Did we have any thoughts on this?

app/date-select/controller.js
solidgoldpig: As discussed the errors get stashed on the sessionModel for that specific wizard config - it doesn't effect the other, er, journeys

locales/en/actions.json
```diff
@@ -6,6 +6,7 @@
   "previous_week": "Previous week",
   "next_week": "Next week",
   "current_week": "This week",
+  "date_select": "Jump to date",
```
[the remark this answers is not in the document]
solidgoldpig: Disagree

` "current_day": "Today"`

app/date-select/controller.js
solidgoldpig: that would lead to csrf issues if someone had double-tabbed. So why bother?

app/date-select/steps.js
```diff
@@ -0,0 +1,13 @@
+const DateSelectController = require('./controller')
+
+const steps = {
+  '/': {
```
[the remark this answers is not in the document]
solidgoldpig: And then you end up with very funky urls.

This causes no strange behaviour with saved value and error values. The only edge case is if someone opens up jump-to-date pages in multiple tabs. Then there will be no referrer value when it comes to submit. Using the above pattern and adding the query forward etc, just leads me to think that we shouldn't use form-wizard for this at all.

common/controllers/form-wizard/index.js
solidgoldpig: I actually think it makes the tests for the form-wizard controller cleaner, so personally I think it should stay for that reason alone, in addition to making it reusable

app/date-select/controllers.test.js
solidgoldpig: this file doesn't even exist any more, so not sure how it showed up in the list of files for you

common/components/pagination/template.njk
```diff
@@ -24,7 +24,7 @@
     {% endif %}
 
     {% for item in params.items %}
-      <li class="app-pagination__list-item">
+      <li class="app-pagination__list-item {{- ' ' + item.classes if item.classes}}"  {%- for attribute, value in item.attributes -%} {{ attribute }}="{{ value }}"{% endfor %}>
```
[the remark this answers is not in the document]
solidgoldpig: Also added document for `items` which was missing

test/e2e/_routes.js
```diff
@@ -26,3 +26,4 @@ export const newAllocation = `${E2E.BASE_URL}/allocation/new`
 export const allocations = `${E2E.BASE_URL}/allocations`
 export const allocationsWithDate = date =>
   `${E2E.BASE_URL}/allocations/week/${date}/outgoing`
+export const singleRequests = `${E2E.BASE_URL}/moves/requested`
```
merlinc, line 29: Is this related to this PR?

common/middleware/collection/set-pagination.js
```diff
@@ -1,5 +1,6 @@
 const { format } = require('date-fns')
 
+const DATESELECT = require('../../../app/date-select/constants')
```
merlinc, line 3: Is this the first instance of `common` reaching into the app directory?

app/date-select/controller.test.js
```diff
@@ -0,0 +1,181 @@
...
+        )
+      })
+    })
+  })
+
+  describe('#successHandler', function () {
+    context('when called', function () {
+      beforeEach(function () {
+        req.sessionModel.toJSON.returns({
+          referrer: '/bar/foo-1234/outgoing',
```
merlinc, line 164: Could these examples use actual dates, to make it clearer as to what is happening.
`2020-01-01` changing to `2020-12-01` is more meaningful that `foo-1234` changing to `stored-date`

common/helpers/field/set-errors.js
```diff
@@ -0,0 +1,26 @@
+/**
```
merlinc, line 1: Interesting way to work around the form-wizard issues - fake it till you make it! :)

merlinc (commented):
This looks like an annoyingly complicated piece of work, and it looks like the need for multiple forms at the same time is causing some fundamental issues to work around.

I'm not see anything obviously wrong to me at the moment, but @teneightfive has a number of open comments, so I'm going to be polite to him and not approve the PR just yet.

common/middleware/collection/set-pagination.js
solidgoldpig: Unfortunately not

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/helpers/field/index.js#L11-L12

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/middleware/locals.js#L3

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/3516af7a7acdf8020b9d4327b71de88edfd794a9/common/presenters/move-to-meta-list-component.js#L3

test/e2e/_routes.js
solidgoldpig: Yep. It provides the url for the single requests page so that e2e tests can check the jumper on it

app/date-select/controller.test.js
solidgoldpig: I wanted it to be clear that we are using a mocked value rather than the actual regex. Have hopefully now made it explicit enough.

locales/en/actions.json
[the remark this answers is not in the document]
teneightfive: That’s why I said “most” 😉

app/date-select/controller.js
teneightfive: The CSRF is no longer tied to the journey. That was the work I had to do previously around extracting it to the user’s global session. 

Before that work the FormWizard actually stored the CSRF token against the session so killing the session would have creates a CSRF issue when using multiple tabs. 

The reason for suggesting cleaning it up is that we won’t be left with old journeys on the global session that way.

common/controllers/form-wizard/index.js
teneightfive: That’s fair.

app/date-select/controllers.test.js
teneightfive: It showed up when I went through the commits. It was attached to the early commit and then removed in a later one.

teneightfive: Once this is ready maybe you can ensure it’s removed at the right commit and then it’ll never enter the history.

teneightfive:
> > Looking at this - this is an instance where entering "next friday" would be really useful for users I think...
> > I wonder if we expand our hint text in two ways:
> > 
> > 1. to be more context (ie. 28 October **2020** - rather than 1990, which is more relevant for DOBs) - can we have this update to display a day +/- 7 days away from the present date? I understand this might be a bit out there... Also good if we did the same for booking move dates...?
> > 2. surface the "next Friday" command
> > 
> > ie.
> > For example 20/11/2020, 20 November 2020 or next Friday ["next Friday"]
> > ?
> 
> Did we have any thoughts on this?

@Andy-Millar I’m not quite sure what you mean about surfacing a +/- 7 day command? 

Do you mean having a weekly navigation alongside the day navigation? 

My opinion would be to start with the simplest thing here of updating the hint text. I think having “Next day” alongside “Next week” could look confusing.

app/date-select/controller.js
```diff
@@ -0,0 +1,47 @@
...
+  checkReferrer(req, res, next) {
+    if (req.query.referrer) {
+      req.sessionModel.reset()
+      req.sessionModel.set('referrer', req.query.referrer)
+      return res.redirect(req.form.options.fullPath)
+    }
+
+    const referrer = req.sessionModel.get('referrer')
+
+    // if someone has opened up the jump-to-date page in 2 tabs then maybe, just maybe, the session model won't have the referrer info. In that case, the best we can do is go back to the start. Or of course, maybe someone enters the url directly without a referrer
```
teneightfive, line 21: Why don’t we just persist the query and always use that in the success handler?

config/i18n/index.js
```diff
@@ -18,6 +18,7 @@ i18next.use(Backend).init({
     'dashboard',
     'default',
     'actions',
+    'date_select',
```
teneightfive, line 21: Is this still needed?

locales/en/actions.json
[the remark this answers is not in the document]
teneightfive: It does feel like you’re basing your addition on the 5% instead of the 95% though. All the other submit action keys follow the pattern of using the value in the key.

app/date-select/controller.js
solidgoldpig: that's exactly what I'm able to do now that I zap the errors just before rendering

solidgoldpig:
> @Andy-Millar I’m not quite sure what you mean about surfacing a +/- 7 day command?

I think Andy means that the date shown should be dynamic and always show a date 7 days ago or from now. 

i18next actually makes this quite tricky.

1. Create the values at run time and inject them into the date hint text

ie. the date will be fixed relative to the deploy time and will stay the same until the next deploy

2. Create the values at request time and pass them as part of the context

We'd either have to pass the values explicitly or intercept `i18n.t` to automatically add them in

And if we did that would these values be pre-formatted or would we add some more formatters to the interpolater to take care of that?


.

app/date-select/controllers.test.js
solidgoldpig: Already done

app/date-select/controller.js
solidgoldpig: Superceded

teneightfive (approved):

locales/en/fields.json
```diff
@@ -1,8 +1,12 @@
 {
   "date": {
-    "hint": "For example 28/10/1990 or 28 October 1990",
+    "hint": "For example 28/11/2020 or 28 November 2020",
```
teneightfive, line 3: So going on what we just discussed, I think we should update this one to be `For example 28/10/2020, 28 October 2020 or next Friday`

And you could adapt the DOB field to use `For example 28/10/1990, 28 October 1990 or next Friday` as the 1990 might be more relevant for historic dates like @Andy-Millar suggested.

I think it _should_ just pickup the override hint field for the dob field. If it's not that simple leave it for now.

solidgoldpig: I don't think there's any need for the 'next friday' in the historic one ;)

PR #1053, opened 2020-11-26
Reverse order of items in move timeline

solidgoldpig:
### What changed

- Reverses order in which timeline items are output
- Adds basic e2e tests for move timeline
  - ensures that timeline view can be rendered
  - ensures that minimal events exist on page

### Why did it change

Reverse order is as per design.

Tests are good, but more to the point, ensuring that the timeline view can be rendered would have caught the infinite loop bug in the model `transformResource` method.

### Issue tracking

n/a

## Screenshots

| After | Before |
| ------ | ----- |
| <kbd>Move_details_for_SCHMITT__ERICH_–_Book_a_secure_move</kbd> | <kbd>Move_details_for_SCHMITT__ERICH_–_Book_a_secure_move</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/test/e2e/move.timeline.test.js
+++ b/test/e2e/move.timeline.test.js
@@ -1,0 +1,17 @@
+import { expectStatusCode } from './_helpers'
+import { createMove } from './_move'
+import { policeUser } from './_roles'
+import { home, getTimeline } from './_routes'
+import { moveTimelinePage } from './pages'
+
+fixture('Move timeline').beforeEach(async t => {
+  await t.useRole(policeUser).navigateTo(home)
+  await createMove()
+})
+
+test('Check timeline exists', async t => {
+  const timelineUrl = getTimeline(t.ctx.move.id)
+  await expectStatusCode(timelineUrl, 200)
+  // there should be the actual proposed event and a corresponding auto-inserted triggered event
+  await t.expect(moveTimelinePage.nodes.timelineItems.count).eql(2)
+})
--- a/test/e2e/pages/move-timeline.js
+++ b/test/e2e/pages/move-timeline.js
@@ -1,0 +1,17 @@
+import { Selector } from 'testcafe'
+
+import MoveDetailPage from './move-detail'
+
+class MoveTimelinePage extends MoveDetailPage {
+  constructor() {
+    super()
+
+    this.nodes = {
+      ...this.nodes,
+      timeline: Selector('.app-timeline'),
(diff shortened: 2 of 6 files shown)


test/e2e/move.timeline.test.js
```diff
@@ -0,0 +1,38 @@
...
+    name: 'STC user',
+    role: stcUser,
+  },
+  {
+    name: 'SCH user',
+    role: schUser,
+  },
+]
+
+users.forEach(user => {
```
teneightfive, line 26: Do we need to e2e test for every user? Just thinking of our overall speed in this scenario.

The main purpose of this test is to ensure the timeline is working right, not that particular users have access to it. The access for roles can be handled by the permissions library unit tests.

test/e2e/pages/move-timeline.js
```diff
@@ -0,0 +1,17 @@
...
+import MoveDetailPage from './move-detail'
+
+class MoveTimelinePage extends MoveDetailPage {
+  constructor() {
+    super()
+
+    this.nodes = {
+      ...this.nodes,
+      timeline: Selector('.app-timeline'),
+      timelineItem: Selector('.app-timeline__item'),
```
teneightfive, line 12: `timelineItems`? I'm guessing there will be many?

teneightfive (approved):

test/e2e/move.timeline.test.js
merlinc: Yeah, I think I introduced that pattern of e2e tests for every user in another PR, and it was not the greatest idea in retrospect. It would be much better served as some non-existent fast & mocked data test.

So, if we're sure that the request will behave the same way for all users, then we could reduce this down to one main user type.

test/e2e/pages/move-timeline.js
solidgoldpig: mmkay

test/e2e/move.timeline.test.js
teneightfive: Yeah, I think we have the smoke tests which log in as the different user types we know, I guess as a last ditch to see if there is anything glaringly wrong with a particular user but I think for areas like this it's sufficient to test it with one user that we know has access to it.

PR #1069, opened 2020-12-01
[P4-2391] Move transformer

teneightfive:
### What changed

This includes some changes to help add support for the youth risk assessment.

It:

- refactors the way the assessment answers and frameworks are rendered on the view template so that they share the same method
- adds an extra property to the move that can be used to determine logic for the youth risk assessment.

### Why did it change

The youth risk assessment is another framework that needs to be supported alongside the existing data that comes from the move request as well as data that comes from the person escort record.

These changes allow this logic to be simplified and allows more re-use of existing patterns.

### Issue tracking

- [P4-2391](https://dsdmoj.atlassian.net/browse/P4-2391)

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/views/_includes/person-escort-record-summary.njk
+++ b/app/move/views/_includes/person-escort-record-summary.njk
@@ -5,26 +5,11 @@
     </h2>
 
     {% if assessment.isCompleted %}
-      {% for panel in assessment.panels %}
-        {% call appPanel(panel) %}
-          {{ appMetaList(panel.metaList) }}
-        {% endcall %}
-      {% else %}
-        {{ appMessage({
-          classes: "app-message--muted govuk-!-margin-top-2",
-          allowDismiss: false,
-          content: {
-            html: t("person-escort-record::flags.empty", {
-              name: assessment.name,
-              url: assessment.url
-            })
-          }
-        }) }}
-      {% endfor %}
+      {{ renderAssessmentComponent(assessment) }}
     {% else %}
       {{ govukInsetText({
         classes: "govuk-inset-text--compact govuk-!-margin-0",
-        text: t("person-escort-record::flags.section_incomplete")
+        text: t("assessment::section_incomplete")
       }) }}
     {% endif %}
 
@@ -36,7 +21,9 @@
       {{ govukDetails({
         html: html,
         classes: "govuk-!-font-size-16 govuk-!-margin-top-2",
-        summaryText: t("person-escort-record::flags.information_provided_when_requested")
+        summaryText: t("assessment::view_previous_assessment", {
+          context: assessment.previousAssessment.context
+        })
       }) }}
     {% endif %}
 
--- a/common/lib/api-client/transformers/move.transformer.js
+++ b/common/lib/api-client/transformers/move.transformer.js
@@ -1,0 +1,11 @@
+const { get } = require('lodash')
+
+module.exports = function moveTransformer(data) {
+  const fromLocationType = get(data, 'from_location.location_type')
+  const isYouthMove = [
+    'secure_training_centre',
+    'secure_childrens_home',
+  ].includes(fromLocationType)
+
+  data._is_youth_move = isYouthMove
+}
--- a/common/presenters/framework-section-to-panel-list.js
+++ b/common/presenters/framework-section-to-panel-list.js
@@ -1,5 +1,7 @@
 const { find, kebabCase, omit } = require('lodash')
 
+const componentService = require('../services/component')
+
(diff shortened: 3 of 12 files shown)


common/lib/api-client/transformers/move.transformer.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function moveTransformer(data) {
+  const isYouthMove = [
+    'secure_training_centre',
+    'secure_childrens_home',
+  ].includes(data.from_location?.location_type)
+
+  return {
+    ...data,
+    _is_youth_move: isYouthMove,
```
teneightfive, line 9: I've been between minds about this pattern. I used it on the [assessment transformer](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/3516af7a7acdf8020b9d4327b71de88edfd794a9/common/lib/api-client/transformers/assessment.transformer.js#L15) to in order to be able to distinguish between properties added within our service and properties that have come from the API resource.

This isn't something we do in the [person transformer](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/3516af7a7acdf8020b9d4327b71de88edfd794a9/common/lib/api-client/transformers/person.transformer.js#L4-L5) (yet) though because when moving that there I didn't want to apply that everywhere.

The other reason for the `_` prefix is to avoid name clashes. The person escort record actually already has a `framework` property which I didn't want to override.

What are other's thoughts on this?

common/presenters/framework-section-to-panel-list.js
```diff
@@ -51,42 +53,46 @@ function frameworkSectionToPanelList({
...
-            // Remove link so that when displayed in the main
-            // body the tag isn't linked
-            tag: omit(tag, ['href']),
-            attributes: { id: kebabCase(tag.text) },
-            metaList: frameworkResponseToMetaListComponent(
-              flagsWithQuestionDescription
-            ),
-          }
-        }),
+      count: panels.length,
```
teneightfive, line 93: I think there is still a large amount of improvement around this presenter but didn't want to tackle it all at once.

I've made [some improvements](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1070) to the way the person escort record gets transformer that will hopefully allow this to be simplified.

app/move/views/_includes/person-escort-record-summary.njk
```diff
@@ -5,26 +5,11 @@
...
-          allowDismiss: false,
-          content: {
-            html: t("person-escort-record::flags.empty", {
-              name: assessment.name,
-              url: assessment.url
-            })
-          }
-        }) }}
-      {% endfor %}
+      {{ renderAssessmentComponent(assessment) }}
```
teneightfive, line 8: The main benefit of this change is that it will allow a youth risk assessment section to be set as a `previousAssessment` and to be rendered correctly.

common/lib/api-client/transformers/move.transformer.js
solidgoldpig: Seems like a good idea.

Are we definitely not using `_` to start properties elsewhere? (Could we use `$` or a more explicit prefix `auto_`, `virtual_`?

Should we use this pattern for all created properties... eg. `person.fullname`? (I guess that's what you're getting at)

And obviously, it should be properly documented…

teneightfive: >Are we definitely not using _ to start properties elsewhere? (Could we use $ or a more explicit prefix auto_, virtual_?

That's a good question. I'm not aware of any API attributes using an underscore prefix. I went with underscore initially as I know it's something express does with the request object. 

>Should we use this pattern for all created properties... eg. person.fullname? (I guess that's what you're getting at)

Any yep, if we think it's a pattern worth using we should do it in all places and document it (as you pointed out).

>And obviously, it should be properly documented…

Agreed!

common/presenters/framework-section-to-panel-list.js
solidgoldpig: Where is `count` actually being used?

solidgoldpig (approved):

common/presenters/framework-section-to-panel-list.js
teneightfive: It's used in [the macro](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/e83f3e9812c497769d2243c37d320cac86ebbc78/app/move/views/view.njk#L262) in the view file. The main reason for this change is so that this same structure can be passed to that macro.

I'm going to look to also simplify this once #1070 is in.

PR #1075, opened 2020-12-02
chore: Add services to request middleware

Nimphal:
### What changed

Adds middleware which instantiates the services on the `req` . Updates how the court hearing service is used as an example.

### Why did it change
This is the first step towards the services getting access to the request context. It transitions the court hearing service to the new approach as an example.

### Issue tracking

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/controllers/create/save.js
+++ b/app/move/controllers/create/save.js
@@ -17,7 +17,7 @@ function filterAnswer(currentAssessment, searchKey) {
 
 class SaveController extends CreateBaseController {
   async saveValues(req, res, next) {
-    const courtHearingService = req.services.courtHearing()
+    const courtHearingService = req.services.courtHearing
 
     try {
       const sessionData = req.sessionModel.toJSON()
--- a/app/move/controllers/create/save.js
+++ b/app/move/controllers/create/save.js
@@ -1,7 +1,6 @@
 const { get, omit, capitalize, flatten, values, some } = require('lodash')
 
 const analytics = require('../../../../common/lib/analytics')
-const courtHearingService = require('../../../../common/services/court-hearing')
 const moveService = require('../../../../common/services/move')
 const profileService = require('../../../../common/services/profile')
 const filters = require('../../../../config/nunjucks/filters')
@@ -18,6 +17,8 @@ function filterAnswer(currentAssessment, searchKey) {
 
 class SaveController extends CreateBaseController {
   async saveValues(req, res, next) {
+    const courtHearingService = req.services.courtHearing()
+
     try {
       const sessionData = req.sessionModel.toJSON()
       const assessment = sessionData.assessment
--- a/common/middleware/set-services.js
+++ b/common/middleware/set-services.js
@@ -1,19 +1,11 @@
 const services = require('../services')
 
 const setServices = (req, res, next) => {
-  const servicesCache = {}
-
-  const servicesWithReq = Object.keys(services).reduce((acc, serviceKey) => {
-    if (!servicesCache[serviceKey]) {
-      servicesCache[serviceKey] = services[serviceKey]
-    }
-
-    acc[serviceKey] = () => servicesCache[serviceKey]
+  req.services = Object.keys(services).reduce((acc, serviceKey) => {
+    acc[serviceKey] = services[serviceKey]
 
     return acc
   }, {})
-
-  req.services = servicesWithReq
 
   next()
 }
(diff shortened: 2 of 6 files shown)


common/middleware/set-services.js
```diff
@@ -0,0 +1,13 @@
+const services = require('../services')
+
+const setServices = (req, res, next) => {
```
Nimphal, line 3: The original suggestion for this had a `servicesCache` in here. I left it out only because I was not 100% sure it would be safe when we do end up passing the req through. The [suggested approach](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/783cf9797e93d64757f32a1789c570ff4eff566e/common/middleware/set-services.js) looked like it would potentially cache a service instantiated with some req and then wouldn't update it when a different req came along, which would cause all sorts of data problems.

Very happy to for someone to explain why I'm wrong, it's easily added back in.

common/services/index.js
```diff
@@ -1,5 +1,6 @@
 const allocation = require('./allocation')
 const courtHearing = require('./court-hearing')
+const document = require('./document')
```
Nimphal, line 3: Discovered it was missing while messing about. We'd need to make sure new services are added here if they stand a chance of getting the req.

common/middleware/set-services.js
solidgoldpig: The services cache was within the middleware so only lived for the life cycle of the request so what you feared might happen, wouldn't ;)

solidgoldpig: The previous code allowed for the instantiation of the service if and only when it was actually needed, using the `req` object in the closure within `setServices`. NB. you'll have to address how to ensure that services called by the serviceified services also get the `req` context passed on to them.

Nimphal: Yeah, once all services have been refactored in some fashion to take on the req, it will be possible to pass it on from anywhere that calls the service. Maybe. Looking back, that seems like a bold statement.

common/services/index.js
merlinc: So does this mean that we need to add in all the other ones like `referenceData` and `locationsFreeSpaces`?

app/move/controllers/create/save.js
```diff
@@ -17,6 +16,8 @@ function filterAnswer(currentAssessment, searchKey) {
 
 class SaveController extends CreateBaseController {
   async saveValues(req, res, next) {
+    const courtHearingService = req.services.courtHearing()
```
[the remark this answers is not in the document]
teneightfive: I think it might be a cleaner transition to not require the services to be instantiated in each controller/middleware, rather we do that within the `set-services` middleware once the services support this way of being constructed.

What do you think?

common/services/index.js
Nimphal: Ah, good spot for locationsFreeSpaces! The reference data is already there :)

common/middleware/set-services.js
```diff
@@ -0,0 +1,21 @@
...
+  const servicesCache = {}
```
teneightfive, line 4: I'm not sure I fully understand the benefit of this cache...is there a case where this middleware would be called twice during a request?

What is the cache trying to prevent?

common/middleware/set-services.js
```diff
@@ -0,0 +1,21 @@
...
+
+  const servicesWithReq = Object.keys(services).reduce((acc, serviceKey) => {
+    if (!servicesCache[serviceKey]) {
+      servicesCache[serviceKey] = services[serviceKey]
+    }
+
+    acc[serviceKey] = () => servicesCache[serviceKey]
```
teneightfive, line 11: Could this be `acc[serviceKey] = servicesCache[serviceKey]` so that the services don't require instantiation in controllers/middleware?

app/move/controllers/create/save.js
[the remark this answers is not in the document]
teneightfive: >I'm guessing this is being used as an example of how we can move to this approach?

I should've read the description properly 🤦

common/middleware/set-services.js
solidgoldpig: Making sure that `req.services.foo()` returns the same 'instance' no matter how many times it's called

Nimphal: I think this is resolved with the approach of instantiating the services here so the controllers always access whatever is set here, no service is 'called' in multiple places.

This isn't implemented here yet though! Currently the services are still just objects with nothing passed.

teneightfive (approved):

PR #1076, opened 2020-12-02
Use mutation in API client transform layer

teneightfive:
### What changed

The API client library uses a mutation method when deserializing response
data and storing it into its local cache.

This will now mutate the data passed to the transform methods and
also doesn't need to mess with the library's cache.

### Why did it change

The initial implementation of the local transformers were built to
return a fresh set of data rather than mutate the original object as
it is often a better approach and causes less side-effects.

However, because the original library mutates and the stores nested
resources within a cache layer, when our application would return fresh
objects it would lead to incorrect data being returned when parent
objects were given a transformer.

### DIFF
--- a/common/lib/api-client/README.md
+++ b/common/lib/api-client/README.md
@@ -1,0 +1,70 @@
+# API client
+
+The internal API client is based on top of the [JSON:API devour client](https://github.com/twg/devour).
+
+This module exports a singleton to avoid creating a new devour instance with each use.
+
+## Middelware
+
+This library uses [middleware modules](./middleware) to take care of things like authentication and data
+transformation in a single layer.
+
+## Models
+
+[Models](./models) are defined as objects that are passed to the `devour.define`  method. The take the following form:
+
+```javascript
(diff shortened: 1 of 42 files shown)


locales/en/events.json
```diff
@@ -28,7 +28,7 @@
   "incident_location_vehicle_false": "{{location.title}}",
   "incident_escaped": "$t(events::person) escaped from $t(events::select_incident_location_vehicle) $t(events::incident_reported_classified)",
   "no_events": "No events",
-  "person": "{{person.fullname, uppercase}}",
+  "person": "{{person.fullname}}",
```
solidgoldpig, line 31: Shouldn't this be `{{person._fulname}}` ??

common/lib/api-client/README.md
```diff
@@ -0,0 +1,70 @@
+# API client
```
solidgoldpig, line 1: 🥇

solidgoldpig (approved):

PR #1077, opened 2020-12-02
P4 2462 add breadcrumb

merlinc:
## Proposed changes

Add a breadcrumb to the population screens so the user has a better sense of where they are.

Note: The deeplinking can look a little weird due to deep linking into the edit state, and the lack of the Population Location page which is scheduled for future development.
  
### What changed

This commit is based upon a number of different pieces of prior art:

- Previous attempt in Book a Secure move
  - https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/compare/person-escort-record/manifest
- Previous usage in Data Hub -
  - https://github.com/uktrade/data-hub-frontend/blob/master/src/middleware/breadcrumbs.js
  - https://github.com/uktrade/data-hub-frontend/blob/master/src/middleware/__test__/breadcrumbs.test.js

### Issue tracking

- [P4-XXXX]()

## Screenshots

<kbd>Screenshot_2020-12-02 Population Details - create_population(6)</kbd>

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/population/controllers/edit.js
```diff
@@ -50,6 +51,14 @@ class DetailsController extends FormWizardController {
     }
   }
 
+  setBreadcrumbs(req, res, next) {
+    res.breadcrumb({
```
solidgoldpig, line 55: Why put it on `res` rather than `req`? (I believe the "pattern" we're following is stash everything on `req` and then update `res.locals` prior to rendering)

merlinc: @solidgoldpig I've updated the ticket with a little more context - basically this is slightly tweaked reusing of an unmerged PR of @teneightfive 's

It sounds like our development patterns have changed in that time, so refactoring this to match the other code is a good idea

teneightfive: The reason for this being on `res` instead of `req` is because it is concerned with the response rather than the original request.

The major benefit is that it can be chained with other `res` methods like `res.render`.

For example:

```
res
  .breadcrumb(...)
  .render(...)

```

teneightfive: It's also a different use case to the previous things we have attached to the request object.

app/population/controllers/edit.js
```diff
@@ -50,6 +52,19 @@ class DetailsController extends FormWizardController {
     }
   }
 
+  setBreadcrumbs(req, res, next) {
+    const breadcrumbText = req.population
```
teneightfive, line 56: This looks identical to the condition in `setPageTitle`. Could this be extracted?

teneightfive:
Something about the second item not being a link just doesn't feel right somehow...

app/population/controllers/edit.js
merlinc: It's only the 2nd instance of duplication - I find prematurely abstracting to be an anti-pattern and leads to more confusing code. Once the code is repeated 3 or 4 times, I think there's an argument for reuse.

app/population/controllers/edit.js
merlinc: I hadn't seen this followup discussion, and in the meantime I'd switched most usages to run off `req` as per my discussion with Alex. It's only a single commit, so can be easily dropped.

merlinc:
> Something about the second item not being a link just doesn't feel right somehow...

Yeah, the 2nd item not being a link is because that should go to a weekly location view that doesn't exist yet. Even if I add that, the deep linking direct into `create` when `editing` means that the 3rd link, date view,  is currently bypassed. We did do some work on what the `create` screen should look like so that could be linked up as well.

Some form of breadcrumbing is needed to give the user context as to how they navigate around these screens, so I think this is an improvement.

I'm working on the weekly location view at the moment, so I'm going to hold off on merging this until that page exists, as I think it will make the interaction using this breadcrumb clearer.

app/population/controllers/edit.js
teneightfive: I think it would be good to keep the original implementation on `res`. It serves a different purpose to how we've attached things to the `req`.

The breadcrumb feels more part of the response that the request too and I think the benefit of being able to chain `breadcrumb()` and `render()` together is a nice pattern.

app/population/middleware/set-breadcrumb.js
```diff
@@ -0,0 +1,26 @@
+const { format, parseISO } = require('date-fns')
+
+const { DATE_FORMATS } = require('../../../config/index')
+
+function setBreadcrumb(req, res, next) {
+  const { date, locationName } = req
+
+  res.breadcrumb({
```
teneightfive: Because `breadcrumb` returns the `res` object these can be chained to:

```javascript
res
  .breadcrumb({
    text: 'Population Overview',
    href: `/population/week/${date}`,
  })
  .breadcrumb({
    text: locationName,
    href: '',
  })
  .breadcrumb({
    text: format(parseISO(req.date), DATE_FORMATS.WITH_DAY),
    href: req.population ? req.baseUrl : '',
  })
```

common/middleware/breadcrumbs.js
```diff
@@ -81,12 +81,12 @@ function setHome(options = {}) {
...
 
   return function (req, res, next) {
-    const homeBreadcrumb = find(res.breadcrumb(), breadcrumb => {
+    const homeBreadcrumb = find(req.breadcrumb(), breadcrumb => {
       return breadcrumb._home
     })
 
     if (!homeBreadcrumb) {
-      res.breadcrumb({
+      req.breadcrumb({
```
teneightfive, line 89: I think if we're deciding we do want this approach and to give up being able to chain `breadcrumb` and `render` together then we could in fact switch to just using this [express-breadcrumbs](https://www.npmjs.com/package/express-breadcrumbs) package.

I now have a vague memory of us writing this version for the uktrade app so that we can could `res.breadcrumb().render()`.

Without it, there's probably no point supporting that code ourselves.

merlinc: @solidgoldpig 's original point was that the pattern we're using is put things on `req` and then fill `res.locals` just before rendering.

In the `Basm` app do we have any other examples of this `res` chaining?

merlinc: re: [express-breadcrumbs](https://www.npmjs.com/package/express-breadcrumbs) - I really dislike reinventing the wheel if someone else already has, however that code hasn't been updated in 7 years, and talks about supporting Jade (not even Pug!), and has no tests, so I'm a little wary of adopting it!

teneightfive: Yes, I think the lack of tests may have been why we wrote it ourselves in the UK Trade app.

>In the Basm app do we have any other examples of this res chaining?

I see this as a different use-case than the other instances though. The other instances are objects that we set and use later in the request lifecycle. "Request" being the key point there. 

The breadcrumbs I felt were more attached to the response, much like `res.render()` render a template. Probably a similar reason why they didn't go with `req.render()` in express.

I'll leave the call up to you on this one though.

app/population/middleware/set-breadcrumb.js
merlinc: Done!

common/middleware/breadcrumbs.js
merlinc: I've removed the commit that switched from `res` to `req`

app/population/middleware/set-breadcrumb.js
```diff
@@ -0,0 +1,40 @@
...
+const {
+  formatDateWithRelativeDay,
+} = require('../../../config/nunjucks/filters')
+
+function setBreadcrumb(req, res, next) {
+  const { date, locationName } = req
+
+  let dateSuffix = ''
+
+  if (isCurrentWeek(date)) {
```
merlinc, line 15: For help with navigation, I want the breadcrumb to show Last / This / Next week - but we don't have anything that does this via the i18n files - the currently implementation is inside the `nunjucks/filters` - and hardcodes the string value in place. Also, I'd like to use some brackets.

Any suggestions on a better place for this part of the code to live are welcome.

common/helpers/date.test.js
```diff
@@ -3,6 +3,9 @@ const {
   getCurrentDayAsRange,
   getRelativeDate,
   getDateRange,
+  isCurrentWeek,
```
merlinc, line 6: There is a `_isCurrentWeek` function in `nunjucks/filters` but it is "private" and untested, and I also need PreviousWeek and NextWeek

common/helpers/date.js
```diff
@@ -7,23 +9,21 @@ const {
   endOfWeek,
   isValid: isValidDate,
   isDate,
+  isSameWeek,
 } = require('date-fns')
 
 const { DATE_FORMATS } = require('../../config/index')
+const DEFAULT_DATE_OPTIONS = { weekStartsOn: DATE_FORMATS.WEEK_STARTS_ON }
```
merlinc, line 16: All the functions were sharing the same date options, it made sense to pull them out here

common/helpers/date.js
```diff
@@ -72,4 +68,28 @@ module.exports = {
       format(dateTo, DATE_FORMATS.URL_PARAM),
     ]
   },
+
+  isCurrentWeek(date) {
```
merlinc, line 72: I'm not sure if it is clearer to have slightly duplicated code, or have to have a function that adds `-1`, `0` and `1` to the current week for last, current and next week calculations.

app/population/controllers/edit.js
merlinc: I've dropped the commit that used `req` and left it as per the original implementation.

app/population/middleware/set-breadcrumb.js
teneightfive: It looks like that filter [uses the locales]() for the current week?

I think the implementation for week navigation was to only specifically use `This week` and not next/previous.

Maybe this could use [`formatDateRangeWithRelativeWeek`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/955759046b2ed6ea625949833d74d6e8497a9254/config/nunjucks/filters.js#L228)?

common/helpers/date.js
teneightfive: I think date-fns already has this method: [`isThisWeek`](https://date-fns.org/v2.16.1/docs/isThisWeek)

app/population/middleware/set-breadcrumb.js
merlinc: re: filters & locales - my mistake, I was getting `current_week` confused with `Today` - https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/config/nunjucks/filters.js#L185-L195

So yeah, the existing week navigation is only to specifically use `This week` and not `Next/Previous`. I discovered through my working with the dashboard - and after discussion with Andy - that there was a use for seeing this information as it keeps the user grounded while navigating around the populations.

That being said, that's not a good reason to hold this up, so I'll separate that part out, and see if I can get back to it later.

common/helpers/date.js
merlinc: Related to the comments above, yes it does have `isThisWeek`, but it doesn't have `isPreviousWeek` or `isNextWeek` - thus my question about how best to code this up so it's easier to understand.

As per above, I'll just remove it for the time being.

app/population/middleware/set-breadcrumb.js
merlinc: Also, `formatDateRangeWithRelativeWeek` does not work on single dates very well. And giving it two dates the same, results in `30 to 30 Nov 2020` - which is also what's not wanted.

`formatDateWithRelativeDay` adds in the year, which is also not what's wanted.

And  as an aside, our version of `_isCurrentWeek` take a single date and turns it into an array of dates for comparison, which seems odd.

I can't help but feel that the filters / date utils needs an overhaul, but I definitely don't have time for that.

I'm going to introduce some local date formatting code in this breadcrumb for now, and see about coming back and cleaning this up later.

teneightfive (approved):

PR #1080, opened 2020-12-03
build: Add tasks folder to Dockerfile

teneightfive:
The tasks need to be called in some environments to do actions like
clearing cache.

To have access to this we need this copied across in the container.


Nimphal (approved):
Nice!

merlinc (approved):

PR #1096, opened 2020-12-07
Show important event flags on profile and dashboard [P4 2506]

solidgoldpig:
**SHOULD NOT BE MERGED UNTIL CONFIRMED THAT BE CHANGES FOR EVENT CLASSIFICATION IN PLACE**

### What changed

- Add important events as tags to 
  - move profile/timeline header
  - dashboard cards

- Event tags link to corresponding event in timeline view

- Events of the same type have a count appended to them

- Event tags are visually distinguished from other tags by having a black pling sign on a yellow background at the start - provided by the new flag attribute for the `tag` component which inserts some hidden assistive text in the dame way that the `govukWarning` element does.

#### Examples

- Timeline with important event flags
  https://booksecuremove-pr-1007.herokuapp.com/move/AAAAAAAA-AAAA-AAAA-AAAA-AAAAAAAAAAAA/timeline
- Move profile/view with important event flags
  https://booksecuremove-pr-1007.herokuapp.com/move/AAAAAAAA-AAAA-AAAA-AAAA-AAAAAAAAAAAA
- Moves list with important event flags
  https://booksecuremove-pr-1007.herokuapp.com/moves/day/2020-12-04/0ee995d1-9390-4e85-9f5a-c6a436716234/outgoing

(NB. must have Bishopsgate as chosen location to be able to view the dashboard list)

### Why did it change

MVP design

### Issue tracking

- [P4-2506 - Show Red Flag & Medical events as flags on Dashboard & Profile page](https://dsdmoj.atlassian.net/browse/P4-2506)

## Screenshots

| After | Before |
| ------ | ----- |
| <kbd>Move_details_for_FANG__SIMON_–_Book_a_secure_move</kbd> | <kbd>Move_details_for_FANG__SIMON_–_Book_a_secure_move</kbd> |
| <kbd>Move_details_for_FANG__SIMON_–_Book_a_secure_move</kbd> | <kbd>Move_details_for_FANG__SIMON_–_Book_a_secure_move</kbd> |
| <kbd>Outgoing_moves_for_Monday_7_Dec_2020__Today_</kbd> | <kbd>Outgoing_moves_for_Monday_7_Dec_2020__Today_</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/presenters/move-to-card-component.js
```diff
@@ -30,6 +32,9 @@ function moveToCardComponent({
       href,
     })
 
+    // TODO: only include this is showTags is true???
```
solidgoldpig, line 35: @teneightfive I'll double-check with Andy and Jason, but I'm guessing that if we're not showing the other tags, then the same applies for the important events too... Or am I overlooking all many of ph34rsome dashboard combos?

common/components/flag/macro.njk
```diff
@@ -0,0 +1,3 @@
+{% macro appFlag(params) %}
```
teneightfive, line 1: What's the difference between this component and the "tag" component? It feels like it would be better to adapt that component to support this variation rather than introduce another component with a similar name.

merlinc: I went to the components library to see what the actual difference was, and it looks like there's a rendering error - https://booksecuremove-pr-1096.herokuapp.com/components/flag - I dont know if this is in the component, or the library route

common/components/card/template.njk
```diff
@@ -63,6 +63,16 @@
       </ul>
     {% endif %}
 
+    {% if params.importantEventsTagList.length %}
```
teneightfive, line 66: Can we not re-use the existing functionality that this card already has to display tags.

These are being rendered using the same markup so I think it would be better to re-use them. There is already a way to visually distinguish them so we can do without this.

common/presenters/move-to-card-component.js
teneightfive: Yes, I'd say that's correct. We use this option to hide tags from display's like the person lookup where we don't need to render tags.

common/components/flag/macro.njk
teneightfive: Good spot! It looks like there's some form of position relative missing?

teneightfive: In my mind if feels more like a class modifier that can be applied to `app-tag` to produce the different visual style rather than a separate component.

teneightfive: Or maybe even an attribute on the component rather than a class. It feels very similar to how the GOV.UK Design System handles the [Start button variation](https://design-system.service.gov.uk/components/button/#start-buttons) with the `isStartButton` option.

common/services/move.js
```diff
@@ -81,11 +81,11 @@ const moveService = {
   format(data) {
     const booleansAndNulls = ['move_agreed']
     const relationships = [
-      'to_location',
       'from_location',
       'person',
       'prison_transfer_reason',
       'supplier',
+      'to_location',
```
merlinc, line 88: Sorting! :heart:

common/services/move.test.js
```diff
@@ -639,12 +639,13 @@ describe('Move Service', function () {
         expect(moveService.getAll).to.be.calledOnceWithExactly({
           isAggregation: false,
           include: [
+            'from_location',
```
merlinc, line 642: Sorting! :heart: :brain:

common/components/flag/macro.njk
solidgoldpig: There was quite a bit of difference. It was a flag to put some content inside the tag, not the tag itself. Anyhow, the `tag` component now has a `flag` attribute and the `flag` component is gone.

common/services/move.js
solidgoldpig: I knew Dom was hacking these too - just made sense for us to both sort them so that nothing got overlooked in the merge

common/assets/scss/components/_all.scss
```diff
@@ -1,22 +1,22 @@
-@import "autocomplete";
-@import "focus-table";
-@import "icons";
-@import "sortable-table";
-@import "stack-trace";
-@import "sticky-sidebar";
+@import 'autocomplete';
```
teneightfive, line 1: Any reason for these quote changes? I feel like every other `scss` file in our app uses double quotes so this one would now stick out.

common/components/tag/_tag.scss
```diff
@@ -10,8 +12,8 @@
   outline: 2px solid transparent;
   outline-offset: -2px;
 
-  color: govuk-colour("white");
-  background-color: govuk-colour("blue");
+  color: govuk-colour('white');
```
teneightfive, line 15: I'm guessing this might be some autosave feature on your IDE with the quotes?

It would be good if we keep them as is and consistent with the rest of the files for now.

common/components/timeline/_timeline.scss
```diff
@@ -2,79 +2,82 @@
    #TIMELINE
    ========================================================================== */
 
-   .app-timeline {
-    margin-bottom: govuk-spacing(4);
-  }
+.app-timeline {
```
teneightfive, line 5: Are there any changes to this file? It's a little hard to see with the whitespace noise.

common/lib/api-client/transformers/move.transformer.js
```diff
@@ -9,3 +9,43 @@ module.exports = function moveTransformer(data) {
 
   data._is_youth_move = isYouthMove
 }
+
+const addImportantEvents = data => {
+  // timeline_events and important_events are mutually exclusive at the BE
+  // copy any event with a non-default classification to .important_events
+  // maintaining the same order
+  if (data.timeline_events && data.timeline_events.length) {
```
teneightfive, line 17: Could this be a little safer and check that important events doesn't exist before creating/overwriting it?

common/lib/api-client/transformers/move.transformer.js
```diff
@@ -9,3 +9,43 @@ module.exports = function moveTransformer(data) {
...
+    })
+    const multiEventTypes = Object.keys(eventTypes).filter(
+      key => eventTypes[key] > 1
+    )
+
+    multiEventTypes.forEach(eventType => {
+      data.important_events
+        .filter(event => event.event_type === eventType)
+        .forEach((event, index) => {
+          event._eventCount = index + 1
```
teneightfive, line 41: What about `event._index` instead of `event._eventCount`? 

(Taking guidance from the naming of this _thing_ in loops).

common/presenters/event-to-tag-component.js
```diff
@@ -0,0 +1,23 @@
+const eventHelpers = require('../helpers/events/event')
+
+const eventToTagComponent = (event, moveId) => {
+  const { id } = event
+
```
teneightfive, line 5: Any reason for the line between all these declarations? It's not something we've done in many other places.

common/helpers/events/event.js
```diff
@@ -0,0 +1,85 @@
...
+  const description = i18n
+    .t(`${eventTypePrefix}.description`, details)
+    .replace(/^(\s*<br>)+/, '')
+  // Some strings that get added together are prefixed with a <br>.
+  // This enables them to be used as the first item or where a previous string doesn't get output
+  // and not insert additional space without having to make the <br> conditional
+  return description
+}
+
+module.exports = {
```
teneightfive, line 78: Can we either structure this as a folder with each helper as a separate file (if there are too many) or have them stored under `helpers/events.js`.

This will match the pattern we have used elsewhere and it is to aid being able to start with a single file (`events.js`) and then break that out into separate files with an index loader as the app scales.

common/lib/api-client/transformers/move.transformer.js
solidgoldpig: Switching the check `!data.important_events? && data.timeline_events?.length`

However
a) They are mutually exclusive.
b) Even if both properties were returned, the routine would just filter the timeline_events to create exactly the same important_events. Think about it...

common/helpers/events/event.js
solidgoldpig: `helpers/events/index.js` and individual files created

common/presenters/event-to-tag-component.js
solidgoldpig: eslint rules pleeze 🙏

common/components/tag/_tag.scss
solidgoldpig: I'm using VSCode with the ESlint and Prettier ESLint extensions. Pretty sure it didn't reformat these files like that before. - maybe the rules have changed a bit? Anyhow, the singles are doubles again

common/components/timeline/_timeline.scss
solidgoldpig: Yep. Should be clearer now.

common/assets/scss/components/_all.scss
solidgoldpig: As noted in the next comment...  (tldr; no, not really)

common/components/timeline/_timeline.scss
solidgoldpig: Really should be clearer now ;)

common/lib/api-client/transformers/move.transformer.test.js
```diff
@@ -1,3 +1,5 @@
+const { expect } = require('chai')
```
teneightfive, line 1: Isn't this already globally available?

teneightfive (approved):

common/lib/api-client/transformers/move.transformer.test.js
solidgoldpig: GAH!  VSC inserting it automagickally again

PR #1097, opened 2020-12-07
Add initial development tools

teneightfive:
### What changed

This adds a sup-app that is only available when in development or enabled using an environment variable.

### Why did it change

The first tool within this app allows the permissions to be set based on the roles that exist within the frontend application.

## Screenshots

<kbd>localhost_3000_tools_permissions</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


Nimphal (approved):
Nice!

solidgoldpig (approved):

config/index.js
```diff
@@ -104,6 +104,8 @@ module.exports = {
   },
   ENABLE_COMPONENTS_LIBRARY:
     IS_DEV || /true/i.test(process.env.ENABLE_COMPONENTS_LIBRARY),
+  ENABLE_DEVELOPMENT_TOOLS:
```
solidgoldpig, line 107: Is there any way we could ensure that this could never be set on production. (ie. currently, a naughty person with access could add the env var to the k8s config and simply apply it)

app/_development-tools/index.js
```diff
@@ -0,0 +1,23 @@
...
+const { updatePermissions, renderPermissions } = require('./controllers')
+
+// Define routes
+router.route(routes.permissions).get(renderPermissions).post(updatePermissions)
+
+// Export
+module.exports = {
+  router,
+  routes,
+  mountpath: '/tools',
```
merlinc, line 20: I think this is a decently named path for these sorts of things. :+1:

config/index.js
merlinc: Any environment variable is susceptible to that sort of manipulation.

Best option I can think of involves denying this path in the ingress configuration.

Only other two options I can think off are either not including this module in our production builds, which is not a great solution, because then what is staging actually representing. Alternatively, if this was a separate microservice stitched together at the Ingress layer we wouldn't need to worry about this - but that's effectively the same as denying it at that later anyway.

app/_development-tools/controllers.js
```diff
@@ -0,0 +1,27 @@
+const { uniq } = require('lodash')
```
merlinc, line 1: `_development-tools` does not thrill me, however it does make it obvious that its a "non-functional requirement" module.

Only alternative to this I can think of is ripping this out as a separate `npm module` and including it that way - but I think it's too early for us to know what that would look like just yet.

config/nunjucks/globals.js
```diff
@@ -4,10 +4,15 @@ const {
   ANALYTICS,
   ASSETS_HOST,
   ENABLE_COMPONENTS_LIBRARY,
+  ENABLE_DEVELOPMENT_TOOLS,
   FEEDBACK_URL,
   PERSON_ESCORT_RECORD_FEEDBACK_URL,
   SUPPORT_EMAIL,
 } = require('../')
+const {
+  mountpath: toolsUrl,
```
merlinc, line 13: This is a path, not a full url

app/_development-tools/controllers.js
```diff
@@ -0,0 +1,27 @@
...
+const { permissionsByRole } = require('../../common/lib/permissions')
+
+function renderPermissions(req, res) {
+  res.render('_development-tools/views/permissions', {
+    activeRoles: req.session.activeRoles || [],
+    roles: permissionsByRole,
+  })
+}
+
+function updatePermissions(req, res) {
```
merlinc, line 12: Future enhancement idea - supporting updating by both form and JSON.

We used this on session injection to give humans an easy form, but then used the JSON updating as a more reliable approach for automated browser tests

docs/environment-variables.md
```diff
@@ -39,7 +39,8 @@
...
 | LOG_LEVEL                           | Level of logs to output                                                                                                                                                                                                            | production: `error`, development: `debug`                                           |
-| ENABLE_COMPONENTS_LIBRARY           | Whether to enable the component library and allowed it to be browsed on this environment                                                                                                                                           | false                                                          |
+| ENABLE_COMPONENTS_LIBRARY           | Whether to enable the component library and allow it to be browsed on this environment                                                                                                                                           | false                                                          |
```
merlinc, line 42: :books: :heart:

teneightfive:
This could help address #804

config/index.js
teneightfive: I did also toy with this but the reason I created the env was to actually potentially allow it on some "production-like" environments like Heroku or Staging.

I think some of the tools we may add to this will be useful in those environments, like pre-filling a PER. 

However, you're right that this does potentially open up security concerns if things like permissions can be manipulated.

Maybe it is safer at this stage that it can **only** be done in development environments and we remove support for the environment variable altogether.

@merlinc's suggestion is also something we thought about with the original development only middleware. We could potentially add a npm task that cleans up after a deploy to remove such files?

app/_development-tools/controllers.js
teneightfive: Yep, it didn't thrill me either...I also toyed with `./dev`, `./development-only`, `./_tools`....the prefix sort of helped set it out above the rest of the "real" apps. But maybe that's not necessary as we don't do that with the component app which is only available in dev or on certain environments.

What about simple `./tools`?

app/_development-tools/controllers.js
teneightfive: Yep, that's a really nice idea! Nothing preventing us extending this to support a JSON body, or in fact just a `POST` within the e2e tests?

PR #1098, opened 2020-12-07
[IM-28] Protect assign journey with extra checks

teneightfive:
### What changed

This change includes two fixes to prevent a circular structure error. It:

- prevents the assign journey being completed when a move already has a profile
- handles the deciding of the next move to assign using a redirect route instead of on page load of the allocation

### Why did it change

See Jira ticket for more information.

### Issue tracking

- [IM-28](https://dsdmoj.atlassian.net/browse/IM-28)

Fixes BOOK-A-SECURE-MOVE-FRONTEND-T
https://sentry.io/organizations/ministryofjustice/issues/1920836995

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/allocation/controllers/assign-to-allocation.js
+++ b/app/allocation/controllers/assign-to-allocation.js
@@ -1,0 +1,12 @@
+function assignToAllocation(req, res) {
+  const { allocation } = req
+  const movesWithoutProfile = allocation.moves.filter(move => !move.profile)
+
+  if (movesWithoutProfile.length) {
+    return res.redirect(`/move/${movesWithoutProfile[0].id}/assign`)
+  }
+
+  res.redirect(`/allocation/${allocation.id}`)
+}
+
+module.exports = assignToAllocation
--- a/app/move/controllers/assign/base.test.js
+++ b/app/move/controllers/assign/base.test.js
@@ -10,8 +10,8 @@ const controller = new UpdateBaseController({ route: '/' })
 describe('Assign controllers', function () {
   describe('Assign base controller', function () {
     describe('#middlewareChecks()', function () {
-      it('should inherit middlewareChecks from CreateBaseController', function () {
-        expect(controller.middlewareChecks).to.exist.and.equal(
+      it('should not inherit middlewareChecks from CreateBaseController', function () {
+        expect(controller.middlewareChecks).to.exist.and.not.equal(
           BaseProto.middlewareChecks
         )
       })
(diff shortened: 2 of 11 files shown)


app/move/controllers/assign/base.test.js
```diff
@@ -111,6 +142,58 @@ describe('Assign controllers', function () {
...
+        controller.checkNoProfileExists(req, res, nextSpy)
+      })
+
+      it('should redirect back to allocation assign route', function () {
+        expect(res.redirect).to.have.been.calledOnceWithExactly(
+          '/allocation/__allocationId__/assign'
+        )
+      })
+
+      it('should call next', function () {
```
Nimphal, line 177: Minor, but I think this should be 'should not call next'?

app/allocation/controllers/assign-to-allocation.js
```diff
@@ -0,0 +1,12 @@
+function assignToAllocation(req, res) {
+  const { allocation } = req
+  const movesWithoutProfile = allocation.moves.filter(move => !move.profile)
+
+  if (movesWithoutProfile.length) {
+    return res.redirect(`/move/${movesWithoutProfile[0].id}/assign`)
+  }
+
+  res.redirect(`/allocation/${allocation.id}`)
```
Nimphal, line 9: Does the page handle no more moves left? I'm sure it does, I just don't think I've seen that scenario :D

Nimphal (approved):

app/allocation/controllers/assign-to-allocation.js
teneightfive: Yep, that's what this catch is for. 

So when all moves are filled, this will catch someone clicking the "Add person" button in an old tab. It redirects back to the allocation (which will now show as filled) rather than `/move/undefined/assign` where they'd get a `404`.

solidgoldpig:
When I open multiple tabs to add to the allocation (using this branch, not running with browsersync), I get 'Journey expired' - the session/journey is getting shared between the tabs.

teneightfive:
> When I open multiple tabs to add to the allocation (using this branch, not running with browsersync), I get 'Journey expired' - the session/journey is getting shared between the tabs.

As discussed on over a quick call, that will actually still be an issue when using a single button to assign the next move. 

It was agreed that this fix will improve the current situation and remove the error where a user can end up assigning a person to a move that already has a profile, which could also cause cases of the move being overridden before being moved.

The ideal solution would be to pick the move at the end but that would require a lot more extra work at this stage so we'll merge this change and look to add some monitoring around multiple tab use. If this is true user behaviour we may have to think about the current design to accommodate users opening moves in different tabs.

PR #1102, opened 2020-12-08
Use correct content for framework assessment

teneightfive:
### What changed

This set of changes moves some of the Person Escort Record specific content keys to the generic namespace so that they can be shared between the two.

### Why did it change

Some of the Youth Risk Assessment pages are currently using the Person Escort Record content when they need to be using
youth risk assessment specific content.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/templates/framework-section.njk
+++ b/common/templates/framework-section.njk
@@ -4,7 +4,7 @@
   {{ t("person-escort-record::section_overview", {
     section: sectionTitle
   }) }}
-  – {{ t("person-escort-record::page_title") }}
+  – {{ t("assessment::page_title", { context: i18nContext }) }}
   – {{ SERVICE_NAME }}
 {% endblock %}
 
--- a/common/templates/framework-section.njk
+++ b/common/templates/framework-section.njk
@@ -49,7 +49,7 @@
             {{ step.pageTitle }}
           </h2>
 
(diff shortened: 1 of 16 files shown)


solidgoldpig (approved):

common/templates/framework-section.njk
```diff
@@ -49,7 +49,7 @@
             {{ step.pageTitle }}
           </h2>
 
-          {% if canAccess('person_escort_record:update') and isEditable %}
```
merlinc, line 52: The rest of this PR is effectively a simple rename - why is the `canAccess` being removed here?

teneightfive: Is part of [this commit](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1102/commits/ec05faf2418613d6fa113ddbffaf25a1061ae49a) which makes this check dynamic based on the assessment type rather than just check for person escort record permissions.

merlinc: You've rebased, so that commit has vanished forever!

But that makes sense :)

merlinc (approved):

PR #1108, opened 2020-12-08
refactor: Update controllers to use PER service from req

Nimphal:
### What changed
Make use of the new setServices middleware for the PER service. Part of the work to enable request context in services.

### Why did it change
All services need to be moved to this pattern in preparation for being refactored to take the request context.

### Issue tracking

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/person-escort-record/index.js
+++ b/app/person-escort-record/index.js
@@ -13,7 +13,6 @@ const {
   setRecord,
 } = require('../../common/middleware/framework')
 const { protectRoute } = require('../../common/middleware/permissions')
-const personEscortRecordService = require('../../common/services/person-escort-record')
 
 const confirmApp = require('./app/confirm')
 const newApp = require('./app/new')
@@ -26,7 +25,7 @@ router.use(newApp.mountpath, newApp.router)
 
 // Define shared middleware
 router.use(protectRoute('person_escort_record:view'))
-router.use(setRecord('personEscortRecord', personEscortRecordService.getById))
+router.use(setRecord('personEscortRecord', 'personEscortRecord'))
 router.use(setAssessment('personEscortRecord'))
 
 // Define sub-apps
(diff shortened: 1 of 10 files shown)


app/person-escort-record/index.js
```diff
@@ -26,7 +25,7 @@ router.use(newApp.mountpath, newApp.router)
 
 // Define shared middleware
 router.use(protectRoute('person_escort_record:view'))
-router.use(setRecord('personEscortRecord', personEscortRecordService.getById))
+router.use(setRecord('personEscortRecord', 'personEscortRecord'))
```
Nimphal, line 28: This needed a bit of rethinking, happy for someone to point out a better approach.

teneightfive: Gah. Because this method doesn't expose the request object...

I think this seems like a good approach on the surface. I suppose the only thing we're relying on is that both services (or any others we may create) always have a `getById` method...

Nimphal: Yes, that is the downside, though if it doesn't, things will error out fairly quickly which prompt some resolution either to this middleware or to the service.

solidgoldpig: If we can't rely on the method being called `getById` (though I think that is a worthwhile convention to enforce), something like

`setRecord('personEscortRecord', req.services.personEscortRecord, 'getById')`
or
`setRecord('personEscortRecord', 'personEscortRecord', 'getById')` ?

common/middleware/framework/set-record.js
```diff
@@ -1,4 +1,4 @@
-function setRecord(key, getMethod) {
+function setRecord(key, serviceName) {
```
solidgoldpig, line 1: `function setRecord(key, serviceName, methodName) {`

as above

app/person-escort-record/index.js
Nimphal: we can definitely pass it - I reckon we can make that change if it becomes necessary though.

solidgoldpig (approved):

PR #1116, opened 2020-12-09
Catch API errors that are generated in the end-to-end tests

teneightfive:
### What changed

Catch API errors that occur when running end-to-end tests and log them to Sentry.

### Why did it change

Currently in some scenarios the API may return a `422` when we
create a dynamic fixture.

It's hard to debug to see what extactly went wrong so this change
adds the ability to catch those errors and log them to Sentry so that
they can be investigated.

## Screenshots

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

### DIFF
--- a/test/e2e/_helpers.js
+++ b/test/e2e/_helpers.js
@@ -2,6 +2,7 @@ import { unlinkSync } from 'fs'
 import { homedir } from 'os'
 import { join } from 'path'
 
+import * as Sentry from '@sentry/node'
 import { format } from 'date-fns'
 import faker from 'faker'
 import glob from 'glob'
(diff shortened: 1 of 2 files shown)


test/e2e/_helpers.js
```diff
@@ -14,8 +15,30 @@ import personService from '../../common/services/person'
...
+}
+
+function errorHandler(body) {
+  return err => {
+    Sentry.withScope(scope => {
+      err.errors.forEach((apiErr, idx) => {
+        scope.setContext(`error_${idx}`, apiErr)
+      })
+      scope.setContext('body', body)
+      Sentry.captureException(err)
```
teneightfive, line 35: I'm not sure what level we want to capture these as: `error`, `warning`, `info`. The default is `error`.

They will be filtered out already by the `environment` value and won't create noise in the main alerts channel so maybe `error` is okay.

test/e2e/_helpers.js
```diff
@@ -14,8 +15,30 @@ import personService from '../../common/services/person'
...
+  })
+}
+
+function errorHandler(body) {
+  return err => {
+    Sentry.withScope(scope => {
+      err.errors.forEach((apiErr, idx) => {
+        scope.setContext(`error_${idx}`, apiErr)
+      })
+      scope.setContext('body', body)
```
teneightfive, line 34: This does send the body data to Sentry but in our cases it is fake data generated using faker so won't contain anything sensitive/PII.

test/e2e/_helpers.js
solidgoldpig: Sounds reasonable

solidgoldpig (approved):

test/e2e/_helpers.js
merlinc: Yeah, let's see how noisy `error` is, and work from there.

test/e2e/move.update.sch.test.js
```diff
@@ -47,10 +46,6 @@ test.meta('hasDocument', 'true')(
   }
 )
 
-test('User should be able to cancel move', async () => {
```
merlinc, line 50: These were added in by me, as this functionality was specified in the original story - but if they're unduly slowing down the tests, they're probably one of the safest tests to remoe.

merlinc (approved):

PR #1124, opened 2020-12-10
refactor: Update setServices to check if service is function and pass req if so

Nimphal:
### What changed
The serServices middleware checks if the service is a function, and if so, passes the api client which has been initiated on the req. The end result for the controllers is the same.

Court hearing service updated as an example of what would need to change for each service.

### Why did it change
This work is part of sending the request context to the backend.

### Issue tracking

- [P4-1776]()

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/middleware/set-services.js
+++ b/common/middleware/set-services.js
@@ -3,7 +3,7 @@ const services = require('../services')
 const setServices = (req, res, next) => {
   req.services = Object.keys(services).reduce((acc, serviceKey) => {
     if (services[serviceKey] instanceof Function) {
-      acc[serviceKey] = services[serviceKey](req)
+      acc[serviceKey] = services[serviceKey](req.apiClient)
     } else {
       acc[serviceKey] = services[serviceKey]
     }
--- a/common/middleware/set-services.js
+++ b/common/middleware/set-services.js
@@ -3,7 +3,7 @@ const services = require('../services')
 const setServices = (req, res, next) => {
   req.services = Object.keys(services).reduce((acc, serviceKey) => {
     if (services[serviceKey] instanceof Function) {
-      acc[serviceKey] = new services[serviceKey](req.apiClient)
+      acc[serviceKey] = new services[serviceKey](req)
     } else {
       acc[serviceKey] = services[serviceKey]
     }
--- a/common/middleware/set-services.js
+++ b/common/middleware/set-services.js
@@ -3,7 +3,7 @@ const services = require('../services')
 const setServices = (req, res, next) => {
   req.services = Object.keys(services).reduce((acc, serviceKey) => {
     if (services[serviceKey] instanceof Function) {
-      acc[serviceKey] = services[serviceKey](req.apiClient)
+      acc[serviceKey] = new services[serviceKey](req.apiClient)
     } else {
       acc[serviceKey] = services[serviceKey]
     }
--- a/common/middleware/set-services.js
+++ b/common/middleware/set-services.js
@@ -2,7 +2,11 @@ const services = require('../services')
 
 const setServices = (req, res, next) => {
   req.services = Object.keys(services).reduce((acc, serviceKey) => {
-    acc[serviceKey] = services[serviceKey]
+    if (services[serviceKey] instanceof Function) {
+      acc[serviceKey] = services[serviceKey](req)
+    } else {
+      acc[serviceKey] = services[serviceKey]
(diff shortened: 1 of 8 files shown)


common/services/court-hearing.js
```diff
@@ -13,19 +13,24 @@ const courtHearingService = {
...
+  const create = (data, disableSaveToNomis = false) => {
     const params = {}
 
     if (disableSaveToNomis) {
       params.do_not_save_to_nomis = true
     }
 
     return apiClient
-      .create('court_hearing', courtHearingService.format(data), params)
+      .create('court_hearing', format(data), params)
```
solidgoldpig, line 26: Leaving aside the fact that we don't appear to check that we call `format` on the data before passing it to the api client here (for other services we do have such tests), not using the form `courtHearingService.format` (which I presume is because we're now returning an instance) is going to make this hard (if not impossible) to test since the `format` is within the closure.

Maybe this should now be a good old class that you simply pass the req object to? And then we can use `this.format` instead.

solidgoldpig:
This drip-by-drip approach is still leaving me rather unclear as to how this will actually all join up in the end.

Yes, the service now has the `req` object, but how are we going to get that into the api client?

Will we pass a instance of the api client...

`const apiClient = req.apiClient`

or as part of the req object or will we instantiate the client as needed?

`const apiClient = new ApiClient(req)`
(potentially caching the api client using the transactionId as the key?)
(sorry following the convention of lower case for instance and sentence cased for class - I know Dom hates that)

It would be nice if the services still worked without having to have a `req` context - it's been a life saver being able to use the services in standalone temporary scripts or directly in the node console to see what's happening

`const apiClient = this.req ? req.apiClient || new ApiClient()`

common/middleware/set-services.js
```diff
@@ -2,7 +2,11 @@ const services = require('../services')
 
 const setServices = (req, res, next) => {
   req.services = Object.keys(services).reduce((acc, serviceKey) => {
-    acc[serviceKey] = services[serviceKey]
+    if (services[serviceKey] instanceof Function) {
```
Nimphal, line 5: This is temporary while we transition all the services to the new format.

solidgoldpig:
> Good questions, @solidgoldpig ! I do have an approach in mind for the api client - the service doesn't actually need the whole req object, it just needs a way to ensure it's using an api client which as access to it. 

I disagree. This means we couldn't be able to attach a logger instance to the request and then use it within the services. And I think that would be a very important addition which would finally give us the visibility the project actually needs. (Obviously the api client with the req context would have access to such a logger so it wouldn't be of no benefit, but I think it would be that much more powerful if the services could log with the req context)

> I think we're making the step of always having the req context when using the services, so any scripts should probably build that up in some fashion, but if not desired, all they'd have to do is pass an api client.

I think that is a mistake. YMMV.

`const moveService = require('./common/services/move')(require('./common/lib/api-client')())`

I have had to do a lot of investigatory debugging on the various environments and it was hard enough without having extra hurdles.

If the services become classes, then we create a BaseService class and extend from that

then either set an `apiClient` property in the constructor or have a `getApiClient` method. In both cases checking whether there is a request context and an apiClient attached to it and instantiating one from scratch if not.

Nimphal:
> I disagree. This means we couldn't be able to attach a logger instance to the request and then use it within the services. And I think that would be a very important addition which would finally give us the visibility the project actually needs. (Obviously the api client with the req context would have access to such a logger so it wouldn't be of no benefit, but I think it would be that much more powerful if the services could log with the req context)

The services logging the whole request context is a new requirement on me. I am generally wary at passing things around that are not used, and currently we're not doing that logging and I don't think there are plans to do it any time soon. That being said, the amount of refactor needed to enable it once this work is done is comparatively minimal. Happy to hear what others think about this!

> If the services become classes, then we create a BaseService class and extend from that
> 
> then either set an `apiClient` property in the constructor or have a `getApiClient` method. In both cases checking whether there is a request context and an apiClient attached to it and instantiating one from scratch if not.

I couldn't decide whether having a Base Service was of value here, the only thing in common is the contructor. Probably yes if we end up doing the whole logic around the api client, otherwise maybe not. Not tied to any one approach and happy to implement.

solidgoldpig (approved):

PR #1129, opened 2020-12-11
refactor: Extract date table as decorator

merlinc:
## Proposed changes

The population overview grid needs to be reused on another page, but with a different first column, and additional rows.

This lead to thinking about what the table is doing - and that's two things:
1. Displaying population data
2. Labelling column headers and highlighting cells matching a focusDate (currently always today).

This PR extracts out the second option into a [decorator](https://en.wikipedia.org/wiki/Decorator_pattern).

It works by having the `populationToTable` function add a `date` property onto the component data structure. If the dateTable decorator is applied over the top of that, styling and column headers will be changed. In either case, when eventually rendered as a GDS table, the unknown `date` property is dropped, and not displayed in the final markup.

This should mean that conceptually any table with a date property could be formatted in this way, and with a little bit of work, even something like a calendar could be shown. 

### Issue tracking

- [P4-1423](https://dsdmoj.atlassian.net/browse/P4-1423)

## Screenshots

None needed, because nothing should have changed.

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/population/middleware/set-results.population-table.js
```diff
@@ -1,23 +1,32 @@
...
-const locationsFreeSpacesService = require('../../../common/services/locations-free-spaces')
+const {
+  decorateAsDateTable,
+} = require('../../../common/presenters/date-table/date-table-decorator')
+const {
+  locationsToPopulationTable,
+} = require('../../../common/presenters/date-table/locations-to-population-table-component')
+const {
+  getPrisonFreeSpaces,
+} = require('../../../common/services/locations-free-spaces')
```
Nimphal, line 9: This should be changed to use the service on the when #1128 has been merged. cc @Thource

teneightfive (approved):

PR #1135, opened 2020-12-14
refactor: API client user header and support for custom header config

Nimphal:
### What changed
The API client can pass the req obejct to the requestHeaders middleware which extracts the needed info. Also add a config option for additional headers.

### Why did it change
As part of the work to enable passing the request context to the backend API for audit purposes.

### Issue tracking

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

- [x] No environment variables were added or changes

### Other considerations


common/lib/api-client/index.js
```diff
@@ -18,7 +18,7 @@ const models = require('./models')
 
 let instance
 
-module.exports = function () {
+module.exports = function (req) {
   if (instance) {
```
solidgoldpig, line 22: Won't this mean that whatever is passed as `req` on first instantiation will get cached/used for ever?

Nimphal: Probably :D Didn't fix that, we don't need to check anymore, it will get instantiated on the req and that's what we want. Nice catch, will fix!

solidgoldpig:
```js
const cache = {}

module.exports = function (req = {}) {
    const {transactionId} = req
    if (cache[transactionId]) { return cache[transactionId] }
    const instance = new JsonApi(
    ....
    cache[transactionId] = instance
    return instance
}
````

Comment (solidgoldpig) [reply common/lib/api-client/index.js:22]: Ah - ignore that bit ^^^

File: common/lib/api-client/request-headers.js [hunk 2]
```diff
@@ -3,13 +3,21 @@ const uuid = require('uuid')
...
 
-module.exports = (format = 'application/vnd.api+json') => {
-  return {
+module.exports = (req, format = 'application/vnd.api+json') => {
+  const headers = {
     'User-Agent': `hmpps-book-secure-move-frontend/${APP_VERSION}`,
     Accept: `${format}; version=${VERSION}`,
     'Accept-Encoding': 'gzip',
     'Idempotency-Key': uuid.v4(),
+    ...CUSTOM_HEADERS,
```
solidgoldpig, line 15: Struggling to see the value of this rather than just adding the new headers to this file ¯\_(ツ)_/¯

Nimphal: Unsure myself, following comments from @merlinc regarding a possible configuration. Happy either way.

common/lib/api-client/request-headers.js
```diff
@@ -3,13 +3,21 @@ const uuid = require('uuid')
...
   }
+
+  if (req && req.user) {
```
merlinc, line 18: So my thinking on dynamic headers is more like what's happening here.

For example if we had `scenario-id` or `k8-ingress-id` or `x-api-debug` on the `req`, and we were configured to allow them through, then we'd add them onto the headers of any downstream requests. But these would be dynamic, and not necessarily present on every request so couldn't entirely be backed into the static config.

common/lib/api-client/request-headers.js
merlinc: I think my initial suggestion was a little unclear, so I've detailed it further below.

However, I have the opposite perspective to @solidgoldpig - all of these variables apart from the `format` and the `Idempotency-Key` are static, so could live in the `config` instead.

common/lib/api-client/request-headers.js
```diff
@@ -3,13 +3,21 @@ const uuid = require('uuid')
 const {
   API: { VERSION },
   APP_VERSION,
+  CUSTOM_HEADERS,
 } = require('../../../config')
 
-module.exports = (format = 'application/vnd.api+json') => {
-  return {
+module.exports = (req, format = 'application/vnd.api+json') => {
```
merlinc, line 9: How many calls do we currently make that aren't `json`?

solidgoldpig: The csv download. That's it

common/lib/api-client/request-headers.js
```diff
@@ -3,13 +3,21 @@ const uuid = require('uuid')
...
+  const headers = {
     'User-Agent': `hmpps-book-secure-move-frontend/${APP_VERSION}`,
     Accept: `${format}; version=${VERSION}`,
     'Accept-Encoding': 'gzip',
     'Idempotency-Key': uuid.v4(),
+    ...CUSTOM_HEADERS,
   }
+
+  if (req && req.user) {
+    headers['X-Current-User'] = req.user.username
```
teneightfive, line 19: I was wondering if we want to store the person's fullname instead/as well?

common/lib/api-client/request-headers.js
teneightfive: I'm not sure I see the benefit in them living in "config" either. Why not just import the custom headers module directly? Or even just define them here?

Are they likely to be use elsewhere? And are they always likely to be static?

It feels more complicated to define them in multiple different places if they're only used in one. With this structure I'd need to end up jumping through 2 files to see what other headers are set.

teneightfive (commented):
This approach looks like a good starting point. Left some comments about config vs internal module definition.

common/lib/api-client/request-headers.js
Nimphal: It can't hurt sending it if we want, but Martin B was quite keen for us to not store people's names to leave that to HMPPS Auth. The current usecase is happy with just the username being shown, and we can extract more info from the Auth service if we need it.

There's nowhere to store it in the backend now as well :D

solidgoldpig: Ideally we'd just send the user credentials and let the BE do what it wants with them...  validate the request...  extract the user details as required...

But in the absence of that, shouldn't we just send all the user details and let the BE choose what to use?

Nimphal: I reckon this will evolve over time - currently the BE only knows about username as a first pass, but yes, in the future we will probably send the token and do all sorts of nice things.

common/lib/api-client/index.js
Nimphal: Removed the instance bit. Tests don't like it yet, but they will!

common/lib/api-client/request-headers.js
Nimphal: I've removed this from this PR - it's easy to iterate over the headers now that we have the req.

solidgoldpig (approved):

PR #1144, opened 2020-12-15
[P4-2438] Add support for additional features from framework

teneightfive:
### What changed

This adds support for:
- prefixes and suffixes
- hint text for add another fields

### Why did it change

Content changes to the framework have added support for these new features. This means the frontend needs to
also support importing and rendering these features.

### Issue tracking

- [P4-2438](https://dsdmoj.atlassian.net/browse/P4-2438)

## Screenshots

<kbd>image</kbd>
<kbd>image</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/components/add-another/template.njk
+++ b/common/components/add-another/template.njk
@@ -10,6 +10,8 @@
 {% set items = params.items %}
 
 <div class="moj-add-another">
+  {{ params.hint.html | safe if params.hint.html else params.hint.text }}
+
   {% for item in items %}
     {% if item.html or item.text %}
       {% call govukFieldset({
--- a/common/components/add-another/template.test.js
+++ b/common/components/add-another/template.test.js
@@ -32,6 +32,38 @@ describe('Add another component', function () {
     })
   })
 
+  context('with hint', function () {
+    context('when html is passed to text', function () {
+      it('should render escaped html', function () {
+        const $ = renderComponentHtmlToCheerio('add-another', {
+          title: { text: 'Title' },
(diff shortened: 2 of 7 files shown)


common/components/add-another/template.test.js
```diff
@@ -32,6 +32,40 @@ describe('Add another component', function () {
...
+        const $component = $('body')
+        expect($component.html().trim()).to.contain(
+          '&lt;span&gt;Reference&lt;/span&gt;'
+        )
+      })
+    })
+
+    context('when html is passed to html', function () {
+      it('should render unescaped html', function () {
+        it('should render escaped html', function () {
```
merlinc, line 54: I think this is an accidental nesting

common/components/add-another/template.njk
```diff
@@ -10,6 +10,8 @@
 {% set items = params.items %}
 
 <div class="moj-add-another">
+  {{ params.hint.html | safe if params.hint.html else params.hint.text }}
```
solidgoldpig, line 13: Should this not be a heading or even, a legend in another uber fieldset that wraps all the other ones?

teneightfive: That's probably a good point. Although, in our current use it does get wrapped in a `p`.

If we wrap this component in a heading that's probably not ideal...

solidgoldpig (approved):

PR #1145, opened 2020-12-15
P4 1776 update move service

Nimphal:
### What changed
Converts move service to class which takes the req object on instantiation. Tweaks batch request so that the right api client is used for the requests. Updates the single request service constructor so it can access the move service from req.

### Why did it change
Enables the service to access the api client that was instantiated in the request cycle and has access to the req object.

### Issue tracking

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/services/batch-request.js
+++ b/common/services/batch-request.js
@@ -21,6 +21,8 @@ const batchRequests = async (clientMethodCall, props = {}, propPaths = []) => {
       propPathValue = propPathValue.split(',')
     }
 
+    const { apiClient } = props
+
     const chunks = chunk(propPathValue, LOCATIONS_BATCH_SIZE).map(id =>
       id.join(',')
     )
@@ -29,7 +31,7 @@ const batchRequests = async (clientMethodCall, props = {}, propPaths = []) => {
       chunks.map(chunk => {
         const batchedProps = cloneDeep(props)
         set(batchedProps, propPath, chunk)
-        return clientMethodCall(batchedProps)
+        return clientMethodCall({ ...batchedProps, apiClient: apiClient })
       })
     ).then(response => {
       if (props.isAggregation) {
--- a/common/services/move.js
+++ b/common/services/move.js
@@ -1,9 +1,9 @@
 const dateFunctions = require('date-fns')
 const { mapValues, omitBy, isUndefined, isEmpty } = require('lodash')
 
-const apiClient = require('../lib/api-client')()
 const restClient = require('../lib/api-client/rest-client')
 
+const BaseService = require('./base')
 const batchRequest = require('./batch-request')
 
 const defaultInclude = [
@@ -38,6 +38,7 @@ const defaultInclude = [
 ]
 
 function getAll({
+  apiClient,
   filter = {},
   combinedData = [],
   page = 1,
@@ -66,6 +67,7 @@ function getAll({
       }
 
       return getAll({
+        apiClient,
         filter,
         combinedData: moves,
         page: page + 1,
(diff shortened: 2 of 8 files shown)


common/services/batch-request.js
```diff
@@ -29,7 +31,7 @@ const batchRequests = async (clientMethodCall, props = {}, propPaths = []) => {
       chunks.map(chunk => {
         const batchedProps = cloneDeep(props)
         set(batchedProps, propPath, chunk)
-        return clientMethodCall(batchedProps)
+        return clientMethodCall({ ...batchedProps, apiClient: apiClient })
```
Nimphal, line 34: Happy to be told off for this, but essentially, we don't want to use the cloned api client (mocks don't work, for example). We want to use the same client that was passed in.

common/services/move.js
```diff
@@ -38,6 +38,7 @@ const defaultInclude = [
 ]
 
 function getAll({
+  apiClient,
```
Nimphal, line 41: This is outside of the service scope and doesn't have access to the req. It could be moved inside in which case this passing of apiClient won't be necessary, but kept outside for minimum changes. Happy to revisit.

common/services/single-request.js
```diff
@@ -2,10 +2,16 @@ const dateFunctions = require('date-fns')
...
-const moveService = require('./move')
+const MoveService = require('./move')
 
 const noMoveIdMessage = 'No move ID supplied'
 class SingleRequestService extends BaseService {
+  constructor(req) {
+    super()
+    this.moveService =
+      req?.services?.move || new MoveService({ apiClient: this.apiClient })
+  }
```
Nimphal, line 13: A bit of service-ception going on here, we want the single request service to use the right move service. There are many ways to go about this, but they involve enhancing the constructor. I think these cases are limited, so I opted to not update the base service.

common/services/move.js
solidgoldpig: It think it would be a good idea to move it inside and to perhaps change the name to something less confusing than the duped `getAll`

Previously the biggest problem we had with service calls was batched requests overwhelming the BE - 
If we don't "pass" the req somehow, all the calls that end up using it won't be able to send the transactionId or user details or (future handwavey) do useful logging.

common/services/single-request.js
solidgoldpig: I think this is totally acceptable. But I'd suggest 

```js
class BaseService {
  constructor(req) {
    this.req = req || {}
    this.apiClient = req.apiClient || new ApiClient()
  }
}
```

then it's just

`new MoveService(this.req)`

(sorry, it was an oversight on my part before not to have suggested stashing the request context on the instance)

common/services/batch-request.js
solidgoldpig: This whole method smells :(

common/services/move.js
solidgoldpig: But maybe you're right. Maybe the correct solution is to get to the point where the batching can simply be removed.

common/services/single-request.js
Nimphal: I suppose it's actually a further step towards that mythical logging we've been chatting about! :D

common/services/move.js
Nimphal: There is a comment in the batch file about removing it once services know about a user's locations, which after this refactor, they will be able to, so there's probably a case for tackling that properly soon after this work is complete.

common/services/batch-request.js
Nimphal: I am hoping the comment at the beginning of it about it being able to be removed once the services know the user's locations is true and we can do that in the new year.

solidgoldpig (approved):

PR #1153, opened 2020-12-16
[P4-2557] Create logic to determine if the person is serving a youth sentence

teneightfive:
### What changed

This set of changes introduces logic within the move creation journey to ask if a prisoner is serving their sentence in the youth estate.

This check only occurs when the location is marked as a `prison` and also a `young_offender_institution` and the person is between 18 and 19. 

This moves the logic for determining if a youth risk assessment should be shown to the profile, where the attribute `requires_youth_risk_assessment` is set when creating the move.

In some scenarios this can be determined automatically, in others the extra question needs to be asked.

#### Logic

- if coming from `secure_childrens_home` - **true**
- if coming from `secure_training_centre` - **true**
- if coming from `prison`
  - if `young_offender_institution`
    - if under 18 - **true**
    - if over 18 - **false**
    - if serving on secure youth estate - **true**
    - if serving **not** on secure youth estate - **false**
  - else - **false**
- else - **false**

### Why did it change

Currently, all establishment that can hold young offenders (HMYOI or HMP/YOI) are marked as a `prison` location type within the API. This means in certain circumstances we can't tell if the person being moved is serving their sentence in the youth estate.

That requires us to ask the user creating the move in this scenarios.

### Issue tracking

#### YOI

When over 18

<kbd>overage-prison</kbd>

When under 18

<kbd>underage-yoi</kbd>

When 18 or 19

<kbd>agecheck-yoi</kbd>

### Secure Childrens Home or Secure Training Centre

<kbd>youth-estate</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/move/app/new/controllers/base.js
+++ b/app/move/app/new/controllers/base.js
@@ -188,6 +188,38 @@ class CreateBaseController extends FormWizardController {
 
     return false
   }
+
+  requiresYouthAssessment(req) {
+    const {
+      person = {},
+      from_location_type: fromLocationType,
+      is_young_offender_institution: isYoungOffenderInstitution,
+      serving_youth_sentence: servingYouthSentence,
+    } = req.sessionModel.toJSON()
+    const age = filters.calculateAge(person.date_of_birth)
+
+    if (
+      ['secure_training_centre', 'secure_childrens_home'].includes(
+        fromLocationType
+      )
+    ) {
+      return true
+    }
+
+    if (
+      fromLocationType === 'prison' &&
+      isYoungOffenderInstitution &&
+      age < 18
+    ) {
+      return true
+    }
+
+    if (servingYouthSentence === 'yes') {
+      return true
+    }
+
+    return false
+  }
 }
 
 module.exports = CreateBaseController
--- a/app/move/app/new/controllers/base.js
+++ b/app/move/app/new/controllers/base.js
@@ -1,5 +1,6 @@
 const FormWizardController = require('../../../../../common/controllers/form-wizard')
 const presenters = require('../../../../../common/presenters')
+const filters = require('../../../../../config/nunjucks/filters')
 
 class CreateBaseController extends FormWizardController {
   middlewareSetup() {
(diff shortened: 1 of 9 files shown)


app/move/app/new/controllers/base.js
```diff
@@ -133,14 +134,36 @@ class CreateBaseController extends FormWizardController {
...
+    const isYoungOffenderInstitution = req.sessionModel.get(
+      'is_young_offender_institution'
+    )
+    const { date_of_birth: dateOfBirth } = req.sessionModel.get('person') || {}
+    const age = filters.calculateAge(dateOfBirth)
+
+    if (
+      fromLocationType === 'prison' &&
+      isYoungOffenderInstitution &&
+      age >= 18 &&
```
solidgoldpig, line 159: Would this make more sense if the age ranges lived in the config. (Same goes for `app/moves/middleware/set-body-request-filters/age.js` - and anywhere else???)

app/move/app/new/steps.js
```diff
@@ -137,14 +138,28 @@ module.exports = {
...
     ],
     fields: ['people'],
   },
   '/personal-details': {
     controller: PersonalDetails,
     pageTitle: 'moves::steps.personal_details.heading',
-    next: 'move-details',
+    next: [
+      {
+        fn: Base.prototype.shouldAskYouthSentenceStep,
```
solidgoldpig, line 156: Could we not just set a value in the journey based on the feature flag value and simply have a corresponding next step for the 2 possibilities?

app/move/app/new/controllers/base.js
teneightfive: The request filters are a separate thing I feel.

But in terms of this, yeah maybe it might be better to define this in the config to be more specific.

app/move/app/new/steps.js
teneightfive: I was struggling where to do it really, and I didn't want to set that value on every save. 

The logic is a little more complicated than the value submitted by a form and I felt this was the neatest way to extract that logic.

By using a method it means it will not be run with every request. 

If we were to set a value during the "save" it would also need to be done both via the person search results and creating a new person and that then felt a little harder to follow.

Nimphal (approved):
Nice!

PR #1154, opened 2020-12-16
[P4-2557] Use profile attribute to determine whether to show youth assessment

teneightfive:
### What changed

Use API attribute to determine whether or not to show the youth risk assessment for a move.

### Why did it change

The logic for showing a youth risk assessment has been changed so that
the frontend now determines this.

This means that the logic for showing the banner needs to be updated to
use this new attribute rather than base it just on if the move is from
a youth establishment.

### Issue tracking

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


common/helpers/move/get-assessments.test.js
```diff
@@ -162,237 +162,237 @@ describe('Move helpers', function () {
           answers: [],
         })
       })
+    })
```
merlinc, line 165: With whitespace ignored, this comes up as a single change - just checking that you want to close off the `by default` block here

teneightfive: Yes, I realised that the indentation was originally wrong. The `by default` should be separate to the `with Youth Risk Assessment`.

Nimphal (approved):
Nice!

PR #1171, opened 2020-12-18
[P4-2557] fix: Check for profile in base create controller

teneightfive:
### What changed

Moves profile creation to the base controller.

Fixes issue where changing the person wouldn't create a new profile.

### Why did it change

Currently we check if a profile exists within the assessment controller.

This is done as part of setting the previous assessment.

However, we are introducing different flows to the journey which will
mean these steps are never seen, and therefore the profile is never
created.

This leads to an error when saving the move as the profile doesn't
exist.

This also addresses a bug where if a user tries to change the person,
a new profile is not created as we don't validate that the profile is
for the currently selected person.

### Issue tracking

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/move/app/new/controllers/base.test.js
+++ b/app/move/app/new/controllers/base.test.js
@@ -42,6 +42,8 @@ describe('Move controllers', function () {
       beforeEach(function () {
         sinon.stub(FormWizardController.prototype, 'middlewareSetup')
         sinon.stub(controller, 'use')
+        sinon.stub(controller, 'setModels')
+        sinon.stub(controller, 'setProfile')
 
         controller.middlewareSetup()
       })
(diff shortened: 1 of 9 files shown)


app/move/app/new/controllers/save.test.js
```diff
@@ -46,11 +42,17 @@ const mockValues = {
...
   person: {
+    id: '__person_1__',
     first_names: 'Steve',
     last_name: 'Smith',
   },
   profile: {
-    id: '__profile',
+    id: '__profile_3__',
+    person: {
+      id: '__person_1__',
```
merlinc, line 52: Small thing, but could the profile id and the person id be similar so its more obvious they're connected?

merlinc: Not that they're used anywhere else, so this is probably not a useful change

app/move/app/new/controllers/base.test.js
```diff
@@ -57,8 +59,166 @@ describe('Move controllers', function () {
         )
       })
 
+      it('should call set models method', function () {
```
merlinc, line 62: This should say `call set profile method`

app/move/app/new/controllers/save.test.js
solidgoldpig: `person.id` and `profile.person.id` are the same, aren't they? `profile.id` is a different beast entirely…

solidgoldpig (approved):

app/move/app/new/controllers/save.test.js
teneightfive: This was to show some scenarios where the profile may exist, but the person on that profile can be different (ie a person goes back a step and selects a different person).

>person.id and profile.person.id are the same, aren't they?

Not necessarily in all cases. That's what this also fixes.

PR #1175, opened 2020-12-18
Ensure profile not null in move get-locals helper

solidgoldpig:
### What changed

Put a guard to ensure `move.profile` is not null

### Why did it change

`person` is retrieved from the `profile` object

### Issue tracking

https://sentry.io/organizations/ministryofjustice/issues/2100728538/?project=2014813&query=is%3Aunresolved#exception

Fixes BOOK-A-SECURE-MOVE-FRONTEND-23

## Screenshots

n/a

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/helpers/move/get-locals.js
+++ b/common/helpers/move/get-locals.js
@@ -17,7 +17,7 @@ const getUpdateUrls = require('./get-update-urls')
 function getLocals(req) {
   const { move, canAccess } = req
   const userPermissions = req.session?.user?.permissions
-  const { profile } = move
+  const profile = move.profile || {}
   const { person } = profile
 
   // person
--- a/common/helpers/move/get-locals.test.js
+++ b/common/helpers/move/get-locals.test.js
@@ -296,6 +296,71 @@ describe('Move helpers', function () {
       })
     })
 
+    context('when move doesn’t have a profile', function () {
+      let params
+
+      beforeEach(function () {
+        req.move = {
+          id: 'moveId',
+          status: 'requested',
+        }
+
+        params = getLocals(req)
+      })
+
+      it('should still return the expected locals', function () {
+        expect(Object.keys(params).sort()).to.deep.equal([
+          'additionalInfoSummary',
+          'assessments',
+          'canCancelMove',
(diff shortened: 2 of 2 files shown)


merlinc (approved):

common/helpers/move/get-locals.test.js
```diff
@@ -296,6 +296,38 @@ describe('Move helpers', function () {
       })
     })
 
+    context('when move doesn’t have a profile', function () {
+      let params
+
+      beforeEach(function () {
+        req.move = {
+          id: 'moveId',
+          status: 'requested',
```
teneightfive, line 305: Can we make this test for `profile: null`. I don't think this will prevent the error we've seen because of this:

>Default values in destructuring assignement only work if the variables either don't exist or their value is set to undefined . Any other value, including null , false and 0 , bypasses the default values in the destructuring statement. You can combine default values with renaming in the destructuring assignment.

I think it will still come through as `null` instead of an empty object and cause the error.

common/helpers/move/get-locals.js
```diff
@@ -17,7 +17,7 @@ const getUpdateUrls = require('./get-update-urls')
 function getLocals(req) {
   const { move, canAccess } = req
   const userPermissions = req.session?.user?.permissions
-  const { profile } = move
+  const { profile = {} } = move
   const { person } = profile
```
teneightfive, line 21:
```suggestion
  const { person } = profile || {}
```

common/helpers/move/get-locals.js
```diff
@@ -17,7 +17,7 @@ const getUpdateUrls = require('./get-update-urls')
 function getLocals(req) {
   const { move, canAccess } = req
   const userPermissions = req.session?.user?.permissions
-  const { profile } = move
+  const { profile = {} } = move
```
teneightfive, line 20:
```suggestion
  const { profile } = move
```

teneightfive (approved):

PR #1185, opened 2020-12-22
Store transaction ID in Sentry and pass to API calls

teneightfive:
### What changed

This uses the request ID header to store against sentry issues and also
passes that same value to API calls.

### Why did it change

This allows a particular frontend request can be traced through to particular API calls that might have
errorred.


Nimphal (approved):
How exciting!

PR #1202, opened 2020-12-29
chore: Add support for split pagination modifier

teneightfive:
### What changed

Add a modifier to the pagination component to allow it to be split across the container element.

Also includes a tidy up to the examples. Examples should only show visual changes. Custom options are document in the yaml files and should be tested within creating visual examples that will be rendered in the component explorer.

### Why did it change

This allows pagination to be displayed on either side of the container
element and will be used to pagination results in a collection.

## Screenshots

<kbd>image</kbd>

[**DEMO**](https://booksecuremove-pr-1202.herokuapp.com/components/pagination)

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/components/pagination/pagination.yaml
+++ b/common/components/pagination/pagination.yaml
@@ -73,13 +73,6 @@ examples:
       next:
         href: "/next-day"
         text: "Next day"
-  - name: inline variation
-    data:
-      classes: app-pagination--inline
-      previous:
-        href: "/previous-page"
-      next:
-        href: "/next-page"
   - name: with items
     data:
       previous:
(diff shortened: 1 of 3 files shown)


common/components/pagination/pagination.yaml
```diff
@@ -93,38 +86,33 @@ examples:
           text: "two"
         - href: "/page-3"
           text: "three"
-  - name: with items with class
```
teneightfive, line 96: These were removed from the examples as each example is rendered in the components explorer. These two variations had no visual difference so that's why they've been removed from here and are now tested separately within the module test file.

common/components/pagination/template.test.js
```diff
@@ -184,60 +184,72 @@ describe('Pagination component', function () {
       expect($itemLink.attr('href')).to.equal('/page-3')
     })
   })
-})
 
-context('with items with class', function () {
-  let $, $component, items
+  context('with items with class', function () {
```
teneightfive, line 188: The indentation of these tests was previously incorrect, they were sat outside the containing block that describes this component. That's why this change is showing some whitespace changes.

cwrw (approved):

PR #1203, opened 2020-12-29
refactor: Rename pagination to be date specific

teneightfive:
### What changed

This change makes the date pagination used for rendering resources
within a collection specific to the type of pagination it is - date.

### Why did it change

There is some future work to introduce pagination to cycle through
pages so this separation will allow that work to be introduced more
easily.


cwrw (approved):

cwrw:
looks good, I think we're missing a rename for pagination in `app/population/controllers/dashboard.js` and its corresponding test

PR #1208, opened 2020-12-31
[P4-2366] Paginate allocations and single requests

teneightfive:
### What changed

This adds pagination to the single request and allocations views.

It also removes the use of the batch request library which was responsible for splitting requests with large querystrings.

### Why did it change

These pages often result in timeouts in production, particularly when asking for multiple locations. This is because we are asking for **all** results at once, and sometimes batching requests and paginating each batch.

This change will reduce the page load times for these pages and limit results to 20 at a time.

### Issue tracking

- [P4-2366](https://dsdmoj.atlassian.net/browse/P4-2366)
- [Sentry timeouts](https://sentry.io/organizations/ministryofjustice/issues/1946186432)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd>localhost_3000_moves_week_2021-01-15_requested_status=pending (1)</kbd> | <kbd>localhost_3000_moves_week_2021-01-15_requested_status=pending</kbd> |
| <kbd>localhost_3000_allocations_week_2021-01-15_outgoing_status=unfilled sortBy=date sortDirection=asc (1)</kbd> | <kbd>localhost_3000_allocations_week_2021-01-15_outgoing_status=unfilled sortBy=date sortDirection=asc</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


common/services/allocation.js
```diff
@@ -2,47 +2,6 @@ const dateFunctions = require('date-fns')
 const { mapValues, pickBy, set } = require('lodash')
 
 const BaseService = require('./base')
-const batchRequest = require('./batch-request')
```
Nimphal, line 5: :dancers:

Nimphal (approved):

app/allocations/middleware/set-results.allocations.js
```diff
@@ -9,12 +9,11 @@ async function setResultsAllocations(req, res, next) {
   }
 
   try {
-    const results = await req.services.allocation.getByDateAndLocation(
+    req.results = await req.services.allocation.getByDateAndLocation(
```
merlinc, line 12: Elsewhere we "namespace" these results variables with names like `resultsAsTables` or `resultsAsCards` - is there a more descriptive name we could use for this?

app/moves/middleware/set-results.single-requests.js
```diff
@@ -4,11 +4,10 @@ async function setResultsSingleRequests(req, res, next) {
   const singleRequestService = req.services.singleRequest
 
   try {
-    const singleRequests = await singleRequestService.getAll(req.body.requested)
-
+    req.results = await singleRequestService.getAll(req.body.requested)
```
merlinc, line 7: Are we using `req.results` later on in the request handling lifecycle?

common/middleware/collection/set-pagination.js
```diff
@@ -0,0 +1,53 @@
...
+          }
+        ),
+        text: i18n.t('pagination.next_page'),
+        label: i18n.t('pagination.page_of', {
+          current: nextPageNumber,
+          total: totalPages,
+        }),
+      }
+
+      req.pagination = {
```
merlinc, line 43: It would be good to get tests for this result, both within and outside of the page number boundaries.

common/middleware/collection/set-pagination.test.js
```diff
@@ -0,0 +1,39 @@
...
+        expect(mockReq).not.to.have.property('pagination')
+      })
+
+      it('should call next', function () {
+        expect(nextSpy).to.be.calledOnceWithExactly()
+      })
+    })
+
+    context('with results', function () {
+      context('without existing page', function () {})
```
merlinc, line 34: Tests to come :)

common/services/allocation.js
```diff
@@ -140,34 +99,68 @@ class AllocationService extends BaseService {
...
-        'filter[status]': status,
-        'filter[from_locations]': fromLocations.join(','),
-        'filter[to_locations]': toLocations.join(','),
-        'filter[locations]': locations.join(','),
-        'filter[date_from]': moveDateFrom,
-        'filter[date_to]': moveDateTo,
-        'sort[by]': sortBy,
-        'sort[direction]': sortDirection,
-      }),
+      filter: {
```
merlinc, line 113: Does this mean that other code that uses the `filter[property]` pattern should be updated to use this approach once this is merged in?

common/services/allocation.js
```diff
@@ -140,34 +99,68 @@ class AllocationService extends BaseService {
...
+    include,
+  } = {}) {
+    return this.apiClient
+      .all('allocation')
+      .all('filtered')
+      .post(
+        {
+          filter: this.removeInvalid(filter),
+        },
+        this.removeInvalid({
```
merlinc, line 144: Is it a git-diff formatting issue, or is `removeInvalid` a new function?
... 
*looks 1 file down*

Ah, yes, it's new :)

common/services/base.js
```diff
@@ -1,10 +1,18 @@
...
+
 const ApiClient = require('../lib/api-client')
 
 class BaseService {
   constructor(req) {
     this.req = req || {}
     this.apiClient = (req && req.apiClient) || new ApiClient()
   }
+
+  removeInvalid(input) {
```
merlinc, line 11: Is this really part of the `BaseService` or is it more of a general utility method? It looks like it could be used elsewhere quite easily.

common/controllers/collection/render-as-table.test.js
```diff
@@ -0,0 +1,122 @@
+const controller = require('./render-as-table')
```
merlinc, line 1: Is this the first test for `render-as-table`? :scream: :heart:

merlinc (changes requested):

merlinc:
FYI - the `locations-free-spaces` service also uses the `load all batches` approach, however as a pure stats endpoint it has substantially better performance, and currently only loads 2 pages at most.

app/allocations/middleware/set-results.allocations.js
teneightfive: The reason for this is that this is a different structure of the object and is used by a different middleware than the other properties.

It's used by the [`set-pagination`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/ea1555b5e1f0e5e5d9cc5efabd0775f6bf90ae42/common/middleware/collection/set-pagination.js) middleware.

common/middleware/collection/set-pagination.js
teneightfive: Yep, as mentioned on slack, this was the last module I was finishing tests for. But I wanted to get feedback on the structure :)

common/services/allocation.js
teneightfive: Only if it uses the new `POST` endpoints within the API. If you're calling a `GET` endpoint they should ideally still follow this structure.

But this is to solve a bigger long term issue of the frontend doing the location permissions...

common/services/allocation.js
teneightfive: Yep, it's new. And added some tests :)

common/services/base.js
teneightfive: I toyed with where this should live....the reason I ended up putting it here is because it made it easier to finish the tests for this set of changes and also it's more specifically linked to these services.

common/controllers/collection/render-as-table.test.js
teneightfive: Yep... :/

app/allocations/middleware/set-results.allocations.js
merlinc: Ok, so I guess we have a pattern where we put the raw result on the `req` as `results` and then any modified variants with some `flavour` of namepace, e.g. `resultsAsTables`

common/services/allocation.js
merlinc: This now introduces 2 different ways of using filters with the services, a useful future cleanup would be to bring them in alignment, no matter what backend implementation is being used.

common/services/base.js
merlinc: It feels more like a https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/master/common/helpers to me than a core `BaseService` method - but that does make it a little more annoying to use

app/allocations/middleware/set-results.allocations.js
teneightfive: Essentially, yes. This still just follows the same approach of placing things on the `req` object and then using that in subsequent middleware/controllers.

common/services/allocation.js
teneightfive: So, I don't think it's about the difference in the service. This new implementation just has to adjust how this service interacts with the API client layer.

Previously we were able to use query params to filter results through a `GET` (or the `findAll` method in devour). The problem we encountered with this was we would hit the querystring limit when passing lots of UUIDs into particular filters like locations.

So, for now, the way we've had to implement this is to create a new `POST` endpoint in the API so that we are not restricted by character limit. But in order to use the new endpoint we needed to update the way in which the service interacts with the devour and the API. @cwrw added a [wiki file](https://dsdmoj.atlassian.net/wiki/spaces/PECS4/pages/2777907242/POST+Moves+and+Allocations+Endpoints+for+large+query+strings) on why we've had to do this.

The `filter[property]` pattern is what is documented in the [JSON:API documentation](https://jsonapi.org/format/#fetching-filtering) and is what we should use by default. 

But this change is trying to help solve the wider issue of the API not knowing about a user and their caseload.

common/services/base.js
teneightfive: Yes, I think that was the decision I took at the time. This work has taken quite a while and I didn't want to waste longer refactoring things again too much. 

The other reason for placing it here is that it's only related to these specific services and we didn't currently have any service related "helpers" yet either so it wasn't going to create two places for the same thing.

teneightfive:
> FYI - the `locations-free-spaces` service also uses the `load all batches` approach, however as a pure stats endpoint it has substantially better performance, and currently only loads 2 pages at most.

@merlinc Do you have a link to some code that uses it? When I was making these changes I searched for anywhere else using the `batch-response` and couldn't find any other uses.

app/allocations/middleware/set-results.allocations.js
merlinc: Ok, so I've probably tied myself up in knots avoid using a raw `results` in some of my PMU work then :)

:+1:

merlinc:
> > FYI - the `locations-free-spaces` service also uses the `load all batches` approach, however as a pure stats endpoint it has substantially better performance, and currently only loads 2 pages at most.
> 
> @merlinc Do you have a link to some code that uses it? When I was making these changes I searched for anywhere else using the `batch-response` and couldn't find any other uses.

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/services/locations-free-spaces.js#L24-L34

PMU works differently so doesnt have the same performance problems as the `move` centric endpoints. Also, comparing all prisons in an area in one screen is one of the use cases for PMU, so moving to pagination would degrade functionality rather than improve it.

(I was mostly pointing it out so that we're aware that some parts of the app still use the old method, but for a specific reason)

merlinc (approved):

PR #1213, opened 2021-01-04
test: Add E2E tests for population dashboard

merlinc:
## Proposed changes

- Add choose random region for PMU users, as they mostly work on regions, not on single locations
- Add basic navigation test on population dashboard
- Add focus test for population dashboard
- Add test for editing of first cell in a region

Note: At the moment there are the following limitations with these tests:

1. Region is randomly chosen, but only the first location is edited
2. Once a location has been edited, the URL is manually constructed, rather than followed from the details page
3. The numbers entered on the edit screen are not rechecked on the details screen

### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

- [] Added unit tests to cover changes

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/test/e2e/_roles.js
+++ b/test/e2e/_roles.js
@@ -38,6 +38,11 @@ export const ocaUser = Role(home, async t => {
 
 export const pmuUser = Role(home, async t => {
   await page.signIn(E2E.ROLES.PMU)
+  await page.chooseLocation()
+})
+
+export const pmuUserWithRegions = Role(home, async t => {
+  await page.signIn(E2E.ROLES.PMU)
   await page.chooseRegion()
 })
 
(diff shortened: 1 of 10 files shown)


teneightfive (approved):

test/e2e/_roles.js
```diff
@@ -41,6 +41,11 @@ export const pmuUser = Role(home, async t => {
   await page.chooseLocation()
 })
 
+export const pmuUserWithRegions = Role(home, async t => {
```
merlinc, line 44: @teneightfive So it turns out that many of our tests use the PMU user, but some need a location to be chosen, and some will need a region to be chosen.

The only place we select a location is within these `_roles` methods.

An alternative to the approach here would be to pass in a "choose location / choose region" variable, which also feels a bit hacky.

teneightfive (approved):

PR #1217, opened 2021-01-05
Prevent double click error when confirming a Person Escort Record

teneightfive:
### What changed

This change adapts the save method to check for a particular error
and redirect a user when it is seen.

### Why did it change

We have seen errors in the API where confirmations get prevented.

This appears to be when uses double click the confirmation button
causing two submissions.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


common/controllers/framework/confirm-assessment.js
```diff
@@ -17,6 +19,28 @@ class ConfirmAssessmentController extends FormWizardController {
...
+    }
+
+    try {
+      await services[assessment.framework.name].confirm(assessment.id)
+
+      next()
+    } catch (err) {
+      const apiErrorCode = get(err, 'errors[0].code')
+
+      if (err.statusCode === 422 && apiErrorCode === 'taken') {
```
teneightfive, line 36: This is blocked until we know the code that the API will return so that we can catch it.

cc @cwrw

common/templates/error.njk
```diff
@@ -4,6 +4,8 @@
 
 {% block organisationSwitcher %}{% endblock %}
 
+{% block beforeContent %}{% endblock %}
```
merlinc, line 7: What is this empty block hiding from the error page?

common/controllers/framework/confirm-assessment.js
```diff
@@ -17,6 +19,28 @@ class ConfirmAssessmentController extends FormWizardController {
...
+        return this.successHandler(req, res)
```
merlinc, line 37: If I remember correctly, if we get a 422 and continue onto the next step, we'll show a 'taken' message on that screen?

So this code is only catching the known taken error, and other errors should show up as errors?

common/templates/error.njk
teneightfive: This prevents the breadcrumb being shown on the default error page, I popped a comment in the commit description about this change :)

common/controllers/framework/confirm-assessment.js
teneightfive: Those are currently only picked up in the [save controllers](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/a45e499e51f00792e709644487d3468299200b20/app/move/app/new/controllers/save.js#L128-L153) are specific to creating a duplicate resource.

Although it might be better to follow that pattern of catching it in the `errorHandler` instead of directly when saving?

Also, its worth noting that this code isn't going to be `taken`, I'm waiting for @cwrw to confirm what the code will be. This error occurs when an assessment is already confirmed and we're attempting to patch it with the same status.

common/controllers/framework/confirm-assessment.test.js
```diff
@@ -81,6 +81,121 @@ describe('Framework controllers', function () {
...
+        it('should not call parent error handler', function () {
+          expect(FormWizardController.prototype.errorHandler).not.to.be.called
+        })
+
+        it('should call success handler', function () {
+          expect(controller.successHandler).to.be.calledOnceWithExactly({}, {})
+        })
+      })
+
+      context('with any other error', function () {
```
merlinc, line 183: :+1:

common/controllers/framework/confirm-assessment.test.js
```diff
@@ -81,6 +81,121 @@ describe('Framework controllers', function () {
...
+    })
+
+    describe('#errorHandler', function () {
+      let nextSpy
+
+      beforeEach(async function () {
+        nextSpy = sinon.spy()
+      })
+
+      context('when assessment is already confirmed', function () {
```
merlinc, line 158: :+1:

merlinc (approved):

PR #1223, opened 2021-01-08
refactor(pmu): Use grouped fields on edit page

merlinc:
## Proposed changes

Apply field grouping to the edit screen to more accurately represent the mockups. This is the pre-refactoring to more easily support P4-2463

This was done dynamically to support the "automatic layout" approach that the rest of the application uses. This is done by using a new property on the steps called `fieldGroups` which separates all the fields into separate groups.

### Issue tracking

- [P4-2463](https://dsdmoj.atlassian.net/browse/P4-2463)

## Screenshots

| Before | After |
| ------ | ----- |
| Screenshot_2021-01-08 Change numbers - create_population |  Screenshot_2021-01-08 Change numbers - create_population(1) |
#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/population/views/edit/edit.njk
+++ b/app/population/views/edit/edit.njk
@@ -1,0 +1,34 @@
+{% extends "form-wizard.njk" %}
+
+{% macro displayFieldGroup(group) %}
+  {% set fields = options.fieldGroups[group] %}
+
+  {% set key = "population::daily_view.labels." + group %}
+
+  <div class="govuk-grid-column-full govuk-!-margin-bottom-2">
+    <h2 class="govuk-heading-m">{{ t(key) }}</h2>
+
+    {% for fieldKey in fields %}
+      {% set field = options.fields[fieldKey] %}
+
+      {% if field.component and not field.skip %}
+        {{ callAsMacro(field.component)(field) }}
+      {% endif %}
+    {% endfor %}
+
+    {{ caller() if caller }}
+  </div>
+{% endmacro %}
+
+{% block fields %}
+  <div class="govuk-grid-row">
+    {% call displayFieldGroup('total_space') %}
+    {% endcall %}
+
+    {% call displayFieldGroup('unavailable_space') %}
+    {% endcall %}
+
+    {% call displayFieldGroup('available_space') %}
+    {% endcall %}
+  </div>
+  {% endblock %}
--- a/app/population/steps/edit.js
+++ b/app/population/steps/edit.js
@@ -21,6 +21,12 @@ module.exports = {
       'out_of_area_courts',
       'discharges',
     ],
+    fieldGroups: {
+      total_space: ['operational_capacity', 'usable_capacity'],
+      unavailable_space: ['unlock', 'bedwatch', 'overnights_in'],
+      available_space: ['overnights_out', 'out_of_area_courts', 'discharges'],
+    },
     buttonText: 'actions::save_and_calculate',
+    template: 'edit',
   },
 }
--- a/locales/en/fields.json
+++ b/locales/en/fields.json
@@ -454,21 +454,21 @@
     "label": "Bedwatch"
   },
   "overnights_in": {
-    "label": "Overnights In"
+    "label": "Overnights in"
   },
   "overnights_out": {
-    "label": "Overnights Out"
+    "label": "Overnights out"
   },
   "out_of_area_courts": {
-    "label": "Out Of Area Courts"
+    "label": "Out of area courts"
   },
   "discharges": {
-    "label": "Discharges"
+    "label": "Discharges including immigration"
   },
   "operational_capacity": {
-    "label": "Operational Capacity"
+    "label": "Operational capacity"
   },
   "usable_capacity": {
-    "label": "Usable Capacity"
+    "label": "Usable capacity"
   }
 }


app/population/views/edit/edit.njk
```diff
@@ -0,0 +1,32 @@
+{% extends "form-wizard.njk" %}
+
+{% macro fieldGroup(name) %}
+  {%  set key = "population::daily_view.labels." + name %}
+  {%  set label = t(key) %}
+  <h2 class="govuk-heading-m">{{ label}}</h2>
```
teneightfive, line 6:
```suggestion
  <h2 class="govuk-heading-m">{{ label }}</h2>
```

app/population/views/edit/edit.njk
```diff
@@ -0,0 +1,32 @@
...
+      {{ callAsMacro(fieldsOptions.component)(fieldsOptions) }}
+    {% endif %}
+  {% endfor %}
+{% endmacro %}
+
+
+{% block fields %}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full govuk-!-margin-bottom-2">
```
teneightfive, line 20: I wonder if this could be tackled without the need for that filter:

```nunjucks
{% extends "form-wizard.njk" %}

{% block fields %}
  <div class="govuk-grid-row">
    {% for group, fields in options.fieldGroups %}
      {% set key = "population::daily_view.labels." + group %}

      <div class="govuk-grid-column-full govuk-!-margin-bottom-2">
        <h2 class="govuk-heading-m">{{ t(key) }}</h2>

        {% for fieldKey in fields %}
          {% set field = options.fields[fieldKey] %}

          {% if field.component and not field.skip %}
            {{ callAsMacro(field.component)(field) }}
          {% endif %}
        {% endfor %}
      </div>
    {% endfor %}
  </div>
{% endblock %}
```

merlinc: I like the look of that - much less code to think about!

I think it will still need to be a macro though, as the subsequent PR adds in the Transfers In / Transfers Out - so we'd need to explicitly call each group to be able to add them into the bottom of each `unavailable` / `available` block.

teneightfive: Or you just manually add the markup for those elements as it looks like they're custom anyway?

merlinc: So I've dealt with it by using your code as a `macro` - but in #1224 , I'll invoke the macros using `call` - that way I can fill in the body content inside each block, keeping everything contained.

Nimphal (approved):

app/population/views/edit/edit.njk
```diff
@@ -0,0 +1,40 @@
...
+  <div class="govuk-grid-row">
+    {% call display_field_group('total_space') %}
+    {% endcall %}
+
+    {% call display_field_group('unavailable_space') %}
+    {% endcall %}
+
+    {% call display_field_group('available_space') %}
+    {% endcall %}
+
```
teneightfive, line 36:
```suggestion
```

app/population/views/edit/edit.njk
```diff
@@ -0,0 +1,40 @@
...
+    {% for fieldKey in fields %}
+      {% set field = options.fields[fieldKey] %}
+
+      {% if field.component and not field.skip %}
+        {{ callAsMacro(field.component)(field) }}
+      {% endif %}
+    {% endfor %}
+
+    {{ caller() if caller }}
+
```
teneightfive, line 20:
```suggestion
```

app/population/views/edit/edit.njk
```diff
@@ -0,0 +1,40 @@
+{% extends "form-wizard.njk" %}
+
+{% macro display_field_group(group) %}
```
teneightfive, line 3: Wonder if this should be camel case instead of snake case as we use camel case elsewhere (for example components).

app/population/views/edit/edit.njk
```diff
@@ -0,0 +1,40 @@
...
+    {{ caller() if caller }}
+
+  </div>
+{% endmacro %}
+
+
+{% block fields %}
+
+  <div class="govuk-grid-row">
+    {% call display_field_group('total_space') %}
```
teneightfive, line 28: If there's no data to display inside these you can call them without the `call` I believe:

```nunjucks
{{ display_field_group('total_space') }}
```

app/population/views/edit/edit.njk
teneightfive, line 3:
```suggestion
{% macro renderFieldGroup(group) %}
```

merlinc: Sure - I followed the `nunjucks` styleguide by default, forgetting we had a different standard in our app :)

app/population/views/edit/edit.njk
merlinc: Yeah, that's why I wrote the `if` block around the `content` - I clean this up in the subsequent PR, #1224 - I just wanted to be kinder to my future self and make it easier to keep track of what I was doing.

app/population/views/edit/edit.njk
merlinc: All the various whitespace issues have been tightened up.

PR #1225, opened 2021-01-08
[P4-2595] Extend time component to support relative times

teneightfive:
### What changed

This extends the time component with:

- ability to show relative time and update on the fly
- show time as a tag component

### Why did it change

This time component will be used on the incoming and outgoing dashboards to display the relative time of when a van will be arriving.

### Issue tracking

### As seen on outgoing dashboard

image

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/components/data/_data.scss
+++ b/common/components/data/_data.scss
@@ -27,3 +27,8 @@
     @include govuk-font($size: 80, $weight: bold);
   }
 }
+
+.app-data .govuk-tag {
+  position: relative;
+  top: -2px;
+}
--- a/common/components/data/_data.scss
+++ b/common/components/data/_data.scss
@@ -8,27 +8,39 @@
   @include govuk-font($size: 36, $weight: bold);
   display: block;
   margin: 0;
+
+  .govuk-tag {
+    position: relative;
+    top: -6px;
+  }
 }
 
 .app-data--s {
   .app-data__value {
     @include govuk-font($size: 24, $weight: bold);
+
+    .govuk-tag {
+      top: -2px;
+    }
   }
 }
 
 .app-data--l {
   .app-data__value {
     @include govuk-font($size: 48, $weight: bold);
+
+    .govuk-tag {
+      top: -10px;
+    }
   }
 }
 
 .app-data--xl {
   .app-data__value {
     @include govuk-font($size: 80, $weight: bold);
-  }
-}
 
-.app-data .govuk-tag {
-  position: relative;
-  top: -2px;
+    .govuk-tag {
+      top: -22px;
+    }
+  }
 }
--- a/common/components/time/template.test.js
+++ b/common/components/time/template.test.js
@@ -4,8 +4,17 @@ const {
 } = require('../../../test/unit/component-helpers')
 
 const examples = getExamples('time')
+const mockTimerDate = '2017-08-10'
 
 describe('Time component', function () {
+  beforeEach(function () {
+    this.clock = sinon.useFakeTimers(new Date(mockTimerDate).getTime())
+  })
+
+  afterEach(function () {
+    this.clock.restore()
+  })
+
   context('by default', function () {
     let $component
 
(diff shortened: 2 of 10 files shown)


common/components/time/time.yaml
```diff
@@ -11,6 +11,18 @@ params:
...
+    required: false
+    description: Whether to display the time element within a GOV.UK Tag component
+  - name: relative
+    type: boolean
+    required: false
+    description: Whether to show a relative time
+  - name: imminentOffset
+    type: number
+    required: false
+    description: The offset (in minutes) at which to use a "yellow" tag instead of green. Inside this offset the tag will be yellow, outside of it green, and if the time has past, it will be red.
```
merlinc, line 25: Could we get some examples in the Component library of green / yellow / red styling?

package.json
```diff
@@ -147,6 +147,7 @@
     "slashify": "1.0.0",
     "sticky-sidebar": "3.3.1",
     "stream-mock": "2.0.5",
+    "timeago.js": "4.0.2",
```
merlinc, line 150: > Timeago is a jQuery plugin that makes it easy to support automatically updating fuzzy timestamps

common/components/time/template.test.js
```diff
@@ -60,4 +69,80 @@ describe('Time component', function () {
...
+    beforeEach(function () {
+      const $ = renderComponentHtmlToCheerio('time', {
+        datetime: '2019-01-10',
+        relative: true,
+      })
+
+      $component = $('time')
+    })
+
+    it('should contain a datetime attribute', function () {
```
merlinc, line 85: This isn't dependent on the `relative` property, so doesn't need to be placed in this `context` block.

Or is this an extra check just to make sure things stay together properly?

common/components/time/template.test.js
```diff
@@ -60,4 +69,80 @@ describe('Time component', function () {
...
+      const $datetimeAttr = $component.attr('datetime')
+      expect($datetimeAttr).to.equal('2019-01-10')
+    })
+
+    it('should contain a display relative attribute', function () {
```
merlinc, line 112: `should contain an imminent offset attribute`

common/components/time/template.test.js
```diff
@@ -60,4 +69,80 @@ describe('Time component', function () {
...
+      expect($component.attr('data-imminent-offset')).to.equal('120')
+    })
+  })
+
+  context('shown as a tag', function () {
```
merlinc, line 117: The other two contexts are of the form `with display as tag` - it might be clearer to have all these contexts match.

common/components/time/time.js
```diff
@@ -0,0 +1,74 @@
+const { differenceInMinutes, isPast, parseISO } = require('date-fns')
```
merlinc, line 1: There's a few lines of complicated code here - do we have any easy way to test this at the moment?

common/components/time/time.yaml
teneightfive: Unfortunately not all states...I can show red and green. But yellow is based on an offset and we can't set a dynamic date within the yaml file that will always fall within that offset.

Although the only way might be to set a really high number and a date in the future that we _might_ not reach.

package.json
teneightfive: I think that might be a different package? [`timeago.js`](https://timeago.org/) has no jQuery dependency :)

common/components/time/template.test.js
teneightfive: Yeah, I think I just kept it in there to ensure this attribute is also still there in this instance.

common/components/time/template.test.js
teneightfive: copy/paste fail

common/components/time/time.js
teneightfive: I was struggling a little with it to be honest. I don't think the helpers that load the components also load in the javascript...I'll have a think how we could do it...

common/components/data/_data.scss
```diff
@@ -8,22 +8,39 @@
...
+    }
   }
 }
 
 .app-data--xl {
   .app-data__value {
     @include govuk-font($size: 80, $weight: bold);
+
+    .govuk-tag {
+      top: -22px;
```
Nimphal, line 43: I assume there isn't a nice govuk variable we could be using for this?

common/components/time/time.yaml
Nimphal: Happy with red and green for now, the colour itself is not a super high risk feature.

common/components/time/time.js
```diff
@@ -0,0 +1,74 @@
...
+  },
+
+  onChangeHandler(mutationsList, observer) {
+    for (let index = 0; index < mutationsList.length; index++) {
+      const mutation = mutationsList[index]
+
+      if (mutation.type === 'childList') {
+        const parsedDate = parseISO(this.settings.datetime)
+        const overdue = isPast(parsedDate)
+        const imminent =
```
Nimphal, line 49: Scary variable name is scary!

Nimphal (approved):
Nice!

common/components/data/_data.scss
teneightfive: I don't _think_ so. I added these to correct the position as the label is always a fixed font size, so if they're ever used with other data sizes (we don't currently have a use case ourselves) the position of the label won't look too out of place.

common/components/time/time.js
teneightfive: I struggled with the name of this actually...it's to define the value of when to show the yellow version of the tag...

merlinc (approved):

PR #1226, opened 2021-01-08
[P4-2595] Group moves by vehicle on incoming and outgoing dashboards

teneightfive:
### What changed

This adds the ability to view the incoming and outgoing dashboards to group moves by the van that the people will be getting on.

Previously groupings were done by destination or pickup location.

This change has been adapted so it still supports the existing view by default. This grouping can currently be seen by changing the query string parameter `group_by`. 

Once we are confident there is sufficient data within production we can either enable this by default or allow users to choose their preferred view.

### Why did it change

This change is possible by the supplier's sending the van for each move.

Grouping moves by van allows users within prison reception or police custody suites to better manage their workload. It allows them to know which vans are arriving in which order and who needs to get on them. They can time when they need to prepare people more efficiently.

It also allows them to see when people are coming into their establishment and allows them to start preparing any necessary check-in processes.

### Issue tracking

## Screenshots

[**OUTGOING**](https://booksecuremove-br-moves-group.herokuapp.com/moves/day/2021-01-12/b8dec44a-fd67-4969-a4ff-046e862609c7/outgoing?status=active&group_by=vehicle)
[**INCOMING**](https://booksecuremove-br-moves-group.herokuapp.com/moves/day/2021-01-12/b8dec44a-fd67-4969-a4ff-046e862609c7/incoming?status=active&group_by=vehicle)

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/presenters/moves-by-vehicle.js
+++ b/common/presenters/moves-by-vehicle.js
@@ -1,0 +1,95 @@
+const { isToday, parseISO } = require('date-fns')
+const { isEmpty, get, groupBy, orderBy, sortBy } = require('lodash')
+
+const i18n = require('../../config/i18n')
+const filters = require('../../config/nunjucks/filters')
+const componentService = require('../services/component')
+
+const moveToCardComponent = require('./move-to-card-component')
+
+module.exports = function movesByVehicle({
+  moves,
+  context = 'outgoing',
+  showLocations = false,
+} = {}) {
+  const groups = []
+  const grouped = groupBy(moves, '_vehicleRegistration')
+  const expectedTimeProperty =
+    context === 'outgoing' ? '_expectedCollectionTime' : '_expectedArrivalTime'
+
+  Object.entries(grouped).forEach(([group, groupedItems]) => {
+    const vehicleRegistration = groupedItems[0]._vehicleRegistration
+    const sortedByTime = sortBy(groupedItems, expectedTimeProperty)
+    const expectedTime = get(sortedByTime[0], expectedTimeProperty)
+    const isComplete = groupedItems.every(
+      move => move[context === 'outgoing' ? '_hasLeftCustody' : '_hasArrived']
+    )
+    const showRelativeTime = !isComplete && isToday(parseISO(expectedTime))
+    const items = sortBy(groupedItems, 'profile.person._fullname').map(
+      moveToCardComponent({
+        showToLocation: showLocations || context === 'outgoing',
+        showFromLocation: showLocations || context === 'incoming',
+      })
+    )
+    const timeData = {
+      datetime: expectedTime,
+      text: filters.formatTime(expectedTime),
+    }
+    const timeComponent = componentService.getComponent('appTime', timeData)
+
+    let timeHTML = expectedTime
+      ? timeComponent
+      : i18n.t('collections::labels.no_expected_time')
+
+    if (showRelativeTime) {
+      timeHTML += componentService.getComponent('appTime', {
+        ...timeData,
+        relative: true,
+        displayAsTag: true,
+        imminentOffset: 60,
+      })
+    }
+
+    if (isComplete) {
+      timeHTML += componentService.getComponent('govukTag', {
+        html: i18n.t('collections::labels.complete', {
+          context: context,
+        }),
+        classes: 'govuk-tag--green',
+      })
+    }
+
+    groups.push({
+      expectedTime,
+      items,
+      isComplete,
+      hasVehicle: !isEmpty(vehicleRegistration),
+      vehicleRegistration,
+      header: [
+        {
+          label: i18n.t('collections::labels.vehicle_registration'),
+          value:
+            vehicleRegistration ||
+            i18n.t('collections::labels.awaiting_vehicle'),
+          classes: 'govuk-grid-column-one-quarter',
+        },
+        {
+          label: i18n.t('collections::labels.expected_time', { context }),
+          value: timeHTML,
+          classes: 'govuk-grid-column-one-half',
+        },
+        {
+          label: i18n.t('people'),
+          value: items.length,
+          classes: 'govuk-grid-column-one-quarter',
+        },
+      ],
+    })
+  })
+
+  return orderBy(
+    groups,
+    ['isComplete', 'hasVehicle', 'expectedTime', 'vehicleRegistration'],
+    ['asc', 'desc', 'asc', 'asc']
+  )
+}
--- a/locales/en/collections.json
+++ b/locales/en/collections.json
@@ -49,6 +49,13 @@
   "download_csv_incoming_moves": "Download moves (.csv)",
   "download_csv_single_requests": "Download moves (.csv)",
   "labels": {
+    "vehicle_registration": "Vehicle registration",
+    "awaiting_vehicle": "Awaiting vehicle",
+    "expected_time_incoming": "Expected arrival time",
+    "expected_time_outgoing": "Expected collection time",
+    "no_expected_time": "Not provided",
+    "complete_incoming": "Arrived",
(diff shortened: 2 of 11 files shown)


common/presenters/moves-by-vehicle.js
```diff
@@ -0,0 +1,77 @@
...
+const setMissingToLocation = require('../helpers/move/set-missing-to-location')
+
+const moveToCardComponent = require('./move-to-card-component')
+
+module.exports = function movesByVehicle(data, locationKey = 'to_location') {
+  const locations = []
+  const groupedByLocation = groupBy(data, 'meta.vehicle_registration')
+
+  Object.entries(groupedByLocation).forEach(([group, moves], index) => {
+    const vehicleRegistration = get(moves[0], 'meta.vehicle_registration')
```
Nimphal, line 13: I assume every accepted move has a vehicle registration?

common/presenters/moves-by-vehicle.js
```diff
@@ -0,0 +1,77 @@
...
+      }
+
+      return move
+    })
+
+    const sortedByTime = sortBy(moves, `meta.${expectedTimeMetaKey}`)
+    const expectedTime = get(sortedByTime[0], `meta.${expectedTimeMetaKey}`)
+    const hasLeftCustody =
+      moves.filter(move => ['in_transit', 'completed'].includes(move.status))
+        .length === moves.length
```
Nimphal, line 48: Would something like `every` work here? I think this is checking that every move is in these states. I think this could be:
`moves.every(move => ['in_transit', 'completed'].includes(move.status))` which returns a boolean

common/presenters/moves-by-vehicle.js
```diff
@@ -0,0 +1,77 @@
...
+    const hasArrived =
+      moves.filter(move => ['completed'].includes(move.status)).length ===
+      moves.length
```
Nimphal, line 51: Same as the one above.

locales/en/collections.json
```diff
@@ -49,6 +49,14 @@
...
   "download_csv_single_requests": "Download moves (.csv)",
   "labels": {
+    "vehicle_registration": "Vehicle registration",
+    "awaiting_vehicle": "Awaiting vehicle",
+    "expected_time_incoming": "Expected arrival time",
+    "expected_time_outgoing": "Expected collection time",
+    "no_expected_time": "Not provided",
+    "complete_incoming": "Arrived",
+    "complete_outgoing": "Left custody",
+    "people": "people",
```
Nimphal, line 59: Should this be lowercase? Looking at the screenshots, I think it's upper, but very likely missed something.

teneightfive: I think this might be an old one that is no longer used...

common/helpers/move/set-missing-to-location.js
```diff
@@ -0,0 +1,21 @@
...
+    to_location: {
+      key: `unknown__${move.move_type}`,
+      title: hasOwnLabel.includes(move.move_type)
+        ? i18n.t(`fields::move_type.items.${move.move_type}.label`)
+        : i18n.t('fields::move_type.items.unknown.label'),
+    },
+  }
+}
+
+module.exports = setMissingToLocation
```
merlinc, line 21: Note: This is only a draft, so tests to come

common/presenters/move-to-card-component.js
```diff
@@ -9,21 +9,47 @@ function moveToCardComponent({
...
+      from_location: fromLocation,
+    } = move
     const href = profile ? `/move/${id}${hrefSuffix}` : ''
     const excludedBadgeStatuses = ['cancelled']
 
     showTags = isCompact ? false : showTags
 
+    const moveMetaItems = []
+
+    if (showToLocation) {
```
merlinc, line 32: What should happen if `toLocation` or `toLocation.title` is undefined? Should it show a label with no text, or not show it at all?

common/presenters/moves-by-vehicle.js
merlinc: Does `get` support the `[0]` property as the start of the path?

common/presenters/moves-by-vehicle.js
```diff
@@ -0,0 +1,77 @@
...
+  const groupedByLocation = groupBy(data, 'meta.vehicle_registration')
+
+  Object.entries(groupedByLocation).forEach(([group, moves], index) => {
+    const vehicleRegistration = get(moves[0], 'meta.vehicle_registration')
+    const expectedTimeMetaKey =
+      locationKey === 'to_location'
+        ? 'expected_collection_time'
+        : 'expected_time_of_arrival'
+
+    // TODO: Remove dummy times below
```
merlinc, line 19: Do these sample times need to be removed before this PR can be merged?

common/presenters/moves-by-vehicle.js
```diff
@@ -0,0 +1,77 @@
+const { addMinutes, formatISO, isToday, parseISO } = require('date-fns')
+const { isEmpty, get, groupBy, orderBy, sortBy } = require('lodash')
+
+const setMissingToLocation = require('../helpers/move/set-missing-to-location')
+
+const moveToCardComponent = require('./move-to-card-component')
+
+module.exports = function movesByVehicle(data, locationKey = 'to_location') {
```
merlinc, line 8: Note: This is only a draft, so tests to come

common/presenters/moves-by-vehicle.js
teneightfive: I think it could be used in either instance in this case. But what this is doing is looking for the property path `meta.vehicle_registration` within the first move.

This is based on the assumption that because these moves have already been grouped by this property, that they all contain the same registration. So within this block I'm assigning the value from the first move to this variable.

teneightfive: >I assume every accepted move has a vehicle registration?

@Nimphal not necessarily. However, that's where the fallback comes in. So this line is just taking whatever the value is and assigning it to a variable.

However, thinking about it, I could use `group` which should provide the same value.

common/helpers/move/set-missing-to-location.js
teneightfive: Yep, the review was more for feedback on the approach so if anything needed majorly refactoring I didn't have to spend lots of time changing all the tests too :)

common/presenters/move-to-card-component.js
teneightfive: In those scenarios, it won't show the meta. 

The [`profileToCardComponent`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/presenters/profile-to-card-component.js#L54) will remove any meta that doesn't have a value:

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/2e1fd309d03239e7b30cd14c1fa2bfd21621acb0/common/presenters/profile-to-card-component.js#L54

However, there is also [a helper](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/2e1fd309d03239e7b30cd14c1fa2bfd21621acb0/common/helpers/move/set-missing-to-location.js#L3) to populate moves with the correct to location if it doesn't exist - for example Prison recall moves.

common/presenters/moves-by-vehicle.js
teneightfive: Yep :) I left the `TODO` in there to pick that up.

You can ignore that block as part of the initial review. This PR is mostly still draft but I wanted to get early eyes in case there was any glaring things that need addressing.

Nimphal (approved):

merlinc (approved):

PR #1234, opened 2021-01-12
[P4-2595] Add support for new API meta data and handle missing to locations

teneightfive:
### What changed

This set of changes extends the move transformer to add properties for a vehicle or expected times for a move.

It also moves the helper to set missing to locations for a move to this part of the codebase so whenever a move is accessed elsewhere it will always have a to location.

### Why did it change

This refactor is needed to support grouping moves by vehicle and being able to show expected times.

### Issue tracking

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


Nimphal (approved):
Nice!

common/lib/api-client/transformers/move.transformer.js
```diff
@@ -1,6 +1,38 @@
...
+
+const setExpectedTime = data => {
+  data._expectedCollectionTime =
+    get(data, 'meta.expected_collection_time') || undefined
+  data._expectedArrivalTime =
+    get(data, 'meta.expected_time_of_arrival') || undefined
+}
+
+const setExtraStates = data => {
+  data._hasLeftCustody = ['in_transit', 'completed'].includes(data.status)
```
merlinc, line 32: You've separated out `hasOwnLabel` into it's own variable - but you haven't here. Any specific reasoning for that?

merlinc: And, depending upon the answer to the above question, these status here also don't really need to be scoped inside the function, as they're groups of statuses, which are independent of these functions.

i.e. are the `hasLeftCustody` statuses always going to be these two?

teneightfive: >You've separated out hasOwnLabel into it's own variable - but you haven't here. Any specific reasoning for that?

Not quite sure what you mean by this?

The reason for creating these properties on each move is to keep them within one layer when we transform a resource from the API layer.

These properties can then be used to do sorting/grouping/filtering without needing to define the set of statuses that make up this "status" in multiple places within the app.

merlinc: Ok, so we have 3 arrays of statuses here:
` const hasOwnLabel = ['prison_recall', 'video_remand']`
`  data._hasLeftCustody = ['in_transit', 'completed'].includes(data.status)`
` data._hasArrived = ['completed'].includes(data.status)`

More clearly my questions were:
1. Why is hasOwnLabel separated out into it's own property, and the other two are inlined?
2. Are these arrays of statuses really only relevant inside each function closure or does the concept of "hasArrived" always have those statuses, in which case, should these values be living in config somewhere?

teneightfive: >Why is hasOwnLabel separated out into it's own property, and the other two are inlined?

Ah, gotcha. In short, different developer style/time of writing 😅 the `hasOwnLabel` is actually taken from a previous implementation within the `moves-by-location.js` presenter.

Happy to make these more consistent in this implementation though :) 

>Are these arrays of statuses really only relevant inside each function closure or does the concept of "hasArrived" always have those statuses, in which case, should these values be living in config somewhere?

I would say these "statuses" would always be based on those API statuses. The reason for moving them here is that anytime anything needs to do something with a move based on this value is can know it is consistent rather than having to repeat the status check. If the API introduced another status in the future we would only need to update it here once.

As for config...I don't mind either way. I didn't want to overload the config and this is an isolated enough place to define that property. It's also similar to what we already do with the assessment transformer and person transformer.

merlinc (approved):

common/lib/api-client/transformers/move.transformer.js
merlinc: Cool :)

I'm not opposed to anything in the PR, just pointing out this area of note - I've approved the PR as there's no reason to hold it up for these tweaks

merlinc (approved):

PR #1236, opened 2021-01-13
chore: Expand error logging detail in Sentry

teneightfive:
### What changed

This change improves the way 422 errors are passed to Sentry.

This change improves the way that the error is sent with the exception
so that more detail, like the source of meta of an error can be seen.

### Why did it change

Previously errors were sent as additional data, however, this meant that
only the first two levels of properties could be seen.

## Screenshots

| Before | After |
| ------ | ----- |
|  | <kbd>image</kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/controllers/form-wizard/index.js
+++ b/common/controllers/form-wizard/index.js
@@ -179,7 +179,20 @@ class BaseController extends Controller {
 
     if (err.statusCode === 422) {
       Sentry.withScope(scope => {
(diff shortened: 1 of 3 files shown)


Nimphal (approved):
Yes please!

merlinc (approved):

PR #1242, opened 2021-01-13
Catch error when redirecting moves without a current location

teneightfive:
### What changed

This uses optional chaining so if the parent object doesn't exist the value will return `undefined` rather
than throw a "Cannot find property" error.

### Why did it change

The test for this catch was using the wrong permission name which meant
that it didn't catch the scenario where the current location didn't
exist and the app would through a "Cannot read property" error.

### Issue tracking

https://sentry.io/organizations/ministryofjustice/issues/2149703392/?project=2014813&query=is%3Aunresolved

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/moves/middleware/redirect-base-url.js
+++ b/app/moves/middleware/redirect-base-url.js
@@ -6,7 +6,7 @@ function redirectBaseUrl(req, res) {
 
   if (
     canAccess('moves:view:proposed') &&
-    currentLocation.location_type === 'prison'
+    currentLocation?.location_type === 'prison'
   ) {
     return res.redirect(`${baseUrl}/requested`)
   }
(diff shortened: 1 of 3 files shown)


app/moves/middleware/redirect-base-url.js
```diff
@@ -6,7 +6,7 @@ function redirectBaseUrl(req, res) {
 
   if (
     canAccess('moves:view:proposed') &&
-    currentLocation.location_type === 'prison'
+    currentLocation?.location_type === 'prison'
```
Nimphal, line 9: is there no e2e test for this? If there is it will fail until we get babel 7 going with testcafe :(

Nimphal (approved):

app/moves/middleware/redirect-base-url.js
teneightfive: This file isn't included by the e2e tests so should be fine to use optional chaining. It's only when we include files directly in the e2es, like the services that create resources for test scenarios that we get caught out.

This middleware is only used within the app so should be fine :)

PR #1259, opened 2021-01-19
Revert back to old way of deploying staging temporarily

Nimphal:
### What changed

Move back to old day of releasing to unblock the flow of work as the two setups are incompatible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,7 +344,7 @@ workflows:
           name: helm_lint
           env: 'staging'
       # Docker builds
-      - hmpps/build_docker:
+      - build-docker-image:
           <<: *only_master
           name: build-docker-image-master
       - build-docker-image:
@@ -354,11 +354,10 @@ workflows:
           preprod: 'true'
           production: 'true'
       # Deployment
-      - hmpps/deploy_env:
+      - deploy:
           <<: *only_master
           name: deploy-staging
           env: 'staging'
-          retrieve_secrets: 'none'
           requires:
             - build-docker-image-master
             - end-to-end-test-master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -347,6 +347,7 @@ workflows:
       - build-docker-image:
           <<: *only_master
           name: build-docker-image-master
+          staging: 'true'
       - build-docker-image:
           <<: *only_deploy_tags
           name: build-docker-image-release


.circleci/config.yml
```diff
@@ -344,7 +344,7 @@ workflows:
           name: helm_lint
           env: 'staging'
       # Docker builds
-      - hmpps/build_docker:
+      - build-docker-image:
           <<: *only_master
           name: build-docker-image-master
```
teneightfive, line 349: Do we also need to add back in the `staging: 'true'` line that was removed by [this commit](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/aa62571d3e07e6f5a22eb9c1518a61f376396cf6#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L347)?

teneightfive (changes requested):

.circleci/config.yml
Nimphal: Yes, great catch! Added back!

.circleci/config.yml
```diff
@@ -344,21 +344,21 @@ workflows:
...
           name: build-docker-image-release
           uat: 'true'
           preprod: 'true'
           production: 'true'
       # Deployment
-      - hmpps/deploy_env:
+      - deploy:
           <<: *only_master
           name: deploy-staging
           env: 'staging'
```
teneightfive, line 361: I think this (`env: 'staging'`) was also added as part of the change to `hmpps/deploy_env` right? But it shouldn't effect switching back to `deploy`, right?

teneightfive (approved):

.circleci/config.yml
Nimphal: it looks to have been there since before, this is the diff of the first commit I did on this work https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/44dfb92fe482565f84ded4eb3d1f8334c06542e7#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47

teneightfive: Ahhh. Yep, I think I confused it with this change: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/commit/0e39af15191b9d049e830061ecf631abb47c4bfe#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47

PR #1265, opened 2021-01-22
[P4-2580] Populate on add numbers

merlinc:
## Proposed changes

When adding new numbers, pre-populate the form with the previously entered numbers, together with the unlock and discharges from NOMIS. 

### What changed

- Loading of population data into the session has been extracted out into a hidden step
- Pre-populating population data is another hidden step
- Switching logic has been added into the `steps` config
- Hint text will be dynamically added if the data is pre-populated

### Issue tracking

- [P4-2580](https://dsdmoj.atlassian.net/browse/P4-2580)
- [P4-2577](https://dsdmoj.atlassian.net/browse/P4-2577)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/services/population.js
```diff
@@ -51,6 +51,25 @@ class PopulationService extends BaseService {
...
+
+    if (!location) {
+      return Promise.reject(new Error(noLocationIdMessage))
+    }
+
+    if (!date) {
+      return Promise.reject(new Error(noDateMessage))
+    }
+
+    return this.apiClient
```
merlinc, line 65: The endpoint to populate new data isn't JSON api compliant - see #1269 for further details.

In order to fetch the data, I've needed to drop down into the base `request` method.

app/population/controllers/edit/details.js
```diff
@@ -80,5 +73,19 @@ class DetailsController extends FormWizardController {
 
     next()
   }
+
+  setHintText(req, res, next) {
+    if (!req.population) {
```
merlinc, line 78: If there is no population, then this is "Add numbers" - so we'll need to dynamically change the hint text.

The alternative would be to set the hint text by default, and delete it when in "Update numbers" mode.

app/population/controllers/edit/details.test.js
```diff
@@ -225,6 +196,39 @@ describe('Population controllers', function () {
       })
     })
 
+    describe('setBreadcrumbs', function () {
```
merlinc, line 199: Added tests for the breadcrumbs, as that was missed out of a previous PR

app/population/steps/edit.js
```diff
@@ -7,10 +11,29 @@ module.exports = {
     resetJourney: true,
     skip: true,
     forwardQuery: true,
+    next: [
+      {
+        fn: req => {
```
merlinc, line 16: I'd usually prefer to run branching logic off variables, but we don't have any form variables to work with, which is the point of the next two steps.

common/services/population.js
teneightfive: What was the reason for just a bespoke endpoint? I had a look at [1269](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1269) on the API but couldn't see anything around making this a normal REST endpoint.

Also, [the documentation](https://hmpps-book-secure-move-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk/api-docs/index.html#operations-Populations-get_populations_new) suggests that it does contain a JSON:API payload?

app/population/controllers/edit/details.js
```diff
@@ -9,14 +9,7 @@ class DetailsController extends FormWizardController {
     this.use(this.setBreadcrumbs)
     this.use(this.setCancelUrl)
     this.use(this.setPageTitle)
-  }
-
-  setInitialValues(req, res, next) {
```
teneightfive, line 14: So are the initial values no longer added? Would this not effect the "edit" state?

merlinc: This has been extracted out into separate controllers now.

On entry to the form, we check if the `req.population` exists. If it does, then we call `LoadPopulationController` and use the population data, if it doesn't we call `PopulateController` which invokes the populate method to get data - both of these set the returned population onto the session - this is then used by the `DetailsController` to display.

common/services/population.js
merlinc: The problem I encountered was because I wasn't specifying an ID, `JSON` api **helpfully** pluralised the word for me so made a request to `/population/news` not `/population/new`

teneightfive: Ah, I see, it was the endpoint and not the payload. Devour does this based on the global setting for [pluralisation](https://github.com/twg/devour#initializer) as the expectation is that endpoints are consistently named.

You do however, have two other options in this instance which might be neater:

1. In the model options set `pluralize` to `false`
2. In the model options use a custom `collectionPath` with the singular form (`/population/new`). We do a similar thing for the [reference data models](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/c35022a6b2924599878028a1e8740039c6cfbf69/common/lib/api-client/models/gender.js#L11)

app/population/controllers/edit/details.js
teneightfive: Yes, I did see that later on. Okay, if it works then it's all good. I was a little confused at first.

common/services/population.js
merlinc: Ok, I'll give that a try!

I do need a combination of pluralised (`/populations/` / `/populations/{id}`) and non-plurarlised (`/populations/new`) endpoints so hopefully some thing better than what I've got will work :)

teneightfive: Ah, I see what you mean. In fact, maybe what you have is actually easiest then.

teneightfive (approved):

common/lib/api-client/models/populate.js
```diff
@@ -0,0 +1,9 @@
+const population = require('./population')
+module.exports = {
```
teneightfive, line 2: If this also works then it's fine, but I think this method actually feels more confusing when looking at it. 

Can we guarantee that we always want to include all the fields from the population model? I just wonder if changing that model in the future would have any side-effects on this now that we're extending it.

common/services/population.js
```diff
@@ -51,6 +51,25 @@ class PopulationService extends BaseService {
...
+      .findAll('populate', {
```
teneightfive, line 66: I think on reflection, the previous implementation was a little easier to understand what was going on.

However, if you don't want to spend any time reverting it, I understand.

teneightfive (approved):

PR #1273, opened 2021-01-26
Revert pagination

teneightfive:
### What changed

This reverts the commits that started using the `POST` endpoints for filtering moves and allocations in the API.

### Why did it change

These new endpoints have caused a major increase in the number of `Bad gateway` errors being received which seem linked to `POST` requests. This reverts back to using the existing `GET` methods for returning a list of moves.


common/services/base.test.js
```diff
@@ -1,77 +0,0 @@
-const proxyquire = require('proxyquire')
```
merlinc, line 1: Removing this keeps the revert commit simple, but I'd vote for this to be restored in the meantime, whatever the eventual resolution is.

merlinc (approved):

common/services/base.test.js
teneightfive: Not sure what you mean?

PR #1296, opened 2021-02-02
refactor: Remove unused app param middleware

merlinc:
## Proposed changes

Remove unused app param middleware.

### Why did it change

This function is not actually doing anything.

https://expressjs.com/en/4x/api.html#app.param
> Param callback functions are local to the router on which they are
> defined. They are not inherited by mounted apps or routers. Hence, param
> callbacks defined on app will be triggered only by route parameters
> defined on app routes.

We do not define our routes on the main app, therefore this means that this
middleware is never used, as params are never configured at this level.

A similar function will need to be applied at each level of router that
requires it

This was tested by adding a `throw new Error('App Param Check')` and running the E2E tests. There were no failures, and no entries in the log files, so this function is not being used.

### Issue tracking

- [P4-XXXX]()

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/middleware/set-locations.js
+++ b/common/middleware/set-locations.js
@@ -8,16 +8,6 @@ const pipe = functions => data => {
   }, data)
 }
 
-const getParamsLocation = data => {
-  const { locationId } = data.req.params
-
-  if (locationId) {
-    data.locations = [{ id: locationId }]
-  }
-
-  return data
-}
-
 const getCurrentLocation = data => {
   const currentLocation = data.req.session?.currentLocation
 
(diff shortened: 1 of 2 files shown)


common/middleware/set-locations.js
```diff
@@ -8,16 +8,6 @@ const pipe = functions => data => {
   }, data)
 }
 
-const getParamsLocation = data => {
```
teneightfive, line 11: I feel like there are some tests associated with this that might need cleaning up?

But maybe it doesn't cater for the way express uses this functionality.

merlinc: Yeah, I forgot to remove the tests for this, in this branch.

It's definitely not used anywhere, as I said, I put a `throw new Error` and ran the tests, and clicked around - didn't see anything.

The only way this could work, is if we set an `app` route of `.*/location/REGEX/.*` and then had every other router inherit params. The addition of location is important, as that would allow us to target the regex specifically - otherwise `allocation` or `population` ids would match as well.

teneightfive (approved):

PR #1298, opened 2021-02-02
[P4-1892] Digital handover

teneightfive:
### What changed

This set of changes will amend how a Person Escort Record (PER) is confirmed and handed over. 

It allows the PER to be printed sooner and but still uses the same logic behind the scenes to prevent editing the resource once it has been confirmed (or handed over).

During confirmation more data is now collected about which parties completed the handover, and when. This data is stored against the `person_escort_record` in the API.

### Why did it change

Currently users must "confirm" the PER before being able to print the record. This means that if any changes occur after this point they cannot make any amendments.

With these changes they will be able to print the record sooner and record the handover separately at a later point in time.

The record will also no longer be editable once the supplier starts the move (moving it to `in_transit`).

### Issue tracking

- [P4-1892](https://dsdmoj.atlassian.net/browse/P4-1892)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/move/controllers/timeline.js
```diff
@@ -1,8 +1,8 @@
 const moveHelpers = require('../../../common/helpers/move')
 const presenters = require('../../../common/presenters')
 
-module.exports = async function view(req, res) {
-  const { move } = req
+module.exports = function view(req, res) {
+  const { move = {} } = req
```
merlinc, line 5: Are there corresponding tests that need updating due to these changes?

app/person-escort-record/app/confirm/handover-details.njk
```diff
@@ -0,0 +1,56 @@
+{% extends "form-wizard.njk" %}
+
+{% macro renderField(name) %}
+  {{ callAsMacro(options.fields[name].component)(options.fields[name]) }}
```
merlinc, line 4: I think this is at least our 2nd or 3rd instance of this function (there's something similar in population) - so next time we need this functionality, we should try to cleanup any tech debt

app/person-escort-record/controllers/print-record.js
```diff
@@ -127,6 +138,7 @@ function printRecord(req, res) {
     personalDetailsSummary,
     personEscortRecordSections,
     personEscortRecordTagList,
+    timestamp,
```
merlinc, line 141: This function could do with some unit tests

app/person-escort-record/app/confirm/controllers.js
```diff
@@ -0,0 +1,97 @@
...
+        const otherTime = new Date(2020, 1, 1, timeParts[0], timeParts[1])
+        const otherDate = parseISO(values.handover_other_date)
+        const newDate = formatISO(otherDate, {
+          representation: 'date',
+        })
+        const newTime = formatISO(otherTime, {
+          representation: 'time',
+        })
+
+        values.handover_occurred_at = `${newDate}T${newTime}`
```
merlinc, line 63: This parsing of the time into a separate date object, which is then rendered along side the original date is a little confusing.

Perhaps it could be cleaned up using [date-fns/set](https://date-fns.org/v2.17.0/docs/set)?

```set(otherDate, {hours: timeParts[0], seconds: timeParts[1]})``` would give you a date object with the time in it

merlinc (commented):
It's looking ok so far, nothing is really standing out for me as an issue.

Obviously you're still working on the unit tests for the new code.

teneightfive:
This has been replaced by #1319 and #1321

PR #1301, opened 2021-02-03
fix: Prevent loop when accessing the timeline

teneightfive:
### What changed

**This is a temporary fix to remove the lookup for missing resources
from the frontend.**

It replaces the population loop with Sentry logging for missing resources within the events structure.

### Why did it change

Currently the API doesn't return the full "eventable" resource within
the `included` property of the response payload. This means that the
frontend API client can't deserialize this resource.

Code was implemented to do this extra loading from the frontend and
to loop over each eventable to try and get that resource and deserialize
it.

This has caused some unexpected infinite looping in production so for
now this population is being removed and replaced with a Sentry warning
so that we can get an idea of scale of missing resources.

It might be better to implement this retrieval directly on the API too
so any major changes are being postponed until a decision on that.

This may lead to some missing title/context for some events, but we are logging those to
get a better understanding of scale and which type of resources are frequently missing.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/move/middleware.js
+++ b/app/move/middleware.js
@@ -21,7 +21,7 @@ module.exports = {
 
     try {
       const move = await req.services.move.getByIdWithEvents(moveIdWithEvents)
-      await populateResources(move.timeline_events, req)
+      populateResources(move.timeline_events, req)
       req.move = move
       next()
     } catch (error) {
--- a/common/lib/populate-resources.js
+++ b/common/lib/populate-resources.js
@@ -1,37 +1,32 @@
+const Sentry = require('@sentry/node')
+const { pick } = require('lodash')
+
 const findUnpopulatedResources = require('./find-unpopulated-resources')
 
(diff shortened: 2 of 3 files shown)


app/move/middleware.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
 
     try {
       const move = await req.services.move.getByIdWithEvents(moveIdWithEvents)
-      await populateResources(move.timeline_events, req)
```
merlinc, line 24: `populateResources` is no longer asynchronous :+1:

teneightfive: Nope. It may well have to be in the future but for now I've removed the extra API calls. They appear to be what was breaking production. I think we need to collect some data about instances where the loops never ends before we re-implement this code.

Nimphal (approved):

PR #1328, opened 2021-02-12
Contract delivery manager view

Nimphal:
### What changed

Adds check if the user is a contract delivery manager and if so, loads all available locations.

### Why did it change

Contract delivery managers need an overview of the everything that's happening.
### Issue tracking

- [P4-2639](https://dsdmoj.atlassian.net/browse/P4-2639)

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -24,6 +24,21 @@ async function locations(req, res, next) {
     req.session.user.locations = userLocations
   }
 
+  if (userPermissions.includes('contract_delivery_manager')) {
+    const suppliers = await req.services.referenceData.getSuppliers()
+    const supplierLocations = await Promise.all(
+      suppliers.map(async supplier => {
+        return await req.services.referenceData.getLocationsBySupplierId(
+          supplier.id
+        )
+      })
+    )
+
+    userLocations = supplierLocations.flat()
+
+    req.session.user.locations = userLocations
+  }
+
   const locations = sortBy(userLocations, 'title')
 
   res.render('locations/views/locations.njk', {
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -24,7 +24,7 @@ async function locations(req, res, next) {
     req.session.user.locations = userLocations
   }
 
-  if (userPermissions.includes('contract_delivery_manager')) {
+  if (userPermissions.includes('locations:contract_delivery_manager')) {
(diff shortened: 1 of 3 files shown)


common/lib/permissions.js
```diff
@@ -25,6 +25,18 @@ const policePermissions = [
   'person_escort_record:confirm',
 ]
 
+const contractDeliveryManagerPermissions = [
+  'allocations:view',
+  'contract_delivery_manager',
```
merlinc, line 30: The pattern we've got for permissions is `area:permission` or `area:sub:permission` - would something like `locations:contract_delivery_manager` make any sense here? Just seeing `contract_delivery_manager` on its own looks weird compared to the other permissions

app/locations/controllers.js
```diff
@@ -24,6 +24,21 @@ async function locations(req, res, next) {
     req.session.user.locations = userLocations
   }
 
+  if (userPermissions.includes('contract_delivery_manager')) {
```
merlinc, line 27: If the user has a supplier ID and the `contract_delivery_manager` does that make the call above uneeded?

Nimphal: They won't have supplier Ids, those come from the group (so Serco PECS or Geoamey PECS), and CDMs won't have that group. I don't know if it's worth complicating the per supplier call with a check for CDMs, there are very few CDMs and the chance of them being accidentally added to a supplier group is even lower.

common/lib/permissions.js
```diff
@@ -25,6 +25,18 @@ const policePermissions = [
   'person_escort_record:confirm',
 ]
 
+const contractDeliveryManagerPermissions = [
+  'allocations:view',
```
teneightfive, line 29: Do we want to give them `dashboard:view` access too?

teneightfive (approved):

common/lib/permissions.js
Nimphal: The ticket says just moves and stuff for now, and dashboard later, and I suspect we will be tweaking these permissions when people actually have a play anyway. I think I'll merge it like this, have product have a play and then update with whatever is missing.

PR #1338, opened 2021-02-15
Remove environment variable dependency for mock auth server

teneightfive:
### What changed

This change removes the extra bit of configuration and simplifies the
implementation of the test suite.

### Why did it change

Currently the mock auth server relies on environment variables being
set to define the username and password.

This is a legacy implementation from when we used to hit the real
auth service to run end-to-end tests.

Since mocking has been introduced there is no reason to keep these
values in environment variables anymore.

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


config/index.js
```diff
@@ -189,42 +189,70 @@ module.exports = {
...
-      },
+    USERS: {
       COURT: {
-        username: process.env.E2E_COURT_USERNAME,
-        password: process.env.E2E_COURT_PASSWORD,
+        locations: ['ABDRCT', 'BDFRCT'],
+        name: 'End-to-end Court',
+        authorities: ['ROLE_PECS_COURT'],
+        username: 'E2E_COURT',
+        password: 'ZZDBm6Bf',
```
merlinc, line 198: If we are steering into have completely different local passwords to the ones in environments, can we make them memorable? This way when we're running locally we won't have to scroll to line `194`-`255` of a `config` file to figure out the correct password.

mocks/auth-server.js
```diff
@@ -124,15 +70,10 @@ app.get('/auth/logout', (req, res) => {
   res.redirect(req.query.redirect_uri)
 })
 
-const getUserFromToken = req => {
```
merlinc, line 127: Function hasn't been deleted, Github has just misplaced the move. :+1:

mocks/auth-server.js
```diff
@@ -3,76 +3,22 @@
...
 
 const { decodeAccessToken } = require('../common/lib/access-token')
 const {
-  E2E: { ROLES },
+  E2E: { USERS },
 } = require('../config')
 
-const roleConfig = {
-  COURT: {
-    locations: ['ABDRCT', 'BDFRCT'],
```
merlinc, line 14: We're already pulling in config from the main app (see line `9`) so there's not much difference in where this config lives.

I'm not trilled about having test config in the main app config, but I'm not seeing a cleaner way of dealing with this at the moment.

teneightfive: Yeah, good point. I did think of moving this just to the auth server but some of the config is pulled in in various places in the e2e tests. So config still seemed the most logic place...

config/index.js
teneightfive: Yep, not against that. Could just use a simple format seeing as it's only for mock test purposes anyway.

merlinc (approved):

PR #1372, opened 2021-02-24
fix: De-duplicate list of locations for Contract Delivery Managers

Nimphal:
### What changed
The list of locations shown to Contract Delivery Managers now only contains unique locations.

### Why did it change
Sme locations have changed suppliers in the past, and for historic reasons appear in the list of locations for both suppliers. When the lists are merged for the Contract Delivery Managers, these locations appeared twice.

### Issue tracking

- [P4-2639](https://dsdmoj.atlassian.net/browse/P4-2639)

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -34,7 +34,7 @@ async function locations(req, res, next) {
       })
     )
 
-    userLocations = supplierLocations.flat()
+    userLocations = [...new Set(supplierLocations.flat())]
 
     req.session.user.locations = userLocations
   }
--- a/app/locations/controllers.test.js
+++ b/app/locations/controllers.test.js
@@ -34,6 +34,10 @@ const mockSecondSupplierLocation = [
     id: '0987654-bc17-4ea1-bae3-4ef429f9081e',
     title: 'b location',
   },
+  {
+    id: '70923762-bc17-4ea1-bae3-4ef429f9081e',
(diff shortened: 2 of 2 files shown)


app/locations/controllers.js
```diff
@@ -34,7 +34,7 @@ async function locations(req, res, next) {
       })
     )
 
-    userLocations = supplierLocations.flat()
+    userLocations = [...new Set(supplierLocations.flat())]
```
merlinc, line 37: https://caniuse.com/mdn-javascript_builtins_set

Nice, it's covered by all the browsers I think we're supposed to support.

However, I just noticed this - https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/.browserslistrc#L5

Our browser list config says we support last 2 versions of all browsers, and all versions of IE < 11 - which seems odd.

merlinc (approved):

app/locations/controllers.js
Nimphal: yeah, we do, there's some people using edge in IE7 mode or something scary like that. However, this is on the server side, so I don't think the browser cares. It is, apparently, the most vanilla way to do it, so we should safe either way.

PR #1373, opened 2021-02-24
feat: Add journey map as template include

merlinc:
## Proposed changes

Add map as a template include. [Openlayer library](openlayers.org/) included via CDN. Both of these decisions are to avoid unnecessarily increasing page size for all users given a very small use case. To implement it as a component, we will need to introduce code splitting, so that every page doesn't need to have this JavaScript included.

Currently viewable as a demo by using #1406

### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| | Screenshot_2021-03-04 Move for Person to be added – Book a secure move(1) |
|  | Screenshot_2021-03-04 Move for Person to be added – Book a secure move |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/templates/includes/map.njk
```diff
@@ -0,0 +1,220 @@
...
+  <div id="map" class="map" class="govuk-grid-column-full" style="height: 600px"></div>
+  <div id="message" class="govuk-grid-column-full govuk-heading-l" style="position: absolute; top: 100px; height: 200px; text-align: center"></div>
+</div>
+
+{#  <script src="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.js" integrity="sha512-O7kHS9ooekX8EveiC94z9xSvD/4xt10Qigl6uEKvspYykdux3Ci5QNu5fwi4ca0ZkZI/oCgx5ja8RklWUEqzxQ==" crossorigin="anonymous"></script>#}
+{#  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.css" integrity="sha512-D2c/nmSVlwoZk29SLwJK8GkLyCyPuNbZ/vZQL/lNzw+o+yXzsbnJt7lK/iCfdBmgyLRDi0qx27gpE8Zb1C6I7g==" crossorigin="anonymous" />#}
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol-debug.css" integrity="sha512-hBSieZLd5rse9gdkfv4n0pDU4D04SxpqBtwDzRy/QiXRBhczDyfCTDTnHCada73ubNqiQv6BLgCRXHAJPUwC5w==" crossorigin="anonymous" />
+<script src="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol-debug.js" integrity="sha512-RGRzp3Gfp9l+vkKXNjm45LxXEujXIZBIur2l2CRqIYDiZ+QLe2hprjK7YZBKo1RrfbF/pBvg2431gHU7pjjJ1A==" crossorigin="anonymous"></script>
```
Nimphal, line 9: Do we know if the people using this will be on some kind of restricted network? Like Quantum, etc? I am unsure if Quantum will allow a connection to cloudflare.

Nimphal (approved):

common/templates/includes/map.njk
merlinc: Urgh, that's a good point - I'd forgotten about restricted networks.

As I said in the description, I've done it this way so that we don't have to immediately implement Webpack [Code Splitting](https://webpack.js.org/guides/code-splitting/) and [Lazy Loading](https://webpack.js.org/guides/lazy-loading/) as that can be a bit of a mine field for development.

@martin-ballhatchet-moj Thoughts on the above?

martin-ballhatchet-moj: Hi. So - I'm fine with Cloudflare for loading static assets like this. I would hope that Quantum doesn't block Cloudflare domains. It would be good to know what the user would see if that is blocked though - to help troubleshooting later.
I'm a little worried that we may deploy this to "production" with the debug versions of the CSS and JS - would be good to have a way to prevent that (even if it's a JIRA ticket to switch them over!)

merlinc: @martin-ballhatchet-moj re: `debug` - That's a good point - I'll flip around the prod/debug includes for this PR. 

I'll still keep the debug lines, but commented out, as it's not immediately obvious how to get those URLS - we should only be needing debug on local machines anyway.

docs/environment-variables.md
```diff
@@ -41,6 +41,14 @@
 | ENABLE_COMPONENTS_LIBRARY           | Whether to enable the component library and allow it to be browsed on this environment                                                                                                                                           | false                                                          |
 | ENABLE_DEVELOPMENT_TOOLS            | Whether to enable the development tools and allow it to be browsed on this environment                                                                                                                                           | false                                                          |
 
+## External Services
```
teneightfive, line 44: Wonder whether this needs a separate heading? You could say that the API and the Auth service are also "external services". Wonder if it's easier just to leave them in the main table for consistency.

app/map/controllers.js
```diff
@@ -0,0 +1,36 @@
...
+      },
+      url: authUrl,
+      auth: {
+        username: apiKey,
+        password: apiSecret,
+      },
+      data: qs.stringify({ grant_type: 'client_credentials' }),
+    })
+    .then(response => {
+      res.type('json')
```
teneightfive, line 23: Do we need `.type('json')` as well as `.json()`?

These can also be chained:

```javascript
res
  .status(200)
  .json(response.data)
```

common/templates/includes/map.njk
```diff
@@ -0,0 +1,219 @@
+<div class="govuk-grid-column-full" style="height: 600px; position: relative">
+  <div id="map" class="map" class="govuk-grid-column-full" style="height: 600px"></div>
+  <div id="message" class="govuk-grid-column-full govuk-heading-l" style="position: absolute; top: 100px; height: 200px; text-align: center"></div>
+</div>
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.js" integrity="sha512-O7kHS9ooekX8EveiC94z9xSvD/4xt10Qigl6uEKvspYykdux3Ci5QNu5fwi4ca0ZkZI/oCgx5ja8RklWUEqzxQ==" crossorigin="anonymous"></script>
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.css" integrity="sha512-D2c/nmSVlwoZk29SLwJK8GkLyCyPuNbZ/vZQL/lNzw+o+yXzsbnJt7lK/iCfdBmgyLRDi0qx27gpE8Zb1C6I7g==" crossorigin="anonymous" />
+{#<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol-debug.css" integrity="sha512-hBSieZLd5rse9gdkfv4n0pDU4D04SxpqBtwDzRy/QiXRBhczDyfCTDTnHCada73ubNqiQv6BLgCRXHAJPUwC5w==" crossorigin="anonymous" />#}
```
teneightfive, line 8: Rogue comments?

teneightfive (approved):

common/templates/includes/map.njk
merlinc: These are not rogue comments - as mentioned in https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1373#discussion_r587460110 I've left the debug URLs commented out for now, as they are not obvious to determine and will probably be useful in the future.

If we decide we want some more dynamic way of using them, we could build something off a query string in the future.

docs/environment-variables.md
merlinc: tbh I was wondering about breaking out the other API keys into their own sections too. Having `LOCATIONS_BATCH_SIZE`, `SESSION_NAME`, `API_SECRET` and `ORDNANCE_MAP_API_URL` in the same section really did seem like a very very general grouping (which I know is the heading)

app/map/controllers.js
merlinc: Ah, I didn't notice this in the documentation:
> This method sends a response (**with the correct content-type**) 

Reason why they weren't chained was to make the tests easier, otherwise I'd have to return a thing that returned a spy, and that was much more fragile than having two separate spies that were checked.

docs/environment-variables.md
merlinc: I'll make that reordering a separate cleanup PR :)

PR #1391, opened 2021-03-01
fix: Improve missing person ID error when creating a move

teneightfive:
### What changed

This change adds some extra logic around to cater for this scenario.

It also caters for a scenario where a person is selected using the
person lookup, but then the user decides to create a new person instead.

In the instance this does still occur we will now be capturing the move ID for the move
that was created and also the expected profile ID.

### Why did it change

We have seen a number of errors in Sentry where a profile can't
be updated because it is "missing a Person ID".

Some investigation allowed this error to be re-created when going back
to the personal details page and making a change during the move creation
journey.

### Issue tracking

- https://sentry.io/organizations/ministryofjustice/issues/2179979853

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/move/app/new/controllers/personal-details.test.js
+++ b/app/move/app/new/controllers/personal-details.test.js
@@ -1,12 +1,7 @@
-const FormController = require('hmpo-form-wizard').Controller
-
 const fieldHelpers = require('../../../../../common/helpers/field')
 const referenceDataHelpers = require('../../../../../common/helpers/reference-data')
-const personService = {
-  create: sinon.stub(),
-  update: sinon.stub().resolves(),
-}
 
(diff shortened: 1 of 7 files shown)


app/move/app/new/controllers/personal-details.test.js
```diff
@@ -1,4 +1,4 @@
-const FormController = require('hmpo-form-wizard').Controller
+// const CreateBaseController = require('hmpo-form-wizard').Controller
```
Nimphal, line 1: Maybe not needed anymore?

Nimphal (approved):

PR #1399, opened 2021-03-02
Allow multiple moves and allocation to be created at the same time

teneightfive:
### What changed

This set of changes adds support to create moves and allocations within the same session using multiple tabs.

It uses the existing unique form wizard to create a new wizard for each journey. This prevents journeys being reset when a new one is opened in a new tab.

### Why did it change

The Form Wizard that is used to power a lot of the form journeys was original designed to only support one journey and session at a time.

However, as our system is a repeat use system we have started to see errors where the journey is missing pre-requisite steps.

The assumption based on the data from Sentry is that users are using multiple tabs to try and complete multiple moves at the same time. This behaviour makes sense when you think of how people want to use this system.

### Issue tracking

- https://sentry.io/organizations/ministryofjustice/issues/2080838686

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/server.js
+++ b/server.js
@@ -193,6 +193,12 @@ app.use(
           'www.googletagmanager.com',
           'www.google-analytics.com',
         ],
+        connectSrc: [
+          "'self'",
+          "'unsafe-inline'",
+          'www.googletagmanager.com',
+          'www.google-analytics.com',
+        ],
         imgSrc: ["'self'", 'www.google-analytics.com'],
         fontSrc: ["'self'", 'fonts.googleapis.com'],
       },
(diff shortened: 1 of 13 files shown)


Nimphal (approved):

server.js
```diff
@@ -193,6 +193,12 @@ app.use(
           'www.googletagmanager.com',
           'www.google-analytics.com',
         ],
+        connectSrc: [
```
merlinc, line 196: Does this block need to be added for this PR to work correctly?

teneightfive: It was an error I'd noticed and made sense to bring it in to the GA changes. But in theory, no...

However, I weighed this up vs more PRs and more e2e tests, etc...

merlinc (approved):

PR #1498, opened 2021-04-13
chore: Upgrade webpack to version 5

Nimphal:
### What changed
Upgrade to webpack 5. Remove node as target.

### Why did it change
The upgrade enables us to stay up to date on security issues. However, webpack 5 does not currently support both web and node as targets.

#### Manual testing

- [x] Chrome
- [x] Firefox
- [x] Edge
Screenshot 2021-04-15 at 13 58 31

- [x] Internet Explorer
Screenshot 2021-04-15 at 13 59 48


.browserslistrc
```diff
@@ -1,5 +1,4 @@
 last 2 version
 > 2%
-maintained node versions
```
Nimphal, line 3: Webpack 5 does not currently support multiple target types, so it's either web or node. I was not sure why we had node in here so removed as a test and it worked. If we do need node then we can't update to 5 until they've resolved this problem. More info here https://webpack.js.org/blog/2020-10-10-webpack-5-release/#improved-target-option

merlinc (commented):
I can't see anything obvious here, so I think this is good to merge.

merlinc (approved):

PR #1546, opened 2021-04-21
chore: Remove second query string library

teneightfive:
### What changed

We currently use two libraries for working with querystrings:

- [qs](https://www.npmjs.com/package/qs)
- [query-string](https://www.npmjs.com/package/query-string)

This change removes the use of `query-string` which was used on fewer
occasions and doesn't currently support parsing of objects which we
use on some occasions.

### Why did it change

Having two dependencies doing very similar things seemed unnecessary. The benefit of moving to one will be easier maintenance moving forward too.

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/filters/views/filters.njk
+++ b/app/filters/views/filters.njk
@@ -17,10 +17,8 @@
 
   <form method="POST">
     <input type="hidden" name="referrer" value="{{filters.referrer.url}}">
-    {% for name, values in filters.referrer.values %}
-      {% for value in values %}
-        <input type="hidden" name="{{name}}" value="{{value}}">
-      {% endfor %}
+    {% for name, value in filters.referrer.values %}
+      <input type="hidden" name="{{name}}" value="{{value}}">
     {% endfor %}
     {% for key, filterComponent in filters.components %}
       {% if filterComponent.component and not filterComponent.skip %}
--- a/common/controllers/filters.js
+++ b/common/controllers/filters.js
@@ -1,3 +1,5 @@
+const { omit } = require('lodash')
+
 const getDefaultValues = fields => {
   return Object.keys(fields).reduce((values, key) => {
     if (fields[key].defaultValue !== undefined) {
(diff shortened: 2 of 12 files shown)


app/filters/views/filters.njk
```diff
@@ -17,10 +17,8 @@
 
   <form method="POST">
     <input type="hidden" name="referrer" value="{{filters.referrer.url}}">
-    {% for name, values in filters.referrer.values %}
-      {% for value in values %}
-        <input type="hidden" name="{{name}}" value="{{value}}">
-      {% endfor %}
+    {% for name, value in filters.referrer.values %}
+      <input type="hidden" name="{{name}}" value="{{value}}">
```
Nimphal, line 21: Nice!

common/presenters/filters-to-categories-list-component.js
```diff
@@ -1,5 +1,11 @@
+const queryString = require('qs')
+
 const i18n = require('../../config/i18n')
-const request = require('../lib/request')
+
+function getUrl(page, args) {
+  const stringified = queryString.stringify(args, { arrayFormat: 'comma' })
+  return `${page}${stringified ? `?${stringified}` : ''}`
+}
```
Nimphal, line 8: I am slightly confused by this - I can't see this function being deleted from the `request` lib, but it looks like it's been moved to here? Was that because it wasn't used anywhere else?

common/controllers/filters.js
```diff
@@ -22,17 +24,9 @@ const removeDefaultFilterValues = (fields, values) => {
...
-  })
-  Object.keys(referrerValues).forEach(key => {
-    if (!Array.isArray(referrerValues[key])) {
-      referrerValues[key] = [referrerValues[key]]
-    }
-  })
+  const omittedKeys = Object.keys(fields)
 
-  return referrerValues
+  return omit(values, omittedKeys)
```
Nimphal, line 29: Also very nice!

common/presenters/filters-to-categories-list-component.js
teneightfive: So, the method from the request lib is still being used elsewhere using the default settings so I didn't want to change that. However, in this file, switching to `qs` changed the default behaviour for handling values that are an array so needs to use `arrayFormat: 'comma'` to behave how it previously did.

I didn't want to get into re-writing this module too much with this change so I opted to having a private method to this module with that particular setting. 

Although whilst writing this, I wonder if the `getUrl` could support that option being passed in instead? What do you think?

Nimphal: ah, that makes sense! I don't mind it being here and not using the request one, it's fine if it needs custom behaviour. And since this is the only place where the custom behaviour is used, I'd say there's limited value in refactoring the original method. Cheers! 👍

PR #1556, opened 2021-04-23
chore: Run app in strict mode

### LINKED ISSUE #32: Minilog Octal literal error when running node in strict mode
## Problem

Using the `--use-strict` flag to run `node start.js` in this app causes an error to be thrown by one of the sub-dependencies. The error traces back to [this issue](https://github.com/mixu/minilog/pull/43) within the [minilog](https://github.com/mixu/minilog) library, a dependency of the [devour](https://github.com/twg/devour) library used to format API requests/responses.

## Temporary solution

The temporary fix is to remove `--use-strict` from the start command.

## Permanent solution

Long term we need to wait for the fix to be applied to the minilog library and then update the version to include the fix.

teneightfive:
### What changed

An issue was created to re-enable strict mode when running the app.

This change re-enabes [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). There does seem to be a current [issue](https://github.com/nodejs/node/issues/30039) with Node v14 and strict mode but this change will support the use of it once the bug is fixed.

Closes #32

### Issue tracking

- #32 

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/package.json
+++ b/package.json
@@ -246,7 +246,7 @@
     "release:github": "conventional-github-releaser -p angular",
     "release:tags": "git push --follow-tags origin HEAD:main",
     "security_audit": "npx audit-ci --config audit-ci.json",
-    "start": "node start.js",
+    "start": "node --use-strict start.js",
     "test": "NODE_ENV=test LOG_LEVEL=silent mocha './app/**/*.test.js' './common/**/*.test.js' './config/**/*.test.js'",
     "test-e2e": "node test/e2e/parallel-runner.js",
     "test-e2e:browserstack:ie": "testcafe 'browserstack:IE@11.0:Windows 10' test/e2e/*.test.js",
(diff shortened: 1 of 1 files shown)


teneightfive:
@Nimphal on second thoughts, I'm going to close this as the [minilog](https://github.com/mixu/minilog) library still hasn't been updated or removed from devour. 

So if strict mode gets fixed the app will error again and could create a little minefield to debug. What are your thoughts on closing this and just leaving without strict mode?

Nimphal:
I agree with closing this down. The gains from strict mode right now are marginal compared to the risks of errors which seem quite likely.

teneightfive:
Closing this based on the comments above.

Closes #32

PR #1558, opened 2021-04-23
feat: Align header with HMPPS Digital Services

Thource:
### What changed

The styling of the header has changed to match the HMPPS Digital Services header.

### Why did it change

This changed because it's nicer if everything matches style, also because there was no way to navigate back to HMPPS Digital Services from BaSM.

### Issue tracking

- [P4-2626](https://dsdmoj.atlassian.net/browse/P4-2626)

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
-$govuk-header-background: govuk-colour("black");
+$govuk-header-background: govuk-colour('black');
```
teneightfive, line 1: Could we revert these quote changes please. I suspect it might be an IDE setting you have. Although we don't have some form of scss linting applied yet the standard approach up until now has been double quotes so it would be good to stick to that and reduce the noise in this PR.

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
 $govuk-header-nav-item-border-color: #2e3133;
 
 .app-header {
-  @include govuk-font($size: 16);
-
-  color: $govuk-header-text;
-  background: $govuk-header-background;
-}
-
-.app-header__container--full-width {
```
teneightfive, line 17: Are we removing support for this modifier with these changes? Might be nice to see if we can still support it.

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
 .app-header__link {
-  text-decoration: none;
+  @include govuk-font($size: 19);
 
   &:link,
-  &:visited {
-    color: $govuk-header-link;
+  &:visited,
+  &:active {
+    color: #fff;
```
teneightfive, line 93: Do we need this change? The value of `$govuk-header-link` defined on line 5 is already white. I think it would be nice to continue to support the configuration approach.

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
-  font-size: 30px; // We don't have a mixin that produces 30px font size
-  line-height: 1;
-
-  @include mq ($from: desktop) {
-    display: inline-block;
-  }
-
-  &:link,
-  &:visited {
+    outline: 3px solid transparent;
```
teneightfive, line 102: All of these definitions are actually being applied by using `@include govuk-focused-text;` so maybe let's revert back to that so that if that changes in the future we'll benefit from it.

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
-  }
-}
-
-.app-header__content {
-  @include mq ($from: desktop) {
-    width: 50%;
-    padding-left: $govuk-gutter-half;
-    float: left;
+    svg {
+      fill: $govuk-header-background;
```
teneightfive, line 109: Could we use this (from line 43):
```suggestion
      fill: currentColor;
```

This will mean that it will benefit from whatever the color defined is rather than having to be set separately.

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
-    &:visited {
-      color: govuk-colour("black");
-    }
+.app-header__link--service-name {
+  height: 100%;
+  align-items: center;
+  display: flex;
+  margin-right: govuk-spacing(4);
+  padding-right: govuk-spacing(4);
+  border-right: 1px solid #fff;
```
teneightfive, line 175: Are we able to clean up the use of hardcoded colours from these definitions. Maybe we can use one of the variables defined at the start of this file?

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
 
-    // Do not append link href to GOV.UK link when printing (e.g. '(/)')
```
teneightfive, line 292: These statements might be worth keeping so that the links for "sign out" aren't printed for example

common/components/internal-header/_internal-header.scss
```diff
@@ -1,311 +1,182 @@
...
-      display: none;
-    }
+  &:last-child {
+    margin-right: 0;
+    border-right: 0;
+    padding-right: 0;
   }
 }
-
-// Begin adjustments for font baseline offset
```
teneightfive, line 299: I think it would also be good to keep these at the bottom of the file as overrides. They're things currently required to work with the font we have and these allow those fixes to stand out quite nicely.

common/components/internal-header/template.njk
```diff
@@ -1,9 +1,9 @@
 <header class="app-header {{ params.classes if params.classes }}" role="banner" data-module="header"
         {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-  <div class="app-header__container {{ params.containerClasses | default('app-width-container') }}">
+  <div class="app-header__container {{ params.containerClasses if params.containerClasses }}">
```
teneightfive, line 3: You can shorten this to:
```suggestion
  <div class="app-header__container {{ params.containerClasses }}">
```

If it has no value nunjucks will not output anything.

common/components/internal-header/template.njk
```diff
@@ -1,9 +1,9 @@
...
 
     <div class="app-header__logo">
-      <a href="{{ params.homepageUrl | default('/') }}" class="app-header__link app-header__link--homepage">
+      <a href="{{ AUTH_BASE_URL }}/auth" class="app-header__link app-header__organisation-name">
```
teneightfive, line 6: I'm a bit conscious of having this variables within the component as now they are tied specifically to the app configuration and are no longer independent. 

I think a nice approach would be to update the value of `homepageUrl` to `AUTH_BASE_URL + "/auth"` in [base.njk](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/d3b9972af9a0c322de782586255d9a57c05adc06/common/templates/layouts/base.njk#L96) where we define the header. Then the value is tied specifically to our implementation of this component and when testing the component you don't need to worry about setting global nunjucks values (with I think might be causing some of the unit test failures).

common/components/internal-header/template.njk
```diff
@@ -54,14 +49,18 @@
...
-    <button type="button" role="button" class="app-header__menu-button js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
     <nav>
       <ul id="navigation" class="app-header__navigation {{ params.navigationClasses if params.navigationClasses }}" aria-label="Top Level Navigation">
         {% for item in params.navigation %}
           {% if item.href and item.text %}
             <li class="app-header__navigation-item{{ ' app-header__navigation-item--active' if item.active }}">
               <a class="app-header__link" href="{{ item.href }}"{% for attribute, value in item.attributes %} {{attribute}}="{{value}}"{% endfor %}>
                 {{ item.text }}
+                {% if item.subtext %}
+                  <span class="app-header__link__sub-text">
```
teneightfive, line 60: I think this should be:
```suggestion
                  <span class="app-header__link-sub-text">
```

to better fit with the [BEM](http://getbem.com/) naming convention.

common/components/internal-header/template.test.js
```diff
@@ -1,3 +1,4 @@
+const { AUTH_BASE_URL } = require('../../../config/index.js')
```
teneightfive, line 1: I've left a comment on the template file as to how we can remove the need to include our app configuration from this component. I think it would be a much cleaner separation to keep the two separate.

config/nunjucks/globals.js
```diff
@@ -15,6 +15,7 @@ const {
   routes: toolsRoutes,
 } = require('../../app/tools')
 const i18n = require('../i18n')
+const { AUTH_BASE_URL } = require('../index.js')
```
teneightfive, line 18: This file is already included on line 3 so you can add a definition to that instead of including it again.
```suggestion
```

test/e2e/#smoke.test.js
```diff
@@ -53,10 +58,12 @@ users.forEach(user => {
...
   })(`As ${user.name}`, async t => {
     await t
-      .expect(page.nodes.appHeader.exists)
+      .expect(page.nodes.appHeaderOrganisation.exists)
+      .ok()
+      .expect(page.nodes.appHeaderProduct.exists)
       .ok()
       .expect(page.nodes.username.innerText)
-      .eql(user.username)
+      .eql(`${user.displayName}\nManage your details`)
```
teneightfive, line 66: This could be changed to `.contains` so you don't need to worry about it including the manage link.

teneightfive (changes requested):
I've left quite a few comments around the big changes to the styles and layout. Let me know if it's easier to go through them together.

I've also looked at the changes using different breakpoints and the mobile view seems to be missing some padding that I think we should address:

<kbd>image</kbd>

compared to:

<kbd>image</kbd>

I've also got a slight concern that with the changes to smaller screen widths there is now no way to access the homepage for our service as we've removed the service name. Previously it stacked the service name and I wonder if we should continue that:

<kbd>image</kbd>

@elizabethgoddard what do you think?

Another wider thing @elizabethgoddard and I talked about it making the container size increase with this same change. Currently this would make the header now wider than our content which we felt looked a little odd:

<kbd>image</kbd>

So instead, we could look to override the value for `$govuk-page-width` and `$moj-page-width` in our [application.scss](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/d3b9972af9a0c322de782586255d9a57c05adc06/common/assets/scss/application.scss#L6) file. Which will produce something more like this:

<kbd>image</kbd>

common/components/internal-header/_internal-header.scss
Thource: From what I could see, it wasn't used anywhere in the project, the styles that it added are included in `.app-header__container` now (except for border-color, which had no effect outside of `container--with-border`, which also wasn't used anywhere)

common/components/internal-header/_internal-header.scss
Thource: I knew I'd have missed one, I copied the styles from hmpps-auth and converted them to scss

common/components/internal-header/_internal-header.scss
Thource: Hmm, the whole header is being hidden during print now (same behaviour as HMPPS DS), do we want the header to be visible?

common/components/internal-header/_internal-header.scss
Thource: I believe I may have fixed the issue that required these styles to be necessary? I changed the SVG path and attrs to be the same as HMPPS DS and from what I can see, the text in the header and the crest are lined up perfectly? (same as HMPPS DS)

ElizabethGoddard:
Hey @Thource and @teneightfive, sorry for the delay was just chatting to Dom and also separately with DPS.

> I've also looked at the changes using different breakpoints and the mobile view seems to be missing some padding that I think we should address

Yes please to fixing the padding :)

> I've also got a slight concern that with the changes to smaller screen widths there is now no way to access the homepage for our service as we've removed the service name. Previously it stacked the service name and I wonder if we should continue that

Had a chat with Dom - I'm not too concerned because we have access via 'Home' in the primary navigation, however police users do not currently. We will probably give police users the home dashboard at some point for consistency though. Hold this thought because I'm just checking with DPS what their other services do before we decide - they are going to get back to me.

> Another wider thing @ElizabethGoddard and I talked about it making the container size increase with this same change. Currently this would make the header now wider than our content which we felt looked a little odd

@Thource, Dom mentioned you looked at DPS and they have a 960 container. Actually this is only on the landing page (they tell me this is owned by someone else but they are looking to update it to the wider grid). If you click into DPS or other services you'll see it actually is the wider grid (1170px). If we could update to match that would be preferable because that landing page will eventually match and we will then also match DPS and associated services. Also like Dom says it looks more balanced with the header and contents in alignment.

Thanks guys!

common/components/internal-header/_internal-header.scss
teneightfive: Ah right, I see. I think the intention of that definition is for `.js` files though as that's also a rule we use for JavaScript linting.

Currently almost all of our `.scss` files use double quotes. I might look to clarify this with either adding an extension override to the `.prettierrc` or adding style linting. 

But if we could revert it as part of this change for now that would keep it consistent with the rest of the files of the same type.

common/components/internal-header/template.njk
Thource: Replaced `AUTH_BASE_URL + "/auth"` with `organisationHomepageUrl`, homepageUrl is already in use for the BaSM homepage

.prettierrc
```diff
@@ -2,5 +2,13 @@
   "trailingComma": "es5",
   "semi": false,
   "singleQuote": true,
-  "arrowParens": "avoid"
+  "arrowParens": "avoid",
+  "overrides": [
+    {
+      "files": "*.scss",
```
teneightfive, line 8: 👍

common/components/internal-header/_internal-header.scss
```diff
@@ -2,310 +2,179 @@ $govuk-header-background: govuk-colour("black");
...
     @include govuk-focused-text;
-  }
 
-  // alphagov/govuk_template includes a specific a:link:focus selector
-  // designed to make unvisited links a slightly darker blue when focussed, so
-  // we need to override the text colour for that combination of selectors.
-  @include govuk-compatibility(govuk_template) {
-    &:link:focus {
-      @include govuk-text-colour;
+    svg {
```
teneightfive, line 106: This looks like its nested within the `@include govuk-compatibility`, is that GitHub diffs playing up or intentional?

I'm guessing we might not need it nested if so?

common/components/internal-header/_internal-header.scss
teneightfive: Yes, that's something we're doing at an app level but I was thinking more at a component level for these styles. If its ever extracted and put into a service where the header isn't removed I think it would be useful to keep these styles to prevent the links being shown.

But maybe that's extra page weight...I'll leave that as your call

common/components/internal-header/_internal-header.scss
teneightfive: Yes, that's true but it was more the approach if settings the adjustments at the end of the page. I think the same minus top position has just been moved inside that statement instead of noted as an adjustment.

It feels like we've strayed enough to maybe not warrant needing them defined like this so fine to leave it. The adjustments were to cater for not using GDS Transport (I think).

test/e2e/#smoke.test.js
```diff
@@ -13,24 +13,28 @@ const users = [
     name: 'Police user',
     role: policeUser,
     username: 'End-to-end Police',
```
teneightfive, line 15: Do we still need both `username` and `displayName`? I feel like it was only used for the smoke tests that you've changed to use `displayName`?

teneightfive (approved):
Couple minor comments but looks good to me 👍

common/components/internal-header/_internal-header.scss
Thource: That's just GitHub diffs playing up, this is nested as `.app-header__link { &:focus { svg {`

PR #1568, opened 2021-04-27
Fix missing catch

Nimphal:
### What changed
The locations and move controllers were updated to wrap calls to the API in try catch blocks.

### Why did it change
If the API returned an error, we weren't handling it. It would be logged, but it wouldn't exit the req cycle and the users could end up seeing partial data.

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -4,39 +4,37 @@ async function locations(req, res, next) {
   const userPermissions = get(req.session, 'user.permissions', [])
 
   let regions = []
-
-  if (userPermissions.includes('allocation:create')) {
-    try {
-      regions = await req.services.referenceData.getRegions()
-    } catch (error) {
-      return next(error)
-    }
-  }
-
   let userLocations = req.userLocations
-
   const supplierId = req.session.user.supplierId
 
-  if (supplierId) {
-    userLocations = await req.services.referenceData.getLocationsBySupplierId(
-      supplierId
-    )
-  }
+  try {
+    if (userPermissions.includes('allocation:create')) {
+      regions = await req.services.referenceData.getRegions()
+    }
 
-  if (userPermissions.includes('locations:contract_delivery_manager')) {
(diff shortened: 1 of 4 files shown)


app/locations/controllers.js
```diff
@@ -4,39 +4,37 @@ async function locations(req, res, next) {
...
+
+      // The locations have been uniqued based on title to prevent
+      // duplicates when multiple suppliers have the same location
+      userLocations = uniqBy(supplierLocations.flat(), 'id')
+    }
+  } catch (error) {
+    return next(error)
   }
 
   req.session.user.locations = userLocations
```
teneightfive, line 40: It won't let me comment on the line but lines 42 to 46 were the ones removed by my PR so be good to just double check they don't creep back in with this. Will likely be resolved with a rebase :)

app/locations/controllers.test.js
```diff
@@ -214,5 +204,25 @@ describe('Locations controllers', function () {
...
       })
     })
+
+    context('when there is an API error', function () {
+      it('should fail gracefully for regions API', async function () {
+        req.session.user.permissions = ['allocation:create']
+        req.services.referenceData.getRegions = sinon.fake.returns(
+          Promise.reject(new Error())
+        )
+        await proxiedController.locations(req, res, nextSpy)
```
teneightfive, line 214: I also made a change to remove the need to use proxyquire here as we weren't mocking any dependencies of this module so it was unnecessary. You can change this to the following once rebase:
```suggestion
        await locationsController(req, res, nextSpy)
```

teneightfive (approved):
Changes look good. Might just want to double check a few once rebase on top of #1567

app/locations/controllers.js
Nimphal: Git was clever enough to not make a fuss about this, so all is as it should be :)

Issue #1062, opened 2020-11-30
Incorrect handling of dropped network connections by api client


solidgoldpig:
We have occasionally seen the BE report 499 (Client closed upstream connection) which appears to correlate with the FE send timeout responses to the end user.

(For context: https://talent-intl.slack.com/archives/G016LJDSRS4/p1606489817036200)

Running locally

- Ask staging for a lot of moves  eg.  moveService.getAll()
- Turn off internet

The api-client waits for the timeout period to be reached and then throws an error 

```
Uncaught Error: timeout of 30000ms exceeded
ECONNABORTED
```

NB. although `ECONNABORTED` looks like a valid error, it appears that `Axios` always sets that when it times out
https://github.com/axios/axios/issues/1543

With a "pure" axios request, it still takes some time for the droppage to be spotted (ie. to the default timeout), but the resulting error is  `ENETDOWN`

In any case we shouldn’t be getting a timeout error, we should be getting a network error (regardless of whether this is fixed or not - https://github.com/axios/axios/issues/1296) and `ENETDOWN` or `ERR_INTERNET_DISCONNECTED` in the best of all possible worlds

I presume that we have done something in the middleware that has faffed things up in this case… but nothing screams out as being an obvious candidate. Also not sure that axios will help us fail faster... `Got` also doesn't react to the dropped connection (and only then if am explicit timeout value it set)

Maybe we should be reducing the api timeout now that the BE is more performant. The only endpoint that would probably still need a safer higher value would be the csv download.

PR #1569, opened 2021-04-27
Replace axios with Got for API client requests

teneightfive:
### What changed

This replaces the underlying HTTP request library used by the API
client, currently axios, for Got.

### Why did it change

Got has some different support which may help us resolve some timeout
handling issues we see when making API requests.

### Issue tracking

- [P4-2709](https://dsdmoj.atlassian.net/browse/P4-2709)
- #1062 

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


common/lib/api-client/middleware/response.js
```diff
@@ -0,0 +1,48 @@
+const { isArray } = require('lodash')
+
+function needsDeserialization(method) {
+  return ['GET', 'PATCH', 'POST'].indexOf(method.toUpperCase()) !== -1
```
Nimphal, line 4: Could we use [].includes here? So it would be `['GET', 'PATCH', 'POST'].includes(method.toUpperCase())`

common/lib/api-client/middleware/response.js
```diff
@@ -0,0 +1,48 @@
...
+
+module.exports = {
+  name: 'app-response',
+  res: function (payload) {
+    /*
+     * Note: This has been overridden from the devour library to support the
+     * use of `Got` instead of `axios` to make requests
+     */
+    const { jsonApi, res, req } = payload
+    const { data, errors, meta, links, included } = res.body || {}
```
Nimphal, line 15: I assume it's this line that means we need to override the response middleware, replacing data with body?

teneightfive: Yep. At this stage there isn't a huge deal more, however, this will be extended once we start looking at moving pagination inside the API client. I have a [wip branch](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/got-pagination) that shows that implementation.

common/lib/api-client/index.js
```diff
@@ -1,38 +1,56 @@
...
+  }
+
   instance.replaceMiddleware('POST', post(FILE_UPLOADS.MAX_FILE_SIZE))
   instance.replaceMiddleware(
     'axios-request',
-    request({
+    requestMiddleware({
       cacheExpiry: API.CACHE_EXPIRY,
       useRedisCache: API.USE_REDIS_CACHE,
+      timeout: API.TIMEOUT,
```
Nimphal, line 48: if the request middleware isn't got-request, would this timeout still be needed/work as expected?

common/lib/api-client/middleware/get-cache.js
```diff
@@ -13,12 +13,14 @@ function getCacheMiddleware({ useRedisCache = false } = {}) {
       }
 
       debug('SEARCHING', cacheKey)
-      const data = await cache.get(cacheKey, useRedisCache)
+      const body = await cache.get(cacheKey, useRedisCache)
 
-      if (data) {
-        debug('RETURNING', cacheKey, data)
+      if (body) {
```
Nimphal, line 18: Is this also different between got and axios?

common/lib/api-client/middleware/post.js
```diff
@@ -25,6 +18,10 @@ module.exports = function post(maxFileSize) {
         return payload
       }
 
+      const originalMiddleware = find(jsonApi._originalMiddleware, {
+        name: 'POST',
+      })
+
```
Nimphal, line 24: Unsure how this change is related to got and if it is, does it need a feature flag?

teneightfive: It's _technically_ not but I made a change to this file originally and tidied up these parts whilst I was there. So seeing as I was still working on that part of the codebase I left these changes.

The early return is related to caching. If a cached response is set the idea was to bypass the middleware. However, the `POST` middleware is never called as we never (and probably should never?) cache a `POST` response.

The other change just moves the find to only when it is needed. I figured less unnecessary computation would be better :)

common/lib/api-client/index.js
teneightfive: Good spot. It's not strictly *needed* by the original request middleware, however, because the options are passed in an object and only the required options are destructured it doesn't have any adverse effects to pass in this extra option to both.

I could have wrapped the two calls in an `if/else` statement but that would also require doing that statement every time we instantiate an instance of the API client. This felt the lesser of the two evils and as (I'm hoping) its only temporary I left it like this.

common/lib/api-client/middleware/get-cache.js
teneightfive: The difference between the two sits on line `20` when we return the response structure.

Got does expect a property called `body` whereas axios would expect `data`. So my thought was to return both properties on the object that is then used in subsequent middleware. This prevents us from having to (and remember to) clear the cache when we toggle the environment variable.

Once we're happy its working okay we can remove support for both and clear the cache to ensure things are all reset but I thought this would be the safest for now.

PR #1584, opened 2021-04-30
feat: Add prison/PNC number on in/out moves

Thource:
### What changed

The incoming/outgoing move screens now display prison/PNC number (dependant on move to location type) on the move cards.

### Why did it change

This changed because prison users want to know the prison numbers of individuals coming into and going out of establishments quickly and easily so that they can be prepared for their departure/arrival.

### Issue tracking

- [P4-2799](https://dsdmoj.atlassian.net/browse/P4-2799)

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -4,12 +4,13 @@ const presenters = require('../../../common/presenters')
 
 function viewAllocation(req, res) {
   const { allocation, canAccess } = req
-  const { moves, status, totalSlots } = allocation
+  const { moves, status, totalSlots, from_location: fromLocation } = allocation
   const bannerStatuses = ['cancelled']
   const movesWithoutProfile = moves.filter(move => !move.profile)
   const movesWithProfile = moves.filter(move => move.profile)
   const moveToCardComponent = presenters.moveToCardComponent({
     showStatus: false,
+    locationType: fromLocation.location_type,
   })
 
   const removeUnassignedMoves = move => {
--- a/app/move/app/new/controllers/person-search-results.js
+++ b/app/move/app/new/controllers/person-search-results.js
@@ -34,9 +34,13 @@ class PersonSearchResultsController extends PersonController {
   }
 
   setPeopleItems(req, res, next) {
+    const { from_location_type: locationType } = req.sessionModel.toJSON()
     const { people } = req.form.options.fields
     people.items = req.people.map(person => {
-      const card = presenters.profileToCardComponent({ showTags: false })({
+      const card = presenters.profileToCardComponent({
+        locationType,
+        showTags: false,
+      })({
         person,
       })
       return {
--- a/app/moves/middleware/set-results.moves.js
+++ b/app/moves/middleware/set-results.moves.js
@@ -19,11 +19,13 @@ function setResultsMoves(bodyKey, locationKey) {
           moves: activeMoves,
           context: locationKey === 'from_location' ? 'incoming' : 'outgoing',
           showLocations: !req.location,
+          locationType: req.location.location_type,
         })
       } else {
         req.resultsAsCards = presenters.movesByLocation(
           activeMoves,
-          locationKey
+          locationKey,
+          req.location.location_type
         )
       }
 
--- a/common/components/card/template.njk
+++ b/common/components/card/template.njk
@@ -42,7 +42,7 @@
           {% if item.label and (item.html or item.text) %}
             <div class="app-card__meta-list-item">
               <dt class="app-card__meta-list-key{% if item.hideLabel %} govuk-visually-hidden{% endif %}">
-                {{ item.label }}
+                {{ item.label.html | safe if item.label.html else item.label.text }}
               </dt>
               <dd class="app-card__meta-list-value">
                 {{ item.html | safe if item.html else item.text }}
--- a/common/presenters/move-to-card-component.js
+++ b/common/presenters/move-to-card-component.js
@@ -12,6 +12,7 @@ function moveToCardComponent({
   showToLocation = false,
   showFromLocation = false,
   hrefSuffix = '',
+  locationType,
 } = {}) {
   return function item(move) {
     const {
@@ -31,14 +32,14 @@ function moveToCardComponent({
 
     if (showFromLocation) {
       moveMetaItems.push({
-        label: i18n.t('fields::from_location.label'),
+        label: { text: i18n.t('fields::from_location.label') },
         text: fromLocation?.title,
       })
     }
 
     if (showToLocation) {
       moveMetaItems.push({
-        label: i18n.t('fields::to_location.label'),
+        label: { text: i18n.t('fields::to_location.label') },
         text: toLocation?.title,
       })
     }
@@ -49,6 +50,7 @@ function moveToCardComponent({
       ? { text: i18n.t(`statuses::${status}`) }
       : undefined
     const personCardComponent = profileToCardComponent({
+      locationType,
       meta: moveMetaItems,
       showImage: isCompact ? false : showImage,
       showMeta: isCompact ? false : showMeta,
--- a/common/presenters/move-to-card-component.test.js
+++ b/common/presenters/move-to-card-component.test.js
@@ -9,6 +9,8 @@ const mockMove = {
   profile: {
     person: {
       _fullname: 'Name, Full',
+      prison_number: 'A1234BC',
+      police_national_computer: 'A1BC2D/E3',
     },
   },
   to_location: {
@@ -60,6 +62,7 @@ describe('Presenters', function () {
               href: '/move/12345',
             })
             expect(profileToCardComponentStub).to.be.calledWithExactly({
+              locationType: undefined,
               meta: [],
               showImage: true,
               showMeta: true,
@@ -130,6 +133,7 @@ describe('Presenters', function () {
           href: '/move/12345/path/to/somewhere',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: true,
           showMeta: true,
@@ -151,6 +155,7 @@ describe('Presenters', function () {
           href: '/move/12345',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: false,
           showMeta: true,
@@ -172,6 +177,7 @@ describe('Presenters', function () {
           href: '/move/12345',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: true,
           showMeta: false,
@@ -193,6 +199,7 @@ describe('Presenters', function () {
           href: '/move/12345',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: true,
           showMeta: true,
@@ -214,6 +221,7 @@ describe('Presenters', function () {
           href: '/move/12345',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: true,
           showMeta: true,
@@ -239,6 +247,7 @@ describe('Presenters', function () {
           href: '/move/12345',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: false,
           showMeta: false,
@@ -274,6 +283,7 @@ describe('Presenters', function () {
           href: '/move/12345',
         })
         expect(profileToCardComponentStub).to.be.calledWithExactly({
+          locationType: undefined,
           meta: [],
           showImage: false,
           showMeta: false,
@@ -395,9 +405,10 @@ describe('Presenters', function () {
             href: '/move/12345',
           })
           expect(profileToCardComponentStub).to.be.calledWithExactly({
+            locationType: undefined,
             meta: [
               {
-                label: '__translated__',
+                label: { text: '__translated__' },
                 text: mockMove.to_location.title,
               },
             ],
@@ -422,9 +433,10 @@ describe('Presenters', function () {
             href: '/move/12345',
           })
           expect(profileToCardComponentStub).to.be.calledWithExactly({
+            locationType: undefined,
             meta: [
               {
-                label: '__translated__',
+                label: { text: '__translated__' },
                 text: mockMove.from_location.title,
               },
             ],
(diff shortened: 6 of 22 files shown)


common/presenters/move-to-card-component.js
```diff
@@ -29,6 +30,20 @@ function moveToCardComponent({
 
     const moveMetaItems = []
 
+    if (showIdentifierNumber && profile && profile.person && move.to_location) {
```
teneightfive, line 33: I'm wondering if we need the `move.to_location` check here? What's the scenario we're trying to prevent with this check?

teneightfive: Could also shorten `profile && profile.person` to `profile?.person`.
```suggestion
    if (showIdentifierNumber && profile?.person && move.to_location) {
```

common/presenters/move-to-card-component.js
```diff
@@ -29,6 +30,20 @@ function moveToCardComponent({
...
+      if (move.to_location.location_type === 'prison') {
```
teneightfive, line 34: Should we ensure that we have a value for prison number here? Otherwise we could end up with just the label.

Thinking about it, it might be neater to remove the `profile` checks from the outer statement as check for the value here:
```suggestion
      if (move.to_location.location_type === 'prison' && profile?.person?.prison_number) {
```

But double check with @elizabethgoddard what she'd prefer to happen when there is no value. Either not sure the meta item or some fallback text like "No prison number".

common/presenters/move-to-card-component.js
```diff
@@ -29,6 +30,20 @@ function moveToCardComponent({
...
+        moveMetaItems.push({
+          label: i18n.t('fields::prison_number.label'),
+          text: profile.person.prison_number,
+        })
+      } else {
```
teneightfive, line 39: Same suggestion here regarding checking a value exists.

common/presenters/move-to-card-component.test.js
```diff
@@ -464,6 +466,66 @@ describe('Presenters', function () {
           })
         })
       })
+
+      context('with identifier number', function () {
+        context('when to_location is a prison', function () {
```
teneightfive, line 471: I think we're missing a few test scenarios with the current code. For example, when no profile exists or no `to_location` exists. But probably best to wait for clarity on what happens when there is no value and then writing the test cases for those scenarios.

common/components/card/template.njk
```diff
@@ -42,7 +42,7 @@
           {% if item.label and (item.html or item.text) %}
             <div class="app-card__meta-list-item">
               <dt class="app-card__meta-list-key{% if item.hideLabel %} govuk-visually-hidden{% endif %}">
-                {{ item.label }}
+                {{ item.label | safe }}
```
teneightfive, line 45: It would be nice to support the pattern used elsewhere with `.text` and `.html` to support unescaped values. But I'm not sure all the places this is being used and whether it would be trickier to change?

common/presenters/moves-by-location.js
```diff
@@ -12,7 +12,7 @@ module.exports = function movesByLocation(data, locationKey = 'to_location') {
     locations.push({
       sortKey: location,
       items: sortBy(moves, 'profile.person._fullname').map(
-        moveToCardComponent()
+        moveToCardComponent({ showIdentifierNumber: true })
```
teneightfive, line 15: There's [another instance](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/f689f5b35c63585384896ad305891c4a7c2df411/common/presenters/moves-by-vehicle.js#L29) of this presenter being used when we display moves by vehicle. Can we update that to also include the identifier because I think it should follow the same behaviour.

ElizabethGoddard:
Hi @Thource couple of comments:
- The three data fields are now quite far apart which isn't so good for readability. Dom suggested setting them to be 4 columns (rather than 3) to bring them closer together, can we try this?

- Where there is no data can we do the following?

When logic assumes prison number will show:
**Prison number**
Not provided

When logic assumes PNC number will show:
**PNC number**
Not provided

Thanks!

common/components/card/template.njk
Thource: I agree that using `.label.text` and `.label.html` would be nicer, from a quick search though it looks like that'd result in over 170 changes, maybe something for the future instead?

teneightfive: Hmm, are all those instances using this particular template? I think the other instances of `label` within a component will all be escaping by default like this one would have been.

I think my concern was more that we're making all instance now unescaped.

From what I can see we'd only need to update it in these locations:
- [court-case-to-card-component](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/0adb469d7a6508dfb7e79f3b394bb4aab1b9834a/common/presenters/court-case-to-card-component.js#L14)
- [profile-to-card-component](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/0adb469d7a6508dfb7e79f3b394bb4aab1b9834a/common/presenters/profile-to-card-component.js#L43)

I'm happy to leave this for now though but maybe at a GH issue to keep track?

common/presenters/move-to-card-component.js
```diff
@@ -29,6 +30,24 @@ function moveToCardComponent({
 
     const moveMetaItems = []
 
+    if (showIdentifierNumber && profile?.person) {
```
teneightfive, line 33: So, after writing the comment above I noticed that we actually have a presenter to handle rendering a [profile as a card](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/0adb469d7a6508dfb7e79f3b394bb4aab1b9834a/common/presenters/profile-to-card-component.js#L43). This is also what's actually used to render the person on the search results page:

<kbd>image</kbd>

Which lead me to think that we should actually move this code to the profile presenter and then we would also be able to see the identifier on the search results too. @ElizabethGoddard should we show it there as well?

You can do away with the flag on this presenter and just pass `locationType` to `profileToCardComponent` and re-use that logic there.

The more I think about it, the identifier is part of the profile and not the move so I do think this code should be located in the other presenter.

ElizabethGoddard: Hey @teneightfive and @Thource if it makes sense to do the code this way then yes we might as well show it on the search. It might not add that much value now but it may one day if we were to increase the scope to search (e.g. by name and then get a listing, the Prison/PNC number would distinguish the results).

app/move/app/new/controllers/person-search-results.js
```diff
@@ -36,7 +36,10 @@ class PersonSearchResultsController extends PersonController {
   setPeopleItems(req, res, next) {
     const { people } = req.form.options.fields
     people.items = req.people.map(person => {
-      const card = presenters.profileToCardComponent({ showTags: false })({
+      const card = presenters.profileToCardComponent({
+        locationType: req.query.filter.prison_number ? 'prison' : 'other',
```
teneightfive, line 40: Could we use `from_location_type` from the session data here instead, I think that would be safer than manually setting the value.

You can get it using:

```javascript
const { from_location_type: locationType } = req.sessionModel.toJSON()
...
const card = presenters.profileToCardComponent({
  locationType,
  showTags: false,
})({ person })
```

common/presenters/move-to-card-component.js
```diff
@@ -49,6 +49,7 @@ function moveToCardComponent({
       ? { text: i18n.t(`statuses::${status}`) }
       : undefined
     const personCardComponent = profileToCardComponent({
+      locationType: toLocation?.location_type,
```
teneightfive, line 52: I don't think this will cater for both incoming and outgoing contexts. I think the location type we're actually concerned with is the location type of the currently selected location, whether than be from the session or the ID in the URL.

I think you may have to pass the `locationType` in to this via the presenters to get the correct context. I know this is making things a little tricky but I don't think the value we care about is related to the move, it comes from a higher level. Does that make sense?

teneightfive: If you look at how we get the [from location type in the creation journey](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/153c6f50b69528575f790e2829c7ccbf9b7897f4/app/move/app/new/controllers/base.js#L162) we use the session value and I think this is what we should be sending to the presenter.

common/presenters/moves-by-location.js
teneightfive, line 15: Is this still being used? I think what we should be passing in is `locationType` instead, but that will also need to be send in from [the middleware](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/153c6f50b69528575f790e2829c7ccbf9b7897f4/app/moves/middleware/set-results.moves.js#L5).

The value to use would be `req.location.location_type` I think. And then just pass that all the way down. I know it's a bit fiddly but because this logic needs to be determined based on the location type of the current location and not each move I can't see another way?

common/presenters/moves-by-vehicle.js
```diff
@@ -29,6 +29,7 @@ module.exports = function movesByVehicle({
       moveToCardComponent({
         showToLocation: showLocations || context === 'outgoing',
         showFromLocation: showLocations || context === 'incoming',
+        showIdentifierNumber: true,
```
teneightfive, line 32: I think the same suggestion here of using `locationType`.

common/presenters/profile-to-card-component.js
```diff
@@ -6,6 +6,7 @@ const filters = require('../../config/nunjucks/filters')
 const frameworkFlagsToTagList = require('./framework-flags-to-tag-list')
 
 function profileToCardComponent({
+  locationType = undefined,
```
teneightfive, line 9: You can remove the default value and it will be set to `undefined` too:
```suggestion
  locationType,
```

app/allocation/controllers/view.js
```diff
@@ -8,8 +8,10 @@ function viewAllocation(req, res) {
   const bannerStatuses = ['cancelled']
   const movesWithoutProfile = moves.filter(move => !move.profile)
   const movesWithProfile = moves.filter(move => move.profile)
+  const { location_type: locationType } = req.session.currentLocation
```
teneightfive, line 11: Could we swap this for [`req.location`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/main/common/middleware/set-location.js). This will prevent the bug that Robin fixed where if you view a location that isn't actually your "current location" it will still function as expected.

It will allow the location to be taken from the URL first and then current location if one doesn't exist.

app/moves/middleware/set-results.moves.js
```diff
@@ -19,11 +19,13 @@ function setResultsMoves(bodyKey, locationKey) {
           moves: activeMoves,
           context: locationKey === 'from_location' ? 'incoming' : 'outgoing',
           showLocations: !req.location,
+          currentLocationType: req.session.currentLocation.location_type,
```
teneightfive, line 22: Same here for using `req.location` instead of the current session. And maybe might be worth updating `currentLocationType` to `locationType`.

app/moves/middleware/set-results.single-requests.js
```diff
@@ -7,6 +7,7 @@ async function setResultsSingleRequests(req, res, next) {
     req.results = await singleRequestService.getAll(req.body.requested)
     req.resultsAsTable = presenters.singleRequestsToTableComponent({
       query: req.query,
+      currentLocationType: req.session.currentLocation.location_type,
```
teneightfive, line 10: We might not need to pass it in to this presenter, unless we're looking to add identifier to the table view below? @ElizabethGoddard?

<kbd>image</kbd>

common/presenters/moves-by-location.js
```diff
@@ -4,15 +4,21 @@ const i18n = require('../../config/i18n')
 
 const moveToCardComponent = require('./move-to-card-component')
 
-module.exports = function movesByLocation(data, locationKey = 'to_location') {
+module.exports = function movesByLocation(
+  data,
+  locationKey = 'to_location',
+  currentLocationType = undefined
```
teneightfive, line 10:
```suggestion
  locationType
```

common/presenters/moves-by-vehicle.js
```diff
@@ -11,6 +11,7 @@ module.exports = function movesByVehicle({
   moves,
   context = 'outgoing',
   showLocations = false,
+  currentLocationType = undefined,
```
teneightfive, line 14:
```suggestion
  locationType,
```

common/presenters/moves-by-vehicle.js
```diff
@@ -29,6 +30,7 @@ module.exports = function movesByVehicle({
       moveToCardComponent({
         showToLocation: showLocations || context === 'outgoing',
         showFromLocation: showLocations || context === 'incoming',
+        locationType: currentLocationType,
```
teneightfive, line 33:
```suggestion
        locationType,
```

common/presenters/single-requests-to-table-component.js
```diff
@@ -24,6 +28,7 @@ function singleRequestsToTable({ isSortable = true, query } = {}) {
             const card = moveToCardComponent({
               isCompact: true,
```
teneightfive, line 29: I think this is effectively the option that will tell `profileToCardComponent` to not show any of the metadata so we can drop needing to include `locationType` in this presenter.

teneightfive (approved):

app/allocation/controllers/view.js
```diff
@@ -8,8 +8,10 @@ function viewAllocation(req, res) {
   const bannerStatuses = ['cancelled']
   const movesWithoutProfile = moves.filter(move => !move.profile)
   const movesWithProfile = moves.filter(move => move.profile)
+  const { location_type: locationType } = req.location
   const moveToCardComponent = presenters.moveToCardComponent({
     showStatus: false,
+    locationType,
```
teneightfive, line 14: After seeing the end-to-end failure you pointing out I think this should in fact be (along with other suggestion above):
```suggestion
    locationType: fromLocation.location_type,
```

app/moves/middleware/set-results.single-requests.js
ElizabethGoddard: Hi @Thource and @teneightfive thanks for bringing this up, it hadn't crossed my mind as I'm less familiar with single requests. I think that actually it makes sense to include it here too ultimately, since it's a key identifier. That said this isn't as much of a priority. Can we accommodate the code so that we can do it later?

ElizabethGoddard:
Hey guys, @teneightfive and @Thource sorry it's taken me so long to look at this but I've just noticed something that isn't great - you know where we have added this to the search? 
Screenshot 2021-05-18 at 17 01 26
We now have three tight columns so Date of Birth is half dropping to another line. Do you think we could either adjust the grid such that the three sections are wider (there is still plenty of space to the right) or drop the (Age 34) to the next line so that it's not broken? As this will probably happen for all the results and would be better for readability. What do you think?

ElizabethGoddard (approved):
Look great, thank you Adam!

PR #1585, opened 2021-04-30
Add extra logging to missing Person ID error

teneightfive:
### What changed

In some instances we see an error when trying to update a profile
because it is missing a Person ID.

It also adds extra Sentry data to errors picked up from this step
including IDs within the session and the move itself.

### Why did it change

The current information in the Sentry errors isn't helping to pinpoint
where that data is being lost from the session.

This adds logging to capture the journey somebody has taken to get
to this point and may help understand the problem better.

### Issue tracking

- [2272958731](https://sentry.io/organizations/ministryofjustice/issues/2272958731/?project=2014813&query=is%3Aunresolved)

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/move/app/new/controllers/save.js
+++ b/app/move/app/new/controllers/save.js
@@ -1,5 +1,13 @@
 const Sentry = require('@sentry/node')
-const { get, omit, capitalize, flatten, values, some } = require('lodash')
+const {
+  get,
+  omit,
+  capitalize,
+  flatten,
+  values,
+  some,
+  snakeCase,
+} = require('lodash')
 
 const { uuidRegex } = require('../../../../../common/helpers/url')
 const analytics = require('../../../../../common/lib/analytics')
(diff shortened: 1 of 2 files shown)


app/move/app/new/controllers/save.js
```diff
@@ -32,6 +40,10 @@ class SaveController extends CreateBaseController {
         'documents',
       ])
       const move = await req.services.move.create(data)
+
+      // set before promises so it can be used for error handling
+      req.sessionModel.set('move', move)
```
Nimphal, line 45: Just looking at the order, setting it after the Promise.all meant that it wouldn't be set if one of the promises errors out. Moving it here, it will be set beforehand. I don't know if that makes a different, the catch block will still fire and more to the error middleware.

Nimphal (approved):

app/move/app/new/controllers/save.js
teneightfive: So I added a couple extra unit tests to capture it but moving it here means that if one of the update promises fails, like court hearings or profile update it will set this value and we can use it in the `errorHandler` of this controller.

It essentially means that if one of those fails we can trace the move that was created for them and any IDs that were in the session.

PR #1634, opened 2021-05-12
chore: Fix security vulnerabilities

Nimphal:
### What changed
Patch updates to some dev dependencies, and updates allowed advisories
in the audit json config. They are dev dependencies with not available
remediation, mostly around the release process and webpack.

### Why did it change
Fix security vulnerabilities.

The allowed advisories are:
- dot-prop https://www.npmjs.com/advisories/1213 which is a dependency of the package we use for changelog generation.
- postcss https://www.npmjs.com/advisories/1693 which is a dependency if a webpack plugin used to minimize our assets and not used at runtime
- underscore https://www.npmjs.com/advisories/1674 which is a dependency of finepack which use to generate the list of contributors when we do a release


audit-ci.json
```diff
@@ -1,5 +1,5 @@
 {
   "low": true,
   "registry": "https://registry.npmjs.org",
-  "allowlist": [1213, 1665]
+  "allowlist": [1213, 1674, 1693]
```
Nimphal, line 4: 1665 no longer affects us

teneightfive (approved):

PR #1644, opened 2021-05-14
[P4-2984] [P4-2829] Validate Police National Computer number

teneightfive:
### What changed

This change adds validation to the PNC number search and input fields
when creating a new move.

**Update**

This also now removes fields validated with the new Police National
Computer (PNC) validator from the form wizard so that the user doesn't
see this error.

It then adds each instance to Sentry as a warning.

### Why did it change

We've had cases where this number has been incorrectly input which
has lead to records not being returned when searched for.

**Update**

This is because we are not fully certain of the impact of introducing
this validation yet. There are some establishments that are using this
field for non-PNC reference numbers and we want to get an understanding
of scale first before enabling it.

These warnings will allow us to see the location types that are
causing this validation to trigger and allow us to decide an action
to take.

### Issue tracking

- [P4-2829](https://dsdmoj.atlassian.net/browse/P4-2829)
- [P4-2984](https://dsdmoj.atlassian.net/browse/P4-2984)

### Form validation

<kbd>image</kbd>

<kbd>image</kbd>

### Sentry issue

<kbd>image</kbd>

<kbd>image</kbd>

<kbd>image</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/move/app/new/controllers/base.js
+++ b/app/move/app/new/controllers/base.js
@@ -1,3 +1,6 @@
+const Sentry = require('@sentry/node')
+const { forEach, omitBy, pickBy } = require('lodash')
+
 const FormWizardController = require('../../../../../common/controllers/form-wizard')
 const presenters = require('../../../../../common/presenters')
 const filters = require('../../../../../config/nunjucks/filters')
@@ -224,6 +227,36 @@ class CreateBaseController extends FormWizardController {
     }
 
     return false
+  }
+
+  // TODO: Temporary data collection to understand impact of
+  // adding PNC number validation
+  validateFields(req, res, callback) {
+    super.validateFields(req, res, errors => {
+      const PNCPredicate = error =>
+        error.type === 'policeNationalComputerNumber'
+
+      forEach(pickBy(errors, PNCPredicate), (error, key) => {
+        Sentry.captureException(new Error('PNC validation failed'), {
+          level: Sentry.Severity.Warning,
+          tags: {
+            'validation_error.key': key,
+            'validation_error.path': error.url,
+            'validation_error.type': error.type,
+          },
+          contexts: {
+            'Form wizard error': {
+              ...error,
+              // `type` is reserved in Sentry so we need to set an
+              // alternative property
+              validationType: error.type,
+            },
+          },
+        })
+      })
+
+      callback(omitBy(errors, PNCPredicate))
+    })
   }
 }
 
--- a/app/move/validators.js
+++ b/app/move/validators.js
@@ -1,6 +1,30 @@
 const { parseISO, isValid, isAfter } = require('date-fns')
 const { Controller } = require('hmpo-form-wizard')
 
+function isPNCNumberValid(value) {
+  value = (value || '').toString().toUpperCase()
+  const regex = /^([0-9]{2}|[0-9]{4})\/[0-9]+[a-zA-Z]$/i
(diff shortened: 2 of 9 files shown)


teneightfive:
☝️ @ElizabethGoddard

app/move/validators.js
```diff
@@ -33,4 +33,13 @@ module.exports = {
       Controller.validators.regex(value, /^[A-Z][0-9]{4}[A-Z]{2}$/i)
     )
   },
+  policeNationalComputerNumber(value) {
+    return (
+      value === '' ||
+      Controller.validators.regex(
```
Thource, line 39: Do we need to check for the checksum too? https://dsdmoj.atlassian.net/wiki/spaces/PECS4/pages/3082551297/PNC+Validation

martin-ballhatchet-moj: Yes please!

app/move/validators.js
```diff
@@ -1,6 +1,26 @@
...
+  let validPNC = false
+
+  if (Controller.validators.regex(value, regex)) {
+    /* Use CJSE Data Standards Catalogue v4.3, section 3.374 */
+    const derivedPNC = value.replace('/', '').substring(2, 11)
+    const checkDigit = value.substring(12)
+    const i = Number.parseInt(derivedPNC) % 23
+
+    if (mod23chars[i] === checkDigit) {
+      validPNC = true
```
Thource, line 17: I know this is copied from Confluence, but I think it'd be nicer to do away with validPNC and just return true here

app/move/validators.js
```diff
@@ -1,6 +1,26 @@
 const { parseISO, isValid, isAfter } = require('date-fns')
 const { Controller } = require('hmpo-form-wizard')
 
+function isPNCNumberValid(value) {
+  value = (value || '').toString().toUpperCase()
+  const regex = /^([0-9]{2}|[0-9]{4})\/[0-9]+[a-zA-Z]$/i
+  const mod23chars = 'Z-A-B-C-D-E-F-G-H-J-K-L-M-N-P-Q-R-T-U-V-W-X-Y'.split('-')
```
Thource, line 7: Definitely just being picky here, but we could remove all of the `-`

app/move/app/new/controllers/base.js
```diff
@@ -225,6 +228,36 @@ class CreateBaseController extends FormWizardController {
...
+      const PNCPredicate = error =>
+        error.type === 'policeNationalComputerNumber'
+
+      forEach(pickBy(errors, PNCPredicate), (error, key) => {
+        Sentry.captureException(new Error('PNC validation failed'), {
+          level: Sentry.Severity.Warning,
+          tags: {
+            'validation_error.key': key,
+            'validation_error.path': error.url,
+            'validation_error.type': error.type,
```
teneightfive, line 245: Should we capture the failed input value (either here or in the context)?

I've left it off initially to get an idea of how often it occurs first, and we can also run a query on the DB to get an idea of failed patterns as a first, next step.

thomasleese: Agreed, I think it's okay to leave it off for now, I'd assume we don't get the same ones appearing over and over anyway.

thomasleese (approved):

PR #1667, opened 2021-05-21
chore: Add read-only user role

Thource:
### What changed

Added a read-only user role.

### Why did it change

Requirement has been confirmed for a PECS Read Only role for the purposes of HMPPS policy audit/assurance team(s). This has been approved by the Head of PECS. 

### Issue tracking

- P4-2842

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### DIFF
--- a/common/lib/permissions.js
+++ b/common/lib/permissions.js
@@ -43,6 +43,21 @@ const contractDeliveryManagerPermissions = [
   'youth_risk_assessment:print',
 ]
 
+const readOnlyPermissions = [
+  'dashboard:view',
+  'allocations:view',
+  'locations:contract_delivery_manager',
+  'locations:all',
+  'moves:view:outgoing',
+  'moves:view:incoming',
+  'moves:view:proposed',
+  'move:view',
+  'person_escort_record:view',
+  'person_escort_record:print',
+  'youth_risk_assessment:view',
+  'youth_risk_assessment:print',
+]
+
 const secureChildrensHomePermissions = [
   'dashboard:view',
   'moves:view:outgoing',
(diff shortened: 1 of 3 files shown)


common/lib/permissions.js
```diff
@@ -43,6 +43,21 @@ const contractDeliveryManagerPermissions = [
   'youth_risk_assessment:print',
 ]
 
+const readOnlyPermissions = [
```
teneightfive, line 46: Could we add a test to the [user library test](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/baec5e26e3dfb411ef2abaa7880f745cf7e437e6/common/lib/user.test.js#L121) (which is currently where these are unit tested).

common/lib/permissions.js
```diff
@@ -43,6 +43,21 @@ const contractDeliveryManagerPermissions = [
...
+  'dashboard:view',
+  'allocations:view',
+  'locations:contract_delivery_manager',
```
teneightfive, line 49: I can't remember exactly what this allows but I'm assuming the users with this role need to be able to see locations in this way? Does it allow them to see *all* locations?

common/lib/permissions.js
Thource: Oops, I searched for "contract_delivery_manager" and concluded that there were no tests :D

common/lib/permissions.js
Thource: `locations:contract_delivery_manager` is in a conditional that leads to getting all locations for all suppliers, `locations:all` gives permission to access the `/all` route

teneightfive: And I'm guessing this is something required for this user type?

teneightfive (approved):

common/lib/permissions.js
Thource: I believe so, from the ticket: This should be a replica of the current PECS CDM role, with access to Incoming and Outgoing moves for all PECS locations, but without the ability to see the maps.

PR #1684, opened 2021-05-26
chore: Make 5xx User axios requests retry once

Thource:
### What changed

User service 5xx errors from axios will now retry once.

### Why did it change

This changed because we currently have no retry after errors, so if for some reason the request failed when it shouldn't have, it'll be retried and succeed.

### Issue tracking

- P4-2653

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing


Nimphal (approved):

common/services/user.js
```diff
@@ -8,11 +8,25 @@ const referenceDataService = new ReferenceDataService()
 
 const getAuthHeader = token => ({ Authorization: `Bearer ${token}` })
 
+function retryOnceOn5xx(fun) {
+  return fun().catch(error => {
+    if (error.statusCode < 500) {
```
Nimphal, line 13: Actually, looking at this again, if the error is malformed and doesn't have the statusCode attribute, this will fail so we will retry. I wonder if we want to change this to have the explicit condition when we retry specified rather then the one when we don't. So something like:
```
if (error.statusCode && error.statusCode >= 500) {
  return fun()
}
  return

Comment (teneightfive) [reply common/services/user.js:13]: What about using the [retry-axios](https://www.npmjs.com/package/retry-axios) library?

File: common/services/user.test.js [hunk 2]
```diff
@@ -1,3 +1,5 @@
+const axios = require('axios')
+const { expect } = require('chai')
```
teneightfive, line 2: I don't think this needs to be imported as it's already defined in the test globals

common/services/user.js
```diff
@@ -1,4 +1,12 @@
 const axios = require('axios')
+const rax = require('retry-axios')
+
+const axiosInstance = axios.create()
+axiosInstance.defaults.raxConfig = {
+  instance: axiosInstance,
+  retries: 1,
```
teneightfive, line 7: Should this be `retry`? The are the [options](https://github.com/JustinBeckwith/retry-axios/blob/97937ce75b8af78d60332ba8d4194df40645925d/src/index.ts#L15) I could see from their code, granted they don't seem to document them too well.

teneightfive: Do we also want to make use of the [`statusCodesToRetry `](https://github.com/JustinBeckwith/retry-axios/blob/97937ce75b8af78d60332ba8d4194df40645925d/src/index.ts#L42) option to ensure it's only for 5xx responses? 

Looks like the default is `[[100, 199], [429, 429], [500, 599]]`, but maybe we're happy to retry on those too?

teneightfive (changes requested):

common/services/user.js
Thource: Oops, yes, this should be `retry`, `retries` makes more sense though.

I think we should be fine with the default status codes, 429 is rate-limiting, 1xx appear to be things revolving around the request not being complete yet.

teneightfive (approved):

PR #1711, opened 2021-06-07
Retry 502 and 504 errors on the API

teneightfive:
### What changed

The API sometimes returns Bad Gateway or Gateway timeout errors.

In those instances, to prevent the request ending on the first attempt,
enabling a retry may avoid these errors for users.

### Issue tracking

- [BOOK-A-SECURE-MOVE-FRONTEND-4T](https://sentry.io/organizations/ministryofjustice/issues/2422455305/?project=2014813&query=is%3Aunresolved&statsPeriod=14d)

#### Automated testing

- [ ] ~Added end-to-end tests to cover any journey changes~

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/lib/api-client/middleware/got-request.js
+++ b/common/lib/api-client/middleware/got-request.js
@@ -38,6 +38,11 @@ function requestMiddleware({
           http: new HttpAgent(),
           https: new HttpsAgent(),
         },
+        retry: {
+          limit: 1,
+          methods: ['GET'],
+          statusCodes: [502, 504],
+        },
         timeout,
       })
         .then(async res => {
--- a/common/lib/api-client/middleware/got-request.test.js
+++ b/common/lib/api-client/middleware/got-request.test.js
@@ -90,7 +90,16 @@ describe('API Client', function () {
           expect(gotStub.args[0][0]).to.have.all.keys({
             ...payload.req,
             agent: {},
+            retry: {},
             timeout: 5000,
+          })
+        })
+
+        it('should set retry options', function () {
+          expect(gotStub.args[0][0].retry).to.deep.equal({
+            limit: 1,
+            methods: ['GET'],
+            statusCodes: [502, 504],
           })
         })
 
@@ -126,6 +135,7 @@ describe('API Client', function () {
           expect(gotStub.args[0][0]).to.have.all.keys({
             ...payload.req,
             agent: {},
+            retry: 2,
             timeout: 5000,
           })
         })


common/lib/api-client/middleware/got-request.js
```diff
@@ -39,6 +39,10 @@ function requestMiddleware({
           http: new HttpAgent(),
           https: new HttpsAgent(),
         },
+        retry: {
+          limit: 1,
+          statusCodes: [502, 504],
+        },
```
Nimphal, line 45: Is there a danger of retrying a post request here? Maybe @martyn-w best placed to advise, but the API throws those errors sometimes from prison API, and the call to that sometimes happens after the API has saved some data into the database. I do not believe API requests are idempotent, but I am not 100% sure. If I am right, then retrying could leave us with duplicate data. Maybe?

teneightfive: That's a good question. 

However, the calls to build the dashboards and incoming/outgoing pages are POST requests also so if we remove `POST`s from the retries we wouldn't retry those which is an error we often see these type of failures.

martin-ballhatchet-moj: Please note - this is another reason why those calls should not be POSTs. I would rather *not* retry these than risk retrying a POST we don't want to carry out twice

martyn-w: I agree - the only calls which can be safely retried are `GET` requests, plus the following three `POST` requests:
* `/api/allocations/filtered`
* `/api/moves/filtered`
* `/api/moves/csv`

PATCH / DELETE / PUT and any other POST should not be retried otherwise there is a risk of duplicate / corrupt data.

The reason for the three POST routes for getting data is to do with the length limitation on URL query parameters: I think the front end needs to send a large number of location IDs for getting filtered moves - the length of this list exceeded the maximum URL length.

teneightfive: >The reason for the three POST routes for getting data is to do with the length limitation on URL query parameters: I think the front end needs to send a large number of location IDs for getting filtered moves - the length of this list exceeded the maximum URL length.

@martyn-w that's correct, we had to move to a `POST` to allow us to not have to batch up these requests and reduce the calls needed to build the dashboards.

@Nimphal @martyn-w @martin-ballhatchet-moj Shall we have a quick 5 mins to discuss this today? Be good to understand the details of what exactly could happen on retrying some of the requests with data and if there's way to prevent duplication within the API. If the request failed downstream and we bubble the error back to the user on the frontend, we currently tell them to "retry again" anyway, so whether its this layer auto-retrying or the user trying again, if there is a risk of duplication it will occur in both instances.

martin-ballhatchet-moj: "Be good to understand the details of what exactly could happen on retrying some of the requests with data" - that's a significant piece of work, for very little benefit. Based upon the general consensus, we can't retry POSTs. In case it helps - I think we were only talking about retrying 5xx errors against HMPPS Auth initially - based upon the advice of the HMPPS Auth team. If there's benefit to retrying 5xx for the API, I'm OK with it being done for GETs, but nothing else.

teneightfive: >that's a significant piece of work, for very little benefit

In the instance of a failed `POST` or `PATCH` with a `502` the user is still able to manually retry by refreshing the page, so if we're worried about duplicate data being created, it can still occur whether its a retry from our request library or the user.

The issue with only retrying `GET`s at this stage is I don't think it will resolve the [Bad Gateway errors](https://sentry.io/organizations/ministryofjustice/issues/2395154994/events/?project=2014813&query=Bad+Gateway&statsPeriod=14d) we've seen in the frontend coming from the API. That majority have come from either the dashboard or incoming/outgoing dashboards which fail on the `POST` filters.

The frequency does seem to be low though, so if we're happy to just accept them occurring and ignoring them we can also do that.

Nimphal (approved):

PR #1716, opened 2021-06-08
chore: Update Sentry perf tracing to ignore ping

Nimphal:
Make sure sentry doesn't pick up 'ping' endpoint for performance sampling

### DIFF
--- a/server.js
+++ b/server.js
@@ -82,7 +82,21 @@ if (config.SENTRY.DSN) {
           ]
         : [],
     // Half of all requests will be used for performance sampling
-    tracesSampleRate: config.SENTRY.ENVIRONMENT === 'production' ? 0.5 : 0,
+    tracesSampler: samplingContext => {
+      const transactionName =
+        samplingContext &&
+        samplingContext.transactionContext &&
+        samplingContext.transactionContext.name
+
+      if (config.SENTRY.ENVIRONMENT !== 'production') {
+        return 0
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -82,7 +82,21 @@ if (config.SENTRY.DSN) {
           ]
         : [],
     // Half of all requests will be used for performance sampling
-    tracesSampleRate: config.SENTRY.ENVIRONMENT === 'production' ? 0.5 : 0,
+    tracesSampler: samplingContext => {
+      const transactionName =
```
Thource, line 86: I think we can do `samplingContext?.transactionContext?.name`

server.js
```diff
@@ -82,7 +82,21 @@ if (config.SENTRY.DSN) {
...
+        samplingContext &&
+        samplingContext.transactionContext &&
+        samplingContext.transactionContext.name
+
+      if (config.SENTRY.ENVIRONMENT !== 'production') {
+        return 0
+      } else if (transactionName && transactionName.includes('ping')) {
+        return 0
+      } else {
```
Thource, line 95: We can get rid of the `else {}` :)

server.js
Thource: Boo! ☹️

Thource (approved):

server.js
Nimphal: It can, I'm sure this is not my last foray in this piece of code, so I'll clean it up next time :)

PR #1736, opened 2021-06-17
[P4-2910] Refactor move details summary

teneightfive:
### What changed

This refactors the way the move details panel, usually found in the right hand column is display.

The name is no longer the main heading and it uses a more consistent way to show the information.

### Why did it change

The move page is being redesigned. This will include some changes to this right hand column. These changes are a bit smaller in visual impact so are being done to both parts of the app, even though the main new design will be a separate URL that will sit alongside the existing view.

### Issue tracking

- [P4-2910](https://dsdmoj.atlassian.net/browse/P4-2910)

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


app/move/app/assign/controllers/base.js
```diff
@@ -33,7 +33,11 @@ class AssignBaseController extends CreateBaseController {
   }
 
   setMoveSummary(req, res, next) {
-    res.locals.moveSummary = presenters.moveToMetaListComponent(req.move)
+    const sessionData = req.sessionModel.toJSON()
```
Nimphal, line 36: From the tests below I can see we're using the session of profile and to location. Would that be impacted if a user has opened more than one move at a time? While things stored on the req are bound to the individual req context, the session is shared so is there a chance of something getting overridden?

teneightfive: Good question, but no, that sessionModel is also bound to the req for that form wizard. Since we introduced individual form wizards for moves to help with new tab behaviour this won't have an impact.

I'm actually trying to move the new/edit base controllers away from using another "model" concept on top of the existing form wizard `sessionModel` concept.

Basically we added a whole other layer on top of something that already existed which I don't think we need anymore.

Nimphal (approved):

PR #1753, opened 2021-06-25
[P4-2914] Add history of events to move page redesign

teneightfive:
### What changed

This adds the content for the history of events timeline tab for the
new move page redesign.

It extends the new structure of containing this code within a sub-app.

### Issue tracking

- [P4-2914](https://dsdmoj.atlassian.net/browse/P4-2914)

## Screenshots

<kbd>image</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

- [ ] ~Update [README](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/README.md) with any new instructions or tasks~

### DIFF
--- a/app/move/app/view/index.js
+++ b/app/move/app/view/index.js
@@ -4,6 +4,7 @@ const router = require('express').Router({ mergeParams: true })
 const { uuidRegex } = require('../../../../common/helpers/url')
 const {
   setMove,
+  setMoveWithEvents,
   setPersonEscortRecord,
   setYouthRiskAssessment,
 } = require('../../middleware')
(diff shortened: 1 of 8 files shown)


app/move/app/view/index.js
```diff
@@ -18,13 +19,14 @@ const { localsActions, localsMoveDetails, localsTabs } = require('./middleware')
...
 // Define shared middleware
-router.use((req, res, next) => {
-  next()
-})
-router.use(setMove)
+
+// For all non-timeline routes use standard move middleware
+router.use(/\/((?!timeline).)*/, setMove)
+// For all timeline route use move events middleware
+router.use('/timeline', setMoveWithEvents)
```
Nimphal, line 26: Out of curiosity, would this work if the routes were defined the other way around? As in, timeline first, so if it matches it does the thing, everything else after, removing the need for the regex? I seem to remember something about that in express routing. Not suggesting we do it, just curious.

Nimphal (approved):

app/move/app/view/index.js
teneightfive: I don't think it would give us the behaviour we need. This is essentially acting as an `if/else`. So if the route contains `/timeline` run `setMoveWithEvents`, for all other routes run `setMove`. Both middleware set the `req.move` property to allow following middleware and controllers to use that property.

The reason for this was because the API request runs a little slower when we ask for all timeline events so we had to split them as separate calls.

If we swapped them around and removed `/\/((?!timeline).)*/` we'd get both of them being run which would mean the last call would override the `req.move` property and leave it without the `timeline_events` property which is what is needed for the timeline route.

PR #1780, opened 2021-07-02
[P4-2973] Fix auto-confirmed PERs inability to handover

Thource:
### What changed

Removed "Confirmed" state for PERs, they're now either unconfirmed or handed over.

### Why did it change

This changed because when PERs are auto-confirmed, they aren't handed over and in the confirmed state, there is no button to handover, this was causing an issue where these auto-confirmed PERs couldn't be handed over.

### Issue tracking

- [P4-2973](New auto-confirm of PER prevents the ability to complete a handover)

## Screenshots

| Before | After |
| ------ | ----- |
| Incomplete |  |
| Completed |  |
| Confirmed and handover **not** recorded |  |
| Confirmed and handover recorded |  |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


common/presenters/message-banner/move-to-message-banner-component.js
```diff
@@ -24,34 +24,32 @@ function moveToMessageBannerComponent({ move = {}, moveUrl, canAccess } = {}) {
...
+      return assessmentToConfirmedBanner({
+        assessment,
+        baseUrl,
+        canAccess,
+        context,
+      })
+    }
   }
 
-  return undefined
```
teneightfive, line 54: Won't removing this mean the unconfirmed banner is always displayed, even when a PER hasn't been started yet?

Thource: I don't think I've changed any behaviour by removing this, previously we were checking against one conditional, then the reverse of that conditional, so I don't think this would've been hittable?

common/presenters/message-banner/assessment-to-unconfirmed-banner.js
```diff
@@ -30,7 +30,10 @@ module.exports = function assessmentToUnconfirmedBanner({
   `
 
   // PER has been completed
-  if (assessment.status === 'completed') {
+  if (
+    assessment.status === 'completed' ||
+    (assessment.status === 'confirmed' && !assessment.handover_occurred_at)
```
teneightfive, line 35: Could we add another test for this too please

common/presenters/message-banner/move-to-message-banner-component.js
teneightfive: Yeah it's odd we didn't have a test for this and I'm trying to rack my brain about why. I think it was if there was no status, which in theory is everything about line 36 so this looks fine.

teneightfive (approved):
Looks good, just a couple of minor comments

teneightfive:
@Thource can we hold off merging this just yet. Double checking something with Adam Luck regarding moves from other establishment types.

PR #1793, opened 2021-07-07
Avoid multiple API calls for location types

### LINKED ISSUE #1028: Switch hospital call to use combined API filter
## Problem

The API currently only supports passing [one location type](https://hmpps-book-secure-move-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk/api-docs/index.html#operations-Locations-get_reference_locations) as a filter.

To support the feature in #1021 the frontend needs to use multiple location types so requires making multiple calls and duplicate sorting which is less efficient.

## Solution

If the API can add support for filtering multiple location types separated by a comma (similar to how [location IDs can be filtered](https://hmpps-book-secure-move-api-staging.apps.live-1.cloud-platform.service.justice.gov.uk/api-docs/index.html#operations-Moves-get_moves)) it will allow us to remove the [duplicated sorting and the need to make multiple requests](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1021/files#r527858541) from the frontend.

thomasleese:
### What changed

Rather than making an API call for each location type and combining the results, we can now make a single API with all of the location types.

### Why did it change

This avoids making multiple unnecessary calls to the API.

### Issue tracking

Closes #1028 

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/middleware/set-location-items.js
+++ b/common/middleware/set-location-items.js
@@ -3,29 +3,6 @@ const { set } = require('lodash')
 const fieldHelpers = require('../helpers/field')
 const referenceDataHelpers = require('../helpers/reference-data')
 
-const getLocations = async (locationTypes, req) => {
-  const locations = (
-    await Promise.all(
-      locationTypes.map(x => req.services.referenceData.getLocationsByType([x]))
-    )
-  )
-    .reduce((acc, val) => acc.concat(val))
-    .sort(sortByTitle)
-  return locations
-}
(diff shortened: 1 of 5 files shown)


teneightfive (approved):
Great work and love the fast turnaround 👍

PR #1799, opened 2021-07-09
Ensure supplier user locations aren't cleared out

thomasleese:
### What changed

We now populate the locations for supply users when the session is generated, rather than when they visit the locations page.

### Why did it change

We think that supplier users are losing their locations when their authentication session needs regenerating because they don't visit the locations page.

### Issue tracking

- [P4-2975](https://dsdmoj.atlassian.net/browse/P4-2975)

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -4,6 +4,7 @@ const { populateSupplierLocations } = require('../../common/services/user')
 
 async function locations(req, res, next) {
   const userPermissions = get(req.session, 'user.permissions', [])
+  const userLocations = get(req.session, 'user.locations', [])
 
   let regions = []
 
@@ -14,8 +15,6 @@ async function locations(req, res, next) {
   } catch (error) {
     return next(error)
   }
-
-  const userLocations = req.userLocations
 
   try {
     await populateSupplierLocations(req.session.user)
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -4,14 +4,19 @@ async function locations(req, res, next) {
   const userPermissions = get(req.session, 'user.permissions', [])
 
   let regions = []
-  let userLocations = req.userLocations
-  const supplierId = req.session.user.supplierId
 
   try {
     if (userPermissions.includes('allocation:create')) {
       regions = await req.services.referenceData.getRegions()
     }
+  } catch (error) {
+    return next(error)
+  }
 
+  const supplierId = req.session.user.supplierId
+  let userLocations = req.userLocations
+
+  try {
     if (supplierId) {
       userLocations = await req.services.referenceData.getLocationsBySupplierId(
         supplierId
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -1,7 +1,5 @@
 const { get } = require('lodash')
 
-const { populateSupplierLocations } = require('../../common/services/user')
-
 async function locations(req, res, next) {
   const userPermissions = get(req.session, 'user.permissions', [])
   const userLocations = get(req.session, 'user.locations', [])
@@ -12,12 +10,6 @@ async function locations(req, res, next) {
     if (userPermissions.includes('allocation:create')) {
       regions = await req.services.referenceData.getRegions()
     }
-  } catch (error) {
-    return next(error)
-  }
-
-  try {
-    await populateSupplierLocations(req.session.user)
   } catch (error) {
     return next(error)
   }
--- a/app/locations/controllers.js
+++ b/app/locations/controllers.js
@@ -1,4 +1,6 @@
-const { get, uniqBy } = require('lodash')
+const { get } = require('lodash')
+
+const { populateSupplierLocations } = require('../../common/services/user')
 
 async function locations(req, res, next) {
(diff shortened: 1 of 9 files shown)


Thource (approved):

Thource:
Looking good :)

common/services/user.js
```diff
@@ -107,8 +108,32 @@ function getNomisLocations(token) {
     )
 }
 
+async function populateSupplierLocations(user) {
+  const { supplierId, permissions } = user
+
+  if (supplierId) {
+    user.locations = await referenceDataService.getLocationsBySupplierId(
```
teneightfive, line 115: I wonder if it would be better to return the locations here rather than mutate the `user` object like it was previously doing?

app/locations/controllers.js
```diff
@@ -1,44 +1,19 @@
...
-      // The locations have been uniqued based on title to prevent
-      // duplicates when multiple suppliers have the same location
-      userLocations = uniqBy(supplierLocations.flat(), 'id')
-    }
   } catch (error) {
     return next(error)
   }
 
-  req.session.user.locations = userLocations
+  const userLocations = req.userLocations
```
teneightfive, line 16: Having looked at the function that is setting `req.userLocations`, I'm not really sure I'm seeing a good reason for it anymore as its not being used anywhere else.

I wonder if its possible to remove that as part of this change to reduce the amount of abstraction and use something similar as line 4?

```javascript
const userLocations = get(req.session, 'user.locations', [])
```

thomasleese: Sounds reasonable! Do you think the canonical way of accessing the user should be `req.session.user` or [`req.user`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/main/common/middleware/set-user.js)?

common/services/user.js
thomasleese: Yeah, I wondered about that too. I decided to go with this method because I wasn't sure whether the return value should be `null` or `user.locations` if the locations isn't being changed (i.e. the user isn't a supplier) and it felt a bit messy. I'll have another think about what we could do here, because I definitely agree a function which mutates the object isn't great either.

app/locations/controllers.js
thomasleese: (My preference would be `req.session.user` since it's more explicit.)

teneightfive: > Do you think the canonical way of accessing the user should be req.session.user or req.user

That's an excellent question. I think this set of code was introduced before we added `req.user`. The addition of `req.user` was to allow Sentry to automatically add user context to issues and since have started to reference `req.user` instead.

I don't have a strong preference though so happy either way.

common/services/user.js
teneightfive: Yes, I see what you mean. I was also wondering if it could have been incorporated into the [`User`](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/a732c065c08d0b8a288f7a0fba63d16c8b3ee208/common/lib/user.js#L13) class similar to [how permissions are set](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/a732c065c08d0b8a288f7a0fba63d16c8b3ee208/common/lib/user.js#L23) but it would involve making that async which could make things a little more complex...?

I wonder if the promises in the auth middleware could actually be updated to contain the logic as that's almost where it should be determined. In these cases where the user is a supplier or has a particular role we wouldn't need to carry out the other locations call either, right?

app/locations/controllers.js
thomasleese: I've made this change in 1c926ad!

common/services/user.js
thomasleese: I agree this is definitely the way forward. I'm concerned that making this change in this PR would make it quite a big change, and I think I'd prefer to have it at least as a separate PR, but maybe even a separate ticket.

I've had a little play around how we could do it differently. The main problem is that because `user.permissions` is done within the `User` model, we need the `user` instance available before we can try to fetch the supplier locations. But for the `user` to exist, we need to have already tried to fetch the `locations` so we can construct a `User` instance. I definitely agree this is problem and ripe for refactoring, but I'm worried doing it in this PR is too much.

At least with the PR as it is the bug itself should be fixed and the behaviour of _how_ the supplier locations hasn't changed, it's just the _when_. Does that make sense?

teneightfive: Yep, I agree with all of this. I think that's a really good approach, and as you say, the main goal was to prevent the locations being lost when a user's session is regenerated without them visiting the locations controller :)

Maybe we can add a GH issue with some of these suggestions and can pick it up when/if there is some capacity?

teneightfive (approved):

common/services/user.js
thomasleese: Sounds good, I'll write those issues this afternoon!

thomasleese: I've written up https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/1805.

PR #1810, opened 2021-07-13
[P4-2913] Extend task list component to support sections

teneightfive:
### What changed

This change adds the ability to group tasks within sections for the
Task List Component.

This change follows a similar pattern to the MOJ Design System but
maintains the flexibility to apply different tags and labels.
https://moj-design-system.herokuapp.com/components/task-list

[**COMPONENT DEMO**](https://booksecuremove-pr-1810.herokuapp.com/components/task-list)

### Why did it change

Previously it only supported showing tasks rather than the ability to
group them.

### Issue tracking

- [P4-2913](https://dsdmoj.atlassian.net/browse/P4-2913)

## Screenshots

<kbd>image</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/components/task-list/template.njk
+++ b/common/components/task-list/template.njk
@@ -1,29 +1,47 @@
 {% from "govuk/components/tag/macro.njk" import govukTag %}
 
-<ul class="app-task-list">
-  {% for item in params.items %}
-    <li class="app-task-list__item">
-      <span class="app-task-list__task-name">
-        {% if item.href %}
-          <a href="{{ item.href }}" aria-described-by="{{ item.text | urlencode }}-status">
+{% macro taskListItems(items) %}
+  <ul class="app-task-list__items">
+    {% for item in items %}
+      <li class="app-task-list__item">
+        <span class="app-task-list__task-name">
+          {% if item.href %}
+            <a href="{{ item.href }}" aria-described-by="{{ item.text | urlencode }}-status">
+          {% endif %}
+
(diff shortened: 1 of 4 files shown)


common/components/task-list/template.njk
```diff
@@ -1,29 +1,47 @@
...
+            classes: (item.tag.classes or "") + " app-task-list__tag"
+          }) }}
         {% endif %}
-      </span>
+      </li>
+    {% endfor %}
+  </ul>
+{% endmacro %}
+
+{% if params.sections %}
```
thomasleese, line 33: I wonder if there's a way we can guarantee that both `sections` and `items` isn't set at the same time - perhaps by raising an exception? Then this would be visible in tests where the component is used.

thomasleese (approved):
Looks good to me! I've left one comment, but not a blocker.

common/components/task-list/template.njk
teneightfive: I have wondered that in the past for some other similar situations, for example where GOV.UK components use `text` and `html` properties on text elements. Their pattern has been to favour one over the other and to document that in the yaml configuration over throwing an exception.

I've updated the yaml to be even more explicit that `sections` is favoured over `items` when both are provided so hopefully that will help.

PR #1812, opened 2021-07-13
Avoid fetching unnecessary locations for supplier locations

### LINKED ISSUE #1805: Non-supplier locations are unnecessarily fetched for supplier users
We fetch the locations for a supplier user in a slightly different way to other kinds of users. However, at the moment we're doing this after we've already tried to fetch the non-supplier locations, which then get replaced. This ends up with making more requests to the API than necessary.

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/47b5c6b080071e0ce5c625454dda836ea895200c/app/auth/middleware.js#L25-L42

It had to done like this because fetching the supplier locations relies on first parsing the permissions of the user, which depends on having an instance of `User` and that relies on fetching the non-supplier locations.

The suggestion is that we should refactor how the `User` model is constructed and move more of the auth middleware code into the `User` model. Unfortunately class constructors can't be `async` so this'll likely require a separate method which populates the locations after the `user` has been constructed. See https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1799#discussion_r667797447 for more background.

thomasleese:
### What changed

This refactors how locations are fetched for users (following #1799) and various other parts of the authentication and user middleware. The refactor is relatively large, but the underlying behaviour should be pretty much unchanged, and we're able to remove around 150 lines of code :smile:. Worth reviewing commit by commit!

### Why did it change

This is to avoid unnecessarily fetching locations for supplier users which are then not used. It should also make the code related to user authentication and fetching easier to follow and understand.

### Issue tracking

Closes #1805

### DIFF
--- a/app/auth/middleware.js
+++ b/app/auth/middleware.js
@@ -1,12 +1,7 @@
 const { get } = require('lodash')
 
 const { decodeAccessToken } = require('../../common/lib/access-token')
-const User = require('../../common/lib/user')
-const {
-  getFullname,
-  getSupplierId,
-  getLocations,
-} = require('../../common/services/user')
+const { loadUser } = require('../../common/lib/user')
 
 function processAuthResponse() {
   return async function middleware(req, res, next) {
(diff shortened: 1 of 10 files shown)


Thource (approved):
Nice work! Just need to fix the failing e2e test, I suspect an API call isn't being mocked somewhere.

teneightfive (approved):
Nice refactor 👍 I like that all the user work is abstracted away and that we set locations/name/etc all in one place.

thomasleese:
> Nice work! Just need to fix the failing e2e test, I suspect an API call isn't being mocked somewhere.

Looks like the tests are flaky, I had to re-run them twice before they went green, each time failing with a different test!

PR #1813, opened 2021-07-13
Tidy the way update URLs are set so they can be easily re-used

teneightfive:
### What changed

This change includes some refactors to the way edit URLs are retrieved and used for moves. These URLs are currently in multiple places including the main move view controller/template and the presenters to render move details as a meta list component. 

It also extracts the logic used to determine if a move can be edited or cancelled to a shared place in the service and to reduce the need to pass around permissions or the `canAccess` method to determine if a particular user can do this action.

### Why did it change

In some instances we need to pass around permissions or permissions check functions between a few different layers and this started to get particularly messy within the meta list presenter. This attempts to try and standardise this away and write simpler, smaller helper methods that can be re-used. 

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


common/services/move.js
```diff
@@ -278,9 +280,17 @@ class MoveService extends BaseService {
...
-      .then(response => response.data)
+    const { canAccess } = this.req
+
+    return this.apiClient.find('move', id, options).then(response => {
+      const move = response.data
+
+      return {
+        ...move,
+        _canCancel: canCancelMove(move, canAccess),
+        _canEdit: canEditMove(move, canAccess),
```
teneightfive, line 291: I wasn't completely sure on this naming convention so would be open to any other better suggestions here. 

I've continued to use an underscore prefix to signify that these properties are created within the app as opposed to coming from the API layer. It [something we also do](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/1c590276f91c00cfd7443f821d1484527f17b8bc/common/lib/api-client/transformers/person.transformer.js#L2) in the resource transformers for the JSON:API serialization layer.

I toyed with a few different names in my head too so would be good to get opinions on what maybe works better: `_canEdit`, `_isEditable`, `_editable` ...?

common/presenters/move-to-meta-list-component.js
```diff
@@ -4,18 +4,26 @@ const moveAgreedField = require('../../app/move/app/new/fields/move-agreed')
 const componentService = require('../../common/services/component')
 const i18n = require('../../config/i18n')
 const filters = require('../../config/nunjucks/filters')
-const getUpdateLinks = require('../helpers/move/get-update-links')
+const mapUpdateLink = require('../helpers/move/map-update-link')
 
+/**
+ * Convert a move into the structure required to render
```
teneightfive, line 10: I've moved this method to accept an entity as the first argument, which will be required, and then to be able to pass options in as an Object as the second argument. That way, if we ever need to extend/reduce options it will have less of an impact on existing implementations that might only want to pass a single option in. How does this approach sound?

common/services/move.js
thomasleese: I'm happy with `canEdit` since it matches the prefix of the `canEditMove` function and implies they're related (which they are!).

I'm unsure about the underscores myself and even whether this is the right place to insert the `canEdit` information. Without knowing more than the function name, I'm not sure you'd expect the data from the API to be transformed to include this information. I suppose this is why we've got the `_` convention, but there's nothing that actually stops an API response from including fields which start with an underscore. I was also a little surprised to see `this.req` being available in what to me seems like a class that shouldn't necessarily know about the current request, but I think this would be a wider change of this service!

Without knowing the architecture of the app too well, is there a placer higher up where we can insert this field?

Since we've already got this convention in the app, I'm also happy to accept this as it is, better to stick with one way of doing things if we're already doing it in lots of places.

common/helpers/move/can-edit-move.js
```diff
@@ -0,0 +1,17 @@
+function canEditMove(move, canAccess = () => false) {
```
thomasleese, line 1:
```suggestion
function canEditMove(move, hasPermission = () => false) {
```

I realise this is named to match `req.canAccess` but at first the name was a little confusing to me, I wonder if something like this is more explicit.

common/helpers/move/get-update-urls.js
```diff
@@ -1,29 +1,16 @@
...
-  return updateUrls
+function getUpdateUrls(move = {}, canAccess = () => false) {
+  return editSteps
+    .filter(() => move._canEdit)
+    .filter(step => canAccess(step.permission))
+    .reduce((acc, current) => {
+      const entryPointUrl = Object.keys(current.steps)[0]
+      acc[current.key] = getMoveUrl(move.id, 'update', { entryPointUrl })
+      return acc
+    }, {})
```
thomasleese, line 13:
```suggestion
  return Object.fromEntries(
    editSteps
      .filter(() => move._canEdit)
      .filter(step => canAccess(step.permission))
      .map(current => {
        const entryPointUrl = Object.keys(current.steps)[0]
        return [current.key, getMoveUrl(move.id, 'update', { entryPointUrl })]
      })
  )
```

I tend to prefer `map` over `reduce` even for objects as I find it easier to understand, but happy with either.

common/helpers/move/can-edit-move.js
thomasleese, line 1: This function is nice!

common/helpers/move/can-cancel-move.js
```diff
@@ -0,0 +1,37 @@
...
+
+function checkCanCancelMoveProposed(canAccess, status) {
+  if (canAccess('move:cancel:proposed')) {
+    if (status === 'proposed') {
+      return true
+    }
+  }
+
+  return false
+}
```
thomasleese, line 21:
```suggestion
function checkCanCancelMoveProposed(canAccess, status) {
  return canAccess('move:cancel:proposed') && status === 'proposed'
}
```

I realise you only renamed the file, but this feels like a nice refactor.

common/helpers/move/can-cancel-move.js
```diff
@@ -0,0 +1,37 @@
...
+  if (canAccess('move:cancel')) {
+    if (['requested', 'booked'].includes(status)) {
+      if (!allocation) {
+        return true
+      }
+    }
+  }
+
+  return false
+}
```
thomasleese, line 11:
```suggestion
function checkCanCancelMove(canAccess, status, allocation) {
  const suitableStatus = ['requested', 'booked'].includes(status)
  return canAccess('move:cancel') && suitableStatus && !allocation
}
```

common/helpers/move/can-edit-move.js
teneightfive: Yeah, I kept the name of this argument inline with the method that is being passed in. 

I ideally wanted to remove the need to send in this function but currently we have some logic which allows particular edit steps to support specific permissions in case we needed to restrict certain user types to only being able to edit certain bits of information, for example risk/health information but not the date.

If we were to rename this I'd prefer to do this as a separate change as I think we should be consistent in how we reference this method and currently its referred to as `canAccess` in a [number of places](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/search?q=canAccess).

common/helpers/move/get-update-urls.js
teneightfive: Yes, that is neater actually. I tend to prefer reading `map`s too but I initially thought it wouldn't be as nice to read transforming an array into an Object but this does read nicely too :)

common/helpers/move/can-cancel-move.js
teneightfive: I agree. Initially I was trying to keep changes down to hopefully make it easier to review but I agree with this refactor :) I'm also more a fan of combined return statements over nested `if`s.

Will also add this in when updating the tests.

common/services/move.js
teneightfive: >I'm happy with canEdit since it matches the prefix of the canEditMove function and implies they're related (which they are!).

👍 

>Without knowing more than the function name, I'm not sure you'd expect the data from the API to be transformed to include this information. I suppose this is why we've got the _ convention, but there's nothing that actually stops an API response from including fields which start with an underscore.

Yes, that's correct, there is nothing stopping the API including an underscore but I suppose its also just a convention the API has stayed away from which is why we adopted it just for this layer. It allowed us to insert extra data that the API didn't need to or couldn't necessarily insert for us.

And that's actually the case with this example because the editability within this frontend layer is determined by a user's permissions as well as the particular state of the move, and with the architecture we currently have the API doesn't know about the current user's permissions. So it has to be done somewhere in this layer. 

>I'm unsure about the underscores myself and even whether this is the right place to insert the canEdit information.

This is the part that follows on from the above, previously we were doing it in the `getUpdateUrls` helper but that required the `canAccess` method to be passed down through various different layers and functions. This is what I was trying to remove to make some of those functions a little simpler.

But the main problem is that to determine whether it can be edited by this particular user we need access to the user's permissions which lives on the session and in turn the `req` object. So we can either do it in lower level middleware or controllers (this was the previous implementation) or somewhere higher like this move service.

>I was also a little surprised to see `this.req` being available in what to me seems like a class that shouldn't necessarily know about the current request, but I think this would be a wider change of this service!

There is a bit of backstory to this change. The services didn't previously have access to the request object but it was a change done to allow all services have access to it so that we could pass relevant information from the request [along with each API request](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/main/common/lib/api-client/request-headers.js), for example a user's name, ID for audit or transaction IDs to allow us to link up frontend and API requests or errors. 

Here's [a change](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1124/) I found related to that but it was done over a series of PRs. In hindsight, it would have been nice to use a particular issue/milestone/project so that we could've looked back at all related changes easier.

With a bit more context how does the current placement within this service sound?

thomasleese: Thanks for the detailed response, that makes a lot of sense. Happy to proceed with the new placement, it's definitely an improvement on how it is now!

thomasleese (approved):
Nice refactor!

PR #1833, opened 2021-07-20
[P4-2908] Add identity bar to new move page design

teneightfive:
### What changed

This adds the new identity bar component to the move page redesign.

[**DEMO**](https://booksecuremove-pr-1833.herokuapp.com/move/cc721283-e508-4635-ad9b-7f8fc80187b0/beta/warnings)

### Why did it change

This component will be re-used by both the new move page and the person page.

### Issue tracking

- [P4-2908](https://dsdmoj.atlassian.net/browse/P4-2908)

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/assets/javascripts/application.js
+++ b/common/assets/javascripts/application.js
@@ -20,6 +20,7 @@ import MultiFileUpload from '../../components/multi-file-upload/multi-file-uploa
 
 const accessibleAutocomplete = require('accessible-autocomplete')
 const { initAll } = require('govuk-frontend')
+const HcSticky = require('hc-sticky')
 const StickySidebar = require('sticky-sidebar/dist/sticky-sidebar')
 
 const AddAnother = require('../../components/add-another/add-another')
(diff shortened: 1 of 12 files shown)


Thource (approved):

common/assets/javascripts/application.js
```diff
@@ -20,6 +20,7 @@ import MultiFileUpload from '../../components/multi-file-upload/multi-file-uploa
 
 const accessibleAutocomplete = require('accessible-autocomplete')
 const { initAll } = require('govuk-frontend')
+const HcSticky = require('hc-sticky')
```
thomasleese, line 23: Without being too much of an expert in CSS, could we use [`position: sticky`](https://developer.mozilla.org/en-US/docs/Web/CSS/position) instead of this library? I'm not sure what browsers we support, but it's only IE where this isn't supported.

teneightfive: I did consider this, but it was a little more than it staying in the same position and the same form.

Ours now works a little similar to how the GH Pull Requests work in that once you scroll the header out of view, it then appears positioned to the top of the page in a condensed format. I think if we just used `position: sticky` I don't think there would be a way to adjust the styling once it's "stuck".

This library was one of the only ones I could find that also creates the placeholder in the right size so that it doesn't feel too jerky. 

Have a play with it on the [new design layout](https://booksecuremove-pr-1833.herokuapp.com/move/cc721283-e508-4635-ad9b-7f8fc80187b0/beta/warnings) and let me know what you think.

thomasleese: Makes sense, it looks good on the preview! 👍🏻

thomasleese (approved):

PR #1834, opened 2021-07-21
Support 2.2.0 frameworks with NOMIS assessments

thomasleese:
This includes a new type of NOMIS mapping for assessments, a new religious and cultural question and some further grammar changes.

We don't want to start using this version until the suppliers are confident using it themselves, so we'll need to set the `FRAMEWORKS_VERSION` environment variable in production to the previous version.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3015)

## Screenshot

Screenshot 2021-07-26 at 15 33 22


thomasleese:
@ElizabethGoddard https://booksecuremove-pr-1834.herokuapp.com/move/a5b3e471-1b4a-4b85-827a-f0d4cbbcce48/person-escort-record/risk-information/hold-separately is an example of what this looks like in staging.

teneightfive (approved):

thomasleese:
@ElizabethGoddard I'm merging this in now to avoid the PR becoming stale, but I can demonstrate the feature to you when you're back. It won't be live in production until the suppliers have implemented the feature as well, so no rush!

ElizabethGoddard:
> @ElizabethGoddard I'm merging this in now to avoid the PR becoming stale, but I can demonstrate the feature to you when you're back. It won't be live in production until the suppliers have implemented the feature as well, so no rush!

Sounds good!

ElizabethGoddard:
@thomasleese are there any other alerts with more than one date line? If so and they already look like this then we can leave as is but if this is the first we might want to adjust the spacing between the two date lines so that they sit together (we could pair to look at it if you like). Thanks!

thomasleese:
> @thomasleese are there any other alerts with more than one date line? If so and they already look like this then we can leave as is but if this is the first we might want to adjust the spacing between the two date lines so that they sit together (we could pair to look at it if you like). Thanks!

It's not the first, but we could look at adjusting the spacing if we'd prefer it.

PR #1836, opened 2021-07-22
Revert Sentry warnings for a missing location

thomasleese:
This reverts commit a9f1f53.

We've solved the bug which caused us to introduce these warnings and now we're getting them appearing for legitimate cases, so we can stop sending them to Sentry.


teneightfive (approved):
🙌

PR #1837, opened 2021-07-22
Revert "Warn when the auth source cannot be determined"

thomasleese:
Reverts ministryofjustice/hmpps-book-secure-move-frontend#1796

We saw this occur for Delius authentication, which we currently don't support, so we can stop warning about this situation. The original bug that we introduced this warning for has been fixed.

> Delius is the Probation “equivalent” of NOMIS and has been added as an auth provider. We have no expectation at this time that Probation users need access to Book a Secure Move as far as I’m aware, so any attempt to use the service should be blocked as it was in this case.

### DIFF
--- a/common/services/user.js
+++ b/common/services/user.js
@@ -1,4 +1,3 @@
-const Sentry = require('@sentry/node')
 const axios = require('axios')
 const { uniqBy } = require('lodash')
 const rax = require('retry-axios')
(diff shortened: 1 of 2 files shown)


teneightfive (approved):
I'm happy to remove it, but wonder whether its worth leaving in the background as a warning in case it ever crops up we get lots of users trying with unexpected Auth sources in the future and just ignoring the Sentry issue until x occurrences?

thomasleese:
> I'm happy to remove it, but wonder whether its worth leaving in the background as a warning in case it ever crops up we get lots of users trying with unexpected Auth sources in the future and just ignoring the Sentry issue until x occurrences?

That sounds reasonable, I'll close the PR and we can stick with ignoring the issue.

PR #1838, opened 2021-07-22
Add require-await eslint rule

teneightfive:
### What changed

This adds a rule to ensure that asynchronous functions contain
asynchronous code in the form of an `await` or returning a promise.

This also includes the fixes required to pass this new rule.

### Why did it change

We spotted a number of functions, initially within tests, that were using an `async` definition
when there was no need.

Tidying this up ensures we only use `async` for asynchronous functions and means tests are less
likely to hang when failing.

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


thomasleese (approved):
Nice tidy up!

PR #1845, opened 2021-07-23
build(deps): Bump the GOV.UK Frontend dependency

teneightfive:
### What changed

This is a manual dependency bump as it can often include some visual
changes.

In this instance it does include the optional new link styles added
in [`v3.12.0`](https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0).

#### Todo

- [x] Check [new links styles](https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0) don't have any adverse effects
- [x] See if we can use [Group buttons and links](https://github.com/alphagov/govuk-frontend/releases/tag/v3.11.0)
- [x] ~See if we can use [Add a 'none' option and 'or' divider to checkboxes](https://github.com/alphagov/govuk-frontend/releases/tag/v3.13.0) (maybe for the Person Escort Record)~ Not needed for now. Maybe something to keep in mind for future changes/additions to the frameworks repo
- [x] See if we can use [Change the style of table captions](https://github.com/alphagov/govuk-frontend/releases/tag/v3.11.0)
- [x] ~See if we can move to [new banners](https://github.com/alphagov/govuk-frontend/releases/tag/v3.10.0) from `v3.10.0`~ This is a much bigger change and something that may change when we look back at how success messages are rendered. Also something better discussed with @ElizabethGoddard first.

### Why did it change

GOV.UK Frontend includes a number of new additions. In order to keep inline with the style and direction of GOV.UK its best we updated this library.

This change also includes some removal of local code in favour of new GOV.UK Design System approaches. See individual commits for breakdown.

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


.github/dependabot.yml
```diff
@@ -11,8 +11,6 @@ updates:
       interval: 'daily'
     ignore:
       - dependency-name: '@hmpps-book-secure-move-frameworks*'
-      - dependency-name: 'govuk-frontend'
```
teneightfive, line 14: @thomasleese be interesting to get your thoughts on this. Initially this was put in to prevent potential major visual changes. But I think its just lead to us not staying more current.

And also looking at the way GOV.UK Frontend is released, they provide opt-in options for major visual changes, for example the new link styles. 

I personally would prefer us to be notified of these releases, but maybe they come with a deeper review of the automated PRs?

thomasleese: > I personally would prefer us to be notified of these releases, but maybe they come with a deeper review of the automated PRs?

Yeah, I completely agree with this. It's not like we're overwhelmed at the moment with the number of PRs, so I think it's reasonable to reduce the number that we're ignoring. Plus regardless of what is being upgraded, we should spend time reviewing them as appropriate, something which upgrades a small library we use in the tests require less in depth analyse as a major update or `govuk-frontend` for example.

Going on a bit of a tangent, have we ever thought about putting in place some visual regression testing?

teneightfive: >Going on a bit of a tangent, have we ever thought about putting in place some visual regression testing?

Yeah, it has been something we've discussed but the main reason had just been capacity previously. It's maybe something that would be good to include moving forward when we hand over into a maintenance state.

On a related note I spotted that images are no longer showing, I'm guessing after a recent browserify upgrade, and this type of thing would have caught it.

I've used [wraith](https://github.com/bbc/wraith) in the past but not for a little while. Are there any libraries you've had a good experience with?

thomasleese: > I've used [wraith](https://github.com/bbc/wraith) in the past but not for a little while. Are there any libraries you've had a good experience with?

Wraith is the only library I've used in the past too, but I understand that it's not really updated any more.

teneightfive: Yep, unfortunately so. Last update was 3 years ago...

thomasleese (approved):

ElizabethGoddard:
@teneightfive sorry for the delayed reaction on this, I'm not sure I understand it, maybe if we get the chance we can discuss it?

PR #1846, opened 2021-07-26
docs: Improve .env.example defaults

thomasleese:
This adds a few more useful environment variables that need to be set to run the app, re-orders and groups them into sections, and documents where to get a suitable value for a few of them.

Specifically I've removed "v1" from API_PATH as this is determined from API_VERSION now, and I've set it to a meaningful value.


teneightfive (approved):
Nice improvements, thanks!

PR #1848, opened 2021-07-26
fix: Ensure cache is cleared on errors during deserialization

teneightfive:
### What changed

Currently devour, the JSON:API client will build a cached of
deserialized resources. We have added a transformation layer to help
with certain things the frontend requires for these resources.

However, if an error occurs, the cache remains in tact which will
mean that in some cases, on a refresh of the page the app will use
the cached version.

This can lead to some unexpected renders so this fix calls the `clear()`
method on the cache to ensure on the next request it will begin again
from the start.

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd>before</kbd> | <kbd>after</kbd> |

#### Automated testing

#### Manual testing

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/lib/api-client/transformers/transform-resource.js
+++ b/common/lib/api-client/transformers/transform-resource.js
@@ -2,23 +2,28 @@ const { cloneDeep } = require('lodash')
 
 module.exports = function transformResource(transformer) {
   return function deserializer(item, included) {
-    const _this = cloneDeep(this)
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
Nice catch!

PR #1867, opened 2021-08-02
Refactor tab templates

thomasleese:
This refactors how tabs are rendered so there's a common template we can use in as many views as we like. We can't, unfortunately, use the component provided by `govuk-frontend` as it expects all the panels to be rendered at the same time, and then JS is used to switch between the tabs.

I've made this change ahead of needing to use the tabs template for the new profile page, so I decided to split it out into its own PR.


teneightfive (approved):
Nice little refactor 👌

PR #1871, opened 2021-08-03
Refactor person app in preparation for moves tab

thomasleese:
This refactors the person app to make it slightly easier to add the new moves tab as per the designs. This follows a lot of the patterns established in the new move page design.

There are some slight visual changes for users of the page, but nothing major that I think would prevent this PR from being deployed ahead of the new move tab. This has the advantage of making the PR for the move tab smaller as it won't need to include these changes too.

I suspect this PR is easier to review commit by commit!


app/person/middleware.js
```diff
@@ -1,16 +1,29 @@
-module.exports = {
-  setPerson: async (req, res, next) => {
-    const personId = req.params.personId
+async function setPerson(req, res, next) {
```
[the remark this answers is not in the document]
thomasleese: I'm unsure about this. I agree the code is very similar to `setMove`, but it's also a very simple piece of code. Refactoring this out into a function feels like it adds unnecessary complexity - we'd have to be able to pass in the API call, the name of the variable on `req` which gets set, the name of the URL param... not sure it's worth it.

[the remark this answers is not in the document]
teneightfive: I agree, definitely ignore codeclimate on this instance :)

app/person/index.js
```diff
@@ -1,21 +1,20 @@
...
+const { uuidRegex } = require('../../common/helpers/url')
 const breadcrumbs = require('../../common/middleware/breadcrumbs')
 const { PLACEHOLDER_IMAGES } = require('../../config')
 
-const { image, renderPerson } = require('./controllers')
-const { setPerson } = require('./middleware')
+const { image, personalDetails } = require('./controllers')
+const { setPerson, setBreadcrumb } = require('./middleware')
+
+router.get('/', (req, res) => res.redirect(`${req.baseUrl}/personal-details`))
```
teneightfive, line 10: I wonder if we should actually remove this and just serve the `personDetails` controller at `/person/:personId`?

It's something I was considering for the new move changes actually. Would also save a further redirect and might give (a slight) performance gain?

app/person/controllers.js
```diff
@@ -3,52 +3,51 @@ const axios = require('axios')
...
-  image: fallbackImage => {
-    return async (req, res) => {
-      try {
-        const imageUrl = await req.services.person.getImageUrl(
-          req.params.personId
-        )
-        const response = await axios.get(imageUrl, {
-          responseType: 'arraybuffer',
-        })
+function image(fallbackImage) {
```
teneightfive, line 6: Wonder if its actually worth extracting this one layer further into separate files like some of the other larger apps if we know there will be another one (possibly two) controllers?

So something like:

```
/controllers
  image.js
  image.test.js
  personal-details.js
  ...
```

teneightfive (approved):
Another nice little refactor 👍 

Couple of minor comments but no blockers and will leave the choice up to you :)

app/person/controllers.js
thomasleese: Yeah, I wondered about this too. I decided to leave it because the controllers and middleware were quite small, but since you've raised it, I'll add a commit which does this!

app/person/index.js
thomasleese: Interesting that it's something you've been considering removing from the new move page, I quite liked that it gave some structure to the routes where they matched the tabs. I can see both arguments though, and avoiding unnecessary redirects sounds sensible.

teneightfive: Yeah, I have been in two minds about it actually. I totally agree that I liked that the URLs then matched the tab structure (especially seeing as we use a route per tab) but the other side of my head was saying "is it a bit unnecessary...?".

I think its less of a concern here as at least we don't make an API call to get the person each time but if we use `router.param` it would result in multiple calls which felt inefficient. 

Lets leave this as is for now then as it least it matches the other concept. And if we want to investigate it and see if has a big performance impact we can change both instances. At the moment neither of these are actually exposed to users as their part of the new design so it won't have an impact yet anyway.

teneightfive: Added a [GH issue](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/issues/1872) for it to prompt us to pick this up separately and make a decision at some point.

thomasleese: Awesome, that looks great, thanks!

PR #1712, opened 2021-06-07
Prevent Sentry tracing in all non-production environments

teneightfive:
### What changed

We're currently recording tracing metrics for all environments which
is causing us to get close to current limits within the plan.

Related to #1675 

#### Automated testing

- [ ] ~Added unit tests to cover changes~
- [ ] ~Added end-to-end tests to cover any journey changes~

#### Manual testing

Has been tested in the following browsers:

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

- [x] No environment variables were added or changed

- [ ] Documented in the [README](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/README.md)
- [ ] Added to [continous integration](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-book-secure-move-frontend)
- [ ] Added to staging environment (deployment repository)
- [ ] Added to preproduction environment (deployment repository)
- [ ] Added to production environment (deployment repository)

### Other considerations

Commit messages with a `fix` or `feat` type are automatically used to generate the [changelog](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/CHANGELOG.md) and
[GitHub release notes](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/releases) during the release task. Please make sure they will read well on their own in a
summary of changes and that the commit body gives a more detailed description of those changes if necessary.

- [x] No new [Personally Identifiable Information (PII)](https://support.google.com/analytics/answer/6366371?hl=en) is being sent via analytics
- [x] Update [README](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/README.md) with any new instructions or tasks

PR #1874, opened 2021-08-04
Improve Sentry performance tracing

thomasleese:
This reduces the number of traces we record in Sentry, to avoid hitting rate limits. It also sends all traces to Sentry, rather than not just production.

See individual commits for more details on the changes.

Related to https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1529

### DIFF
--- a/server.js
+++ b/server.js
@@ -70,17 +70,12 @@ if (config.SENTRY.DSN) {
     dsn: config.SENTRY.DSN,
     environment: config.SENTRY.ENVIRONMENT,
     release: config.SENTRY.RELEASE,
-    integrations:
-      config.SENTRY.ENVIRONMENT === 'production'
-        ? [
-            // enable HTTP calls tracing
-            new Sentry.Integrations.Http({ tracing: true }),
-            // enable Express.js middleware tracing
-            new Tracing.Integrations.Express({
-              app,
-            }),
-          ]
-        : [],
+    integrations: [
+      // enable HTTP calls tracing
+      new Sentry.Integrations.Http({ tracing: true }),
+      // enable Express.js middleware tracing
+      new Tracing.Integrations.Express({ app }),
+    ],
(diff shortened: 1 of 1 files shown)


teneightfive (approved):
The Sentry environment condition was added in #1712 to also help avoid hitting the rate limits.

I do like the fact that we can see performance data across different environment and potentially spot dangerous changes before being released but will removing this not cause us to potentially hit those limits even though we are reducing the sample rate? (I'm not fully aware of the figures but just wanted to double check)

thomasleese:
> I do like the fact that we can see performance data across different environment and potentially spot dangerous changes before being released but will removing this not cause us to potentially hit those limits even though we are reducing the sample rate? (I'm not fully aware of the figures but just wanted to double check)

I'm not fully aware of the figures either, but I'm working on the assumption that production gets the majority of requests. I did the same in production and we didn't see a large increase in the number of traces. I suspect we should give it a go with 25% of traces and we can always bring the logic back if it is a problem.

Personally, I think it's better that we have a fewer proportion of traces for all environments, than a higher proportion but only for production.

teneightfive:
> > I do like the fact that we can see performance data across different environment and potentially spot dangerous changes before being released but will removing this not cause us to potentially hit those limits even though we are reducing the sample rate? (I'm not fully aware of the figures but just wanted to double check)
> 
> I'm not fully aware of the figures either, but I'm working on the assumption that production gets the majority of requests. I did the same in production and we didn't see a large increase in the number of traces. I suspect we should give it a go with 25% of traces and we can always bring the logic back if it is a problem.
> 
> Personally, I think it's better that we have a fewer proportion of traces for all environments, than a higher proportion but only for production.

Yep, happy with this. The only thing to keep in mind with staging is that we run all our end-to-end tests against it which could consume a lot of the limit, especially given all the automated dependabot PRs.

PR #1908, opened 2021-08-18
Focus on warnings when clicked using anchors

teneightfive:
### What changed

This adds a change to allow panels to be focusable, specifically in the scenario where they are used to display warnings
such as risk and health information.

It also includes a change to offset the scroll position when these elements are clicked so that content isn't hidden by placing
them at the top of the window.

### Why did it change

The alert badges can be clicked either from the listing pages (incoming/outgoing) or the header of a move. Currently there is not focus which means it visually hard to see which has been selected, and also will prevent content being read by assistive technology.

In some cases, like the framework steps a form field will also be highlighted, but the question and hint text can sometimes
be visually out of view.

## Screenshots

| Before | After |
| ------ | ----- |

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/common/components/panel/_panel.scss
+++ b/common/components/panel/_panel.scss
@@ -9,6 +9,10 @@ $_border-width: 2px;
   & + & {
     margin-top: govuk-spacing(2);
   }
+
+  &:focus {
+    outline: $govuk-focus-width solid $govuk-focus-colour;
+  }
 }
 
 // If panel contains a tag component, style based on this parent
--- a/common/components/panel/panel.yaml
+++ b/common/components/panel/panel.yaml
@@ -1,4 +1,8 @@
 params:
+  - name: isFocusable
+    type: boolean
+    required: false
+    description: Determines whether the panel will have a tabindex to allow it to be focused. Defaults to `false`.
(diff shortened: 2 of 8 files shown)


common/assets/scss/core/_elements.scss
```diff
@@ -10,3 +10,7 @@ img {
   display: inline-block;
   max-width: 100%;
 }
+
+// [id] {
+//   scroll-margin-top: 100px; // Approximate height of sticky navigation.
```
teneightfive, line 15: This appears to be causing TestCafe to have issues selecting nested elements, need to investigate.

ElizabethGoddard (approved):
Hey @teneightfive this is great - thank you!

thomasleese (approved):

teneightfive:
I've pulled the commit containing the scroll offset into a [separate PR](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/1921) for now as its causing the end-to-end tests to fail.

PR #1912, opened 2021-08-19
Move page prototype

teneightfive:


thomasleese:
We don't need the prototype anymore.

PR #1917, opened 2021-08-20
Only show warnings once the Person Escort Record has been started

teneightfive:
### What changed

This changes the logic around the warnings tab for the new profile page design. When there is no Person Escort Record (PER) it won't show any warnings, instead a message will be shown saying what needs to happen.

Once the PER has been started and sections become complete, it will show the warnings for this information.

### Why did it change

When people use this section to determine how a person should be moved they should use the latest information from the PER. Therefore, we have reduced the logic around this section to only show once the PER has been started and sections become complete.

### Without a Person Escort Record (PER)

<kbd>Screenshot 2021-08-20 at 16 22 57</kbd>

### With partially completed PER
<kbd>Screenshot 2021-08-20 at 16 23 09</kbd>

### With completed PER

<kbd>Screenshot 2021-08-20 at 16 23 23</kbd>

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### DIFF
--- a/app/move/app/view/middleware/locals.warnings.js
+++ b/app/move/app/view/middleware/locals.warnings.js
@@ -9,35 +9,14 @@ function setWarnings(req, res, next) {
     return next()
   }
 
-  const assessmentAnswers = profile.assessment_answers || []
+  let tagList
   const personEscortRecord = profile.person_escort_record
-  const youthRiskAssessment = profile.youth_risk_assessment
-  const requiresYouthRiskAssessment = profile.requires_youth_risk_assessment
-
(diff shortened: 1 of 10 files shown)


thomasleese (approved):
Nice!

ElizabethGoddard (commented):
This is great @teneightfive. I wonder if we should say 'Warnings will display once a Person Escort Record has been **started**' as opposed to completed? (as we have a partially completed state)

PR #1923, opened 2021-08-24
fix: Cope with result.moves being undefined

thomasleese:
This uses the safe navigation operator to ensure that `result.moves.length` doesn't raise a `TypeError` when `result.moves` is undefined. This appears to be happening occasionally as we've received a Sentry issue for it: https://sentry.io/organizations/ministryofjustice/issues/2582039703/

It looks like the code is only temporary anyway since there's a TODO block around it, but since we're getting Sentry issues for it today I figure there's no harm in fixing it now.


teneightfive (approved):
Nice!

PR #1931, opened 2021-08-25
Add custom feedback link for new move design

thomasleese:
This adds the ability to customise the feedback URL by moving the logic into global middleware. Then we customise the feedback URL when users have opted in to the move pages with the new design. This allows us to tailor the feedback for the new design.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3112)


Thource (approved):

app/move/app/view/index.js
```diff
@@ -41,6 +42,7 @@ router.use(setPersonEscortRecord)
 router.use(setYouthRiskAssessment)
 router.use(setBreadcrumb)
 router.use(localsUrls)
+router.use(localsFeedbackUrl)
```
teneightfive, line 45: Sorry, I think the ticket might have got this a little wrong. Mara wanted us to change this globally through the app and only if the user has previously opted in to see the new move page design.

I think with this change it will update the feedback URL on the move page for users who haven't opted in yet or who have opted out.

teneightfive: So I think this middleware can then sit globally in `server.js`

app/move/app/view/middleware/locals.feedback-url.js
```diff
@@ -0,0 +1,11 @@
+const { MOVE_DESIGN_FEEDBACK_URL } = require('../../../../../config')
+
+function setFeedbackUrl(req, res, next) {
+  if (MOVE_DESIGN_FEEDBACK_URL) {
```
teneightfive, line 4: I think we'll need to update this to also check for the cookie being set to `1` like in [checkPreviewChoice](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/7ca8fd3624f3afa5ad86ccb708069f16eb7b6baa/app/move/middleware.js#L18)

teneightfive (changes requested):

app/move/app/view/index.js
thomasleese: Ahh yes, makes sense, I misread the ticket I think!

thomasleese: I've amended the PR with the correct behaviour now!

teneightfive (approved):

PR #1934, opened 2021-08-25
Record Google Analytics events on opting in or opting out

thomasleese:
When a user opts in or out of the new move page design, we will record an event in Google Analytics so we can do analysis on the data.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3114)


Thource (approved):

app/move/controllers/preview-opt-in.js
```diff
@@ -1,14 +1,20 @@
...
-function previewOptIn(req, res) {
+async function previewOptIn(req, res) {
   const moveId = req.query.move_id
   const cookieName = COOKIES.MOVE_DESIGN_PREVIEW.name(req.user.userId)
 
   res.cookie(cookieName, 1, {
     maxAge: COOKIES.MOVE_DESIGN_PREVIEW.maxAge,
   })
 
+  await sendEvent({
```
teneightfive, line 13: It might be worth adding a `try/catch` around this with a call to `next(error)` if it fails.

app/move/controllers/preview-opt-out.js
```diff
@@ -1,13 +1,19 @@
...
-function previewOptOut(req, res) {
+async function previewOptOut(req, res) {
   const moveId = req.query.move_id
   const cookieName = COOKIES.MOVE_DESIGN_PREVIEW.name(req.user.userId)
 
   res.cookie(cookieName, 0, {
     maxAge: COOKIES.MOVE_DESIGN_PREVIEW.maxAge,
   })
 
+  await sendEvent({
```
teneightfive, line 12: Same with this one about adding a `catch` to handle any errors

PR #1963, opened 2021-09-08
Remove youth_risk_assessment:print permission

thomasleese:
This isn't an action that can be performed on youth risk assessments, so there's no reason to have a permission for it.

By removing this permission it should avoid the "Print Youth Risk Assessment" links from appearing on the frontend as they only render if the user has the permission. This fixes a bug where pressing that link would take the user to a 404 page.

## Before

Screenshot 2021-09-08 at 15 54 25

## After

Screenshot 2021-09-08 at 15 56 03

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3137)

### DIFF
--- a/common/lib/permissions.js
+++ b/common/lib/permissions.js
@@ -42,7 +42,6 @@ const contractDeliveryManagerPermissions = [
   'person_escort_record:view',
   'person_escort_record:print',
   'youth_risk_assessment:view',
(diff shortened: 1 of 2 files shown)


Thource (approved):

ElizabethGoddard (approved):
This is great, thanks Tom!

PR #1966, opened 2021-09-09
Hide moves tab for non-CDM users

thomasleese:
This hides the moves tab on the profile page for non-CDM users, since they can't access the page anyway. I've also change the tag style to the standard MoJ badge until we're confident on the colours we want.

## Without permission

Screenshot 2021-09-09 at 13 27 14

## With permission

Screenshot 2021-09-09 at 14 55 10

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3131)

### DIFF
--- a/app/person/middleware/locals.tabs.js
+++ b/app/person/middleware/locals.tabs.js
@@ -1,14 +1,23 @@
+const { get } = require('lodash')
+
+const { check } = require('../../../common/middleware/permissions')
+
 module.exports = (req, res, next) => {
   const tabs = [
     {
       text: 'person::tabs.personal_details',
       url: `${req.baseUrl}/personal-details`,
     },
-    {
+  ]
+
+  const userPermissions = get(req.session, 'user.permissions')
+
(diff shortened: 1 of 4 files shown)


ElizabethGoddard:
Hey @thomasleese this looks great, just one thing about the tags - it looks like there is letter spacing of 1px which I don't think is on the other tags?

thomasleese:
> Hey @thomasleese this looks great, just one thing about the tags - it looks like there is letter spacing of 1px which I don't think is on the other tags?

Ah yes, you're right. I've fixed that now, should be visible in the Heroku preview in a few minutes.

ElizabethGoddard:
@thomasleese the only other comment is about the padding on the date field but now that you've adjusted that I think we're good to go! As mentioned I'll write a new ticket to address the look of the table for small screens along with other tables in the application. Thanks!

ElizabethGoddard (approved):

Thource (approved):

PR #1980, opened 2021-09-14
Improve framework button navigation

thomasleese:
This improves the framework button navigation to make it easier for users to complete an entire PER/YRA in one go. Specifically this:

- Removes section summary screens from the end of each section and instead takes users straight to the first question of the next section.
- Removes the ‘Save and continue’ button from the final question screen as this is the end of the framework. Instead this is replaced by the existing ‘Save and return to overview button’.

See individual commits for more details as I had to do some refactoring before making the implementation changes.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3142)

## Screenshots

This is mostly not a visual change, except for the button on the last question.

### Old
Screenshot 2021-09-14 at 09 52 56

### New
Screenshot 2021-09-14 at 09 52 47

### DIFF
--- a/common/controllers/framework/framework-section.js
+++ b/common/controllers/framework/framework-section.js
@@ -1,6 +1,7 @@
-const { filter, findIndex, snakeCase, sortBy } = require('lodash')
+const { filter, snakeCase } = require('lodash')
 
 const i18n = require('../../../config/i18n')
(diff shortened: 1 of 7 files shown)


Thource (approved):

ElizabethGoddard (approved):
Hi @thomasleese this is brilliant - thanks so much!

PR #1984, opened 2021-09-16
Improve assessment call to actions

thomasleese:
This improves the journey for how to access parts of an assessment (youth risk assessment and person escort record). Specifically this moves the print button to the overview page, adds a link to the PER in the identity bar and moves the assessment links to a documents section in the details tab.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3136)

### Old

Screenshot 2021-09-16 at 09 26 11

Screenshot 2021-09-16 at 09 26 17

Screenshot 2021-09-16 at 09 26 26

### New

Screenshot 2021-09-16 at 09 25 58

Screenshot 2021-09-16 at 09 25 48

Screenshot 2021-09-16 at 09 25 38

### DIFF
--- a/common/controllers/framework/framework-overview.js
+++ b/common/controllers/framework/framework-overview.js
@@ -4,7 +4,7 @@ const moveHelpers = require('../../helpers/move')
 const presenters = require('../../presenters')
 
 function frameworkOverview(req, res) {
-  const { originalUrl, assessment = {}, move } = req
+  const { originalUrl, assessment = {}, move, canAccess, baseUrl } = req
   const moveId = move?.id
   const profile = move?.profile || assessment.profile
   const fullname = profile?.person?._fullname
@@ -15,12 +15,19 @@ function frameworkOverview(req, res) {
   })
   const i18nContext = snakeCase(assessment.framework?.name || '')
 
+  const canPrint =
+    canAccess(`${i18nContext}:print`) &&
+    (['confirmed', 'completed'].includes(assessment.status) ||
+      !!assessment.handover_occurred_at)
+
   res.render('framework-overview', {
     ...moveHelpers.getMoveSummary(move),
     i18nContext,
     moveId,
     taskList,
     fullname,
+    canPrint,
+    baseUrl,
   })
 }
 
--- a/app/move/app/view/controllers/render-details.js
+++ b/app/move/app/view/controllers/render-details.js
@@ -25,6 +25,7 @@ function renderDetails(req, res) {
 
   const moveSummary = presenters.moveToSummaryListComponent(move, {
     updateUrls,
+    canAccess,
   })
 
   const additionalInfo = presenters.moveToAdditionalInfoListComponent(move)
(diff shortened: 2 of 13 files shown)


common/controllers/framework/framework-overview.js
```diff
@@ -15,12 +15,18 @@ function frameworkOverview(req, res) {
   })
   const i18nContext = snakeCase(assessment.framework?.name || '')
 
+  const printable =
+    assessment.status === 'confirmed' ||
+    assessment.handover_occurred_at ||
+    (assessment.status === 'completed' && canAccess(`${i18nContext}:confirm`))
```
Thource, line 21: Misplaced brackets? Should be `(['confirmed', 'completed'].includes(assessment.status) || assessment.handover_occurred_at) && canAccess(`${context}:print`)`

thomasleese: This logic is actually the same as it was before:

- https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/8c8d8a65a6e927cf0d89e978b4828324642b19b5/common/presenters/move-to-identity-bar-actions/assessment-actions.js#L22
- `OR`
- https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/8c8d8a65a6e927cf0d89e978b4828324642b19b5/common/presenters/move-to-identity-bar-actions/assessment-actions.js#L63

I agree it's a little confusing though, which is why I pulled it out into a `printable` variable, but let me see if I can make it clearer.

thomasleese: Although I do think we can get rid of the `canAccess`.

Thource: Ah, ignore me, I didn't see that you're checking `&& canAccess(`${i18nContext}:print`)` just below :D

thomasleese: I've decided your suggestion is better, I think it was actually the intention, I've pushed up an amended commit which changes it!

Thource (approved):

ElizabethGoddard:
Thanks @thomasleese! This is great :) Just two things:

1. When I click the Print PER link in the PER overview screen I get 'page not found'?
2. Are we able to horizontally centre align the text in the ID bar when it is condensed? I don't think I can add an image here but I can send you one on slack... if this would take much time then don't worry but if it's simple let's try it.

Thanks!

thomasleese:
> 1. When I click the Print PER link in the PER overview screen I get 'page not found'?

Good catch, this should be fixed now!

> 2. Are we able to horizontally centre align the text in the ID bar when it is condensed? I don't think I can add an image here but I can send you one on slack... if this would take much time then don't worry but if it's simple let's try it.

Yep, no problem. It should look like this now:

Screenshot 2021-09-16 at 12 08 15

ElizabethGoddard (approved):
Flippin' brilliant! Thanks Tom!!

PR #1965, opened 2021-09-09
Improve navigation to framework overviews

thomasleese:
This makes two small changes to make it clearer to the user what the overview it by using the phrasing consistently across the site. See individual commits for more details.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3143)

PR #1988, opened 2021-09-17
Revert PER navigation behaviour for old move design

thomasleese:
This ties the button navigation behaviour with whether the user has opted in to the new move page design, and reverts some of the logic if the user has not.

This is in a response to a question of whether we definitely want this design.

See #1965 for the original changes.


Thource (approved):

thomasleese:
Superseded by #2001

PR #2001, opened 2021-09-22
Improve framework button navigation

thomasleese:
This follows on from #1965 following some feedback, while also trying to reduce unexpected changes to users who haven't opted-in to the new move design.

- The "Save and continue" button is not shown on the last possible step of a section (it is shown if there are conditional next steps).
  - This fixes a bug in production right now where the last property page is difficult to fill in.
- Users who haven't opted in to the new move design will be taken to the move record page if they click on "Save and return to overview" rather than the framework overview page

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3198)

## Screenshots

Screenshot 2021-09-22 at 11 53 06

Screenshot 2021-09-22 at 11 53 15

### DIFF
--- a/common/controllers/framework/framework-step.js
+++ b/common/controllers/framework/framework-step.js
@@ -190,10 +190,19 @@ class FrameworkStepController extends FormWizardController {
       isLastStep,
       frameworkSection: { key: currentSection },
       body: { save_and_return_to_overview: goToOverview },
+      moveDesignPreview,
+      assessment: {
+        framework: { name: frameworkName },
+      },
     } = req
 
     if (goToOverview || isLastStep) {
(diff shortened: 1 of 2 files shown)


ElizabethGoddard:
Hey Tom, as mentioned I think the 'Save and continue' on the property page when answering 'Yes' isn't quite right as it's going to the move page rather than bag page. Thanks!

thomasleese:
> Hey Tom, as mentioned I think the 'Save and continue' on the property page when answering 'Yes' isn't quite right as it's going to the move page rather than bag page. Thanks!

That issue should be fixed now, ready to be reviewed again!

ElizabethGoddard (approved):
Amazing, thank you Tom!

PR #2002, opened 2021-09-23
Hide framework from handover breadcrumbs

thomasleese:
This hides the framework assessment item from the breadcrumbs of a page if the user is performing a handover.

### Old

Screenshot 2021-09-23 at 08 13 31

### New

Screenshot 2021-09-23 at 08 13 09


ElizabethGoddard (approved):
Great, thank you Tom!

PR #2128, opened 2021-11-08
chore: Fix unit tests that broke due to DST ending

Thource:
### What changed

Tests broken by DST have been unbroken.

### Why did it change

Due to American DST ending, some of the tests failed, these have been fixed in a way that means that they won't need fixing again.


matt-crowson-moj (approved):
LGTM

matt-crowson-moj:
As discussed I'd question the value of having tests with non UK time zones.

PR #2163, opened 2021-11-18
Explicitly install Python 3 in Docker container

thomasleese:
This should fix an issue where Alpine Linux no longer has a `python` package available and instead requires an explicit version.

See https://github.com/docker-library/docker/issues/240 for more information.


matt-crowson-moj (approved):
LGTM

PR #2164, opened 2021-11-18
fix: Adding right hand padding to the message banner as there was no  spacing when text was wrapped

matt-crowson-moj:
This fixes an issue in the message banner where text was wrapping without any right spacing/padding to the surrounding DIV.

[JIRA P4-3135 ](https://dsdmoj.atlassian.net/browse/P4-3135) 

**Old:**

image

**New:**

image


thomasleese (approved):
Looks good!

PR #2169, opened 2021-11-19
Build Docker container with Python 2

thomasleese:
In 841ad36b359aefe10e063d5ec2ef784bff5cb12d I changed the Python package name from `python` to `python3` to fix the package being missing, but actually it looks like the correct solution would have been to change this to `python2` instead.


matt-crowson-moj (approved):
LGTM

PR #2170, opened 2021-11-19
Downgrade to Alpine 3.13

thomasleese:
There appears to be an issue in Alpine 3.14 related to the version of Docker we're using. There's more information about this issue here: https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2

For now, the suggested fix might be to downgrade to the previous version of Alpine Linux. Once the build is working again, we might think about switching to a more familiar Debian base.


matt-crowson-moj (approved):
LGTM

PR #2172, opened 2021-11-22
fix: Removing the ability to upload documents

matt-crowson-moj:
The ability to upload documents for moves is no longer required.

Historic moves with documents attached can still be downloaded.

**Move with document added (prior to change, as in added move with document attached then took out ability to add):**

image

**New move (without a previously added document):**

image


locales/en/moves.json
```diff
@@ -101,7 +98,7 @@
     "assessment_answer": {
       "label": "{{question}} details"
     },
-    "documents": {
+    "z": {
```
thomasleese, line 101: Is this change intentional?

matt-crowson-moj: Nope.  I must have hit a key or something ...

thomasleese (commented):
Code changes look good, I think it would be worth rebasing (to get rid of the merge commit) and squashing the two other commits together so we don't need the typo fix commit.

thomasleese (approved):
Looks good!

ElizabethGoddard (approved):
Awesome, thanks Matt!

PR #2176, opened 2021-11-24
fix: Fixing wording in identity bar for proposed moves

matt-crowson-moj:
If a move was in proposed state it would not have a move date.  The original summary header message being displayed expected a date to be present ergo it did not read correctly.

**Before any changes:**

image

**After (move pending review):**

image

**After (move reviewed):**

image


thomasleese (approved):

ElizabethGoddard:
Looks great thanks Matt!

PR #2182, opened 2021-11-25
fix: Changes to remove of handover notification banner, information availble in history of events

matt-crowson-moj:
Removal of the handover notification banner (not the flash banner). [JIRA](https://dsdmoj.atlassian.net/browse/P4-3181)

As part of the recent changes a dismissible flash banner was introduced to highlight when a user has recorded the handover. This is now not required. Users can get this information by looking at the History of Events (linked in the flash banner). 

**Removing the following banner on PER completion/handover:**

image

**With it removed:**

image


Thource (approved):

ElizabethGoddard:
Looks great, thanks Matt

PR #2187, opened 2021-11-26
fix: Removing blue line appearing to the right on allocation pages

matt-crowson-moj:
Small change to prevent a (random) blue line from appearing on the allocations pages.  This was being included as part of a person summary being present and in this case there isn't one. [JIRA](https://dsdmoj.atlassian.net/browse/P4-3118)

**Before:**

blue-line-before

**After:**

blue-line-after


thomasleese (approved):

ElizabethGoddard:
Fab - thanks @matt-crowson-moj!

PR #2189, opened 2021-11-29
fix: Removes superfluous back link from the PER

matt-crowson-moj:
Removes superfluous 'Back to move' link from the PER.  You can get back to the move via the breadcrumb trail.

**Before:**

back-link-with

**After:**

back-link-without


thomasleese (approved):

ElizabethGoddard:
This is great, thanks Matt!

PR #2193, opened 2021-11-30
fix: Removes the person details side panel when nothing to show

matt-crowson-moj:
When the personal details have not yet been provided we do not want to show side panel (to avoid clutter). 

**Before:**

person-details-before

**After:**

person-details-after


thomasleese (approved):

ElizabethGoddard:
Brilliant, thanks Matt

PR #2196, opened 2021-12-01
fix: Remove style override on cancel allocation banner so uses default style

matt-crowson-moj:
**Before:**

allocation-cancel-before

**After:**

allocation-cancel-after


thomasleese (approved):

thomasleese (approved):

ElizabethGoddard:
Amazing! Thanks Matt!

PR #2201, opened 2021-12-06
fix: Adds the PER section status tag to the top of section pages

matt-crowson-moj:
Changes shows the PER section status tag on the individual PER section pages. [JIRA](https://dsdmoj.atlassian.net/browse/P4-3157).

**Before:**

per-section-tag-before

**After:**

per-section-tag-after


ElizabethGoddard (approved):
This is perfect thank you Matt!

Thource (approved):
Nice :)

PR #2202, opened 2021-12-06
fix: Removes additional info from side panel when prison remand

matt-crowson-moj:
This no longer displays additional information for prison remand moves as this is repeating what is already on screen.

**Before:**

prison-remand-before

**After:**

prison-remand-after


ElizabethGoddard (approved):
Perfect, thanks Matt!

thomasleese (approved):

PR #2205, opened 2021-12-08
fix: Remove style override on cancel allocation banner so uses default style on new design

matt-crowson-moj:
You will only see this change on the new design.

**Before: (old design)**

image

**After: (new design)**

image

### DIFF
--- a/app/allocation/controllers/view.js
+++ b/app/allocation/controllers/view.js
@@ -3,7 +3,7 @@ const { sortBy } = require('lodash')
 const presenters = require('../../../common/presenters')
 
 function viewAllocation(req, res) {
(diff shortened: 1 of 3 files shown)


thomasleese (approved):

ElizabethGoddard:
Hi @matt-crowson-moj I'm not sure if I can test this? I did try - on staging I cleared my cookies so that I have the old move page design and cancelled an allocation. The old style banner is there so thats good. Then I opted-in to the new page design and then cancelled another allocation, but the old style banner is still there. Should I be seeing the new style in that case? Maybe it's just not possible to test though - not sure?

PR #2206, opened 2021-12-09
fix: Show the flash banner when the to location is changed on a move

matt-crowson-moj:
Prior to this change no flash banner was presented to the end user when the to location was changed.  Now the following is displayed to the user upon changing the to location:

image


thomasleese (approved):

Thource (approved):

ElizabethGoddard:
This is perfect Matt, thank you

PR #2213, opened 2022-01-10
Hide record handover button if move is cancelled

thomasleese:
It doesn't make sense that the handover can be recorded once a move has been cancelled, so the button shouldn't be visible.

### Old

Screenshot 2022-01-10 at 13 30 50

### New

Screenshot 2022-01-10 at 13 31 01


Thource (approved):

matt-crowson-moj (approved):
LGTM

PR #2221, opened 2022-02-02
Add journey dates to journeys view

thomasleese:
This adds the ability to see the journey dates when viewing journeys on a move. This hasn't been designed in particular, but since the view is limited only to CDMs at the moment I think that's okay.

### Before

Screenshot 2022-02-03 at 08 03 22

### After

Screenshot 2022-02-03 at 08 03 06

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3337)

### DIFF
--- a/common/lib/api-client/models/journey.js
+++ b/common/lib/api-client/models/journey.js
@@ -4,6 +4,7 @@ module.exports = {
     billable: '',
     vehicle: '',
     timestamp: '',
+    date: '',
     from_location: {
       jsonApi: 'hasOne',
       type: 'locations',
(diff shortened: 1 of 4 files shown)


matt-crowson-moj (approved):
One comment I do have is around the date format.

Would it be better/nicer to have 1st January 2022 type format?  Looks like there is space for it.

thomasleese:
> Would it be better/nicer to have 1st January 2022 type format? Looks like there is space for it.

Good point, I'll put that in. I was thinking this is just really for debugging at the moment since very few users see this page, but it's reasonable to format it well.

PR #2259, opened 2022-03-04
Handle case when field doesn't have an id

thomasleese:
This is sometimes the case, particularly with a radio buttons where there isn't a single element with a single ID. Instead of having a link which doesn't go anywhere, we don't need to have a link.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3493)

### DIFF
--- a/common/helpers/field/add-errorlist-to-errors.js
+++ b/common/helpers/field/add-errorlist-to-errors.js
@@ -9,7 +9,7 @@ function addErrorListToErrors(errors = {}, fields) {
         ...field,
         ...error,
       }),
-      href: `#${field.id}`,
+      href: field.id ? `#${field.id}` : undefined,
     }
   })
 


common/helpers/field/add-errorlist-to-errors.js
```diff
@@ -9,7 +9,7 @@ function addErrorListToErrors(errors = {}, fields) {
         ...field,
         ...error,
       }),
-      href: `#${field.id}`,
```
Thource, line 12: If field.id is always a string or undefined, we could just have `href: field.id` here

Thource (approved):

common/helpers/field/add-errorlist-to-errors.js
thomasleese: Unfortunately not, we need the `#` at the start of the string to represent the link to the element.

Thource: Oops, I missed the `#`, carry on :D

PR #2261, opened 2022-03-04
Display "What’s new" banner content

georgeschena:
### What changed

We have added a What's new content banner to the application as well as a footer link

### Why did it change

To allow users to get regular updates of changes to the application

### Issue tracking

- [P4-3457](https://dsdmoj.atlassian.net/browse/P4-3457)

## Screenshots

Screenshot 2022-03-04 at 15 06 21

#### Automated testing

#### Manual testing

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer


app/home/controllers.js
```diff
@@ -1,8 +1,10 @@
 const dateHelpers = require('../../common/helpers/date')
+const contentfulService = require('../../common/services/contentful')
 
-function dashboard(req, res) {
+async function dashboard(req, res) {
   const currentWeek = dateHelpers.getCurrentWeekAsRange()
   const today = new Date().toISOString()
+  const content = await contentfulService.fetchEntries()
```
thomasleese, line 7:
```suggestion
  const whatsNewContent = await contentfulService.fetchEntries()
```

I'm thinking this might make it a little clearer which content the variable contains.

common/templates/layouts/base.njk
```diff
@@ -149,6 +150,14 @@
       </div>
     {% endif %}
   {% endblock %}
+  
+  {% if REQUEST_PATH === '/' %}
```
thomasleese, line 154:
```suggestion
  {% if whatsNewContent %}
```

I think having something like this would make the template more flexible, and it makes it easier for us to show banner content on a different page if we wanted to.

common/components/whats-new-banner/template.njk
```diff
@@ -0,0 +1,14 @@
+<div id="whats-new-banner" class="whats-new-banner" data-module="app-whats-new-banner" data-content-date="{{ params.contentDate }}">
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row banner-padding">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">{{ params.contentTitle }}</h1>
+        <p class="govuk-body">{{ params.contentDate }}: {{ params.contentBannerText }}.<p>
```
thomasleese, line 6: I reckon in this component we don't need the `content` prefix as it feels superfluous.

```suggestion
        <h1 class="govuk-heading-l">{{ params.title }}</h1>
        <p class="govuk-body">{{ params.date }}: {{ params.body }}.<p>
```

common/components/whats-new-banner/whats-new-banner.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      if (c.indexOf(name) === 0) {
+        return c.substring(name.length, c.length)
+      }
+    }
+
+    return ''
+  },
+  getContnetDate: function (element) {
```
thomasleese, line 61:
```suggestion
  getContentDate: function (element) {
```

common/components/whats-new-banner/whats-new-banner.js
```diff
@@ -0,0 +1,72 @@
...
+        c = c.substring(1)
+      }
+
+      if (c.indexOf(name) === 0) {
+        return c.substring(name.length, c.length)
+      }
+    }
+
+    return ''
+  },
```
thomasleese, line 60: Code looks good, I wonder if there's a library we can use for this to make it more resilient? In the server we use https://www.npmjs.com/package/cookie-parser but I realise it's not so easy to use a library like that on the client-side.

common/assets/scss/components/_whats-new-banner.scss
```diff
@@ -0,0 +1,17 @@
+.whats-new-banner {
+	width: 100% ; 
+	height: 279px;
+	left: 0px;
+	top: 178px;
```
thomasleese, line 5: Is there a way we can make these numbers more dynamic? I'm a little nervous of what might happen if a user zooms in or has custom font sizes, whether the text might get cut off. cc @ElizabethGoddard

thomasleese (commented):
Nice work on this, overall looks good, I've left a few comments on a few things that came to mind!

common/templates/layouts/base.njk
georgeschena: After speaking to Lizzie, we only want to show the banner on the home page and nowhere else. I guess if we need it on every page we can always refactor it to do so, but I feel for now this is okay?

common/components/whats-new-banner/whats-new-banner.js
georgeschena: I did look around at some libraries to do this. But I have used this exact code snippet in a fair few client side applications and has done well AFAIK, and reads fairly well.

common/assets/scss/components/_whats-new-banner.scss
georgeschena: Good shout, Ill do this now.

app/home/controllers.js
matt-crowson-moj, line 7: I am splitting hairs a little here but the fact we are using Contentful I see more as implementation details. Would this make more sense to name the service after the specific e.g. BannerService or something to that effect.  So instead of fetchEntries you have something more akin to fetchWhatsNewContent() or something like that. 

Just a thought.

common/templates/layouts/base.njk
thomasleese: Hmm, I was thinking it feels wrong to have `REQUEST_PATH === '/'` as it makes the template quite tied to the logic of the controller, and makes an assumption that the dashboard will always be at `/`. It's not a big deal, but it feels nicer to me to not rely on the path in the template and instead rely on whether the data is available.

app/home/controllers.js
thomasleese: Yeah, I think this is worth considering, it's more like a `WhatsNewContentService` I reckon, and that makes things a little clearer down the road if we wanted to put in a `MaintenanceContentService` but pass it to the same banner component or page template.

common/templates/layouts/base.njk
georgeschena: Yeahh I do agree it did feel a tad `dirty` when I was doing it.

app/home/controllers.js
georgeschena: I shall make this change in another PR, as this PR is pretty bulky as it is!

ElizabethGoddard:
Thanks @georgeschena, looking good! The padding top and bottom looks a little off the mocks though, could we try and align to them? So in my mocks I have 50 top and 60 bottom. In the screenshots it looks a little too small at the top and a little too large at the bottom. Thanks! Lizzie

thomasleese (approved):
I'm happy with the code change. The only thing is we should be aware that deploying this will put the banner live in production, so we should make sure there's no test content left in Contentful!

georgeschena:
> I'm happy with the code change. The only thing is we should be aware that deploying this will put the banner live in production, so we should make sure there's no test content left in Contentful!

Just for visibility to others. We will be setting a feature flag for this banner until we have got our first bit of publishable content!

SophieDodds (commented):
@georgeschena The title should be "What's new" with lower case N on "new" but if this is just test content then don't worry. Other than that it looks fine to me thanks (minus the padding issue Lizzie has already flagged - agree it looks a bit padding-heavy on bottom).

PR #2274, opened 2022-03-08
Fix incorrect use of strings which expect data

thomasleese:
We're using strings which expect data to to be passed with them but without actually passing any data. This leads to the placeholder text remaining visible for users.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3546)

### Old

Screenshot 2022-03-08 at 15 04 22
Screenshot 2022-03-08 at 15 04 37

### New

Screenshot 2022-03-08 at 15 04 49
Screenshot 2022-03-08 at 15 04 56

### DIFF
--- a/app/moves/constants.js
+++ b/app/moves/constants.js
@@ -95,7 +95,7 @@ const FILTERS = {
       status: 'rejected',
     },
     {
-      label: 'collections::cancelled',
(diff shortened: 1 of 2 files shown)


Thource (approved):

ElizabethGoddard:
Looks good @thomasleese

PR #2284, opened 2022-03-10
Add WHATS_NEW_BANNER feature flag 

georgeschena:
Add `BANNER_VISIBLE` feature flag  to allow us to have the option to show or hide the whats new banner


thomasleese (commented):
Looks good, my only suggestion would be that we already have a section for feature flags in this config, which I think is worth following that pattern for this: https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/37f18db853732f4cedc7ec548a1fb85a35363c72/config/index.js#L288-L290

georgeschena:
> Looks good, my only suggestion would be that we already have a section for feature flags in this config, which I think is worth following that pattern for this:
> 
> https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/37f18db853732f4cedc7ec548a1fb85a35363c72/config/index.js#L288-L290

I literally just saw this!

config/index.js
```diff
@@ -287,6 +287,7 @@ module.exports = {
   },
   FEATURE_FLAGS: {
     GOT: /true/i.test(process.env.FEATURE_FLAG_GOT),
+    BANNER_VISIBLE: /true/i.test(process.env.FEATURE_FLAG_BANNER_VISIBLE),
```
thomasleese, line 290:
```suggestion
    WHATS_NEW_BANNER: /true/i.test(process.env.FEATURE_FLAG_WHATS_NEW_BANNER),
```

I think this is a clearer name, but I don't really mind since we'll get rid of this feature flag when we release!

thomasleese (approved):

PR #2291, opened 2022-03-11
Rename contentfulService to whatsNewContentService

georgeschena:
Rename contentfulService to whatsNewContentService for better readability.

### DIFF
--- a/common/services/whats-new-content.js
+++ b/common/services/whats-new-content.js
@@ -36,7 +36,7 @@ const service = {
     space: CONTENTFUL_SPACE_ID,
     accessToken: CONTENTFUL_ACCESS_TOKEN,
   }),
-  fetchEntries: async () => {
+  fetchWhatsNewContent: async () => {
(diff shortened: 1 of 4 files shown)


matt-crowson-moj (approved):

common/services/whats-new-content.js
```diff
@@ -36,7 +36,7 @@ const service = {
     space: CONTENTFUL_SPACE_ID,
     accessToken: CONTENTFUL_ACCESS_TOKEN,
   }),
-  fetchEntries: async () => {
+  fetchWhatsNewContent: async () => {
```
thomasleese, line 39: I reckon we could call this "fetch", since we're already in the what's new service, it can only be what's new content.

thomasleese (approved):
I've left a comment, but not major.

common/services/whats-new-content.js
georgeschena: Good shout

PR #2324, opened 2022-03-19
Refactor how we retrieve content/entries from Contentful

georgeschena:
This PR has refactored the way we pull in content from Contentful, as well as how we create content. What we have done now is allow someone to create multiple bits of content independently, so instead of having one large bit of content we now split it into many, this allows us to have greater flexibility of how we can present the content to the end user.


app/whats-new/whats-new.njk
```diff
@@ -1,19 +1,35 @@
...
+      What's new
+    </h1>
+  </header>
+
+  <p class="govuk-!-margin-bottom-5">
+    The latest updates and improvements to Book a Secure Move.
+  </p>
+
+  {% for post in posts %}
+    {% if post.fields.body and post.fields.title %}
```
thomasleese, line 19: Is this `if` statement necessary? I guess it feels like the body and the title will always be set, even if it's an empty string?

common/services/whats-new-content.js
```diff
@@ -49,22 +49,30 @@ const service = {
       return null
     }
 
+    for (let i = 0; i < entries.items.length; i++) {
+      const currentItem = entries.items[i]
+      currentItem.fields.body = service.convertToHTMLFormat(
+        currentItem.fields.body
+      )
+      currentItem.fields.date = service.formatDate(currentItem.fields.date)
+    }
```
thomasleese, line 58:
```suggestion
    const entries = entries.items.forEach(({ fields: { title, body, date } }) => (
      {
        title: title,
        body: service.convertToHTMLFormat(body),
        date: service.formatDate(date)
      }
    ))
```

I wonder if something like this is slightly nicer, allows us to be more explicit on which fields we want to pass through to the template as well?

common/templates/layouts/base.njk
```diff
@@ -153,9 +153,9 @@
   
   {% if FEATURE_FLAGS.WHATS_NEW_BANNER and REQUEST_PATH === '/' %}
     {{ appWhatsNewBanner({
-      title: whatsNewContent.title,
-      body: whatsNewContent.bannerText,
-      date: whatsNewContent.date
+      title: whatsNewContent.bannerContent.title,
+      body: whatsNewContent.bannerContent.body,
+      date: whatsNewContent.bannerContent.date
```
thomasleese, line 158:
```suggestion
      whatsNewContent.bannerContent
```

I think we can do `appWhatsNewBanner(whatsNewContent.bannerContent)` now without the need to pass the variables individually.

thomasleese (commented):
Thanks for making this change so quickly! I've got a few comments, but nothing major.

app/whats-new/whats-new.njk
```diff
@@ -1,19 +1,35 @@
...
+      </h2>
+
+      <hr>
+      
+      <h3 class="govuk-heading-l govuk-!-margin-top-5">
+        {{ post.fields.title }}
+      </h3>
+        
+      {{ post.fields.body | safe }}
+      {% else %}
```
matt-crowson-moj, line 31: This might be a daft question but what purpose does the `{% else % }` serve?   i.e. is it needed?

matt-crowson-moj (approved):

app/whats-new/whats-new.njk
georgeschena: Was going to originally have a message if we had no content to show, but think thats changed, so the else is not needed now

common/services/whats-new-content.js
georgeschena: I do like this, nicer to read 🚀

thomasleese (approved):

Thource (approved):

PR #2331, opened 2022-03-22
feat: If contetnful is down dont render the banner

georgeschena:
### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/components/whats-new-banner/template.njk
+++ b/common/components/whats-new-banner/template.njk
@@ -1,14 +1,16 @@
-<div id="whats-new-banner" class="whats-new-banner" data-module="app-whats-new-banner" data-content-date="{{ params.date }}">
-  <div class="govuk-width-container">
-    <div class="govuk-grid-row banner-padding">
-      <div class="govuk-grid-column-two-thirds">
-        <h1 class="govuk-heading-l">{{ params.title }}</h1>
-        <p class="govuk-body">{{ params.date }}: {{ params.body }}.<p>
-        <a href="/whats-new" id="whats-new-page-link" class="govuk-link">Read more about these changes.</a>
-      </div>
-      <div class="govuk-grid-column-one-third">
-        <a href="#" id="dismiss-banner-button" class="govuk-link dismiss-button">Dismiss</a>
+{% if not params.error %}
+  <div id="whats-new-banner" class="whats-new-banner" data-module="app-whats-new-banner" data-content-date="{{ params.date }}">
+    <div class="govuk-width-container">
+      <div class="govuk-grid-row banner-padding">
+        <div class="govuk-grid-column-two-thirds">
+          <h1 class="govuk-heading-l">{{ params.title }}</h1>
+          <p class="govuk-body">{{ params.date }}: {{ params.body }}.<p>
+          <a href="/whats-new" id="whats-new-page-link" class="govuk-link">Read more about these changes.</a>
+        </div>
+        <div class="govuk-grid-column-one-third">
+          <a href="#" id="dismiss-banner-button" class="govuk-link dismiss-button">Dismiss</a>
+        </div>
       </div>
     </div>
   </div>
-</div>
+{% endif %}
--- a/common/services/whats-new-content.js
+++ b/common/services/whats-new-content.js
@@ -45,6 +45,10 @@ const service = {
   fetch: async () => {
     const entries = await service.client.getEntries()
 
+    if (!entries.items && entries.sys.id === 'ServerError') {
+      return { error: 'ServerError', errorMessage: entries.message }
+    }
+
     if (!entries.items?.length) {
       return null
     }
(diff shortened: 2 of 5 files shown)


Thource (approved):

common/services/whats-new-content.js
```diff
@@ -45,6 +45,10 @@ const service = {
   fetch: async () => {
     const entries = await service.client.getEntries()
 
+    if (!entries.items && entries.sys.id === 'ServerError') {
+      return { error: 'ServerError', errorMessage: entries.message }
+    }
+
```
thomasleese, line 51: Do we know for sure this is what happens if there's an error from the client?

I'm just wandering [because the documentation suggests that it raises an exception](https://github.com/contentful/contentful.js#your-first-request):

Screenshot 2022-03-23 at 07 31 31

common/components/whats-new-banner/template.njk
```diff
@@ -1,14 +1,16 @@
...
-  <div class="govuk-width-container">
-    <div class="govuk-grid-row banner-padding">
-      <div class="govuk-grid-column-two-thirds">
-        <h1 class="govuk-heading-l">{{ params.title }}</h1>
-        <p class="govuk-body">{{ params.date }}: {{ params.body }}.<p>
-        <a href="/whats-new" id="whats-new-page-link" class="govuk-link">Read more about these changes.</a>
-      </div>
-      <div class="govuk-grid-column-one-third">
-        <a href="#" id="dismiss-banner-button" class="govuk-link dismiss-button">Dismiss</a>
+{% if not params.error %}
```
thomasleese, line 1: I wonder if we could do `if params.title && params.date && params.body` here, which would support the error case but also allow us to better handle the case when there are no entries, and therefore no banner to show?

app/whats-new/index.js
```diff
@@ -4,8 +4,15 @@ const router = require('express').Router()
...
 // Define routes
-router.get('/', async (req, res) => {
+router.get('/', async (req, res, next) => {
   const content = await whatsNewContentService.fetch()
+
+  if (content.error === 'ServerError') {
+    const error = new Error('Contentful servers are down')
+    error.statusCode = 500
+    return next(error)
+  }
```
thomasleese, line 14: If we get this to raise an exception rather than return something, this code might end up slightly nicer I think.

```js
try {
  const content = await whatsNewContentService.fetch()
} catch (e) {
  return next(e)
}
```

This also means we can move the responsibility for raising the exception in the whats new content service.

thomasleese (commented):
I realise this is already merged in, but I've got a few comments.

app/whats-new/index.js
thomasleese: In fact, I'm not even sure we need the `catch (e) { return next(e) }` part, we could get Express to catch the exception for us and propagate it up and render the error page.

common/services/whats-new-content.js
georgeschena: From reading [Contentful Errors](https://www.contentful.com/developers/docs/references/errors/), they say 

`Errors share the same common attributes as other API resources. They have a sys property, where sys.type is always "Error" and a sys.id that identifies the exact error code (see table below) that occurred. They always have a message property which will be a short description of what went wrong.`

And when I curl the endpoint with an error I get this format back

```
{
  "sys": {
    "type": "Error",
    "id": "NotFound"
  },
  "message": "The resource could not be found.",
  "requestId": "c08a4c02-cb9e-43ea-9c61-79be44a0786c"
}
```

thomasleese: Yeah, I'm just wondering if the Node client automatically handles that API response for us and raises an exception, rather than returning it. The documentation above with the `client.getEntry(...).catch(err => console.log(err))` line suggests that it does, but it isn't entirely clear and explicit about it.

Have you tried requesting that endpoint with the Node client rather than with `curl`?

georgeschena: So when I do it in with Node I get:

```
await contentful.createClient({accessToken: "sdsdsds", space: "m5k1kmk3zqw"}).getEntries().then(entry => console.log(entry))
Uncaught:
NotFound: {
  "status": 404,
  "statusText": "Not Found",
  "message": "The resource could not be found.",
  "details": {},
  "request": {
    "url": "entries",
    "headers": {
      "Accept": "application/json, text/plain, */*",
      "Content-Type": "application/vnd.contentful.delivery.v1+json",
      "X-Contentful-User-Agent": "sdk contentful.js/9.1.10; platform node.js/v16.14.0; os macOS/v16.14.0;",
      "Authorization": "Bearer ...sdsds",
      "user-agent": "node.js/v16.14.0",
      "Accept-Encoding": "gzip"
    },
    "method": "get"
  },
  "requestId": "f39c6bfd-d02e-4b19-b746-441922e210cf"
}
```

Not clear at all in the error docs. But okay at least we know for certain we need to catch it.

PR #2340, opened 2022-03-23
Refactor Contenful error handling

georgeschena:


app/whats-new/index.js
```diff
@@ -5,12 +5,12 @@ const whatsNewContentService = require('../../common/services/whats-new-content'
 
 // Define routes
 router.get('/', async (req, res, next) => {
```
thomasleese, line 7:
```suggestion
router.get('/', async (req, res) => {
```

Since we don't use this. I'm also wondering if this is why Express wasn't catching the error for us...

common/components/whats-new-banner/template.njk
```diff
@@ -1,4 +1,4 @@
-{% if not params.error %}
+{% if params.title and params.date and params.body %}
```
thomasleese, line 1: I'm wondering whether we need this now and instead should modify where it's called:

https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/07b7dedbbeb3f67a01527a18e4a7f1195b9ba322/common/templates/layouts/base.njk#L154-L156

To include the check for the `whatsNewContent`:

```
  {% if FEATURE_FLAGS.WHATS_NEW_BANNER and REQUEST_PATH === '/' and whatsNewContent %}
    {{ appWhatsNewBanner(whatsNewContent.bannerContent) }}
  {% endif %}
```

Because I'm wondering what happens right now when `whatsNewContent` is set to `null` (either because there are no entries or because Contentful is down) and the template tries to do `whatsNewContent.bannerContent`, would that cause the entire dashboard to fail to render?

thomasleese (commented):
Nice improvement, I've just got a couple of comments.

app/whats-new/index.js
Thource: `next` is used at line 13

thomasleese: Oh yes, that was a poorly worded comment.

I was still trying to figure out why we needed to call `next` explicitly and why Express wasn't catching the error for us (as this is something we do in numerous other places) and wondered if this could be the reason.

Thource: Ah, I see what you mean now. @georgeschena what happens without the try/catch?

common/components/whats-new-banner/template.njk
Thource: Adding to this, it feels better to have the conditional outside of the template too.

common/components/whats-new-banner/whats-new-banner.yaml
```diff
@@ -11,10 +11,11 @@ params:
...
+      title: 'Title Here'
+      date: '4th March 2022'
+      body: 'Some random banner text'
   - name: withError
     data:
-      error: 'ServerError'
-      errorMessage: 'Server error occured.'
+      title: null
+      date: null
+      body: null
```
thomasleese, line 21: I don't think we need this anymore since we'd always expect these to be set for this component.

thomasleese (approved):
Nice, this looks good! I'd suggest perhaps squashing some of the commits together, but not a big deal.

PR #2342, opened 2022-03-24
Add lockout warning to listing

georgeschena:
### What changed

When the move has an event of `MoveLockout` present in its `timeline_events` we show a warning message.

### Why did it change

So we can show the user that a move is locked out

### Issue tracking

- [P4-3340](https://dsdmoj.atlassian.net/browse/P4-3340)

## Screenshots

Screenshot 2022-03-24 at 15 22 58

### DIFF
--- a/app/moves/middleware/set-results.moves.js
+++ b/app/moves/middleware/set-results.moves.js
@@ -14,6 +14,12 @@ function setResultsMoves(bodyKey, locationKey) {
       const args = omitBy(req.body[bodyKey], isUndefined)
       const activeMoves = await req.services.move.getActive(args)
 
+      for (let i = 0; i < activeMoves.length; i++) {
+        const move = activeMoves[i]
+        const eventsForMove = await req.services.move.getByIdWithEvents(move.id)
+        move.timeline_events = eventsForMove.timeline_events
+      }
+
       if (groupBy === 'vehicle') {
         req.resultsAsCards = presenters.movesByVehicle({
           moves: activeMoves,
--- a/common/components/card/template.njk
+++ b/common/components/card/template.njk
@@ -1,6 +1,7 @@
 {% from "tag/macro.njk" import appTag %}
 {% from "moj/components/badge/macro.njk" import mojBadge %}
 {% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
 
 {% set headingLevel = params.headingLevel if params.headingLevel else 2 %}
 
@@ -68,5 +69,15 @@
     {% if params.insetText %}
       {{ govukInsetText(params.insetText) }}
     {% endif %}
+
+    {% if params.isLockOut %}
+      <div class="card-warning-text">
+        {{ govukWarningText({
+          text: "Prison lockout",
+          iconFallbackText: "Warning"
+        }) }}
+      </div>
+    {% endif %}
+
   </div>
 </div>
--- a/common/presenters/move-to-card-component.js
+++ b/common/presenters/move-to-card-component.js
@@ -9,6 +9,7 @@ function moveToCardComponent({
   showMeta = true,
   showTags = true,
   showStatus = true,
+  showIsALockout = false,
   showToLocation = false,
   showFromLocation = false,
   hrefSuffix = '',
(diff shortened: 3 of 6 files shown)


app/moves/middleware/set-results.moves.js
```diff
@@ -14,6 +14,12 @@ function setResultsMoves(bodyKey, locationKey) {
       const args = omitBy(req.body[bodyKey], isUndefined)
       const activeMoves = await req.services.move.getActive(args)
 
+      for (let i = 0; i < activeMoves.length; i++) {
+        const move = activeMoves[i]
+        const eventsForMove = await req.services.move.getByIdWithEvents(move.id)
+        move.timeline_events = eventsForMove.timeline_events
+      }
```
thomasleese, line 21:
```suggestion
      activeMoves.forEach((move) => {
        const eventsForMove = await req.services.move.getByIdWithEvents(move.id)
        move.timeline_events = eventsForMove.timeline_events
      })
```

I'm a little concerned about the performance implications of this though, it means we'd have to make an API call for every move, which [can lead to N+1 problems](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) where the site gets slower and slower with the more moves you need to render.

Is there a way we could change the API to include this information already when you request the list of moves? I'm thinking either the `timeline_events` could be included directly, or if that in itself is quite expensive, could we add a field which marks whether the move has a `MoveLockout` event or not? In this way the API is doing the query against the database directly and should be more performant.

common/components/card/template.njk
```diff
@@ -68,5 +69,15 @@
     {% if params.insetText %}
       {{ govukInsetText(params.insetText) }}
     {% endif %}
+
+    {% if params.isLockout %}
+      <div class="card-warning-text">
+        {{ govukWarningText({
+          text: "Prison lockout",
```
thomasleese, line 76: I think we should move this text in to a [locales](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/tree/main/locales/en) file.

common/presenters/move-to-card-component.js
```diff
@@ -44,6 +45,17 @@ function moveToCardComponent({
...
 
+    if (move.timeline_events) {
+      for (let i = 0; i < move.timeline_events.length; i++) {
+        const eventType = move.timeline_events[i].event_type
+
+        if (eventType === 'MoveLockout') {
+          showIsALockout = true
+          break
+        }
+      }
```
thomasleese, line 56:
```suggestion
      showIsALockout = move.timeline_events.some(({ event_type }) => event_type === 'MoveLockout')
```

[`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is a method on arrays and returns true if one of the elements of the array passes the function provided to it.

thomasleese (commented):
Screenshot looks good, I have some concerns about the performance of this way of doing it in the frontend, and think it's worth considering moving the logic in to the backend.

common/presenters/move-to-card-component.js
georgeschena: Niceee!

app/moves/middleware/set-results.moves.js
georgeschena: Yeah I fully agree, would be good so when we call `req.services.move.getActive(args)` that it also returns the `timeline_events` with it.

georgeschena: For visibility we have gone with the approach of sending over the `is_lockout` attribute from the api in this PR: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1795

Thource (approved):

common/presenters/move-to-card-component.js
```diff
@@ -9,6 +9,7 @@ function moveToCardComponent({
   showMeta = true,
   showTags = true,
   showStatus = true,
+  showIsALockout = false,
```
thomasleese, line 12: I'm not sure we need this argument, since we set it to true if the move is a lockout anyway.

I think it's working differently to the other `show` arguments, where they decide whether or not to include this information in the component, whereas with `showIsALockout` if you set it to `false`, it'll show it regardless if the move is a lockout.

thomasleese: Since we have no need to hide the lockout message, I think we shouldn't give people the option, at least for now.

georgeschena: Ahh yes this makes sense, I forgot to update these tests to reflect the api change a little better

common/components/card/template.njk
```diff
@@ -68,5 +69,15 @@
     {% if params.insetText %}
       {{ govukInsetText(params.insetText) }}
     {% endif %}
+
+    {% if params.isLockout %}
+      <div class="card-warning-text">
```
thomasleese, line 74:
```suggestion
      <div class="<div class="govuk-!-padding-top-3"></div>">
```

We can use the padding styles from the design system here: https://design-system.service.gov.uk/styles/spacing/

thomasleese (approved):

ElizabethGoddard (approved):
Looks great, thank you @georgeschena!

PR #2357, opened 2022-03-29
Change flag colour of 'DEATH IN CUSTODY' from red to grey

georgeschena:
### What changed

When an `event_type` is `PersonMoveDeathInCustody` we use the `app-tag--inactive` class instead of the `app-tag--destructive` class

### Why did it change

The purpose of this to change the colour of the flag ‘DEATH IN CUSTODY’ from red to grey. This is because it is **NOT** a risk, but rather ‘general information’.

### Issue tracking

- [P4-3520](https://dsdmoj.atlassian.net/browse/P4-3520)

## Screenshots

Screenshot 2022-03-29 at 11 46 57
Screenshot 2022-03-29 at 11 47 23

### DIFF
--- a/locales/en/events.json
+++ b/locales/en/events.json
@@ -362,7 +362,7 @@
     "description": ""
   },
   "PersonMoveDeathInCustody": {
-    "flag": "red",
+    "flag": "grey",
     "heading": "Death in custody",
     "description": "{{supplier_personnel_numbers, oxfordJoin}} reported a death in custody at {{reported_at, formatTime}} at {{location.title}}<br>$t(events::select_fault_classification)<br>$t(events::select_incident_vehicle)"
   },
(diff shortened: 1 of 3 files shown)


locales/en/events.json
```diff
@@ -362,7 +362,7 @@
     "description": ""
   },
   "PersonMoveDeathInCustody": {
-    "flag": "red",
+    "flag": "grey",
```
thomasleese, line 365: I'm actually wondering if these are still used, I noticed some of the events don't have one.

thomasleese (approved):

locales/en/events.json
georgeschena: Yeah I thought changing that would have done it, but it did not. I changed the flag value anyway just in case it causes confusion in the future.

georgeschena: I know the `heading` and `description` are used defo being used

ElizabethGoddard (approved):
Looks great, thank you @georgeschena!

PR #2361, opened 2022-03-30
Update identity bar at relevant points in the move

georgeschena:
### What changed

When a move is locked out we show a prison lockout banner warning message, and we show all the journeys which are not cancelled or rejected to the user.

### Why did it change

As a user I want to know where and when a move is stopping overnight for a lockout so that I know the whereabouts of the detainee.

### Issue tracking

- [P4-3342](https://dsdmoj.atlassian.net/browse/P4-3342)

## Screenshots

Screenshot 2022-04-04 at 13 54 20
Screenshot 2022-04-04 at 13 57 15

### DIFF
--- a/common/components/identity-bar/template.njk
+++ b/common/components/identity-bar/template.njk
@@ -12,6 +12,18 @@
           {{ params.heading.html | safe if params.heading.html else params.heading.text }}
         </h1>
 
+        {% if params.journeys | length %}
+          <p class="app-identity-bar__summary">
+            {% for journey in params.journeys %}
+              {% if loop.first %}
+                {{ t("moves::detail.page_heading_summary", journey) | safe }}
+              {% else %}
+                <br />{{ t("moves::detail.page_heading_summary_and", journey) | safe }}
+              {% endif %}
+            {% endfor %}
+          </p>
+        {% endif %}
+
         {% if params.summary.html or params.summary.text %}
           <p class="app-identity-bar__summary">
             {{ params.summary.html | safe if params.summary.html else params.summary.text }}
--- a/common/components/identity-bar/template.njk
+++ b/common/components/identity-bar/template.njk
@@ -18,7 +18,7 @@
               {% if loop.first %}
                 {{ t("moves::detail.page_heading_summary", journey) | safe }}
               {% else %}
-                <br />{{ t("moves::detail.page_heading_summary_and", journey) | safe }}
+                {{ t("moves::detail.page_heading_summary_and", journey) | safe }}
               {% endif %}
             {% endfor %}
           </p>
--- a/app/move/app/view/index.js
+++ b/app/move/app/view/index.js
@@ -8,6 +8,7 @@ const {
   setMoveWithEvents,
   setPersonEscortRecord,
   setYouthRiskAssessment,
+  setJourneys,
 } = require('../../middleware')
 
 const {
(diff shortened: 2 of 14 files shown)


app/move/middleware.js
```diff
@@ -29,8 +29,16 @@ module.exports = {
 
     try {
       const move = await req.services.move.getByIdWithEvents(moveId)
+      const journeys = await req.services.journey.getByMoveId(moveId)
+      journeys.shift()
```
georgeschena, line 33: This `shift()` method is used as we already show the latest move in the identity bar. So when we render all the Journeys we remove the latest one, as it's already shown

common/components/identity-bar/template.njk
```diff
@@ -16,6 +16,26 @@
...
+          {% if params.isLockedOut %}
+            {% for journey in params.journeys %}
+              <p class="app-identity-bar__summary">
+                and from <b>{{ journey.from_location.title }}</b>
+                to
+                {% if journey.to_location.title %}
+                  <b>{{ journey.to_location.title }}</b>
+                {% else %}
+                  <b> (awaiting destination) </b>
+                {% endif %}
```
thomasleese, line 29: A journey will always have a `to_location` and a `from_location`, what we might see instead in the case of a lockout is that the journeys itself may not exist yet. The supplier might be aware of the lockout, but won't have scheduled the journeys yet.

So I think the condition here should be, if it's a lockout, but the move has still only got 0 or 1 journeys, to show `(awaiting destination)` and `(awaiting date)`.

georgeschena: Okay, that makes sense, wasn't sure of the flow for a supplier. This makes sense though!

ElizabethGoddard:
Hey @georgeschena could we remove the line break and see how it looks both in the normal header and the condensed header? It hadn't crossed my mind that with the line break it would look like this. Thank you!

app/move/app/view/middleware/locals.identity-bar.js
```diff
@@ -28,6 +28,7 @@ function localsIdentityCard(req, res, next) {
         date: filters.formatDate(move.date),
       }),
     },
+    isLockout: move.is_lockout,
```
thomasleese, line 31: I think we can remove this from the identity bar now.

app/move/middleware.js
```diff
@@ -29,8 +29,11 @@ module.exports = {
 
     try {
       const move = await req.services.move.getByIdWithEvents(moveId)
+      const journeys = await req.services.journey.getByMoveId(moveId)
+
       populateResources(move.timeline_events, req)
       req.move = move
+      req.journeys = journeys
```
thomasleese, line 36: I think by having this in `setMoveWithEvents` it won't work on tabs which only use the `setMove` middleware (specifically the details tab).

georgeschena: Ahh yes, the details tab does not show the journeys, ill add journeys to `setMove`

app/move/middleware.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
 
     try {
       req.move = await req.services.move.getById(moveId)
+      req.journeys = await req.services.journey.getByMoveId(moveId)
```
thomasleese, line 17: I'm wondering if this belongs in its own middleware, `setJourneys` for example, as then it can be tested independently.

We could also then refactor the code a little bit so [the journeys controller](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/main/app/move/controllers/journeys.js) uses the same middleware.

thomasleese (approved):
Nice!

ElizabethGoddard (approved):
This looks awesome thanks George!

PR #2384, opened 2022-04-06
Update side panel at relevant points in the move

georgeschena:
### What changed

We now add information to the side panel to inform users where and when a move is stopping overnight for a lockout 

### Why did it change

So that as user knows the whereabouts of the detainee throughout the move

### Issue tracking

- [P4-3343](https://dsdmoj.atlassian.net/browse/P4-3343)

## Screenshots

Screenshot 2022-04-06 at 11 05 24

### DIFF
--- a/common/presenters/move-to-meta-list-component.js
+++ b/common/presenters/move-to-meta-list-component.js
@@ -1,4 +1,4 @@
-const { get, isNil } = require('lodash')
+const { isNil } = require('lodash')
 
 const moveAgreedField = require('../../app/move/app/new/fields/move-agreed')
 const componentService = require('../../common/services/component')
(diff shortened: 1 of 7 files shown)


common/presenters/move-to-meta-list-component.test.js
```diff
@@ -160,93 +173,6 @@ describe('Presenters', function () {
       })
     })
 
-    context('with only `date from`', function () {
```
thomasleese, line 163: I think we need to keep this behaviour for moves which don't have a date, in that case we don't even need to show the journeys summary as the `move.date` will be `null`.

common/presenters/move-to-meta-list-component.test.js
```diff
@@ -348,53 +277,16 @@ describe('Presenters', function () {
         })
       })
 
-      context(
```
thomasleese, line 351: I'm also wondering about this logic and whether we need to keep it.

ElizabethGoddard:
Hi George, I assume much like the other work that the (awaiting destination) and (awaiting date) appear when relevant?

Assuming so, consider this approved from me. Thanks, Lizzie

georgeschena:
> Hi George, I assume much like the other work that the (awaiting destination) and (awaiting date) appear when relevant?
> 
> Assuming so, consider this approved from me. Thanks, Lizzie

Yeah exactly!

thomasleese (approved):

Thource:
Looking at the code, this will state `Monday 28 Mar 2022 to Tuesday 29 Mar 2022 to Wednesday 30 Mar 2022 to... (etc etc)` would `and` work better?

common/presenters/move-to-meta-list-component.js
```diff
@@ -105,7 +118,7 @@ function moveToMetaListComponent(move, { updateUrls = {} } = {}) {
         text: i18n.t('fields::date_type.label'),
       },
       value: {
-        text: filters.formatDateWithRelativeDay(date),
+        text: journeysSummary.map(({ date }) => date).join(' to '),
```
Thource, line 121: It looks like this will state `Monday 28 Mar 2022 to Tuesday 29 Mar 2022 to Wednesday 30 Mar 2022 to... (etc etc)` would and work better?

georgeschena: I am pretty neutral tbh, I am going off the content designs. As my content/design outlook is pretty bad haha

thomasleese: Although this code is generic, I'm not aware of any situation where they'll be more than 2 journeys anyway, so I don't think this will be a big deal.

Thource (approved):

PR #2385, opened 2022-04-06
Add support link to footer

georgeschena:
Add Support footer link

Screenshot 2022-04-06 at 13 22 32

### DIFF
--- a/config/nunjucks/globals.js
+++ b/config/nunjucks/globals.js
@@ -8,6 +8,7 @@ const {
   FEEDBACK_URL,
   SUPPORT_EMAIL,
   FEATURE_FLAGS,
+  IS_PRODUCTION,
 } = require('../')
 const { mountpath: componentsUrl } = require('../../app/components')
 const i18n = require('../i18n')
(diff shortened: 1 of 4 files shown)


config/nunjucks/globals.js
```diff
@@ -8,18 +8,27 @@ const {
...
 const { mountpath: componentsUrl } = require('../../app/components')
 const i18n = require('../i18n')
 const logger = require('../logger')
 const { manifest: manifestPath } = require('../paths')
 
 let webpackManifest = {}
+
+const supportLink = IS_PRODUCTION
+  ? 'https://support.hmpps.service.justice.gov.uk/feedback-and-support/book-secure-move'
+  : 'https://support-dev.hmpps.service.justice.gov.uk/feedback-and-support/book-secure-move'
```
thomasleese, line 22: I'm thinking we should move this logic in to the `config/index.js` file, in a similar way to how we have `FEEDBACK_URL` coming directly from the config.

It's not a big deal, but it means we can use this in multiple places if we needed to without needing to extra the logic out of this file.

thomasleese (approved):

config/nunjucks/globals.js
georgeschena: Good shout

Thource (approved):

thomasleese (approved):

ElizabethGoddard:
Hi George, the review button has gone for me but I am happy to approve this. Thanks, Lizzie

ElizabethGoddard:
Actually @georgeschena I just realised I can't check where the actual link goes from a screenshot - could you let me re-review once this is in staging? Thanks!

PR #2390, opened 2022-04-07
Add police custody form

georgeschena:
### Issue tracking

- [P4-XXXX]()

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations


app/move/app/view/middleware/locals.move-details.js
```diff
@@ -5,8 +5,25 @@ function localsMoveDetails(req, res, next) {
...
+
+  move.important_events.forEach(event => {
+    if (event.event_type === 'MoveLodgingStart') {
+      moveLodgingStarted = true
+      moveLodgingEnded = false
+    } else if (event.event_type === 'MoveLodgingEnd') {
+      moveLodgingStarted = false
+      moveLodgingEnded = true
+    }
+  })
```
thomasleese, line 23:
```suggestion
  res.locals.moveLodgingStarted = move.important_events.some({ event_type } => event_type === 'MoveLodgingStart')
  res.locals.moveLodgingEnded = move.important_events.some({ event_type } => event_type === 'MoveLodgingEnd')
```

app/move/app/view/views/_layout.njk
```diff
@@ -30,11 +30,20 @@
...
+      <div class="govuk-width-container prison-lockout-banner-content">
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-two-thirds">
+            {{ govukWarningText({
+              text: t("moves::is_lockout"),
+              iconFallbackText: "Warning"
+            }) }}
+          </div>
+          <div class="govuk-grid-column-one-third">
+            {% if moveLodgingStarted === true and moveLodgingEnded === false %}
```
thomasleese, line 42:
```suggestion
            {% if moveLodgingStarted and !moveLodgingEnded %}
```

common/services/event.js
```diff
@@ -0,0 +1,85 @@
...
+      let type
+      let relationshipsId
+      let supplierPersonnelNumber
+
+      const eventDescription = lockoutEvents[event]
+
+      if (PER_EVENTS.includes(event)) {
+        type = 'person_escort_records'
+        relationshipsId = move.profile.person_escort_record.id
+        supplierPersonnelNumber = 'supplier_personnel_number'
```
thomasleese, line 38:
```suggestion
        supplierPersonnelNumber = 'police_personnel_number'
```

As we're sending this up as police, it should be the police personnel number.

common/services/event.js
```diff
@@ -0,0 +1,85 @@
...
+      'PersonMoveUsedForce',
+      'PersonMoveDeathInCustody',
+    ]
+
+    const todaysDate = new Date()
+    const events = []
+
+    events.push(lockoutEvents.events)
+
+    events.flat().map(async event => {
```
thomasleese, line 28: Do we need to `await` this to get the results? I'm wondering what happens if one of these `POST`s fails, does it stop immediately or does it continue to the next one, and what does the user see?

common/services/event.js
```diff
@@ -0,0 +1,85 @@
...
+      'PerMedicalAid',
+      'PerMedicalDrugsAlcohol',
+      'PerMedicalMentalHealth',
+      'PerPropertyChange',
+      'PerGeneric',
+    ]
+    const PERSON_MOVE_EVENTS = [
+      'PersonMoveUsedForce',
+      'PersonMoveDeathInCustody',
+    ]
```
thomasleese, line 21: I'd be tempted to pull these out of the function and to the top of the file or in the classs.

common/services/event.js
```diff
@@ -0,0 +1,85 @@
...
+          relationships: {
+            eventable: {
+              data: {
+                type: type,
+                id: relationshipsId,
+              },
+            },
+            location: {
+              data: {
+                id: `${move.from_location.id}`,
```
thomasleese, line 70: This should actually be the location of where the event occurred, which in this case would be where the person is held over night.

app/police-custody-form/police-custody-form.njk
```diff
@@ -0,0 +1,430 @@
...
+          </h3>
+
+          {{ appMetaList(moveDetails) }}
+
+          <p class="govuk-!-margin-top-3 govuk-!-margin-bottom-0 govuk-!-font-size-16">
+            <a href="{{ urls.move.details }}#tabs">
+              {{ t("actions::view_or_edit_move_details") }}
+            </a>
+          </p>
+        {% endcall %}
```
thomasleese, line 420: I'm wondering if this form should actually be within the `move` app. It has a few benefits, mostly that we can use the same layout template so this wouldn't be required, and it means we can use the same `setMove` middleware.

thomasleese (commented):
Generally looks good, I've left a few comments!

common/services/event.js
Thource: Yeah I think this should be `return Promise.all(events.flat().map(async event => {...`

app/move/app/view/views/_layout.njk
georgeschena: I don't think NJK supports the `!` operator on a variable

common/services/event.js
georgeschena: Yeah agreed adding `Promise.all` will help if any of the events sent to the api errors out. 

There is a separate ticket to deal with errors on the police custody form here [P4-3515](https://dsdmoj.atlassian.net/browse/P4-3515). We can create some nice error messaging if the submission to the API fails, telling the user to retry. We will need Sophie and Lizzie to potentially create some error messaging for this type of error as I can't see any designs for a failed submission.

app/move/app/view/views/_layout.njk
Thource: NJK supports `not` instead of `!`, so maybe this should be `if moveLodgingStarted and not moveLodgingEnded`

common/services/event.js
georgeschena: Could we use the `move.to_location.id` in this case then? as the [incident events git repo](https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki/Incident-Events) says: 

`location - The location the incident occurred or the destination the vehicle was headed`

So we could use the original `to_location` as that would be the destination the vehicle was originally headed too?

app/police-custody-form/police-custody-form.njk
georgeschena: I did think about moving it to the `move` app, and I agree the benefit of the `setMove` middleware. 

I found that the layout for the police custody form page is slightly different, as it has a few less components on the side panel, it doesn't include the links to edit or view move details or the cancel move link. 

I wanted to align it in a similar way to how the person escort record form is laid out in order to have some consistency with these forms.

Thource (approved):

georgeschena:
Just want to leave this comment here. 

As Tom is off till Friday and we don't want to hold this PR back as other tickets are dependent on this PR going through. Adam and I have decided to merge this. 
Happy to have a convo around the comments Tom had left once he's back! As no one will be using this page yet so we can make changes once this PR is merged no issue. Just wanted to leave this comment here as didn't want it to come across as me ignoring the PR comments :)

ElizabethGoddard:
Hey @georgeschena let's pair on this as it's easiest but just adding a list here for our reference to go through:

- [ ] Full stop after 'Select all that apply'
- [ ] 40px between each group and buttons at the bottom
- [ ] Padding on events button
- [ ] Check padding on 'Give details'
- [ ] Discuss box width
- [ ] Discuss inset text option from discussion with Tom

common/services/event.js
thomasleese: I think ideally we'd use the police lockout location, but if that's impossible then `move.to_location.id` is better, however it wouldn't be entirely accurate (users might be confused if they see the event happened at the start location when actually it happened during a lockout).

app/move/app/view/views/_layout.njk
thomasleese: Oh nice!

app/police-custody-form/police-custody-form.njk
thomasleese: I think in this case then it's worth moving the `setMove` middleware in to the common middlewares since it's useful in multiple apps, and means we can avoid unnecessary duplication.

PR #2399, opened 2022-04-11
Reduce default field length

thomasleese:
For single line top level text fields the length ends up being the full length of the page. This looks a little odd, and it's inconsistent with the text area fields. We'd like to change the default so it looks better.

### Old

Screenshot 2022-04-11 at 14 45 24

### New

Screenshot 2022-04-11 at 14 47 23


Thource (approved):

ElizabethGoddard:
Looks good thank you @thomasleese

PR #2432, opened 2022-04-22
Use the journeys to_location for lockout events payload

georgeschena:


app/police-custody-form/controllers.js
```diff
@@ -5,6 +5,7 @@ exports.addEvents = async function (req, res) {
   const user = req.user
   const move = req.move
   const moveId = req.move.id
+  const journeys = req.journeys
```
thomasleese, line 8: I know it's not related to this PR, but we could refactor this:

```suggestion
  const { user, move, journeys } = req
  const moveId = move.id
```

thomasleese (approved):

PR #2433, opened 2022-04-22
Create new 'Lockouts events added successfully' flash banner for when events have been added

georgeschena:
### What changed

When a user submits lockout events successfully, a flash banner will appear notifying them it has been successful. 

### Why did it change

To allow the user to know that their lockout events have been submitted

### Issue tracking

- [P4-3510](https://dsdmoj.atlassian.net/browse/P4-3510)

## Screenshots

Screenshot 2022-04-22 at 14 33 19

### DIFF
--- a/app/police-custody-form/controllers.js
+++ b/app/police-custody-form/controllers.js
@@ -23,7 +23,13 @@ exports.addEvents = async function (req, res) {
 
   await req.services.event.postEvents(lockoutEvents, move, journeys, user)
 
-  req.session.successBanner = true
+  const fullName = move.profile.person._fullname
+
+  req.flash('success', {
+    title: req.t('messages::events_added.heading'),
+    content: req.t('messages::events_added.content', { fullName }),
+  })
+
   return res.redirect(`/move/${moveId}/timeline`)
 }
 
--- a/app/police-custody-form/controllers.js
+++ b/app/police-custody-form/controllers.js
@@ -23,6 +23,7 @@ exports.addEvents = async function (req, res) {
 
   await req.services.event.postEvents(lockoutEvents, move, journeys, user)
 
+  req.session.successBanner = true
   return res.redirect(`/move/${moveId}/timeline`)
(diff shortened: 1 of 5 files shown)


app/police-custody-form/controllers.js
```diff
@@ -23,6 +23,7 @@ exports.addEvents = async function (req, res) {
 
   await req.services.event.postEvents(lockoutEvents, move, journeys, user)
 
+  req.session.successBanner = true
```
thomasleese, line 26: `req.flash` should store the message in the session, so you should be able to call it here directly before the redirect and avoid the logic in the timeline view. This also means we can change the redirect without having to change that logic, and keeps it close to where the success happens.

locales/en/messages.json
```diff
@@ -70,5 +70,9 @@
     "content":"",
     "content_youth_risk_assessment":"The $t({{context}}) for <strong>{{name}}</strong> has been successfully confirmed. You can now start the Person Escort Record.",
     "content_person_escort_record":"The handover for <strong>{{name}}</strong> has been recorded. You can view the details recorded in the <a href=\"{{timelineUrl}}\">history of events</a>."
+  },
+  "events_added": {
+    "heading": "Lockout events added successfully",
+    "content": "You have added new lockout events for <b>{{fullName}}</b>. You can view the details in the <a href='#tab-history-of-events' class='govuk-link'>history of events</a>."
```
thomasleese, line 76:
```suggestion
    "content": "You have added new lockout events for <strong>{{fullName}}</strong>. You can view the details in the <a href='#tab-history-of-events' class='govuk-link'>history of events</a>."
```

I think generally we use a `strong` tag as it's less about the presentation and instead that's controlled in CSS - and it's better for accessibility.

app/police-custody-form/controllers.js
georgeschena: Ohh nice!

PR #2467, opened 2022-04-29
Handle when move has no important events

thomasleese:
It seems like sometimes the move won't have any important events, so we should make sure to handle this situation.

[Sentry Issue](https://sentry.io/organizations/ministryofjustice/issues/3229345386/?referrer=slack)


Thource:
:rocket: PR Preview successfully deployed at https://hmpps-book-secure-move-frontend-pr-2467.apps.live-1.cloud-platform.service.justice.gov.uk

PR #2499, opened 2022-05-11
Improve content for the handover event

thomasleese:
This improves the textual content for the handover event as per content designer request. The main difference is that the location of where the handover occurred is now included, as this supports lockouts where there can be more than one handover event.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3662)

### Old

Screenshot 2022-05-11 at 15 21 32

### New

Screenshot 2022-05-11 at 15 17 44

### DIFF
--- a/locales/en/events.json
+++ b/locales/en/events.json
@@ -153,7 +153,7 @@
   },
   "MoveLodgingStart": {
     "heading": "Lodging started",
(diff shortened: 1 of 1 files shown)


matt-crowson-moj (approved):

Thource (approved):

SophieDodds:
I think this is good to go other than the odd numbers which I know happen automatically and also happen in the events too. We can deal with that issue in another ticket. Happy to merge, thanks!

PR #2512, opened 2022-05-13
Exclude certain events from tags list

thomasleese:
These were added accidentally when we started including these events in db2990edad77481b2e34e4274be7ebac6d40c42b and 5532e88971848e0d77ed60174c4624068d851d7a.

### Old

Screenshot 2022-05-13 at 11 03 39

### New

Screenshot 2022-05-13 at 11 03 43


ElizabethGoddard:
Hi Tom, just to check this will remove any events that aren't warnings (with exclamations), so also lodging ended and anything else, right?

thomasleese:
> Hi Tom, just to check this will remove any events that aren't warnings (with exclamations), so also lodging ended and anything else, right?

It removes the PER handover, lodging start and lodging end events. It doesn't touch the next of kin, or property change events, for example.

Thource (approved):

PR #2636, opened 2022-06-22
Dedicated self contained Contenful links

georgeschena:
### What changed

Allow a content designer to create content, and allow the content to generate a self contained link in order to share this single piece of content to dedicated users.

### Why did it change

So we can target specific users to send dedicated content too

### Issue tracking

- [P4-3759](https://dsdmoj.atlassian.net/browse/P4-3759)


Thource:
@dependabot rebase

PR #2639, opened 2022-06-23
Fix remove property button from ghosting when adding new property element

georgeschena:
### What changed

Updated the scss styles used on the Remove property button

Note: **This issue is around the `add-another` component, which actually affected all pages where you would add another element to a form**

### Why did it change

The old scss stylings caused the button to act in a way which wasn't what we wanted, not allowing the user to remove a property item due to it ghosting.

### Issue tracking

- [P4-3723](https://dsdmoj.atlassian.net/browse/P4-3723)

### DIFF
--- a/common/components/add-another/template.njk
+++ b/common/components/add-another/template.njk
@@ -42,7 +42,7 @@
     {% endif %}
   {% endfor %}
 
-  <div class="app-add-another__add-item {{ 'app-border-top-1 govuk-!-padding-top-5' if items.length > 0 }}">
+  <div class="app-add-another__add-item {{ 'app-border-top-1 govuk-!-padding-top-5 govuk-!-margin-top-4' if items.length > 0 }}">
     {{ govukButton({
       text: t("actions::add_item", {
         context: "with_items" if items.length > 0,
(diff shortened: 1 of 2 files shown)


common/components/add-another/template.njk
```diff
@@ -42,7 +42,7 @@
     {% endif %}
   {% endfor %}
 
-  <div class="app-add-another__add-item {{ 'app-border-top-1 govuk-!-padding-top-5' if items.length > 0 }}">
+  <div class="app-add-another__add-item {{ 'app-border-top-1 govuk-!-padding-top-5 govuk-!-margin-top-4' if items.length > 0 }}">
```
matt-crowson-moj, line 45: Does this mean there is no test coverage for property removal?

matt-crowson-moj (approved):

common/components/add-another/template.njk
georgeschena: There are tests around the `add-another` component here `common/components/add-another/template.test.js`. But not directly testing that it has any styles associated to the component itself it seems. @matt-crowson-moj

georgeschena: The property removal uses this component.

PR #2702, opened 2022-07-18
Allow new content pages to be created in Book a secure move

georgeschena:
### What changed

We have added dedicated content pages where content and designers can add new `dedicatedContent` posts which allow us to generate new url links to display static information.

Within Contentful you now have a new content model where you would create your dedicated content, and then by using the `title` of the post we translate that into a `url slug` which the user can type into the web address bar and be taken to that specific bit of content. 

Once the content has been created all you need to do is visit `content/{URL_SLUG-OF-CONTENT}` and you will be taken to the webpage. 

### Why did it change

As we wanted to send out content designed for specific user groups

### Issue tracking

- [P4-3759](https://dsdmoj.atlassian.net/browse/P4-3759)


app/contnet/content.njk
```diff
@@ -0,0 +1,15 @@
+{% extends "layouts/base.njk" %}
```
Thource, line 1: typo in directory name

georgeschena: Doh

common/services/contentful.ts
```diff
@@ -87,22 +87,37 @@ const service = {
...
-  convertToHTMLFormat: body => documentToHtmlString(body, options),
-  formatDate: date => format(date, DATE_FORMATS.WITH_MONTH),
+  fetchEntryBySlugId: async (slugId: any) => {
+    const entry = await service.client.getEntries({
+      content_type: 'dedicatedContent',
+      'fields.slug[in]': slugId,
+    })
+
+    if (!entry.items?.length) {
+      return []
```
Thource, line 114: I feel like this should be `return undefined` so that the function signature is `Object?` instead of `Object | Array`

Thource (approved):

Thource (approved):

PR #2750, opened 2022-08-05
Add development tool to toggle off the whats new banner

georgeschena:
### What changed

So this change is to add alink in the app-development-tools div when clicked clears the banner-dismissed cookie from the users browser.

### Why did it change

For testing of the banner, we want to bring back the banner after it has been dismissed by the user, so the next user can test out the banner. In order to do this currently you need to go into the browser dev tools and delete the cookie manually, which for no technical users could be a pain. 

### Issue tracking

- [P4-3833](https://dsdmoj.atlassian.net/browse/P4-3833)


app/home/development-tools.ts
```diff
@@ -0,0 +1,27 @@
...
+  cacheEls ($module: any, $document: any) {
+    this.$module = $module
+    this.$document = $document
+  }
+
+  toggleWhatsNewBanner () {
+    const decodedCookie = decodeURIComponent(this.$document.cookie)
+    const cookies = decodedCookie.split(';')
+
+    const result = cookies.filter(cookie => cookie.includes('banner-dismissed'))
```
matt-crowson-moj, line 19: Minor. Rename result to something like `isBannerDismissed`.

matt-crowson-moj (approved):

matt-crowson-moj (dismissed):

matt-crowson-moj (approved):

Thource (approved):

PR #2892, opened 2022-11-08
feat: Add new warning box for incomplete PER when move is today

Thource:
### What changed

A new warning box has been added to the moves overview page when a move is booked for today and the PER is not completed.

### Why did it change

This has changed to make it easier for users to see when a PER needs to be completed.

### Issue tracking

- P4-3931

## Screenshots

| Before | After |
| ------ | ----- |
| <kbd></kbd> | <kbd></kbd> |

#### Automated testing

#### Manual testing

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables


danbenton-mojdt (dismissed):

ElizabethGoddard:
Thanks @Thource! As discussed if you could make the links go directly into the question pages (just for the first time in, so same logic as doing the PER through the other route) for this ticket and the first one that would be great. Other than that, this is approved from my side. Thanks!

danbenton-mojdt (approved):

PR #3159, opened 2023-11-14
Migrating project to use groups in allowlist

andrewrlee:
This PR migrates the project to use groups of IPs in their allowlist.

By referring to groups to IP addresses, we can centralize the definition of groups of ip addresses.
If these lists require changing in the future, we can change the definition once and future deploys across all services will automatically include these new IPs.

0 allowlist(s) have been detected that can be migrated.


andrewrlee:
I'll close as there is no allowlist

PR #3208, opened 2024-08-14
feat: MAP-1503 Highlight when journey dates override move date

danbenton-mojdt:
image

### DIFF
--- a/common/presenters/move-to-meta-list-component.js
+++ b/common/presenters/move-to-meta-list-component.js
@@ -68,6 +68,8 @@ function moveToMetaListComponent(move, journeys, { updateUrls = {} } = {}) {
     }
   }
 
+  const hasJourneys = journeys !== null && journeys.length > 0
+
   const journeysSummary = moveToJourneysSummary(move, journeys, {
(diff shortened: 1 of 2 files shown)


jimbali (approved):

common/presenters/move-to-meta-list-component.js
```diff
@@ -79,6 +81,9 @@ function moveToMetaListComponent(move, journeys, { updateUrls = {} } = {}) {
       })}`
     : journeysSummary.map(({ toLocation }) => toLocation).join(' and ')
 
+  const journeyDates = journeysSummary.map(({ date }) => date).join(' to ')
```
jimbali, line 84: As discussed, we should probably just do "{first journey date} to {last journey date}", but I'm going to address that under MAP-510 anyway

jimbali (approved):

common/presenters/move-to-meta-list-component.js
danbenton-mojdt: 👍

PR #3222, opened 2024-11-20
chore: Move security tasks from CircleCI to GH actions

Thource:


BarryGee-MOJ (dismissed):
LGTM

.github/workflows/security_trivy.yml
```diff
@@ -0,0 +1,12 @@
+name: Security trivy dependency check
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "12 6 * * MON-FRI" # Every weekday
+jobs:
+  security-kotlin-trivy-check:
```
jimbali, line 7: Probably should call this something without Kotlin in the name if we can

jimbali (approved):

PR #3237, opened 2024-12-06
MAP-881: Optional fields

jimbali:
### What changed

- Add " (optional)" to some fields
- Add a negative option to the bottom of some checkbox lists and require a response

### Why did it change

- To meet accessibility recommendations

### Issue tracking

- MAP-881

## Screenshots

Example of negative option:

Screenshot 2024-12-06 at 14 17 38

### DIFF
--- a/app/move/app/new/controllers/health-information.ts
+++ b/app/move/app/new/controllers/health-information.ts
@@ -1,0 +1,15 @@
+import addNegativeOption from '../../../../../common/helpers/field/add-negative-option'
+// @ts-expect-error TODO: convert to TS
+import AssessmentController from '../controllers/assessment'
+
+class HealthInformationController extends AssessmentController {
+  processFields(fields: Record<string, any>) {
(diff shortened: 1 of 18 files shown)


app/move/app/new/controllers/health-information.ts
```diff
@@ -0,0 +1,15 @@
+import addNegativeOption from '../../../../../common/helpers/field/add-negative-option'
+// @ts-expect-error TODO: convert to TS
```
tobyprivett, line 2: Is this file already in typescript?

tobyprivett (approved):

app/move/app/new/controllers/health-information.ts
jimbali: The file it links to is still a .js file but it is a module. This comment will cause a lint error when we convert it to TypeScript, which will remind us to remove the comment.

PR #3241, opened 2024-12-19
Remove references to the old live-1 URL

jimbali:
Remove references to the old live-1 URL. I have also updated the environment variable in CircleCI.


jnioche-jd (approved):
LGTM

PR #3242, opened 2024-12-20
MAP-1976: Remove default URL

jimbali:
We never have a legitimate reason to use the `hmpps-book-secure-move-frontend-production.apps.cloud-platform.service.justice.gov.uk` URL so let's remove it as it is causing SSL cert validation errors.

MAP-1976


GurnankCheema (approved):
LGTM

PR #3256, opened 2025-02-19
feat(MAP-1857): Add Redis cache to contentful api call

BarryGee-MOJ:
### What changed

- Added Redis caching to fetchEntries() to store the data returned from Contentful. This reduces the number of calls made to Contentful. A TTL of 300 seconds / 5 minutes has been added. After which time the application will make a new call to Contentful to retrieve new data.

- Resolved failing unit tests

- Resolved linting errors  

### Why did it change

To reduce the number of calls being made to Contentful. 

### Issue tracking

- MAP-1857

#### Automated testing

#### Manual testing

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

### Other considerations

### DIFF
--- a/common/services/contentful/contentful.ts
+++ b/common/services/contentful/contentful.ts
@@ -148,28 +148,28 @@ export class ContentfulService {
 
     if(cachedEntries) {
       return cachedEntries
-    } else {
-      let entries = (await this.client.getEntries({
-        content_type: this.contentType,
-      })) as contentful.EntryCollection<ContentfulFields>
-
-      const entriesToCache = entries.items?.length ? entries : []
-
-      await set(`cache:entries:${this.contentType}`, 
-        entriesToCache,
-        300,
-        true)
-
-        if (!entries.items?.length) {
-          return []
-        } 
-
-      return entries.items
-        .map(e => this.createContent(e.fields))
-        .sort((a, b) => {
-          return b.date.getTime() - a.date.getTime()
-        })
     }
+
+    const entries = (await this.client.getEntries({
+      content_type: this.contentType,
+    })) as contentful.EntryCollection<ContentfulFields>
+
+    const entriesToCache = entries.items?.length ? entries : []
+
+    await set(`cache:entries:${this.contentType}`, 
+      entriesToCache,
+      300,
+      true)
+
+      if (!entries.items?.length) {
+        return []
+      } 
+
+    return entries.items
+      .map(e => this.createContent(e.fields))
+      .sort((a, b) => {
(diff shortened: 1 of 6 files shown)


common/services/contentful/contentful.ts
```diff
@@ -142,20 +143,32 @@ export class ContentfulService {
...
+
+    if(cachedEntries) {
+      return cachedEntries
+    } else {
+      const entries = (await this.client.getEntries({
+        content_type: this.contentType,
+      })) as contentful.EntryCollection<ContentfulFields>
+
+      if (!entries.items?.length) {
+        return []
```
Thource, line 158:
```suggestion
        entries = []
```

If we return here, we're not setting anything in the cache and then we will still be making lots of calls when there are no entries

common/services/contentful/contentful.ts
```diff
@@ -143,19 +144,32 @@ export class ContentfulService {
...
-    })) as contentful.EntryCollection<ContentfulFields>
-
-    if (!entries.items?.length) {
-      return []
+    let cachedEntries = await get(`cache:entries:${this.contentType}`, true)
+
+    if(cachedEntries) {
+      return cachedEntries
+    } else {
+      let entries = (await this.client.getEntries({
```
Thource, line 152:
```suggestion
      const entries = (await this.client.getEntries({
```

common/services/contentful/dedicated-content.test.ts
```diff
@@ -4,14 +4,19 @@ import sinon from 'sinon'
...
 describe('DedicatedContentService', function () {
   beforeEach(function () {
     service = new DedicatedContentService()
+
     sinon.stub((service as any).client, 'getEntries').resolves([])
   })
 
   it('requests the right content type from Contentful', async function () {
-    await service.fetch()
+    await (service as any).client.getEntries({
```
Thource, line 16: Could we type this, rather than casting to any?

common/services/contentful/downtime.test.ts
```diff
@@ -5,20 +5,47 @@ import { mockDate, unmockDate } from '../../../mocks/date'
...
+    sinon.stub(service, 'fetch').resolves(formattedEntriesMockResponse)
   })
 
   afterEach(function () {
     unmockDate()
   })
 
   it('requests the right content type from Contentful', async function () {
-    await service.fetch()
+    await (service as any).client.getEntries({ content_type: 'downtime' })
```
Thource, line 47: Same here?

Thource (approved):

PR #3259, opened 2025-03-12
[MAP-2156] Restrict NOMIS user access outside of allowed IP addresses

Thource:
### What changed
A middleware has been added to the highest level express router so that if any NOMIS users are logged in outside of the accepted IP addresses, they will be shown a 403 error and be unable to access the service.

### Why did it change
This has changed so that prison users can only use BaSM while at work.

### Issue tracking

- MAP-2156

#### Automated testing

- [x] Added unit tests to cover changes
- [ ] Added end-to-end tests to cover any journey changes

#### Manual testing

Has been tested in the following browsers:

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Internet Explorer

### Environment variables

- [ ] No environment variables were added or changed

- [ ] Documented in the [README](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/README.md)
- [x] Added to [continous integration](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-book-secure-move-frontend)
- [x] Added to Helm config

### Other considerations

Commit messages with a `fix` or `feat` type are automatically used to generate the [changelog](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/CHANGELOG.md) and
[GitHub release notes](https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/releases) during the release task. Please make sure they will read well on their own in a
summary of changes and that the commit body gives a more detailed description of those changes if necessary.

- [x] No new [Personally Identifiable Information (PII)](https://support.google.com/analytics/answer/6366371?hl=en) is being sent via analytics
- [ ] Update [README](ministryofjustice/hmpps-book-secure-move-frontend/blob/main/README.md) with any new instructions or tasks

PR #3268, opened 2025-03-24
MAP-2156 Revert commits to 'prevent prison staff from accessing off network;

tobyprivett:
## Proposed changes

Here we revert the commits that implement Off-Network access restrictions. 
https://dsdmoj.atlassian.net/browse/MAP-2156

On deployment, we learned that various users were unable to log in.
This may be due to their user accounts and/or IP addresses.

In the meantime, we should revert that work and open up the deployment pipeline.

### What changed

Reverts:
- Restrict NOMIS user access outside of allowed IP addresses (#3259)
- Support subnets in checkStaffNetwork (#3263)
- Open up staging allowlist (#3264)
 

[MAP-2156]: https://dsdmoj.atlassian.net/browse/MAP-2156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BarryGee-MOJ (approved):
lgtm

PR #3369, opened 2025-09-17
MAP-2377 Special vehicle interrupt

danbenton-mojdt:
Presents this interrupt when pregnant and/or wheelchair is selected, automatically requesting special vehicle.

This is in order to help prevent moves being requested for custodies that are pregnant without requesting special vehicles

image


dafhoustonmoj (approved):
LGTM

